PR #97, opened 2017-11-30
fix cloning part at installation section at README

strobe:
current clone instruction won't works because  git clone doesn't download submodules  like 'nailgun-server'
therefore it seems like 
'git clone --recursive -j8 https://github.com/scalacenter/bloop.git' should be used (for git >= 1.9)


strobe:
@Duhemm
I fixed `-j8` thing

Issue #101, opened 2017-12-01
No tests run for project which aggregates others


leszekgruchala:
Bloop version: `66018e0a`

I have multi module sbt project.
- core
- web (which depends on core)
- root (which aggregates core and web).

Running tests on root does nothing
```
bloop -- test -p root
bloop -- test -p root-test
```
No output at all.

I have to run tests on project which contains them.

jvican:
Can you not do `--aggregate`?

leszekgruchala:
`bloop -- test -p root-test --aggregate` helped 👍

leszekgruchala:
Can you derive this information from sbt configuratation?

jvican:
We do have all that information from sbt, I guess that what you want is the `--aggregate` is enabled directly for "root" sbt projects, right? If this is what you wanted it wasn't clear :smile:.
  
A better and cleaner option than that one would be to make `--aggregate` the default, since it's the most common anyway. We would add `--isolated` or `--single` to signal that only the single project should be tested.

PR #109, opened 2017-12-01
Add `--aggregate` option to `clean`

Duhemm:
This specifies that all the dependencies of the specified project should
be cleaned too.


jvican:
I'm torn on this one. It's not clear to me we should opt-in for the aggregated version instead of having it as the default, as I discuss in https://github.com/scalacenter/bloop/issues/101. 
  
I think that as the most common case is the aggregated version, we should just default to it and add a flag `--isolated` or `--single` to apply the operation only for one concrete project.
  
Also, we need to clarify whether `--aggregate` wipes up *all* the projects or only the reachable projects from a passed in project. What is the best thing to do in this case? No strong opinion on my side.

jvican:
Closing in favor of #164.

PR #110, opened 2017-12-01
Exract `classpathOptions` in sbt-bloop

Duhemm:


Duhemm:
That's a genuine failure, but I can't reproduce it on my machine...

Duhemm:
I couldn't reproduce the failure on my machine then, and I still can't. Hopefully it'll work now!

jvican:
@Duhemm Did you finally print classpaths here to see how `classpathOptions` is making a difference?

Duhemm:
It's on today's list :)

jvican:
Has the mistery been solved? We may want to find a sorcerer to help us find out what's happening.

Duhemm:
Debug info shows that this is extremely likely to be caused by our integration tests setup: several entries of the classpath and boot classpath do no exists when we do the compilation, because they're downloaded in scripted when we do the setup.

Let's see if sharing the same ivy home for scripted and normal sbt solves the issue.

Duhemm:
I'm pretty sure it didn't run the community build

@jvican Do you know why?

Duhemm:
Nevermind, I had forgotten about your PR that fixes it :)

Duhemm:
test performance please

bloopoid:

bloopoid:

jvican:
Any news on this? :smile:

Duhemm:
Yes, it works. It crashed because it had the wrong configurations (hence the `NPE`s).

frontend/src/test/scala/bloop/tasks/IntegrationTestSuite.scala
```diff
@@ -70,7 +73,11 @@ class IntegrationTestSuite(testDirectory: Path) {
         (state, rootProject)
     }
 
-    val reachable = Dag.dfs(state.build.getDagFor(projectToCompile)).filter(_ != projectToCompile)
+    val reachable =
```
jvican: Why is this necessary?

integrations/maven-bloop/src/main/java/bloop/integrations/maven/BloopMojo.java
```diff
@@ -35,6 +37,21 @@
     @Parameter(property = "bloop.executionFork", defaultValue = "false")
     private boolean bloopExecutionFork;
 
+    @Parameter(property = "bloop.classpathOptions.bootLibrary", defaultValue = "true")
```
jvican: Have you looked if the maven integration already supports all these?

frontend/src/main/scala/bloop/Project.scala
```diff
@@ -96,6 +99,12 @@ object Project {
       ScalaInstance(scalaOrganization, scalaName, scalaVersion, allScalaJars, logger)
     val classpath = toPaths(properties.getProperty("classpath"))
     val classesDir = toPath(properties.getProperty("classesDir"))
+    val classpathOptions = {
+      val values = properties.getProperty("classpathOptions").split(",")
+      val Array(bootLibrary, compiler, extra, autoBoot, filterLibrary) =
```
jvican: The error reporting if this fails is poor here.

bin/detect-community-build.sh
```diff
@@ -2,11 +2,18 @@
...
 if [[ "$PR" ]]; then
-  MONITOR_REPO="scalacenter/bloop"
-  if curl --silent https://api.github.com/repos/"$MONITOR_REPO"/issues/"$PR"/labels | jq -e '.[].name | contains("community-build")'; then
+  if curl --silent https://api.github.com/repos/"$MONITOR_REPO"/issues/"$PR"/labels | jq -e '.[].name | contains("community-build")' > /dev/null 2>&1 ; then
     echo "The community build will be run for this pull request."
-    export RUN_COMMUNITY_BUILD=true
+    exit 0
+  else
+    echo "The community build will not run for this pull request."
+    exit 1
```
jvican: Doesn't this mean that Drone will fail altogether?

backend/src/main/scala/bloop/CompilerCache.scala
```diff
@@ -15,13 +15,14 @@ class CompilerCache(componentProvider: ComponentProvider,
                     logger: Logger,
                     userResolvers: List[Resolver]) {
 
-  private val cache = new ConcurrentHashMap[ScalaInstance, Compilers]()
+  private type CompilerID = (ScalaInstance, ClasspathOptions)
+  private val cache = new ConcurrentHashMap[CompilerID, Compilers]()
```
jvican: Why is this necessary? It's not clear to me why classpath options is part of the compiler key, this is not nice performance wise because every time the classpath options change we need to instantiate a new compiler. Can we avoid this change?

frontend/src/test/scala/bloop/tasks/IntegrationTestSuite.scala
Duhemm: I renamed `state` to `initialState`, and `state` is now the state after cleaning. We need to clean because the project may have already been compiled by a previous test.

integrations/maven-bloop/src/main/java/bloop/integrations/maven/BloopMojo.java
Duhemm: I've been through the plugin but I haven't found how this could be configured, so I assumed it didn't.

jvican: Can you double-check?

bin/detect-community-build.sh
Duhemm: Good catch, the problem was actually in `.drone.yml`

.drone.yml
```diff
@@ -53,10 +53,12 @@ pipeline:
       - git log | head -n 20
       - ./bin/add-coursier.sh
       - ./bin/detect-community-build.sh && export RUN_COMMUNITY_BUILD=true || echo "Not running the community build."
-      - case "$DRONE_COMMIT_MESSAGE" in 
-        *[DOCS]*) echo "Skipping tests for commit that only changes docs." ;;
-        *) ${SBT_RUN}; ./bin/ci-clean-cache.sh ;;
-        esac
+      - if [[ "$DRONE_COMMIT_MESSAGE" == *[DOCS]* ]]; then
```
jvican: This fails in sh, that's why it's implemented via `case`

jvican:
If you want to regenerate the configuration files, you need to update the version number of the configuratino file in our build, you can drop https://github.com/scalacenter/bloop/pull/110/commits/4fe198883f37a9e9fcd467e3d2ddec8fcbe90008

PR #113, opened 2017-12-04
Fix `ExecutionContext` preventing exit

Duhemm:
The `ExecutionContext` that we were using was preventing application
exit because the backing thread pool was never shutdown, thus preventing
the application from exiting.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -6,47 +6,47 @@ import bloop.io.Timer.timed
 import bloop.logging.Logger
 import bloop.reporter.ReporterConfig
 import bloop.tasks.{CompilationTasks, TestTasks}
-import ExecutionContext.threadPool
 import bloop.util.TopologicalSort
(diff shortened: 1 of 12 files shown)


frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -6,47 +6,47 @@ import bloop.io.Timer.timed
...
-      execute(next, logger)
-    case Run(Commands.Test(projectName, aggregate, scalacstyle, cliOptions), next) =>
-      logger.verboseIf(cliOptions.verbose) {
-        val reporterConfig =
-          if (scalacstyle) ReporterConfig.scalacFormat else ReporterConfig.defaultFormat
-        test(projectName, aggregate, cliOptions, reporterConfig, logger)
-      }
-      execute(next, logger)
-  }
+  def execute(action: Action, logger: Logger)(implicit ec: ExecutionContext): ExitStatus =
```
jvican: Can we add curly braces around this patmat please? :smile:

jvican:
We should look into this after benchmarks are in! :smile:.

PR #116, opened 2017-12-04
Disable parallel execution in test

Duhemm:


Duhemm:
The build passed 5 times in a row. What do you think @jvican? It takes about 20% more time to run the build.

jvican:
Let's go for this and see if it stabilizes. When CI times become a problem, we'll get back to this.

PR #118, opened 2017-12-06
Add benchmarks for bloop

Duhemm:


benchmarks/src/main/scala/bloop/HotBloopBenchmark.scala
```diff
@@ -0,0 +1,109 @@
+package bloop
```
jvican: Good that we have them in our own project.

benchmarks/src/main/scala/bloop/HotBloopBenchmark.scala
```diff
@@ -0,0 +1,109 @@
...
+  var tempDir: Path = _
+  var processOutputReader: BufferedReader = _
+  var processInputReader: BufferedWriter = _
+  var output = new java.lang.StringBuilder()
+
+  @Setup(Level.Trial) def spawn(): Unit = {
+    tempDir = Files.createTempDirectory("bloop-")
+    val configDir = Files.createDirectory(tempDir.resolve(".bloop-config"))
+
+    val path = Paths.get(s"../frontend/src/test/resources/projects/$project")
```
jvican: What happens if the project doesn't exist? Shouldn't we report it?

build.sbt
```diff
@@ -32,6 +32,13 @@ val nailgun = project
     hijackScalafmtOnCompile in SbtConfig in NailgunBuild := false,
   )
 
+val compilerBenchmark = project
```
jvican: I propose we call this benchmark bridge.

build.sbt
```diff
@@ -82,6 +90,11 @@ val frontend = project
     parallelExecution in Test := false
   )
 
+val benchmarks = project
```
jvican: Do you add this to `allProjects`?

frontend/src/test/scala/bloop/tasks/IntegrationTestSuite.scala
```diff
@@ -25,25 +25,9 @@ object IntegrationTestSuite extends DynTest {
...
-        javacOptions = Array.empty,
-        sourceDirectories = Array.empty,
-        previousResult = CompilationHelpers.emptyPreviousResult,
-        tmp = classesDir,
-        testFrameworks = Array.empty,
-        origin = None
-      )
-      projects + (rootProjectName -> rootProject)
-    }
+    val projects = ProjectHelpers.loadTestProject(testDirectory.getFileName.toString,
```
jvican: Is it possible to avoid doing these changes? I change these things quite a bit, and looks unrelated to the PR?

project/BuildPlugin.scala
```diff
@@ -68,6 +72,35 @@ object BuildKeys {
     Keys.libraryDependencies += Dependencies.utest,
     Keys.testFrameworks += new sbt.TestFramework("utest.runner.Framework"),
   )
+
+  import sbtassembly.AssemblyKeys
```
jvican: :heart:

project/BuildPlugin.scala
```diff
@@ -68,6 +72,35 @@ object BuildKeys {
...
+  val assemblySettings: Seq[Def.Setting[_]] = List(
+    Keys.mainClass in AssemblyKeys.assembly := Some("bloop.Bloop"),
+    Keys.test in AssemblyKeys.assembly := {}
+  )
+
+  val benchmarksSettings: Seq[Def.Setting[_]] = List(
+    Keys.skip in Keys.publish := true,
+    Keys.javaOptions ++= {
+      def refOf(version: String) = {
```
jvican: You could also do it via dynver, which provides a very similar thing to this.

project/BuildPlugin.scala
```diff
@@ -68,6 +72,35 @@ object BuildKeys {
...
+        }
+      }
+      List(
+        "-DscalaVersion=" + Keys.scalaVersion.value,
+        "-DscalaRef=" + refOf(Keys.scalaVersion.value),
+        "-Dsbt.launcher=" + (sys
+          .props("java.class.path")
+          .split(java.io.File.pathSeparatorChar)
+          .find(_.contains("sbt-launch"))
+          .getOrElse("")),
```
jvican: Are we sure we want the empty string if sbt launch is not there? If this is a corner case, wouldn't it make more sense to throw an exception right away?

project/BuildPlugin.scala
```diff
@@ -68,6 +72,35 @@ object BuildKeys {
     Keys.libraryDependencies += Dependencies.utest,
     Keys.testFrameworks += new sbt.TestFramework("utest.runner.Framework"),
   )
+
+  import sbtassembly.AssemblyKeys
+  val assemblySettings: Seq[Def.Setting[_]] = List(
```
jvican: Will assembly directly add the nailgun server and all the bridges to this mega jar?

benchmarks/src/main/scala/bloop/HotBloopBenchmark.scala
Duhemm: It'd crash a little after that, but let's throw an exception here.

project/BuildPlugin.scala
Duhemm: Yes, the nailgun server is part of the assembly, but not the bridges.

project/BuildPlugin.scala
Duhemm: You only need that when you want to benchmark sbt, so there are situations where it may be okay.

build.sbt
Duhemm: I've changed it

Duhemm:
I think I've addressed all the comments 🎉

PR #121, opened 2017-12-07
Improve bloop internals

jvican:
Our previous internal model was okay for a prototype but I don't think it
could stand the test of time.

This commit creates a better internal model that rests on the concept of
`State`. `State` is immutable and contains all the pieces necessary for
compiling and testing projects.

This model has the advantage that is easy to reason about; it isolates the
previous compilation results from the project, it removes the confusing
`Map[String, Project]`, and it allows for more code reuse and value
caching.

There is one `State` per `AbsolutePath`, meaning that by design we don't
support concurrent compilation of the same project. We don't enforce this
yet, though.

Aside the `State` abstraction, this PR also introduces `Dag`, which
represents a direct acyclic graph that is useful when traversing
dependencies of projects and doing reachability analysis.

As a plus, this commit removes all warnings in our codebase.


build.sbt
```diff
@@ -79,6 +79,32 @@ val frontend = project
     buildInfoKeys := BloopInfoKeys,
     fork in run := true,
     javaOptions in run ++= Seq("-Xmx4g", "-Xms2g"),
+    libraryDependencies += Dependencies.graphviz % Test,
```
jvican: This is for testing dot graph generation.

frontend/src/main/scala/bloop/Cli.scala
```diff
@@ -124,5 +124,37 @@ object Cli {
...
+    import bloop.io.AbsolutePath
+    def getConfigDir(cliOptions: CliOptions): AbsolutePath = {
+      cliOptions.configDir
+        .map(AbsolutePath.apply)
+        .getOrElse(cliOptions.common.workingPath.resolve(".bloop-config"))
+    }
+
+    import bloop.engine.{State, Build}
+    def loadStateFor(configDirectory: AbsolutePath, logger: Logger): State = {
+      State.stateCache.getBuildFor(configDirectory) match {
```
Duhemm: `getBuildFor` returns a `State`?

frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -39,6 +39,7 @@ object Commands {
       @HelpMessage(
         "If set, displays compilation message following scalac's style. Defaults to false.")
       scalacstyle: Boolean = false,
+      @Hidden prependCompile: Boolean = true,
```
Duhemm: 👍

frontend/src/main/scala/bloop/engine/Dag.scala
```diff
@@ -0,0 +1,90 @@
...
+final case class Parent[T](value: T, children: List[Dag[T]]) extends Dag[T]
+
+object Dag {
+  private[bloop] class RecursiveCycle(path: List[Project])
+      extends Exception(s"Not a DAG, cycle detected in ${path.map(_.name).mkString(" -> ")} ")
+
+  def fromMap(projectsMap: Map[String, Project]): List[Dag[Project]] = {
+    val visited = new scala.collection.mutable.HashMap[Project, Dag[Project]]()
+    val visiting = new scala.collection.mutable.LinkedHashSet[Project]()
+    val dependees = new scala.collection.mutable.HashSet[Dag[Project]]()
```
Duhemm: Do you mean `dependents`? The nodes that depend on a given node? I don't think `dependee` exists

frontend/src/main/scala/bloop/engine/tasks/CompileTasks.scala
```diff
@@ -0,0 +1,117 @@
...
+        .getResult(project)
+        .getOrElse(sys.error("Results cache was not initialized"))
+      val inputs = toInputs(project, reporterConfig, previousResult)
+      val result = Compiler.compile(inputs)
+      results.put(project, result)
+      List(project)
+    }
+
+    def compileTask(project: Project) =
+      new Task((_: Seq[Project]) => runCompilation(project), () => ())(Mergeable.SeqMergeable)
```
Duhemm: The implicit is not found otherwise?

frontend/src/main/scala/bloop/engine/tasks/CompileTasks.scala
```diff
@@ -0,0 +1,117 @@
...
+      import scala.collection.JavaConverters._
+      val updatedResults = results.asScala.iterator.foldLeft(state.results) {
+        case (results, (project, result)) => results.updateCache(project, result)
+      }
+      state.copy(results = updatedResults)
+    }
+
+    val taskGraph = constructTaskGraph(state.build.getDagFor(project))
+    Await.result(taskGraph.run()(state.executionContext), Duration.Inf) match {
+      case _: Task.Success[_] => updateState(state)
```
Duhemm: Why are we ignoring the result of the task and putting stuff in `result`? Why is that better than simply getting the result out of the task?

frontend/src/main/scala/bloop/util/JavaCompat.scala
```diff
@@ -20,4 +20,11 @@ object JavaCompat {
   implicit class EnrichOption[T](option: Option[T]) {
     def toOptional: Optional[T] = InterfaceUtil.toOptional(option)
   }
+
+  implicit class EnrichSAM[T, R](f: T => R) {
+    import java.util.function.Function
+    def toJava: Function[T, R] = new Function[T, R] {
+      def apply(t: T): R = f(t)
+    }
+  }
```
Duhemm: Wow why is that required?

frontend/src/main/scala/bloop/engine/Dag.scala
jvican: Yup, I meant that!

frontend/src/main/scala/bloop/engine/tasks/CompileTasks.scala
jvican: Could be found, this is just redundant. I'll change this to a map anyway to address your other comment.

frontend/src/main/scala/bloop/engine/tasks/CompileTasks.scala
jvican: You're right the other one is better, will change

jvican:
Ci is failing because of the changes in gitignore... I'm not gonna fight it, and I'm opening a ticket so that someone else fixes it.

frontend/src/main/scala/bloop/Cli.scala
jvican: No, it returns an option of state, why?

Duhemm: I meant that the name implies that it returns an option of Build. Shouldn’t the method be renamed?

frontend/src/main/scala/bloop/util/JavaCompat.scala
jvican: That was required because there are some limitations in which you cannot convert directly a function type to a SAM, for example when you get a function as a parameter and you place it somewhere where the JVM expects a SAM (`computeIfAbsent`). However, there was only one use site of this, so I've done an in-place function type, and *that one* will indeed be samified. So I'm removing this piece of code.

PR #120, opened 2017-12-07
Add support for `console`

Duhemm:

PR #124, opened 2017-12-08
More integration tests and benchmarks

Duhemm:
With that we're all set to be able to compare sbt, bloop and scalac on the scala library and "mini-better-files" (that's the version that's in scala/compiler-benchmark)

That PR includes the changes to JLine that fix benchmarks. The same changes are part of #120.

I'm opening the PR to make sure everything still works after the rebase.


frontend/src/test/resources/projects/mini-better-files/mini-better-files/src/main/scala/Implicits.scala
```diff
@@ -0,0 +1,263 @@
+package better.files
```
jvican: Can we put this in an independent repo under the Scala Center domain instead of checking in the files?

project/plugins.sbt
```diff
@@ -3,3 +3,4 @@ addSbtPlugin("ch.epfl.scala" % "sbt-release-early" % "2.0.0")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.7.0")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.27")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.6")
+addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.7.0")
```
jvican: This is repeated in the file ^^

frontend/src/test/resources/projects/mini-better-files/mini-better-files/src/main/scala/Implicits.scala
jvican: @Duhemm Can you fix this? :smile:. I can try to merge this when this is done!

Duhemm:
test performance please

bloopoid:

bloopoid:

Benchmarks is based on merging with master ()

Duhemm:
Yay! Let's run it for real now:

test performance please

Duhemm:
test performance please

bloopoid:

bloopoid:

Duhemm:
test performance please

bloopoid:

Duhemm:
test performance please

bloopoid:

bloopoid:
performance test failed:



Error line number: 114

[check  for more information]

Duhemm:
test performance please

bloopoid:

bloopoid:

jvican:
CI is failing here :smile:.

jvican:
See https://github.com/scalacenter/bloop/pull/93.

Duhemm:
__FINALLY__

Duhemm:
test performance please

bloopoid:

benchmarks/src/main/scala/bloop/engine/tasks/TaskBenchmark.scala
```diff
@@ -0,0 +1,64 @@
+package bloop.engine.tasks
+
+import org.openjdk.jmh.annotations.Benchmark
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+import bloop.engine.ExecutionContext.threadPool
+
+object TaskBenchmark {
```
jvican: Are you gonna try with Monix?

benchmarks/src/main/scala/bloop/logging/BloopLoggerBenchmark.scala
```diff
@@ -0,0 +1,40 @@
...
+import java.io.OutputStream
+
+import org.openjdk.jmh.annotations.Benchmark
+
+object BloopLoggerBenchmark {
+  val logger = BloopLogger("benchmark")
+  private val devnull: OutputStream = _ => ()
+  BloopLogger.update(logger, devnull)
+}
+
```
jvican: Is this benchmark valuable at all? It's only testing the cost of registering one log.

benchmarks/src/main/scala/bloop/logging/NoOpLogger.scala
```diff
@@ -0,0 +1,13 @@
+package bloop.logging
+
+object NoOpLogger extends AbstractLogger {
```
jvican: I think `Noop` is an standard name :smile:

benchmarks/src/main/scala/bloop/logging/NoOpLogger.scala
```diff
@@ -0,0 +1,13 @@
...
+  override def name: String = "NoOpLogger"
+  override def ansiCodesSupported(): Boolean = true
+
+  override def debug(msg: String): Unit = ()
+  override def error(msg: String): Unit = ()
+  override def warn(msg: String): Unit = ()
+  override def trace(exception: Throwable): Unit = ()
+  override def info(msg: String): Unit = ()
+  override def verbose[T](op: => T): T = op
```
jvican: Are we sure we need this logger at all? What's the value of benchmarking logging?

build.sbt
```diff
@@ -57,6 +57,15 @@ addCommandAlias(
   ).mkString(";", ";", "")
 )
 
+TaskKey[Unit]("check-clean") := {
```
jvican: Can we use `checkClean`? This syntax is deprecated.

frontend/src/main/scala/bloop/Bloop.scala
```diff
@@ -7,9 +7,13 @@ import bloop.io.AbsolutePath
 import bloop.io.Timer.timed
 import bloop.logging.BloopLogger
 
+import jline.console.ConsoleReader
+
 import scala.annotation.tailrec
 
 object Bloop {
+  private final val reader = consoleReader()
```
jvican: Is it necessary that this is a val? Couldn't it be a local value in `run`? What's the value of having it at the object level?

frontend/src/main/scala/bloop/Bloop.scala
```diff
@@ -52,4 +56,11 @@ object Bloop {
         run(state)
     }
   }
+
+  private def consoleReader(): ConsoleReader = {
```
jvican: I think it would be better to inline this where `reader` is used, since it's only used once.

project/BuildPlugin.scala
```diff
@@ -78,25 +78,30 @@ object BuildKeys {
...
       List(
-        "-DscalaVersion=" + Keys.scalaVersion.value,
-        "-DscalaRef=" + refOf(Keys.scalaVersion.value),
         "-Dsbt.launcher=" + (sys
           .props("java.class.path")
           .split(java.io.File.pathSeparatorChar)
           .find(_.contains("sbt-launch"))
           .getOrElse("")),
-        "-Dbloop.jar=" + AssemblyKeys.assembly.in(sbt.LocalProject("frontend")).value
+        "-DbloopVersion=" + Keys.version.in(sbt.LocalProject("frontend")).value,
```
jvican: Can we hoist up the local project to another place, reusing the same project reference? Also, can we have the id `"frontend"` be reused both by the `BuildPlugin` and `build.sbt`?

project/BuildPlugin.scala
```diff
@@ -222,27 +227,66 @@ object BuildImplementation {
...
          |}
          |TaskKey[Unit]("checkInstall") := {
-         |  Thread.sleep(1000) // Let's wait a little bit because of OS's IO latency
-         |  val mostRecentStamp = (bloopConfigDir.value ** "*.config").get.map(_.lastModified).min
-         |  (System.currentTimeMillis() - mostRecentStamp)
-         |  val diff = (System.currentTimeMillis() - mostRecentStamp) / 1000
-         |  if (diff <= 15) () // If it happened in the last 15 seconds, this is ok!
-         |  else sys.error("The sbt plugin didn't write any file")
+         |  val configFiles = (bloopConfigDir.value ** "*.config").get
+         |  if (configFiles.isEmpty) {
```
jvican: This is not doing the same as the previous piece of code. I understand that forcing resources triggers compilation and makes this fail, but you're only checking if `*.config` files exist. I wonder if there's a better solution? As in, having some kind of synchronization that allows us to assert that the config files were generated in the same sbt run?

project/BuildPlugin.scala
```diff
@@ -222,27 +227,66 @@ object BuildImplementation {
...
+         |    sys.error("The sbt plugin didn't write any file")
+         |  }
          |}
     """.stripMargin
     }
 
+    private val testPluginContents = {
```
jvican: Is there any reason why we have this here and we don't have it in an independent file in `project`? I think that solution would be more elegant because you would make sure that such code always compiles. You can then copy the file from here to whatever place you need. I'm saying this because it's quite huge, and difficult to parse if it's in a string.

project/BuildPlugin.scala
```diff
@@ -222,27 +227,66 @@ object BuildImplementation {
...
         |> registerDirectory
         |> installBloop
+        |> copySources
         |> checkInstall
       """.stripMargin
     }
 
     private val NewLine = System.getProperty("line.separator")
     def scriptedSettings(testDirectory: sbt.SettingKey[File]): Seq[Def.Setting[_]] = List(
-      ScriptedKeys.scriptedBufferLog := false,
```
jvican: Are we sure we want this?

jvican:
#120 is in. Next up is this one. Let me know when you address feedback and rebase the PR :+1:.

bloopoid:
performance test failed:



Error line number: 114

[check  for more information]

benchmarks/src/main/scala/bloop/engine/tasks/TaskBenchmark.scala
Duhemm: Maybe some time later, but I thought it didn't cost much to add a benchmark for the current tasks.

benchmarks/src/main/scala/bloop/logging/BloopLoggerBenchmark.scala
Duhemm: Yeah, but given how complicated our setup with logging is (adding and removing the appenders etc.), I thought it was valuable to have it. There are other configuration options in Log4J that can have an impact (for instance, the layout of the log messages). I'd prefer to know what happens when I change them.

Do you want to get rid of that benchmark?

benchmarks/src/main/scala/bloop/logging/NoOpLogger.scala
Duhemm: Good point. I hope APFS has case sensitive file names...

benchmarks/src/main/scala/bloop/logging/NoOpLogger.scala
Duhemm: It's more about *not* benchmarking logging when we want to benchmark other things.

build.sbt
Duhemm: We agreed to remove that task altogether.

frontend/src/main/scala/bloop/Bloop.scala
Duhemm: Run is recursively called, and I don't want to create a new instance after each command. Do you prefer that I refactor `run`?

project/BuildPlugin.scala
Duhemm: We could have a task that remembers when the scripted test started and check that there are config files newer than that. I'm not sure that's worth the trouble, since we're completely in control of what files are already there, and it's rather unlikely that a project already has a folder named `bloop-config` with `something.config` in it, isn't it?

project/BuildPlugin.scala
Duhemm: It's just that it's much easier to copy to the scripted projects. Having it in `project` means that it would be applied to the projects in our build definition, which I'd rather avoid. Where do you suggest I put it?

project/BuildPlugin.scala
Duhemm: Drone was killing the build after we hit a limit in the size of the log.

Duhemm:
I've made it so that `spark` is compiled up to the `examples` project instead of everything. Compiling the test classes in `sql/core` still fails in CI, for an unknown reason (works on my machine and VMs)...

Duhemm:
test performance please

bloopoid:

bloopoid:

Duhemm:
test performance please

bloopoid:

bloopoid:

Duhemm:
test performance please

bloopoid:

bloopoid:

Duhemm:
test performance please

bloopoid:

project/BuildPlugin.scala
```diff
@@ -243,7 +251,7 @@ object BuildImplementation {
 
     private val NewLine = System.lineSeparator
     def scriptedSettings(testDirectory: sbt.SettingKey[File]): Seq[Def.Setting[_]] = List(
-      ScriptedKeys.scriptedBufferLog := false,
+      ScriptedKeys.scriptedBufferLog := true,
```
jvican: Can we disable it or is it required?

project/BuildPlugin.scala
```diff
@@ -79,25 +82,30 @@ object BuildKeys {
     Keys.test in AssemblyKeys.assembly := {}
   )
 
+  import sbtbuildinfo.BuildInfoKeys
   val benchmarksSettings: Seq[Def.Setting[_]] = List(
     Keys.skip in Keys.publish := true,
+    BuildInfoKeys.buildInfoKeys := Seq[BuildInfoKey](
```
jvican: If you need a reference to the project from here and `javaOptions`, I would add a method in `BuildDefaults` that takes a project as a parameter. Then, I would use that reference to scope whichever settings you need and return a list of settings that you apply from `build.sbt`. I think this is way better than trusting that we use a variable for id and for `LocalProject`, and I think it's also more extensible :smile:

project/BuildPlugin.scala
Duhemm: It's required, otherwise we log too much and the job is killed by Drone.

bloopoid:

Issue #7, opened 2017-11-11
Add support for file watching `~`


jvican:
We should probably add a really fast source file watching for `~compile` and `~test`. We may want to add this directly to Zinc rather than experiment with it directly in blossom. I prefer this approach since it's more generic.

Jasper-M:
Is there something inherently wrong with just using `java.nio.file.WatchService` for this? Not that it has the most beautiful interface...

jvican:
We're using directory-watcher because it claims speedups in OSX and does not do polling. I have a prototype with it I'll PR any time soon.

fsvehla:
> Is there something inherently wrong with just using java.nio.file.WatchService for this? Not that it has the most beautiful interface...

It uses polling under macOS and Linux, which leads to for example SBT using 40% of CPU constantly watching a medium size code base.

jvican:
Does it under LInux? I think under linux it uses fsevents.

fsvehla:
My information re: Linux might be outdated, I am not using it for development.

gmethvin:
The OS differences are basically:
 - Windows JDK `WatchService` does async watching and recursive watching (`ExtendedWatchEventModifier.FILE_TREE`).
 - Linux JDK `WatchService` does async watching but not recursive watching. In practice you can handle the recursive case you can register a new `WatchKey` for directories when they are created.
 - macOS JDK `WatchService` uses polling and at best can be configured to poll every two seconds. To get a decent experience you need to use a separate native program or JNI/JNA (directory-watcher uses JNA).

I think it makes sense to use a higher-level API, perhaps something like `File.Monitor` from better-files. (directory-watcher already implements that: https://github.com/gmethvin/directory-watcher/#better-files-integration-scala). Besides the fact that the `WatchService` API is not very nice, it's also useful to customize the recursive aspect so it uses the most efficient mechanism for each OS.

jvican:
#125 successfully implements this feature with your work, thanks @gmethvin. In the next days we'll experiment with file watching in bloop in larger projects and we'll see how it fares in comparison with the default file watching algorithms in sbt/sbt.

PR #125, opened 2017-12-09
Add file watching to bloop

jvican:
For information on why it's done the way it is, commit messages are rich in details.

This PR adds file watching to bloop and also happens to do a much needed rework of the logger infrastructure to make it more modular, solid and flexible to our needs.

It is still a WIP since I'd like to add one or two more file watching tests, but the one it is right now proves that it works (it's kind of cool we can programmatically test this without emulating a file system of our own).


jvican:
Fixes #7, by the way.

frontend/src/test/scala/bloop/engine/InterpreterSpec.scala
```diff
@@ -2,26 +2,49 @@ package bloop.engine
...
+  def changeOut(state: State): (CliOptions, ByteArrayOutputStream) = {
+    val inMemory = new ByteArrayOutputStream()
+    val newOut = new PrintStream(inMemory)
     val defaultCli = CliOptions.default
-    val memoryStream = new ByteArrayOutputStream()
-    val newOut = new PrintStream(memoryStream)
-    val cliOptions = defaultCli.copy(common = defaultCli.common.copy(out = newOut))
-    val action = Run(Commands.Projects(dotGraph = true, cliOptions), Exit(ExitStatus.Ok))
+    defaultCli.copy(common = state.commonOptions.copy(out = newOut)) -> inMemory
+  }
```
Duhemm: 👍

backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -0,0 +1,99 @@
...
+      updateCache += logger -> out
+      ()
+    }
+
+    updateCache.get(logger) match {
+      case Some(previousOut) if previousOut == out =>
+        logger.debug(s"Update of out ($out) in logger ${logger.name} is cached.")
+      case _ => swapAndCache
+    }
+  }
```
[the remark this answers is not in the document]
jvican: > That's just a suggestion, but I'd have called that currentOutputStreams or loggerToStream maybe.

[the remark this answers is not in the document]
jvican: Good suggestion!

frontend/src/main/scala/bloop/engine/State.scala
```diff
@@ -1,47 +1,78 @@
...
+  private var singleCompilerCache: CompilerCache = null
+  private def getCompilerCache(logger: Logger): CompilerCache = synchronized {
+    if (singleCompilerCache != null) singleCompilerCache
+    else {
+      import sbt.internal.inc.bloop.ZincInternals
+      val provider = ZincInternals.getComponentProvider(Paths.getCacheDirectory("components"))
+      val jars = Paths.getCacheDirectory("scala-jars")
+      singleCompilerCache = new CompilerCache(provider, jars, logger)
+      singleCompilerCache
+    }
```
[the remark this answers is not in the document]
jvican: Yes, this is a tricky part, the compiler cache is not owned by a build in particular, but we still need a logger. As I wanted the first build to trigger the compilation to register the logs of the compiler cache, I did it this way. I agree it's not ideal, but the other solution is to create a global bloop logger that we pass around.
  
However, this will complicate things for not much in return and has some design issues. For example, for every compile command that the server receive we need to swap the output streams. This can cause race conditions and misbehaviours.

frontend/src/main/scala/bloop/io/SourceWatcher.scala
```diff
@@ -0,0 +1,53 @@
...
+
+import bloop.cli.ExitStatus
+import bloop.engine.State
+import bloop.logging.Logger
+
+import scala.collection.JavaConverters._
+import io.methvin.watcher.DirectoryChangeEvent.EventType
+import io.methvin.watcher.{DirectoryChangeEvent, DirectoryWatcher}
+
+final class SourceWatcher(paths0: Seq[Path], logger: Logger) {
```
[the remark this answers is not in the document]
jvican: Yes, I assume so. I mimicked the name that `DirectoryWatcher` uses, but it looks like it only accepts directories.

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -1,41 +1,54 @@
...
+    def updateState(state: State, commonOptions: CommonOptions): State = {
+      State.updateLogger(state.logger, commonOptions)
+      val logger = BloopLogger(state.logger.name)
+      state.copy(commonOptions = commonOptions, logger = logger)
+    }
+
     action match {
-      case Exit(exitStatus) if state.status.isOk => state.copy(status = exitStatus)
-      case Exit(exitStatus) => state
+      case Exit(exitStatus) if state0.status.isOk => state0.copy(status = exitStatus)
```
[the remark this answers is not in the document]
jvican: We need a new logger because log4j forces us to call the logger again to get the changes we just made, that involves calling `LogManager.getLogger` (which is done in the constructor of `BloopLogger`).

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -1,41 +1,54 @@
...
-  def execute(action: Action, state: State): State = {
-    import state.logger
+  def execute(action: Action, state0: State): State = {
+    import state0.logger
     def logAndTime[T](cliOptions: CliOptions, action: => T): T =
-      logger.verboseIf(cliOptions.verbose) { timed(state.logger) { action } }
+      logger.verboseIf(cliOptions.verbose) { timed(state0.logger) { action } }
+    def updateState(state: State, commonOptions: CommonOptions): State = {
+      State.updateLogger(state.logger, commonOptions)
+      val logger = BloopLogger(state.logger.name)
```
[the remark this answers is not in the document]
jvican: We need a new logger because log4j forces us to call the logger again to get the changes we just made, that involves calling LogManager.getLogger (which is done in the constructor of BloopLogger).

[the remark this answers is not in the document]
jvican: You should think of the logger as immutable, and that's why it's important to import it always from the current state.

frontend/src/test/scala/bloop/engine/FileWatchingSpec.scala
```diff
@@ -0,0 +1,148 @@
...
+import bloop.Project
+import bloop.cli.{CliOptions, Commands}
+import bloop.tasks.{CompilationHelpers, ProjectHelpers}
+import bloop.tasks.ProjectHelpers.{RootProject, noPreviousResult, withState}
+import org.junit.Test
+
+class FileWatchingSpec {
+  @scala.annotation.tailrec
+  final def readCompilingLines(target: Int, msg: String, out: ByteArrayOutputStream): Int = {
+    Thread.sleep(100) // Wait 10ms for the OS's file system
```
Duhemm: 100ms?

frontend/src/test/scala/bloop/engine/FileWatchingSpec.scala
```diff
@@ -0,0 +1,148 @@
...
+
+import bloop.Project
+import bloop.cli.{CliOptions, Commands}
+import bloop.tasks.{CompilationHelpers, ProjectHelpers}
+import bloop.tasks.ProjectHelpers.{RootProject, noPreviousResult, withState}
+import org.junit.Test
+
+class FileWatchingSpec {
+  @scala.annotation.tailrec
+  final def readCompilingLines(target: Int, msg: String, out: ByteArrayOutputStream): Int = {
```
Duhemm: I'd rename `readCompilingLines` to `outputContains`, and `target` to `times` or `expected`. WDYT?

jvican: Sounds goooooooood.

frontend/src/test/scala/bloop/engine/FileWatchingSpec.scala
jvican: Yes, to avoid reading all the time until it appears (since it happens concurrently).

Duhemm: Sorry I keep being unclear. The code says 100ms, the comment says 10.

Duhemm: I'll make an effort 💃

frontend/src/main/scala/bloop/engine/Interpreter.scala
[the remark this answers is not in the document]
jvican: We do it because we need to call `LogManager.getLogger` every time we modify it (because apparentely changes are not propagated by log4j).

frontend/src/test/scala/bloop/engine/FileWatchingSpec.scala
jvican: Hahaha, makes sense, removing it :smile:

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -1,41 +1,53 @@
...
+      state.copy(commonOptions = commonOptions, logger = logger)
+    }
+
     action match {
-      case Exit(exitStatus) if state.status.isOk => state.copy(status = exitStatus)
-      case Exit(exitStatus) => state
+      case Exit(exitStatus) if state0.status.isOk => state0.mergeStatus(exitStatus)
+      case Exit(exitStatus) => state0
```
Duhemm: Couldn't we just have `case Exit(exitStatus) => state0.mergeStatus(exitStatus)` ?

jvican: No, because the implementation of mergeStatus is very naive at this point and has to improve. When there are two errors, it's best to at least report the first one and not a weird mix for two. I plan to improve this in a future PR

frontend/src/test/scala/bloop/engine/FileWatchingSpec.scala
```diff
@@ -0,0 +1,147 @@
...
+    assert(projects.forall(p => noPreviousResult(p, state)))
+
+    // The worker thread runs the watched compilation
+    val bloopOut = new ByteArrayOutputStream()
+    val ctx: FileWatchingContext = (state, rootProject, bloopOut)
+    val workerThread = new Thread { override def run(): Unit = workerAction(ctx) }
+    implicit val context = state.executionContext
+    val testFuture = scala.concurrent.Future { testAction(ctx, workerThread) }
+    try Await.ready(testFuture, duration.Duration(15, duration.SECONDS))
+    catch { case t: Throwable => println(s"LOGS WERE ${bloopOut.toString("UTF-8")}"); throw t }
```
Duhemm: Is the `println` a leftover from debugging?

backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  private val currentStreams: scala.collection.mutable.Map[BloopLogger, OutputStream] = {
+    import com.google.common.collect.MapMaker
+    import scala.collection.JavaConverters._
+    new MapMaker().weakKeys().makeMap[BloopLogger, OutputStream]().asScala
+  }
+
+  def update(logger: BloopLogger, out: OutputStream): Unit = {
+    def swapAndCache: Unit = {
+      logger.debug(s"Logger ${logger.name} has not been updated to use $out.")
```
Duhemm: The message reads a bit like a failure, rather than an announcement of what bloop is about to do. Maybe say `logger has not been updated yet` or `updating logger to use`?

frontend/src/test/scala/bloop/engine/FileWatchingSpec.scala
jvican: No, it's something that should help us debug in case these tests fail :smile:.

jvican:
Done, check the last commit :smile:.

PR #126, opened 2017-12-10
Return `EmptyResult` by default in `ResultsCache`

Duhemm:
This is more in-line with the result of doing `reset`, plus it fixes an
issue with compilation tasks complaining about the results cache being
un-initialized.


jvican:
Great work with this, I'm cherry-picking it in #125 :wink:. Closing!

PR #127, opened 2017-12-11
Fix #75: Add configure command to set up global state

### LINKED ISSUE #75: Let users configure the size of the thread pool
#72 introduced a thread pool to execute the task. By default, it's size is the `Runtime.getRuntime.availableProcessors()`. It should be configurable.

jvican:
The `Configure` command is supposed to allow you to set up the global state
of the bloop server, variables like, say, the number of threads.

In the future, I expect this command to grow bigger and support other
global parameters that are similar to the one supported now.

### DIFF
--- a/frontend/src/main/scala/bloop/cli/Commands.scala
+++ b/frontend/src/main/scala/bloop/cli/Commands.scala
@@ -56,7 +56,7 @@ object Commands {
   case class Configure(
       @ExtraName("parallelism")
       @HelpMessage("Set the number of threads used for parallel compilation and test execution.")
-      threads: Int = ExecutionContext.nCPUs,
+      threads: Int = ExecutionContext.executor.getCorePoolSize,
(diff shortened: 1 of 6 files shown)


jvican:
Depends on #125. Will be rebased after that's in.

frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -52,6 +53,13 @@ object Commands {
       @Recurse cliOptions: CliOptions = CliOptions.default
   ) extends Command
 
+  case class Configure(
+      @ExtraName("parallelism")
+      @HelpMessage("Set the number of threads used for parallel compilation and test execution.")
+      threads: Int = ExecutionContext.nCPUs,
```
Duhemm: Shouldn't the default be the current number of threads?

Duhemm:
WDYT @jvican ?

Issue #92, opened 2017-11-30
Nailgun prints out cryptic error messages


olafurpg:
While trying out bloop I get some errors from the nailgun server
```
$. ./bloop-server
Unable to load nailgun-version.properties.
NGServer [UNKNOWN] started on address localhost/127.0.0.1 port 2113.
```

Compared to when I start scalafmt with nailgun
```
$ ./scalafmt_ng
NGServer 0.9.1 started on all interfaces, port 2113.
```

Every bloop command ends with a nailgun exception
```
$ ng bloop.Cli help
bloop f24267cc
Usage: bloop [options] [command] [command-options]


Available commands: about, clean, compile, help, projects, test
Type `bloop 'command' --help` for help on an individual command

Nov 30, 2017 10:59:51 AM com.martiansoftware.nailgun.NGInputStream$1 run
WARNING: Nailgun client read future was interrupted
java.lang.InterruptedException
	at java.util.concurrent.FutureTask.awaitDone(FutureTask.java:404)
	at java.util.concurrent.FutureTask.get(FutureTask.java:204)
	at com.martiansoftware.nailgun.NGInputStream$1.run(NGInputStream.java:91)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

Nov 30, 2017 10:59:51 AM com.martiansoftware.nailgun.NGSession run
INFO: NGSession shutting down
```

I have not seen these errors when running scalafmt with nailgun. However, I have no idea what's causing those errors. I tried

- using the same nailgun version as scalafmt
- redirect out/in/err in case the logging infrastructure was messing it up
```
    Console.withErr(ngContext.err) {
      Console.withIn(ngContext.in) {
        Console.withOut(ngContext.out) {
          val exitStatus = run(cmd, logger)
          ngContext.exit(exitStatus.code)
        }
      }
    }
```

strobe:
I got same error (osx)
```
% bloop projects
Loading 4 projects from '/Users/evgeniy/<project_path>/.bloop-config'...
Projects loaded from '/Users/evgeniy/<project_path>/.bloop-config':
 * macros
[...]

PR #128, opened 2017-12-11
Update to latest nailgun

jvican:
Latest nailgun has a fix wrt reading that may fix scalacenter/bloop#92.
Let's try to see if it does.


jvican:
@Duhemm Merging this to make sure runs of previous PRs don't fail.

Issue #78, opened 2017-11-29
Find out why the jars of scala instance are required for sbt to compile


jvican:
Our sbt integration test fails to compile when we don't add the scala instance jars at the end of the classpath for some misterious reason. I need to figure out why, because these jars should not be there in the first place (they should come from the classpath we get from the `.bloop-config`). See https://github.com/scalacenter/bloop/pull/66/commits/a227c4e9ab0e303ebe9368620d331ec3ea9d6a8a.

PR #130, opened 2017-12-11
Remove scalajars from the classpath

jvican:
I think that #78 was a false positive and the real reason for the CI
failure wasn't this change, so I'm reverting it.


Duhemm:
See also #110

jvican:
Your PR does a better job at this, I'll leave it to you then :smile:.

PR #135, opened 2017-12-15
Add `run` task with support for forking

Duhemm:


Duhemm:
I'm not really happy with the CLI though:
```
bloop ❯ ~/.bloop/bloop-ng.py bloop.Cli -- run -p foo --args wow --args not --args cool
Args: wow, not, cool
Hello, world!
```

I'm not familiar with CaseApp, can we do better?

jvican:
Have you tried separating the arguments with comma?

Duhemm:
@jvican Yes, I doesn't work unfortunately...

```
bloop ❯ ~/.bloop/bloop-ng.py bloop.Cli -- run -p foo --args wow --args not --args cool
Args: 'wow', 'not', 'cool'
Hello, world!
bloop ❯ ~/.bloop/bloop-ng.py bloop.Cli -- run -p foo --args wow, not, cool
Args: 'wow,'
Hello, world!
bloop ❯ ~/.bloop/bloop-ng.py bloop.Cli -- run -p foo --args "wow, not, cool"
Args: 'wow, not, cool'
Hello, world!
bloop ❯ ~/.bloop/bloop-ng.py bloop.Cli -- run -p foo --args wow,not,cool  
Args: 'wow,not,cool'
Hello, world!
```

olafurpg:
I think you can do `--args a b c`. You can also use the "remaining args" feature which gives you the trailing list of arguments which are being discarded here https://github.com/scalacenter/bloop/blob/f828a6a50d01fe1dc4e69b49693d647559903ee3/frontend/src/main/scala/bloop/Cli.scala#L68

frontend/src/test/scala/bloop/logging/RecordingLogger.scala
```diff
@@ -0,0 +1,26 @@
+package bloop.logging
+
+import scala.collection.mutable.Buffer
+
+class RecordingLogger extends AbstractLogger {
```
jvican: Is it necessary to have a new logger for testing? I think there's a more elegant way to do this, and that is redirecting stdout in CLI.

frontend/src/test/scala/bloop/tasks/RunTasksSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  private val projectName = ProjectHelpers.RootProject
+  private val packageName = "foo.bar"
+  private val mainClassName0 = "Main0"
+  private val mainClassName1 = "Main1"
+  private val mainClassName2 = "Main2"
+  private val noDependencies = Map.empty[String, Set[String]]
+  private val args = List("arg0", "arg1", "multiple words")
+  private val EOL = sys.props("line.separator")
```
jvican: `System.lineSeparator()`

frontend/src/main/scala/bloop/engine/tasks/RunTasks.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import bloop.Project
+import bloop.cli.ExitStatus
+import bloop.engine.State
+import bloop.exec.ProcessConfig
+
+import sbt.internal.inc.Analysis
+import xsbt.api.Discovery
+
+object RunTasks {
```
jvican: Do you think it would be better to move this to `CompileTasks` and rename that file to `Tasks`? That file is already misnamed, so I think this is the way to go -- centralizing all the implementations of these core tasks is better IMO.

frontend/src/main/scala/bloop/engine/tasks/RunTasks.scala
```diff
@@ -0,0 +1,38 @@
...
+  def run(state: State, project: Project, className: String, args: Array[String]): State = {
+    val classpath = project.classesDir +: project.classpath
```
jvican: We're doing this a lot in lots of places. Maybe it would be a good idea to create a method that creates the classpath out of the project and update all the call sites.

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -157,6 +161,38 @@ object Interpreter {
     else reportMissing(missing, state)
   }
 
+  private def run(cmd: Commands.Run, state: State): State = {
```
jvican: What about running `run` while file watching?

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -157,6 +161,38 @@ object Interpreter {
...
+    val reporter = ReporterConfig.getDefault(cmd.scalacstyle)
+    def getMainClass(state: State, project: Project): Option[String] = {
+      RunTasks.findMainClasses(state, project) match {
+        case Array() =>
+          state.logger.error(s"No main classes found in project '${project.name}'")
+          None
+        case Array(main) =>
+          Some(main)
+        case mainClasses =>
+          state.logger.error(s"Several main classes were found, specify which one:")
```
jvican: Instead of doing this, can we prettify it and put it all in one string? I'd like to have colors here too.

backend/src/main/scala/bloop/logging/ProcessLogger.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import scala.sys.process
+
+/**
+ *  Wrap a `Logger` to conform to `scala.sys.process.ProcessLogger`. All
+ *  messages are forwarded to the `underlying` logger.
+ *
+ *  @param underlying The logger to forward the messages to.
+ */
+class ProcessLogger(underlying: Logger) extends process.ProcessLogger {
```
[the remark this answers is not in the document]
Duhemm: Removed `scala.sys.process` and moved to `java.lang.ProcessBuilder`

frontend/src/main/scala/bloop/engine/Interpreter.scala
Duhemm: I think that it's easier to read if each main class is on its own line, but I can change it if you really think that it's better on the same line.

I agree about making the output of bloop more colorful. We should probably come up with some "color scheme" and re-use the same colours to mean the same things everywhere. It's probably easier to get something consistent if we take care of improving all outputs in one PR. I'll open an issue.

frontend/src/main/scala/bloop/engine/tasks/RunTasks.scala
Duhemm: Sounds good 👍

frontend/src/main/scala/bloop/engine/Interpreter.scala
jvican: > I think that it's easier to read if each main class is on its own line, but I can change it if you really think that it's better on the same line.
  
I think you mistunderstood me, what I meant is that it's better if we have it all in one string (even if it has newlines to show every entry). As in,putting everything in a triple quote string that afterwards makes it easier to change, rather than printing this and then doing foreach.

Duhemm: Oh, sure 😄

frontend/src/main/scala/bloop/Project.scala
```diff
@@ -36,8 +39,15 @@ case class Project(name: String,
     properties.setProperty("allScalaJars",
                            scalaInstance.allJars.map(_.getAbsolutePath).mkString(","))
     properties.setProperty("tmp", tmp.syntax)
+    properties.setProperty("fork", javaEnv.fork.toString)
+    properties.setProperty("javaHome", javaEnv.javaHome.syntax)
+    properties.setProperty("javaOptions", javaEnv.javaOptions.mkString(";"))
     properties
   }
+
+  lazy val fullClasspath: Array[AbsolutePath] = {
```
jvican: Why is this lazy?

jvican: I would do it the other way around -- I would call this classpath, and then I would call the other version without the classes dir `rawClasspath`. I'm saying this because most of the times we want the `fullClasspath`, not the classpath, and I can see that biting us in the future.

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -45,7 +45,7 @@ object Compiler {
         .distinct
       val classesDir = inputs.classesDir.toFile
       // TODO(jvican): Figure out why `inputs.scalaInstance.allJars` is required here.
-      val classpath = Array(classesDir) ++ inputs.classpath.map(_.toFile) ++ inputs.scalaInstance.allJars
+      val classpath = inputs.classpath.map(_.toFile) ++ inputs.scalaInstance.allJars
```
jvican: Why are you doing this? Looks wrong. Shouldn't it be `inputs.fullClasspath`?

frontend/src/main/scala/bloop/exec/ProcessConfig.scala
```diff
@@ -0,0 +1,116 @@
...
+case class InProcess(classpath: Array[AbsolutePath]) extends ProcessConfig {
+
+  /**
+   * Executes `op` while `System.in` and `System.err` are redirected to `logger`.
+   *
+   * @param logger The whole that will receive messages written to `System.in` and `System.err`.
+   * @param op The operation to perform while those streams are redirected.
+   * @return The result of executing `op`.
+   */
+  private def replaceStandardStreams[T](logger: Logger)(op: => T): T = InProcess.synchronized {
```
jvican: This is too dangerous... I think it's not the way to go. We should have a dynamic thread-sensitive variable that gets us the best stdout for a given thread.

frontend/src/test/scala/bloop/exec/ProcessConfigSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+                       |object $mainClassName {
+                       |  def main(args: Array[String]): Unit = {
+                       |    if (args.contains("crash")) throw new Exception
+                       |    println(s"Arguments: $${args.mkString(", ")}")
+                       |    System.err.println("testing stderr")
+                       |  }
+                       |}""".stripMargin
+  }
+
+  val packageName = "foo.bar"
```
jvican: If this is being used by the previous object, why not have it defined before `ArtificialSources`? If `ArtificialSources` weren't an object, it will be a null pointer exception or a wrong forward reference. It's better to protect ourselves from doing the wrong thing in the future should that change.

frontend/src/test/scala/bloop/logging/RecordingLogger.scala
jvican: Can you explain why `BufferedLogger` doesn't cut it? I don't think we need a new logger... we have plenty of them 😄

frontend/src/test/scala/bloop/tasks/RunTasksSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+                               |class BaseRunnable {
+                               |  def main(args: Array[String]): Unit = {
+                               |    println(s"$mainClassName2: $${args.mkString(", ")}")
+                               |  }
+                               |}
+                               |object $mainClassName2 extends BaseRunnable""".stripMargin
+  }
+
+  private val projectName = ProjectHelpers.RootProject
+  private val packageName = "foo.bar"
```
jvican: Same here.

frontend/src/test/scala/bloop/tasks/RunTasksSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+import org.junit.Test
+import org.junit.Assert.assertEquals
+
+import bloop.cli.Commands
+import bloop.engine.{Interpreter, Run}
+import bloop.engine.tasks.RunTasks
+import bloop.logging.{BloopLogger, ProcessLogger, RecordingLogger}
+import bloop.tasks.ProjectHelpers.{checkAfterCleanCompilation, getProject}
+
+class RunTasksSpec {
```
jvican: Are you not testing forked execution here?

sbt-bloop/src/main/scala/bloop/SbtBloop.scala
```diff
@@ -170,5 +180,27 @@ object PluginImplementation {
         internalClasspath ++ externalClasspath
       }
     }
+
+    private type JavaConfiguration = (Boolean, File, Seq[String])
+
+    /**
+     * Extract the information that we need to configure forking for run or test.
+     */
+    lazy val javaConfiguration: Def.Initialize[Task[JavaConfiguration]] = Def.taskDyn {
```
jvican: Great to see this in an independent task 👍

backend/src/main/scala/bloop/Compiler.scala
Duhemm: No, the `CompileInputs` are constructed with the `fullClasspath`.

frontend/src/main/scala/bloop/Project.scala
Duhemm: Sounds good

frontend/src/test/scala/bloop/logging/RecordingLogger.scala
Duhemm: I use it to test `ProcessConfig`, which doesn't touch the CLI. There's something wrong with logging at the moment, and we can see it with the output being repeated several times in the tests.

When possible, I prefer to explicitly give a logger and inspect its content rather than setting an output stream somewhere and expecting to find some messages in it. When passing a `State` around, it should be enough to just change the underlying logger in the state.

frontend/src/main/scala/bloop/exec/ProcessConfig.scala
Duhemm: Yes, I haven't had time to address that yet.

frontend/src/test/scala/bloop/logging/RecordingLogger.scala
Duhemm: `BufferedLogger` doesn't let us inspect the messages that it contains at the moment, but I'm happy to refactor it if you prefer.

frontend/src/test/scala/bloop/tasks/RunTasksSpec.scala
Duhemm: Duh, good catch.

frontend/src/test/scala/bloop/logging/RecordingLogger.scala
jvican: I agree that's a better approach, but it only works in some of the cases: not all bloop uses loggers to print stuff. If you fix the other stdout issue, then I agree this can stay here :smile:.
  
On the repeated messages in the loggers: I'll look into it.

jvican: Oh, actually I haven't realized that it's better to set up the stdout, since the logger is redirected to the stdout anyway (don't you think this is better than breaking this assumption that things from the logger go to X and things to stdout go to Y?). I think that's simpler because everything in bloop relies on stdout being a nice place where you can redirect it and is accessible to the user. What do you think?

Duhemm: I disagree. I don't think that a logger implementation should be required to eventually write stuff to stdout. Logging to a file, to the network, discarding all messages, etc. are all valid logger implementations in my eyes. Saying "Hey, this is stdout from now on" and expecting to see all output go there sounds much more fragile to me than just passing around the logger that you want to collect the output. It makes the path from message production to the logger much more obvious. It's probably also more friendly to parallel testing when we re-enable that (I'm not sure how much of L4J's global state we're mutating now).

In the future, I imagine that we should try to have one CLI invocation = one logger instance, set the right output stream for that logger, and put everything in the logger. I don't think that we should care about the actual output stream in use, we should just write to the logger and trust it to do the right thing.

Duhemm:
@jvican I think that this one is ready for another round of review.

frontend/src/main/scala/bloop/exec/MultiplexedStreams.scala
```diff
@@ -0,0 +1,118 @@
...
+        registerStreams(id, logger)
+        result = {
+          try Some(op)
+          finally removeStreams(id)
+        }
+      }
+    }
+
+    val future = executor.submit(thread)
+    future.get()
```
jvican: Repeated `future.get`.

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
```diff
@@ -134,3 +141,102 @@ object CompileTasks {
...
+    val exitCode = processConfig.runMain(className, args, state.logger)
+    val exitStatus = {
+      if (exitCode == ProcessConfig.EXIT_OK) ExitStatus.Ok
+      else ExitStatus.UnexpectedError
+    }
+
+    state.mergeStatus(exitStatus)
+  }
+
+  def findMainClasses(state: State, project: Project): Array[String] = {
```
jvican: I'm curious: why do we return an array here?

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
```diff
@@ -134,3 +141,102 @@ object CompileTasks {
...
+        sbt.internal.inc.Analysis.empty
+    }
+
+    val mainClasses = analysis.infos.allInfos.values.flatMap(_.getMainClasses)
+    logger.debug(s"Found ${mainClasses.size} main classes${mainClasses.mkString(": ", ", ", ".")}")
+    mainClasses.toArray
+  }
+}
+
+object TestTasks {
```
jvican: I didn't really mean this when I said putting everything in the `Tasks` class. What I meant is to only have one object `Tasks` that has all these methods. Right now, it doesn't make much sense that we have to remember the prefix of the object where we call `compile`, `test` and `run` to be `CompileTasks`, `TestTasks` and `RunTasks`. I'd rather have `Tasks.compile`, `Tasks.test` and `Tasks.run`. I think it makes for a better API.

frontend/src/main/scala/bloop/exec/IdentifiedThread.scala
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  /** The main loop of this `Thread`. */
+  def work(): Unit
+}
+
+object IdentifiedThread {
+
+  /** An `InheritedThreadLocal` that can be used to identify the currently running `Thread`. */
+  val id = new InheritableThreadLocal[UUID]
```
jvican: Have you looked into reusing `scala,util.DynamicVariable`?

frontend/src/main/scala/bloop/exec/ProcessConfig.scala
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Configuration to execute Java code without starting a new JVM.
+ */
+case class InProcess(classpath: Array[AbsolutePath]) extends ProcessConfig {
+
+  override def runMain(className: String, args: Array[String], logger: Logger): Int = {
+    logger.debug(s"Running '$className' in process.")
+    logger.debug(s"  Classpath: ${classpath.map(_.syntax).mkString(pathSeparator)}")
+    val entries = classpath.map(_.toFile.toURI.toURL)
```
jvican: Can we cache new classloaders that we create? Or maybe you can leave this to me, I think we're not caching the test classloader either. I can do it in another PR.

frontend/src/test/scala/bloop/tasks/CompilationHelpers.scala
```diff
@@ -17,7 +17,7 @@ object CompilationHelpers {
 
   private val ScriptedResolverId = "zinc-scripted-local"
   private val ScriptedResolveCacheDir: File =
-    AbsolutePath(s".ivy2/$ScriptedResolverId").toFile
+    AbsolutePath(s"../.ivy2/$ScriptedResolverId").toFile
```
jvican: Can you add a comment explaining that the reason we need this is because `fork in Test` changes the working directory?

frontend/src/test/scala/bloop/tasks/ProjectHelpers.scala
```diff
@@ -23,7 +24,7 @@ object ProjectHelpers {
     }
 
     project.copy(
-      classpath = project.classpath.map(work),
+      rawClasspath = project.rawClasspath.map(work),
```
jvican: Thanks for doing this 👍

sbt-bloop/src/main/scala/bloop/SbtBloop.scala
jvican: Why is this lazy?

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
Duhemm: You mean as opposed to any other data structure? It's because that's the one we use most often in bloop, but it could be anything else.

jvican: I thought we were using `Seq` as the default. I've only seen `Array` being used when the sbt test interface API has required it and for `bloop.Project`.

frontend/src/main/scala/bloop/exec/IdentifiedThread.scala
Duhemm: Haven't tried it, it doesn't look like it would change many things.

frontend/src/main/scala/bloop/exec/ProcessConfig.scala
Duhemm: Caching comes as part of #136 😄

sbt-bloop/src/main/scala/bloop/SbtBloop.scala
Duhemm: Muscle memory...

frontend/src/main/scala/bloop/exec/IdentifiedThread.scala
jvican: This is ok then :smile:. What I'm most worried is if this will indeed work in concurrent scenarios, and we have supposedly a test for it, so I'll trust it!

Duhemm:
I can confirm that the issues that I had with this PR are fixed by #141

Duhemm:
Tests pass after rebasing 🎉

backend/src/main/scala/bloop/logging/ProcessLogger.scala
```diff
@@ -0,0 +1,87 @@
...
+   * Creates an `OutputStream` from a logging function
+   *
+   * @param logFn A function that logs the given string to a logger.
+   * @return An `OutputStream` that can be used as `System.in` or `System.err`.
+   */
+  def toOutputStream(logFn: String => Unit): OutputStream = {
+    val outputStream = new OutputStream {
+      private val buffer = new ByteArrayOutputStream
+
+      override def write(x: Int): Unit = synchronized {
```
jvican: Doesn't this mean that if we don't end something with a new line we could be missing logs? Isn't there a way we can reuse some JDK utility instead of implementing our own syncrhonized output stream? Or Guava?

backend/src/main/scala/bloop/logging/ProcessLogger.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Creates a `PrintStream` from a logging function
+   *
+   * @param logFn A function that logs the given string to a logger.
+   * @return A `PrintStream` that can be used as `System.in` or `System.err`.
+   */
+  def toPrintStream(logFn: String => Unit): PrintStream = {
+    val outputStream = toOutputStream(logFn)
+    new PrintStream(outputStream, /* autoflush = */ false, encoding)
```
jvican: Why have `autoFlush` as a comment and not use it directly as a named parameter?

frontend/src/test/scala/bloop/tasks/ProjectHelpers.scala
```diff
@@ -104,12 +107,18 @@ object ProjectHelpers {
...
+      javaEnv: JavaEnv
   )(op: State => T): T = {
     withTemporaryDirectory { temp =>
       val projects = projectStructures.map {
         case (name, sources) =>
-          makeProject(temp, name, sources, dependencies.getOrElse(name, Set.empty), scalaInstance)
+          makeProject(temp,
+                      name,
+                      sources,
+                      dependencies.getOrElse(name, Set.empty),
```
jvican: Can we hoist up this to a value and reference to it from here?

frontend/src/main/scala/bloop/Bloop.scala
```diff
@@ -53,6 +53,16 @@ object Bloop {
         val action = Run(command, Exit(ExitStatus.Ok))
         run(Interpreter.execute(action, state))
 
+      case Array("runMain", projectName, mainClass, args @ _*) =>
```
jvican: If the shell is becoming a `CaseApp`, isn't this unnecessary? You may need to adapt it in your other PR, no need to block this one.

backend/src/main/scala/bloop/logging/ProcessLogger.scala
Duhemm: The stream will be flushed when there's a newline and when it's closed. That's the best solution I've found to have the logger reproduce the output of the running application.

At first I relied on `autoFlush` to do the right thing, but I'd end up with duplicated newlines when the application did `println()` for instance.

This means that the output of the application may be delayed until it exits or a newline character is seen, but I don't think that we could lose output that way.

backend/src/main/scala/bloop/logging/ProcessLogger.scala
Duhemm: This is a Java method, named parameters don't work with those AFAIK

frontend/src/main/scala/bloop/Bloop.scala
Duhemm: It's a `CaseApp`, but you still use the shell, so it doesn't affect what you type in. Or am I misunderstanding your comment?

backend/src/main/scala/bloop/logging/ProcessLogger.scala
jvican: > This means that the output of the application may be delayed until it exits or a newline character is seen, but I don't think that we could lose output that way.

frontend/src/main/scala/bloop/Bloop.scala
jvican: Solved in a private discussion

jvican:
Great job.

jvican:
Updating and merging.

PR #136, opened 2017-12-16
Add support for forked test execution

Duhemm:
Builds on top of #135

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
@@ -212,7 +212,8 @@ object Tasks {
           val runner = TestInternals.getRunner(framework, testLoader)
           val tasks = runner.tasks(taskDefs.toArray).toList
           TestInternals.executeTasks(tasks, eventHandler, logger)
-          val _ = runner.done() // TODO(jvican): We've never used this value, what is it?
+          val summary = runner.done()
+          if (summary.nonEmpty) logger.info(summary)
       }
     }
 
--- a/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
@@ -5,7 +5,7 @@ import bloop.engine.{Dag, State}
 import bloop.exec.ProcessConfig
 import bloop.io.AbsolutePath
 import bloop.reporter.{Reporter, ReporterConfig}
-import bloop.testing.TestInternals
+import bloop.testing.{DiscoveredTests, TestInternals}
 import bloop.{CompileInputs, Compiler, Project}
 
 import java.net.URLClassLoader
@@ -203,18 +203,16 @@ object Tasks {
         sbt.internal.inc.Analysis.empty
       }
 
-      val discoveredTests = discoverTests(analysis, frameworks)
-      val allTestNames: List[String] =
-        discoveredTests.valuesIterator.flatMap(defs => defs.map(_.fullyQualifiedName())).toList
-      logger.debug(s"Bloop found the following tests for ${projectName}: $allTestNames.")
-      discoveredTests.iterator.foreach {
-        case (framework, taskDefs) =>
-          val runner = TestInternals.getRunner(framework, testLoader)
-          val tasks = runner.tasks(taskDefs.toArray).toList
-          TestInternals.executeTasks(tasks, eventHandler, logger)
-          val summary = runner.done()
-          if (summary.nonEmpty) logger.info(summary)
+      val discoveredTests = {
+        val tests = discoverTests(analysis, frameworks)
+        DiscoveredTests(testLoader, tests)
       }
+      val allTestNames: List[String] =
+        discoveredTests.tests.valuesIterator
+          .flatMap(defs => defs.map(_.fullyQualifiedName()))
+          .toList
+      logger.debug(s"Bloop found the following tests for ${projectName}: $allTestNames.")
+      TestInternals.executeTasks(processConfig, discoveredTests, eventHandler, logger)
     }
 
     // Return the previous state, test execution doesn't modify it.
@@ -270,9 +268,8 @@ object Tasks {
   private[bloop] val eventHandler =
     new EventHandler { override def handle(event: Event): Unit = () }
 
-  private type Discovered = Map[Framework, List[TaskDef]]
   private[bloop] def discoverTests(analysis: CompileAnalysis,
-                                   frameworks: Array[Framework]): Discovered = {
+                                   frameworks: Array[Framework]): Map[Framework, List[TaskDef]] = {
     import scala.collection.mutable
     val (subclassPrints, annotatedPrints) = TestInternals.getFingerprints(frameworks)
     val definitions = TestInternals.potentialTests(analysis)
--- a/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
@@ -193,8 +193,8 @@ object Tasks {
     val projectsToTest = if (aggregate) Dag.dfs(state.build.getDagFor(project)) else List(project)
     projectsToTest.foreach { project =>
       val projectName = project.name
-      val classpath = constructClasspath(project)
-      val testLoader = new URLClassLoader(classpath, TestInternals.filteredLoader)
+      val processConfig = ProcessConfig(project.javaEnv, project.classpath)
+      val testLoader = processConfig.toClassLoader(Some(TestInternals.filteredLoader))
       val frameworks = project.testFrameworks
         .flatMap(fname => TestInternals.getFramework(testLoader, fname.toList, logger))
       logger.debug(s"Found frameworks: ${frameworks.map(_.name).mkString(", ")}")
@@ -288,8 +288,5 @@ object Tasks {
     }
     tasks.mapValues(_.toList).toMap
   }
-
-  private[bloop] def constructClasspath(project: Project): Array[java.net.URL] =
-    project.classpath.map(_.underlying.toUri.toURL())
 
 }
--- a/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
@@ -194,7 +194,7 @@ object Tasks {
     projectsToTest.foreach { project =>
       val projectName = project.name
       val processConfig = ProcessConfig(project.javaEnv, project.classpath)
-      val testLoader = processConfig.toClassLoader(Some(TestInternals.filteredLoader))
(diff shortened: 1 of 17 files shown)


Duhemm:
Rebased and passing the tests 🎉 

I don't know why GitHub shows the commits in the wrong order though...

jvican:
Yeah, I think that's a bug I've encountered numerous times too :smile:.

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
```diff
@@ -271,7 +270,7 @@ object Tasks {
 
   private type Discovered = Map[Framework, List[TaskDef]]
   private[bloop] def discoverTests(analysis: CompileAnalysis,
-                                   frameworks: Array[Framework]): Discovered = {
+                                   frameworks: Array[Framework]): Map[Framework, List[TaskDef]] = {
```
jvican: Should you either use `Discovered` here or remove the above type alias? Personally I like `Discovered`.

frontend/src/main/scala/bloop/exec/ProcessConfig.scala
```diff
@@ -11,22 +13,43 @@ import bloop.logging.{Logger, ProcessLogger}
...
    */
   def classpath: Array[AbsolutePath]
 
+  /**
+   * Creates a `ClassLoader` from the classpath of this `ProcessConfig`.
+   *
+   * @param parent A parent classloader
+   * @return A classloader constructed from the classpath of this `ProcessConfig`.
+   */
+  final def toClassLoader(parent: Option[ClassLoader]): ClassLoader = {
```
jvican: Can we call this `toExecutionClassLoader`?

frontend/src/main/scala/bloop/exec/ProcessConfig.scala
```diff
@@ -74,23 +101,29 @@ case class InProcess(classpath: Array[AbsolutePath]) extends ProcessConfig {
...
-    logger.debug(s"  classpath   = '${classpath.map(_.syntax).mkString(pathSeparator)}'")
+    logger.debug(s"  classpath   = '${fullClasspath.map(_.syntax).mkString(pathSeparator)}'")
     logger.debug(s"  command     = '${cmd.mkString(" ")}'")
 
     val processBuilder = new ProcessBuilder(cmd: _*)
     val process = processBuilder.start()
     val processLogger = new ProcessLogger(logger, process)
     processLogger.start()
     val exitCode = process.waitFor()
+    logger.debug(s"Forked JVM exited with code: $exitCode")
```
jvican: As far as I see, we're not supporting to run tests and main with a different java home, as sbt allows you to, right? This is only an appreciation, but we may want to support this out of the box in a follow up PR since I think it's quite a popular request: lots of people are stuck in JDK6 and JDK7, and they want to try their 2.11 and 2.10 artifacts there.

frontend/src/main/scala/bloop/testing/TestInternals.scala
```diff
@@ -29,14 +37,83 @@ object TestInternals {
...
+                           discoveredTests: DiscoveredTests,
+                           eventHandler: EventHandler,
+                           logger: Logger): Unit = {
+    logger.debug("Starting forked test execution.")
+
+    val testLoader = fork.toClassLoader(Some(filteredLoader))
+    val server = new TestServer(logger, eventHandler, discoveredTests)
+    val forkMain = classOf[sbt.ForkMain].getCanonicalName
+    val arguments = Array(server.port.toString)
+    val testAgentFiles = DependencyResolution.resolve(sbtOrg, testAgentId, testAgentVersion, logger)
```
jvican: Oh wow... is it possible that we don't resolve in such an important part of the pipeline? Can we not first try to locate and reuse the test agent jar from our current classloader by checking the JDK protection domain (because i think we already have it in bloop's classloader), and only if that fails we try to resolve? I think this is too heavy here, even if coursier is being used.

frontend/src/main/scala/bloop/testing/TestServer.scala
```diff
@@ -0,0 +1,150 @@
...
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under BSD-3-Clause license
+ */
+private final class React(is: ObjectInputStream,
+                          os: ObjectOutputStream,
+                          logger: Logger,
+                          eventHandler: EventHandler) {
+  import ForkTags._
+  @annotation.tailrec
+  def react(): Unit = {
```
jvican: Yup, heavily inspired indeed, I remember this piece of code almost by heart.

frontend/src/main/scala/sbt/SerializableFingerprints.scala
```diff
@@ -0,0 +1,18 @@
...
+import java.io.Serializable
+import sbt.testing.{AnnotatedFingerprint, Fingerprint, SubclassFingerprint}
+
+/**
+ * This object is there only to instantiate the subclasses of Fingerprint that are
+ * expected by the remote test runner (they're package private in sbt)
+ */
+object SerializableFingerprints {
+  // Copied from ForkTests.scala in sbt
+  def forkFingerprint(f: Fingerprint): Fingerprint with Serializable =
```
jvican: You're a hijacker.

frontend/src/test/scala/bloop/tasks/TestTaskTest.scala
```diff
@@ -30,45 +31,58 @@ object TestTaskTest extends DynTest {
...
   }
 
-  import java.net.URLClassLoader
-  private val testLoader: ClassLoader = {
-    val classpath = Tasks.constructClasspath(testProject)
-    new URLClassLoader(classpath, TestInternals.filteredLoader)
+  private def processConfig(fork: Boolean): ProcessConfig = {
+    val javaEnv = JavaEnv.default(fork)
+    val classpath = testProject.classpath
+    ProcessConfig(javaEnv, classpath)
```
jvican: Can we have a more descriptive way? Something along the lines of `ProcessRunnerConfig`?

frontend/src/test/scala/bloop/tasks/TestTaskTest.scala
```diff
@@ -30,45 +31,58 @@ object TestTaskTest extends DynTest {
...
+      TestInternals.getFramework(classLoader, n.toList, testState.logger))
+  }
+
+  private def runTestFramework(frameworkName: String, fork: Boolean): Unit = {
+    testState.logger.quietIfSuccess { logger =>
+      val config = processConfig(fork)
+      val classLoader = testLoader(config)
+      val discovered = Tasks.discoverTests(testAnalysis, frameworks(classLoader)).toList
+      val tests = discovered
+        .map {
```
jvican: I would replace the `map` and `filter` with a `flatMap`.

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
Duhemm: I removed it, but lost the change in the rebase. That's the only place where it's used, and it's confusing now that there's `DiscoveredTests` too, so I'll remove it.

frontend/src/main/scala/bloop/exec/ProcessConfig.scala
Duhemm: `sbt-bloop` already extracts the java home from the config, and it's passed to the `JavaEnv`, so it should work. What do you have in mind? Having a set of different java homes from which we can select before running?

frontend/src/main/scala/bloop/testing/TestInternals.scala
Duhemm: I'll benchmark that to see if it's really that expensive.

frontend/src/main/scala/sbt/SerializableFingerprints.scala
Duhemm: 😎

frontend/src/main/scala/bloop/testing/TestInternals.scala
```diff
@@ -29,14 +37,83 @@ object TestInternals {
...
+                           discoveredTests: DiscoveredTests,
+                           eventHandler: EventHandler,
+                           logger: Logger): Unit = {
+    logger.debug("Starting forked test execution.")
+
+    val testLoader = fork.toExecutionClassLoader(Some(filteredLoader))
+    val server = new TestServer(logger, eventHandler, discoveredTests)
+    val forkMain = classOf[sbt.ForkMain].getCanonicalName
+    val arguments = Array(server.port.toString)
+    val testAgentFiles = DependencyResolution.resolve(sbtOrg, testAgentId, testAgentVersion, logger)
```
Duhemm: It looks like GitHub lost your previous comment... Anyway, I've run a benchmark for that, and the results show that the cost of resolving this dependency with coursier is negligible:

```
[info] Benchmark                                   Mode    Cnt   Score   Error  Units
[info] ResolveBenchmark.resolve                  sample  57525   1.737 ± 0.003  ms/op
[info] ResolveBenchmark.resolve:resolve·p0.00    sample          1.616          ms/op
[info] ResolveBenchmark.resolve:resolve·p0.50    sample          1.698          ms/op
[info] ResolveBenchmark.resolve:resolve·p0.90    sample          1.767          ms/op
[info] ResolveBenchmark.resolve:resolve·p0.95    sample          1.825          ms/op
[info] ResolveBenchmark.resolve:resolve·p0.99    sample          3.074          ms/op
[info] ResolveBenchmark.resolve:resolve·p0.999   sample          4.223          ms/op
[info] ResolveBenchmark.resolve:resolve·p0.9999  sample          7.195          ms/op
[info] ResolveBenchmark.resolve:resolve·p1.00    sample         18.350          ms/op
```

Here's the source of the benchmark:
```scala
package bloop

import java.util.concurrent.TimeUnit

import org.openjdk.jmh.annotations._

@BenchmarkMode(Array(Mode.SampleTime))
@OutputTimeUnit(TimeUnit.MILLISECONDS)
@Fork(value = 1)
@Warmup(iterations = 10, time = 10, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 10, time = 10, timeUnit = TimeUnit.SECONDS)
class ResolveBenchmark {

  @Benchmark
  def resolve(): Unit = {
    DependencyResolution.resolve("org.scala-sbt", "test-agent", "1.0.4", logging.NoopLogger)
  }
}
```

frontend/src/main/scala/bloop/exec/ProcessConfig.scala
jvican: If the Java home can be set by the user it’s fine.

frontend/src/main/scala/bloop/testing/TestInternals.scala
jvican: Okay, if it’s under 20ms is fine for me.

Duhemm:
Then I think I've addressed all of your comments 😄

PR #137, opened 2017-12-18
Allow to run release in PR temporarily

jvican:


allanrenucci:
In Dotty we used to store our keys on the server. I changed it in https://github.com/lampepfl/dotty/pull/3577. Now all the credentials are passed as drone secret. I had issues with sbt not finding the keys because they were not in the default location. Now I simply copy the key to the default location and sbt is happy:
```shell
# Write down PGP secret key to the location expected by sbt-pgp
mkdir -p "$HOME/.sbt/gpg"
echo "$PGP_SECRET" > "$HOME/.sbt/gpg/secring.asc"
```

jvican:
@allanrenucci  I don't know why you did that -- overriding `pgpPublicRing` and `pgpSecretRing` should have fixed your issue :smile:.

allanrenucci:
>  overriding pgpPublicRing and pgpSecretRing should have fixed your issue

Yes that's what we used to do and it worked fine. But in our case having the keys as a drone secret was a better solution:
- We have all the credentials in the same place (drone secrets)
- We can have different keys for different repositories
- The repo doesn't need to be a "trusted" repository
- Overall it turned out to be simpler

jvican:
Yup, we also have the credentials as drone secrets.

jvican:
Ah no, I know what you mean. You're talking about the gpg keys. For us it works now, so there's no rush to change. I don't see how i can make it simpler than the current integration we have.

jvican:
Can you merge @Duhemm ? After this we can merge console, and then benchmarks.

.drone.yml
```diff
@@ -2,10 +2,18 @@ matrix:
   SBT_RUN:
     - sbt "+bridgeIntegration/cachedPublishLocal" \
           "benchmarks/compile" \
+          "git status" \
           "setupTests" \
+          "git status" \
```
Duhemm: Can't we have something that fails if the repository is not clean? That much clearer and easier to debug than having to skim through thousands of lines of logs.

.drone.yml
```diff
@@ -27,22 +35,31 @@ pipeline:
       - /drone/.coursier
       - /drone/.sbt
 
+  build:
```
jvican: This is added back, it was removed in latest master.

.drone.yml
```diff
@@ -2,10 +2,18 @@ matrix:
...
+          "git status" \
           "setupTests" \
+          "git status" \
           "backend/test" \
           "frontend/test"
 
+  SBT_PUBLISH:
+    - sbt "set pgpPublicRing in Global := file(\"/drone/.gnupg/pubring.asc\")" \
+          "set pgpSecretRing in Global := file(\"/drone/.gnupg/secring.asc\")" \
+          "show */*:pgpSecretRing" \
```
Duhemm: I guess that's only for debugging?

build.sbt
```diff
@@ -53,6 +43,7 @@ val benchmarkBridge = project
 /*                            This is the build definition of the wrapper                          */
 /***************************************************************************************************/
 import build.Dependencies
+import ch.epfl.scala.sbt.release.ReleaseEarly
```
Duhemm: Isn't that import unused?

jvican:
Okay, just a comment on why we set `pgpPublicRing` and `pgpSecretRing` manually: when using source dependencies, sbt seems to have no care to apply global settings overrides in the correct order. In our case, the global settings of our build plugin are first applied, and then the global overrides of downstream source dependencies are. This means that our source dependencies have higher priority than our main root build. This is a big bug that has to be fixed upstream, some day.

project/BuildPlugin.scala
```diff
@@ -1,15 +1,17 @@
 package build
 
+import ch.epfl.scala.sbt.release.Feedback
```
Duhemm: I think it's unused too

.drone.yml
jvican: I propose we have this here momentarily and then, in another PR, we override the way Drone clones the repository. There's an option in `git submodule` called `--checkout` that should allow us to force the commit hash that we recorded last time, instead of getting the most recent upstream.
  
All of our problems seem to be because we update the branches and our previous PRs haven't registered that commit as the HEAD. Doing that would solve all our problems.

Duhemm: Not all our problems. Setting up the scripted tests sometimes makes the repositories dirty.

jvican: That's true, forgot about it. But how can we have something that fails if the repository is not clean *while running `scripted`*? The only way I see is to hijack scripted, and it doesn't sound exciting...
  
This problem should not be an issue now that we have stable HEADs in all the projects. But I agree it's frustrating when adding new projects or updating them. I don't think there's a good solution though.

Duhemm: I meant having something like this: https://github.com/scalacenter/bloop/pull/124/commits/14c2fa24834c17cad3d8629c4f76f01e1e5048e6

It doesn't matter, we'll change that if it bites us again.

jvican: But `checkClean` runs before `setupTests`, right? In that case, that will never be able to identify which changes happened after `setupTests` fails. So this use case would be fixed in the future PR in which we force `--checkout`, since by definition it will always be clean then :smile:.

Duhemm: Yeah, but it runs after `scriptedAddSbtBloop` too 😉

PR #138, opened 2017-12-20
Dummy PR to test CI

Duhemm:


jvican:
Yeah, that's expected. The reason why it happens is because `baseDirectory` changes if you do `fork in Test`. Therefore, you need to change the URI of the place where the resolver looks for the compiler bridges.

PR #141, opened 2017-12-22
Allow to specify where `BloopLogger` should write

Duhemm:
This commit makes us define the streams that should be used by an
instance of `BloopLogger` when constructing the logger. As a result, we
have a clearer relation between each CLI invocation and the assigned
logger, making it easier to have output be written to the correct
stream.

It also solves issues that we had with repeated output (output would be
written to the console as many times as we had loggers created).

### DIFF
--- a/frontend/src/main/resources/log4j2.properties
+++ b/frontend/src/main/resources/log4j2.properties
@@ -1,11 +1,1 @@
-name = PropertiesConfig
-
-appender.console.type = Console
-appender.console.name = STDOUT
-appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %highlight{%equals{[%-0.-1level] }{[I] }{}}{FATAL=white, ERROR=bright red, WARN=yellow, INFO=dim blue, DEBUG=green, TRACE=blue}%msg%n
-appender.console.filter.threshold.type = ThresholdFilter
-appender.console.filter.threshold.level = info
-
-rootLogger.level = info
-rootLogger.appenderRef.stdout.ref = STDOUT
+rootLogger.level = all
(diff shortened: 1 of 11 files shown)


frontend/src/main/resources/log4j2.properties
```diff
@@ -1,11 +1 @@
...
-appender.console.type = Console
-appender.console.name = STDOUT
-appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %highlight{%equals{[%-0.-1level] }{[I] }{}}{FATAL=white, ERROR=bright red, WARN=yellow, INFO=dim blue, DEBUG=green, TRACE=blue}%msg%n
-appender.console.filter.threshold.type = ThresholdFilter
-appender.console.filter.threshold.level = info
-
-rootLogger.level = info
-rootLogger.appenderRef.stdout.ref = STDOUT
+rootLogger.level = all
```
jvican: Yay

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -9,43 +9,30 @@ import bloop.Project
 import bloop.logging.BloopLogger
 
 object Interpreter {
-  def execute(action: Action, state0: State): State = {
+  def execute(action: Action, state: State): State = {
     def logAndTime[T](state: State, cliOptions: CliOptions, action: => T): T =
       state.logger.verboseIf(cliOptions.verbose) { timed(state.logger) { action } }
-    def updateState(state: State, commonOptions: CommonOptions): State = {
```
jvican: Great to see this go away.

PR #143, opened 2017-12-26
Make `bloop.Bloop` a CaseApp

Duhemm:


frontend/src/main/scala/bloop/Bloop.scala
```diff
@@ -1,25 +1,29 @@
...
 import bloop.io.AbsolutePath
 import bloop.io.Timer.timed
 import bloop.logging.BloopLogger
 import jline.console.ConsoleReader
+import caseapp.{CaseApp, RemainingArgs}
 
 import scala.annotation.tailrec
 
-object Bloop {
+object Bloop extends CaseApp[CliOptions] {
```
jvican: Bloop is supposed to emulate `CaseApp`'s logic without extending it, so this was on purpose. Why the change?

frontend/src/main/scala/bloop/Bloop.scala
```diff
@@ -1,25 +1,29 @@
...
   private val reader = consoleReader()
-  def main(args: Array[String]): Unit = {
-    val baseDirectory = AbsolutePath(args.lift(0).getOrElse(".."))
-    val configDirectory = baseDirectory.resolve(".bloop-config")
+
+  override def run(options: CliOptions, remainingArgs: RemainingArgs): Unit = {
```
jvican: This is bad because `CaseApp` by defualt uses `sys.exit`. I think we really want to have our custom logic here instead of doing this.

PR #144, opened 2017-12-27
Fix logging with Nailgun

Duhemm:
Using Nailgun, we re-use loggers with the same name. This means that
between sessions, the old appenders will continue to exist for that
logger. Finally, Log4J will not add the appenders that we create if
appenders with the same name already exist for a given logger.

As a result, we were not adding the right appenders when re-using an
existing logger. This commit fixes that by removing all the current
appenders when creating a new instance of `BloopLogger`.

### DIFF
--- a/backend/src/main/scala/bloop/logging/BloopLogger.scala
+++ b/backend/src/main/scala/bloop/logging/BloopLogger.scala
@@ -42,12 +42,22 @@ class BloopLogger(override val name: String, out: PrintStream, err: PrintStream)
   }
 
   private[this] def initialize(): Unit = BloopLogger.synchronized {
+    removeAllAppenders()
+
     outAppender.start()
     errAppender.start()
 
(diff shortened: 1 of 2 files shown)


backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -50,6 +52,14 @@ class BloopLogger(override val name: String, out: PrintStream, err: PrintStream)
     coreLogger.addAppender(errAppender)
   }
 
+  private[this] def removeAllAppenders(): Unit = BloopLogger.synchronized {
```
jvican: This is starting to look quite similar to what we had before 😛

Duhemm: Yeah that's because in the initial version of this PR, I'd create a new logger for every session, but I reverted that later...

Issue #140, opened 2017-12-22
Automatic generation of bloop configuration files


jvican:
The idea behind bloop is that it only understands its own configuration files, which have to be written from several mainstream build tools in order to be used. One small issue with this is that whenever users change the project definitions, they have to run `bloopInstall` or a similar task in other build tools to generate new configuration files.
  
This ticket proposes to change the status quo to perform such generation automatically. At least, Maven and sbt give us the hooks to do so (one via build lifecycle subscription and the other one via `Def.initialize`).
  
The idea is that this process has to be transparent to the user so that the cognitive overhead of using bloop is even smaller. In particular, it has to:
  
* Be blazingly fast (avoiding the project compilation as it does now).
* Be cached (in order to allow us to skip the process if nothing changed).
* Be incremental at the project level (if a project definition changed, we only have to generate the config file for it).

nafg:
Not sure I understand this ticket -- when exactly would automatic generation run?

jvican:
The automatic generation of configuration files a loaded builds and the execution of dependency resolution and source/resource generation are executed. Depending on the build tool, the automatic generation is done one way or another. On Maven, for example, the generation would need to happen in its build lifecycle. On sbt, `Def.initialize` allows us to run any task we want.

Not all build tools would be candidates for this automatic generation. The generation depends on data from the build, and sometimes that data has to be produced (like resolving the dependencies to find a project's classpath). In reality, only users of build tools like bazel, gradle and sbt (in that order) will benefit from the automatic generation because it will be faster than in other build tools.

nafg:
What I mean is, from the user's perspective what actions would cause it to run

jvican:
I already told you! Look into `Def.initialize` or read about Maven's lifecycle. It's done differently depending on the build tool, but the idea is that it will be done as soon as all the build is loaded, automatically.

nafg:
I couldn't find anything about Def.initialize doing something like that.
But it sounds like in short, instead of doing `;reload;bloopInstall` one
would just do `reload`?

jvican:
For sbt users, yeah. In fact, we already do that but for the configuration files of the build project, instead of the projects defined in the build.

jvican:
This is an implicit ticket covered in https://github.com/scalacenter/bloop/issues/316

PR #145, opened 2017-12-28
Replace `discoveredSbtPlugins` in sbt 0.13

Duhemm:
This task always triggers compilation, because it doesn't rely on a
dynamic task as it should. This commit replaces the implementation of
this task in sbt-bloop by the same implementation that is used in sbt
1.0 (which is correct and uses a dynamic task).

Because this task is used in a resource generator that is present by
default, we would always trigger compilation when inspecting the
generated resources. This is no longer the case (except if the project
is an sbt plugin).

### DIFF
--- a/sbt-bloop/src/main/scala-sbt-0.13/bloop/DiscoveredSbtPlugins.scala
+++ b/sbt-bloop/src/main/scala-sbt-0.13/bloop/DiscoveredSbtPlugins.scala
@@ -1,0 +1,16 @@
+package bloop
+
+import sbt.{Compile, Def, PluginDiscovery, Setting}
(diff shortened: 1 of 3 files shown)


Duhemm:
I can confirm that compilation is no longer triggered in spark with this patch 😄

jvican:
Love this, finally we got rid of this annoyance.

jvican:
Since we have this, can we restore the way to test that we indeed generated the config files in the scripted tests? the reason I ask is because once you have config files locally, if there's a change that doesn't generate new config files it should fail (this will help us when we tackle #140

Duhemm:
> Since we have this, can we restore the way to test that we indeed generated the config files in the scripted tests?

Yup, I removed the commit from #124

PR #148, opened 2018-01-02
Add Maven's bloop integration

jvican:
This pull request is still work-in-progress. Concretely, the Maven test
infrastructure is missing. So far, I've been testing it with the Maven build of
Spark, and it has worked great, though we need automatic tests to make sure
this integration will never stop working.

This integration consists of two parts:

1. The sbt maven plugin integration. The part that takes care of creating a
plugin descriptor and doing some custom composition of mojos (imported from the
Scala Maven plugin https://github.com/davidB/scala-maven-plugin/). This part is
the most tricky, since it uses obscure Maven APIs.

2. The maven plugin (bloop mojo), that takes care of reusing the machinery from
Scala maven plugin to generate the bloop configuration files.


jvican:
An excellent property of this plugin is that we don't need to reimplement any Scala logic, we just reuse the Scala Maven plugin setup existing in `pom.xml`. In practice, this also means that we only need two custom configuration file options that do not exist in Scala Maven plugin, and can be independently configured from the Scala logic.

build.sbt
```diff
@@ -102,9 +102,21 @@ val benchmarks = project
   .enablePlugins(JmhPlugin)
   .settings(benchmarksSettings)
 
+lazy val integrationsCore = project
+  .in(file("integrations/core"))
```
Duhemm: Being paranoid with Windows:

`.in(file("integrations") / "core")`

build.sbt
```diff
@@ -102,9 +102,21 @@ val benchmarks = project
...
+    publishLocal := {
+      publishLocal.value
+      publishM2.value
+    }
+  )
+
 import build.BuildImplementation.BuildDefaults
 lazy val sbtBloop = project
-  .in(file("sbt-bloop"))
+  .in(file("integrations/sbt-bloop"))
```
Duhemm: `.in(file("integrations") / "sbt-bloop")`

build.sbt
```diff
@@ -116,7 +128,26 @@ lazy val sbtBloop = project
     BuildDefaults.scriptedSettings(resourceDirectory in Test in frontend),
   )
 
-val allProjects = Seq(backend, benchmarks, frontend, sbtBloop)
+val mavenBloop = project
+  .in(file("integrations/maven-bloop"))
```
Duhemm: `.in(file("integrations") / "maven-bloop")`

build.sbt
```diff
@@ -116,7 +128,26 @@ lazy val sbtBloop = project
...
+  .dependsOn(integrationsCore)
+  .settings(
+    name := "maven-bloop",
+    mavenPlugin := true,
+    publishLocal := publishM2.dependsOn(publishLocal in integrationsCore).value,
+    classpathTypes += "maven-plugin",
+    makePomConfiguration :=
+      makePomConfiguration.value.withIncludeTypes(classpathTypes.value),
```
Duhemm: Maybe move those to `BuildPlugin`?

integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
```diff
@@ -0,0 +1,106 @@
...
+      }
+      val dependencies = project.getProjectReferences().asScala.values.map(_.getArtifactId).toList
+      val allScalaJars = mojo.getAllScalaJars()
+      val tmpDir = new File(classesDir, "tmp-bloop")
+      val fork = mojo.getFork()
+      val javaOptions = launcher.getJvmArgs()
+      val javaHome = mojo.getJavaHome()
+
+      // FORMAT: OFF
+      val compileConfig = BloopConfig(name, baseDirectory, dependencies, mojo.getScalaOrganization,
```
Duhemm: `val config`, no?

build.sbt
jvican: Yup, totally agree :smile:

jvican:
> The MavenPluginIntegration seems terribly complicated... How much more complicated would it be to just use Maven for this integration?
  
Our maven integration does custom stuff that Maven doesn't do by default, like reusing mojo descriptors from downstream projects (`scala-maven-plugin`) so that we can automatically reuse their configuration and not force the user to duplicate it.
  
Personally, I'm surprised that we're able to do all the stuff we're doing with so few lines of code.

integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
```diff
@@ -0,0 +1,106 @@
...
+      val sourceDirs = sourceDirs0.map(_.getCanonicalFile())
+      val classesDir = classesDir0.getCanonicalFile()
+      val classpath = {
+        val cp = classpath0.asScala.toList.asInstanceOf[List[String]].map(new File(_))
+        if (cp.headOption.contains(classesDir)) cp.tail else cp
+      }
+      val dependencies = project.getProjectReferences().asScala.values.map(_.getArtifactId).toList
+      val allScalaJars = mojo.getAllScalaJars()
+      val tmpDir = new File(classesDir, "tmp-bloop")
+      val fork = mojo.getFork()
```
Duhemm: It looks like `fork` from `scala-maven-plugin` actually means whether compilation should happen in a forked JVM? My reading is that `run` is always forked, and that for running the tests, it's the responsibility of another plugin. I'd set it to `true` here too.

https://github.com/davidB/scala-maven-plugin/blob/ef3d064eaf9774859ffec1053962f6fc8117d19e/src/main/java/scala_maven/ScalaMojoSupport.java#L286-L291

project/MavenPluginIntegration.scala
```diff
@@ -0,0 +1,298 @@
...
+    MavenPluginImplementation.projectSettings
+}
+
+object MavenPluginKeys {
+  import sbt.{settingKey, taskKey}
+  val mavenPlugin = settingKey[Boolean](
+    "If true, enables adding maven as a dependency and auto-generation of the plugin descriptor file.")
+  val mavenLogger = settingKey[Logger]("The Plexus logger used for Maven APIs.")
+  val mavenProject =
+    taskKey[MavenProject]("Return the corresponding maven project.")
```
Duhemm: `s/corresponding/current/`?

project/MavenPluginIntegration.scala
```diff
@@ -0,0 +1,298 @@
...
+            annotationExtractor.getMojoDescriptors(classesDir, descriptor, artifacts)
+          annotationMojos.foreach(mojo => descriptor.addMojo(mojo))
+
+          // This is the plugin dependent part of the Maven plugin setup
+          val bloopMojoDescriptor = descriptor.getMojo(BloopGoal)
+          val selector = PluginSelector("scala-maven-plugin", "cc")
+          aggregateParametersFromDependentPlugins(artifacts, selector, bloopMojoDescriptor)
+
+          val generator = new PluginDescriptorGenerator(new SystemStreamLog())
+          val xmlDirectory = Keys.resourceManaged.in(Compile).value./("META-INF/maven")
```
Duhemm: `"META-INF" / "maven"`

project/MavenPluginIntegration.scala
```diff
@@ -0,0 +1,298 @@
...
+        pluginSelector: PluginSelector,
+        mojoDescriptor: MojoDescriptor
+    ): MojoDescriptor = {
+      import sbt.io.IO
+      import sbt.io.syntax.fileToRichFile
+      val dependentMavenPlugins = artifacts.filter(_.getType() == "maven-plugin")
+      val dependencyPluginDescriptors = dependentMavenPlugins.map { (a: Artifact) =>
+        IO.withTemporaryDirectory { tempDir =>
+          val _ = IO.unzip(a.getFile(), tempDir)
+          val pluginXml = tempDir./("META-INF/maven/plugin.xml")
```
Duhemm: `"META-INF" / "maven" / "plugin.xml"`

project/MavenPluginIntegration.scala
```diff
@@ -0,0 +1,298 @@
...
+      def shortenAndToString(f: File): String = baseDir.relativize(f.toPath()).toString()
+      build.setSourceDirectory(shortenAndToString(Keys.sourceDirectory.in(Compile).value))
+      build.setTestSourceDirectory(shortenAndToString(Keys.sourceDirectory.in(Test).value))
+      build.setOutputDirectory(shortenAndToString(Keys.classDirectory.in(Compile).value))
+      build.setTestOutputDirectory(shortenAndToString(Keys.classDirectory.in(Test).value))
+      model.setBuild(build)
+
+      new MavenProject(model) {
+        def fs(f: File): String = f.getAbsolutePath()
+        import scala.collection.JavaConverters._
```
Duhemm: This is already imported a few lines above

project/build.sbt
```diff
@@ -1,8 +1,35 @@
...
+        pluginMainDir / "scala",
+        pluginMainDir / s"scala-sbt-${Keys.sbtBinaryVersion.value}"
+      )
+    },
+    // We need to add libdeps for the maven integration plugin to work
+    libraryDependencies ++= List(
+      "org.apache.maven.plugin-tools" % "maven-plugin-tools-api" % "3.5",
+      "org.apache.maven.plugin-tools" % "maven-plugin-annotations" % "3.5",
+      "org.apache.maven.plugin-tools" % "maven-plugin-tools-generators" % "3.5",
+      "org.apache.maven.plugin-tools" % "maven-plugin-tools-annotations" % "3.5",
```
Duhemm: Factor out the version number?

.gitmodules
```diff
@@ -5,7 +5,7 @@
 [submodule "nailgun"]
 	path = nailgun
 	url = https://github.com/scalacenter/nailgun.git
-	branch = zinc-nailgun
+	branch = bloop-nailgun-coursier
```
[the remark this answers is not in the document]
jvican: Because of sbt-pom-reader, which is classloaded in the same place as the maven plgin integration and caused some misbehaviours and NPEs.

integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  private val emptyLauncher = new AppLauncher("", "", Array(), Array())
+  private val testFrameworks: List[List[String]] = List(
+    List("org.scalacheck.ScalaCheckFramework"),
+    List("org.scalatest.tools.Framework", "org.scalatest.tools.ScalaTestFramework"),
+    List("org.specs.runner.SpecsFramework",
+         "org.specs2.runner.Specs2Framework",
+         "org.specs2.runner.SpecsFramework"),
+    List("com.novocode.junit.JUnitFramework")
```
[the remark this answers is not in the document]
Duhemm: Reading more of the maven scala plugin, I understand better, let's keep it that way.

.gitmodules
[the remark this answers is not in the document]
jvican: Coursier has indeed been removed, this is just an stale name.

integrations/core/src/main/scala/bloop/integrations/BloopConfig.scala
```diff
@@ -0,0 +1,51 @@
+package bloop.integrations
+
+import java.io.{File, FileOutputStream}
+
+case class BloopConfig(
+    name: String,
+    baseDirectory: File,
```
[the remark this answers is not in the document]
jvican: I think this is a copy-paste from the previous code, do you think it's worth the effort?

integrations/core/src/main/scala/bloop/integrations/BloopConfig.scala
```diff
@@ -0,0 +1,51 @@
+package bloop.integrations
+
+import java.io.{File, FileOutputStream}
+
+case class BloopConfig(
```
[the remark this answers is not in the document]
jvican: If we really care about bincompat we should use protobuf :smile: However, bloop should not be used as a library, only as a server, so do you think this will be an issue?

integrations/core/src/main/scala/bloop/integrations/BloopConfig.scala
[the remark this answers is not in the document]
Duhemm: We can leave it for sometime, it's not that important...

integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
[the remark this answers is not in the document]
jvican: The problem here is that the Maven Scala plugin does not have an options for test frameworks! All testing infra goes through another plugin that is Maven specific (to the best of my knowledge).

project/BuildPlugin.scala
```diff
@@ -203,20 +201,9 @@ object BuildImplementation {
...
-    val setupTests = Command.command("setupTests") { state =>
-      s"^sbtBloop/${Keys.publishLocal.key.label}" ::
-        s"sbtBloop/${BuildKeys.scriptedAddSbtBloop.key.label}" ::
-        s"sbtBloop/${ScriptedKeys.scripted.key.label}" ::
-        state
-    }
-
     private def createScriptedSetup(testDir: File) = {
       s"""
+         |import bloop.integrations.sbt.SbtBloop.autoImport.bloopConfigDir
```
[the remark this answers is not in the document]
jvican: It isn't, this is just dead code.

integrations/core/src/main/scala/bloop/integrations/BloopConfig.scala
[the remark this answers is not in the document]
Duhemm: I believe it'd be nice to be able to offer a stable API for integration as a library. We can probably take care of that sometime in the future, when we get closer to a real release.

project/MavenPluginIntegration.scala
jvican: It doesn't have quite the same meaning, why do you prefer `current`?

project/MavenPluginIntegration.scala
jvican: HEHE.

project/MavenPluginIntegration.scala
Duhemm: Because otherwise it doesn't say to what it corresponds. `corresponding to the current project` then?

integrations/core/src/main/scala/bloop/integrations/BloopConfig.scala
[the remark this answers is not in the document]
jvican: Do you think this would be useful? Bloop is a builder server, it should run independently of any tool. What I think we should focus on is in the definition of a nice API :smile:

Duhemm:
test performance please

(I just want to make sure it works on all PRs)

bloopoid:

bloopoid:

project/MavenPluginIntegration.scala
jvican: yes, i'll do that :smile:

PR #153, opened 2018-01-11
Add bsp implementation for bloop

jvican:
The BSP spec is available here: https://scalacenter.github.io/bsp/.

This is an implementation that depends on the awesome lsp4s created in
`language-server`. The implementation is not yet refined and it's a work in
progress that still needs to go through several refinement passes.

There's one test that tests the available functionality as of now, except
displaying test cases, which will be done after the integration with vscode is
polished.


jvican:
Tests are failing because bsp is not yet public.

build.sbt
```diff
@@ -84,6 +89,11 @@ val frontend = project
     fork in run := true,
     fork in Test := true,
     parallelExecution in test := false,
+    libraryDependencies ++= List(
+      "io.monix" %% "monix" % "2.3.2",
+      // Depend on ipcsocket only for testing since it has special changes to use clients
+      "org.scala-sbt.ipcsocket" % "ipcsocket" % "1.0.0" % Test
```
jvican: ipcsocket modifies the local sockets implementations from Nailgun and adds a few utils to spawn clients, we're adding it here only in the tests so that we can create clients for named pipes and unix sockets.

frontend/src/main/resources/logback.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jvican: YES! Another logger configuration file. We need this for lsp4s. Hope we can get rid of it in the future.

project/BuildPlugin.scala
```diff
@@ -155,8 +160,19 @@ object BuildImplementation {
...
 
+  import sbt.{CrossVersion, compilerPlugin}
   final val projectSettings: Seq[Def.Setting[_]] = Seq(
-    Keys.scalacOptions in Compile := reasonableCompileOptions,
+    Keys.scalacOptions := reasonableCompileOptions,
+    Keys.scalacOptions ++= {
+      if (Keys.scalaBinaryVersion.value.startsWith("2.10")) Nil
+      else List("-Yrangepos")
+    },
+    Keys.libraryDependencies ++= {
```
jvican: Let's keep this for language-server, we can remove it if we want in the future.

jvican:
After this pull request is merged, I'll follow up on @Duhemm's awesome work in #170 to propagate all the monix tasks in our interpreter API. When we have that, we can have integration tests for bsp by directly using `Cli.run` instead of `BspServer.run`, allowing us to test all the bsp validation that we have.

frontend/src/main/scala/bloop/Cli.scala
```diff
@@ -139,29 +144,22 @@ object Cli {
...
 
-    val cliOptions = action match {
-      case e: Exit => CliOptions.default
-      case p: Print => CliOptions.default
-      case r: Run => r.command.cliOptions
+    val (cliOptions, logToFile) = action match {
+      case e: Exit => (CliOptions.default, false)
+      case p: Print => (CliOptions.default, false)
+      case Run(cmd: Commands.Bsp, _) => (cmd.cliOptions, false)
+      case r: Run => (r.command.cliOptions, false)
```
Duhemm: `logToFile` is always false?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    override def warn(msg: String): Unit = {
+      callSiteState.logger.warn(msg)
+      Window.showMessage.warn(msg)(client)
+    }
+
+    override def trace(t: Throwable): Unit = {
+      callSiteState.logger.trace(t)
+      Window.showMessage.info(t.toString)(client)
```
Duhemm: Not sure that traces should be forwarded to the clients, it's an even lower log level than debug.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -0,0 +1,142 @@
...
+   * Implements the initialize method that is the first pass of the Client-Server handshake.
+   *
+   * @param initializeBuildParams The params request that we get from the client.
+   * @return An async computation that returns the response to the client.
+   */
+  def initialize(
+      initializeBuildParams: InitializeBuildParams
+  ): MonixTask[Either[JsonRpcResponse.Error, InitializeBuildResult]] = MonixTask {
+    callSiteState.logger.info("request received: build/initialize")
+    bspLogger.info("request received: build/initialize")
```
Duhemm: Why do we need to log that twice? Where are the log messages given to `callSiteState.logger` going? How about `bspLogger`?

frontend/src/main/scala/bloop/bsp/BspServer.scala
```diff
@@ -0,0 +1,79 @@
...
+
+object BspServer {
+  private[bloop] val isWindows: Boolean =
+    System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("windows")
+
+  import java.net.InetSocketAddress
+  private sealed trait ConnectionHandle { def socket: ServerSocket }
+  private case class WindowsLocal(pipeName: String, socket: ServerSocket) extends ConnectionHandle
+  private case class UnixLocal(path: Path, socket: ServerSocket) extends ConnectionHandle
+  private case class Tcp(address: InetSocketAddress, socket: ServerSocket) extends ConnectionHandle
```
Duhemm: Why do we need `address`? How's that different from `socket.getInetAddress`?

frontend/src/main/scala/bloop/bsp/BspServer.scala
```diff
@@ -0,0 +1,79 @@
...
+        state.logger.debug(s"Establishing server connection at pipe $pipeName")
+        WindowsLocal(pipeName, new NGWin32NamedPipeServerSocket(pipeName))
+      case Commands.UnixLocalBsp(socketFile, _) =>
+        state.logger.debug(s"Establishing server connection at $socketFile")
+        UnixLocal(socketFile, new NGUnixDomainServerSocket(socketFile.toString))
+      case Commands.TcpBsp(address, portNumber, name) =>
+        val socketAddress = new InetSocketAddress(address, portNumber)
+        state.logger.debug(s"Establishing server connection via TCP at ${socketAddress}")
+        // Use 0 instead of a concrete number to have an infinite timeout
+        Tcp(socketAddress, new java.net.ServerSocket(portNumber, 0, address))
```
Duhemm: This starts the server rather than establish a connection, right? I think the logs should read something along the lines of `Starting server listening on ...`, WDYT?

frontend/src/main/scala/bloop/bsp/BspServer.scala
```diff
@@ -0,0 +1,79 @@
...
+        case w: WindowsLocal => s"local:${w.pipeName}"
+        case u: UnixLocal => s"local://${u.path.toString}"
+        case t: Tcp => s"tcp://${t.address.getHostString}"
+      }
+    }
+
+    establishConnection(initServer(cmd, state)).flatMap {
+      case Established(socket, handle) =>
+        import state.logger
+        logger.verbose("Bloop has established connection with a client.")
```
Duhemm: The log would be more useful if it read something like `Client connected from [address]`, no?

frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -1,25 +1,84 @@
...
     def cliOptions: CliOptions
   }
 
-  sealed trait CoreCommand extends Command {
+  /** Represents a command that has gone through an initial layer of validation. */
+  sealed trait ValidatedCommand extends Command
+
+  sealed trait ValidatedBsp extends ValidatedCommand
+  case class WindowsLocalBsp(pipeName: String, cliOptions: CliOptions) extends ValidatedBsp
+  case class UnixLocalBsp(socket: Path, cliOptions: CliOptions) extends ValidatedBsp
```
Duhemm: `AbsolutePath`?

backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -98,6 +99,11 @@ object BloopLogger {
   def at(name: String, out: PrintStream, err: PrintStream): BloopLogger =
     new BloopLogger(name, out, err)
 
+  def atFile(name: String, logFile: AbsolutePath): BloopLogger = {
+    val outputForFile = new PrintStream(new FileOutputStream(logFile.toFile))
+    BloopLogger.at(name, outputForFile, outputForFile)
```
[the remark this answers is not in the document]
jvican: I have to remove all this part, this is a remainder of the previous stdin/stdout communication.

frontend/src/main/scala/bloop/Cli.scala
jvican: Yup, because of the above reason :smile:.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: They shouldn't, this was for debugging :smile:.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: Because `callSiteState.logger` logs to a file while `bspLogger` logs to the BSP client. We probably shouldn't do `bspLogger` here. I'll remove it.

frontend/src/main/scala/bloop/bsp/BspServer.scala
jvican: Good catch bro, `socket.getInetAddress` is better.

jvican:
Tests are failing because of a very interesting thing I didn't know: Unix socket paths have a size constraint.
  
```
[error] Test bloop.bsp.BspLocalProtocolSpec.TestBuildTargets failed: java.io.IOException: Cannot fit name [/drone/src/github.com/scalacenter/bloop/integration-tests/integration-projects/utest/bloop-config/test-73a3c6d1.socket] in maximum unix domain socket length, took 1.009 sec
```
  
Will fix this.

jvican:
Internet seems to give us a good answer:
  
```
If v is a path (path-string?), then the current platform must be either Linux or Mac OS X, and the length of v’s corresponding absolute path must be less than or equal to the platform-specific length (108 bytes on Linux, 104 bytes on Mac OS X). Example: "/tmp/mysocket".
```
  
BSD has the same limit as Linux.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  // Internal state, think how to make this more elegant.
+  @volatile private final var currentState: State = null
+
+  /**
+   * Get the latest state that can be reused and cached by bloop so that the next client
+   * can have access to it.
+   *
+   * @return
```
Duhemm: Unfinished comment?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
Duhemm: Didn't you want to remove `bspLogger`?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  def initialized(
+      initializedBuildParams: InitializedBuildParams
+  ): Unit = {
+    bspLogger.info("BSP initialization handshake complete.")
+  }
+
+  def compile(params: CompileParams): MonixTask[Either[JsonRpcResponse.Error, CompileReport]] = {
+    MonixTask {
+      // TODO(jvican): Naive approach, we need to implement batching here.
```
Duhemm: Can you open an issue for that if you don't want to take care of it before we merge?

frontend/src/main/scala/bloop/bsp/BspServer.scala
```diff
@@ -0,0 +1,110 @@
...
+    System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("windows")
+  private[bloop] val isMac: Boolean =
+    System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("mac")
+
+  import java.net.InetSocketAddress
+  private sealed trait ConnectionHandle { def serverSocket: ServerSocket }
+  private case class WindowsLocal(pipeName: String, serverSocket: ServerSocket)
+      extends ConnectionHandle
+  private case class UnixLocal(path: Path, serverSocket: ServerSocket) extends ConnectionHandle
+  private case class Tcp(address: InetSocketAddress, serverSocket: ServerSocket)
```
Duhemm: I thought that we agreed that `address` was not necessary? Is it in the end?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -0,0 +1,141 @@
...
+      Window.showMessage.error(msg)(client)
+    }
+
+    override def warn(msg: String): Unit = {
+      callSiteState.logger.warn(msg)
+      Window.showMessage.warn(msg)(client)
+    }
+
+    override def trace(t: Throwable): Unit = {
+      Window.showMessage.info(t.toString)(client)
```
Duhemm: Do you want to show the traces to the client? My understanding was that you wanted to log them but not forward them to the client.

frontend/src/main/scala/bloop/bsp/BspServer.scala
```diff
@@ -0,0 +1,110 @@
...
+      extends ConnectionHandle
+
+  import Commands.ValidatedBsp
+  import monix.{eval => me}
+  private def initServer(cmd: ValidatedBsp, state: State): me.Task[ConnectionHandle] = {
+    cmd match {
+      case Commands.WindowsLocalBsp(pipeName, _) =>
+        state.logger.debug(s"Establishing server connection at pipe $pipeName")
```
Duhemm: It's not really establishing the connection, it's waiting for a connection, no?

frontend/src/main/scala/bloop/bsp/ProjectUris.scala
```diff
@@ -0,0 +1,25 @@
...
+      query.headOption.flatMap {
+        case Array("id", projectName) => state.build.getProjectFor(projectName)
+        case _ => None
+      }
+    }
+  }
+
+  def toUri(projectBaseDir: AbsolutePath, id: String): URI = {
+    new URI(s"file:///${projectBaseDir.syntax}?id=${id}")
+  }
```
Duhemm: Is this Windows friendly? Do you know the rules to make something Windows-friendly?

frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -1,25 +1,84 @@
...
+  case class TcpBsp(host: InetAddress, port: Int, cliOptions: CliOptions) extends ValidatedBsp
+
+  /** Represents a command that is used by the cli and has no user input validation. */
+  sealed trait RawCommand extends Command
+
+  sealed trait CompilingCommand extends RawCommand {
```
Duhemm: `CompilingCommand` makes it sound a little bit like we're pushing command dependencies inside the hierarchy, but I really can't find a better name.

frontend/src/main/scala/bloop/cli/validation/Validate.scala
```diff
@@ -0,0 +1,85 @@
...
+
+    cmd.protocol match {
+      case BspProtocol.Local if isWindows => validatePipeName
+      case BspProtocol.Local => validateSocket
+      case BspProtocol.Tcp => validateTcp
+    }
+  }
+}
+
+object Feedback {
```
Duhemm: Would be cool in the future to extract that to another file and use it for all the feedback that Bloop has to offer.

frontend/src/test/scala/bloop/bsp/BspClientTest.scala
```diff
@@ -0,0 +1,115 @@
...
+    val common = cmd.cliOptions.common.copy(workingDirectory = cwd.syntax)
+    val cliOptions = cmd.cliOptions.copy(configDir = Some(configDir.underlying), common = common)
+    cmd match {
+      case cmd: Commands.WindowsLocalBsp => cmd.copy(cliOptions = cliOptions)
+      case cmd: Commands.UnixLocalBsp => cmd.copy(cliOptions = cliOptions)
+      case cmd: Commands.TcpBsp => cmd.copy(cliOptions = cliOptions)
+    }
+  }
+
+  val scheduler: Scheduler = Scheduler(java.util.concurrent.Executors.newFixedThreadPool(4),
```
Duhemm: Why not use the default thread pool? I think that's what we do in other tests

build.sbt
```diff
@@ -153,6 +163,7 @@ addCommandAlias(
   Seq(
     s"+${bridgeIntegration.id}/$publishLocalCmd",
     s"+${zincIntegration.id}/$publishLocalCmd",
+    s"${bspIntegration.id}/$publishLocalCmd",
```
Duhemm: While I'm thinking about it, is there a released version of `bsp` somewhere? We'll need it before the next release :)

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -0,0 +1,141 @@
+package bloop.bsp
+
+import bloop.cli.Commands
+import bloop.engine.{Interpreter, State}
+import bloop.io.AbsolutePath
+import ch.epfl.`scala`.bsp.schema._
+import monix.eval.{Task => MonixTask}
```
Duhemm: How about using `me.Task` like you do in the other file?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -0,0 +1,141 @@
...
+   * Creates a logger that will forward all the messages to the underlying bsp client.
+   * It does so via the replication of the `window/showMessage` LSP functionality.
+   */
+  private final val bspForwarderLogger = new bloop.logging.AbstractLogger() {
+    override def name: String = "bsp-logger"
+    override def verbose[T](op: => T): T = op
+    override def ansiCodesSupported(): Boolean = true
+
+    override def debug(msg: String): Unit = {
+      callSiteState.logger.debug(msg)
```
Duhemm: Do we want this logger to always be verbose?

frontend/src/main/scala/bloop/bsp/ProjectUris.scala
```diff
@@ -0,0 +1,25 @@
...
+import bloop.Project
+import bloop.engine.{Dag, State}
+import bloop.io.AbsolutePath
+
+object ProjectUris {
+  def getProjectDagFromUri(projectUri: String, state: State): Option[Project] = {
+    if (projectUri.isEmpty) None
+    else {
+      val uri = new URI(projectUri)
+      val query = uri.getQuery().split("&").map(_.split("="))
```
Duhemm: I think that you need `getRawQuery` here, otherwise, if a value in the query contains `&` you'll split it. Same problem with `=`.

frontend/src/main/scala/bloop/engine/ExecutionContext.scala
```diff
@@ -12,4 +12,10 @@ object ExecutionContext {
   // If you change the backing of this thread pool, please make sure you modify users of `executor`
   implicit val threadPool: scala.concurrent.ExecutionContext =
     scala.concurrent.ExecutionContext.fromExecutorService(executor)
+
+  import monix.execution.Scheduler
+  implicit lazy val bspScheduler: Scheduler = Scheduler {
+    // TODO: Revisit this.
```
Duhemm: Can you open an issue?

frontend/src/main/scala/bloop/engine/State.scala
```diff
@@ -97,4 +96,22 @@ object State {
...
     state
   }
+
+  import bloop.Project
+  import bloop.io.AbsolutePath
+  def loadStateFor(
+      configDirectory: AbsolutePath,
+      commonOptions: CommonOptions,
+      logger: Logger
+  ): State = {
```
Duhemm: I'm wondering if this shouldn't be renamed or refactored. It makes it look like we're loading a state for this `configDirectory`, `commonOptions` and `logger`...

frontend/src/main/scala/bloop/engine/State.scala
```diff
@@ -97,4 +96,22 @@ object State {
...
+    State.stateCache.getStateFor(configDirectory) match {
```
Duhemm: Shouldn't we just use `computeIfAbsent`?

frontend/src/main/scala/bloop/cli/Commands.scala
jvican: With this name I wanted to express the fact that all these commands can trigger compilation.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: Yes

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: I removed the bad line.

frontend/src/main/resources/logback.xml
jvican: No please, no Groovy. Prefer to keep it this way, I just copy pasted it from somewhere, and it's meant to be that way. We can do a whole refactoring of the loggers in a follow up PR.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: I changed `verbose` to delegate to `callSiteLogger`.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: https://github.com/scalacenter/bloop/issues/217

frontend/src/main/scala/bloop/bsp/BspServer.scala
jvican: `address` is necessary because otherwise it's not possible to know the port from `serverSocket.getAddress`, since it returns an `InetAddress`, its super class.

frontend/src/main/scala/bloop/bsp/ProjectUris.scala
jvican: I removed one `/` and now it is windows friendly. However, this should not be important since this URI is only used to uniquely identify every workspace, the user is not supposed to parse it.

frontend/src/main/scala/bloop/engine/State.scala
jvican: Let's call it `loadActiveStateFor`.

frontend/src/main/scala/bloop/engine/State.scala
jvican: Actually, `addIfMissing` relies on `computeIfAbsent`, so I can just use that.

frontend/src/test/scala/bloop/bsp/BspClientTest.scala
jvican: Because I want to constrain it by default. I don't want the thread pool for bsp to be reused for the tests because I'm making the tests more restrictive (less threads per pool) to make sure we don't get any race condition. Should be therefore independent.

project/Dependencies.scala
```diff
@@ -39,4 +41,7 @@ object Dependencies {
   val mavenPluginApi = "org.apache.maven" % "maven-plugin-api" % mavenApiVersion
   val mavenPluginAnnotations = "org.apache.maven.plugin-tools" % "maven-plugin-annotations" % mavenAnnotationsVersion % Provided
   val mavenScalaPlugin = "net.alchim31.maven" % "scala-maven-plugin" % mavenScalaPluginVersion
+
+  val ipcsocket = "org.scala-sbt.ipcsocket" % "ipcsocket" % ipcsocketVersion
+  val monix = "io.monix" %% "monix" % monixVersion
```
Duhemm: Thanks for adding that here, but I think you forgot to update `build.sbt` afterwards?

jvican: LOL, yes.

PR #156, opened 2018-01-12
Move integration tests to `/scripted-tests`

Duhemm:
They were in the resources, which means that sbt was copying them all
before running each tests, slowing down the whole thing. It also look
like sbt was looking at them to determine whether stuff was changed.

sbt feels a bit faster now.

PR #157, opened 2018-01-12
Switch from uTest to JUnit

Duhemm:
The 2 reasons for this change are
 - Focus on a single test framework.
 - Categorize tests between slow and fast. This will become useful especially when we run the tests on AppVeyor, where we'll likely exclude the slow tests.


jvican:
I love this PR, thanks for taking your time to do it.

frontend/src/test/scala/bloop/tasks/IntegrationTestSuite.scala
```diff
@@ -1,41 +1,43 @@
...
 import bloop.io.AbsolutePath
 
-object IntegrationTestSuite extends DynTest {
-  val projects = Files.list(getClass.getClassLoader.getResources("projects") match {
-    case res if res.hasMoreElements => Paths.get(res.nextElement.getFile)
-    case _ => throw new Exception("No projects to test?")
-  })
+object IntegrationTestSuite {
+  val projects = Files
+    .list(getClass.getClassLoader.getResources("projects") match {
```
jvican: Can we not reuse `loadTestProject` here? I think this logic could be moved somewhere better than the test case.

frontend/src/test/scala/bloop/tasks/TaskTest.scala
```diff
@@ -1,121 +1,127 @@
...
-      addFive.dependsOn(addFour)
-
-      addFive.run().map {
-        case Task.Failure(partial, reasons) =>
-          assert(partial == Set(1, 2))
-          assert(reasons.map(_.getMessage) == List("Boom!"))
-        case Task.Success(_) => sys.error("The task failure did not report partial results!")
-      }
+  @Test
+  def resultsAreCombined = {
```
jvican: Remember that you can use symbols (as in `this test checks bla`) and then have junit print it correctly with a flag :wink:

frontend/src/test/scala/bloop/tasks/IntegrationTestSuite.scala
Duhemm: Yup, this has improved in #156 :)

frontend/src/test/scala/bloop/tasks/TaskTest.scala
Duhemm: I've been quickly playing with that, and as far as I can tell, it makes it harder to select individual method to run, so I'd prefer to keep it like it is now...

PR #160, opened 2018-01-14
Use Nailgun's `ThreadLocalPrintStream`

### LINKED ISSUE #159: `MultiplexedOutputStream` kills Nailgun from the second `run` on when not forking
It turns out that Nailgun replaces `System.out` and `System.err` with `ThreadLocalPrintStream` (which happen to do the *exact* same thing as `MultiplexedStreams`), and it expects to find `ThreadLocalPrintStream` instances in `System.out` and `System.err`.

Because we replace those streams in the first run, Nailgun will crash after that because of a `ClassCastException`.

The solution would be to use Nailgun's `ThreadLocalPrintStream` instead of our own implementation. I'm doing that now.

Duhemm:
It is doing the same thing as our implementation in
`MultiplexedStreams`. Unfortunately, using our implementation was
breaking Nailgun because it expected `System.out` and `System.err` to be
instances of `ThreadLocalPrintStream`. We were, however, replacing them
with another implementation, thus resulting in `ClassCastException`s in
Nailgun.

This commit removes `IdentifiedThread` (which was helping to distinguish
between the threads), and adapts `MultiplexedStreams` to use the
`ThreadLocalPrintStream`. `MultiplexedStreams` is kept to offer helpers
to initialize the streams and replace the streams to execute some code.

Fixes #159


frontend/src/main/scala/bloop/exec/MultiplexedStreams.scala
```diff
@@ -2,44 +2,29 @@ package bloop.exec
...
+import com.martiansoftware.nailgun.ThreadLocalPrintStream
 
 /**
  * Helper object to redirect bytes written to `System.out` and `System.err` to a given logger.
  */
 object MultiplexedStreams {
 
-  /** A combination (stdout, stderr) */
-  private case class Streams(out: PrintStream, err: PrintStream)
+  _initialize()
```
jvican: Can we not do a more sensible thing than side-effecting in the constructor? Is it possible to have this done explicitly somewhere?

frontend/src/main/scala/bloop/exec/MultiplexedStreams.scala
```diff
@@ -2,44 +2,29 @@ package bloop.exec
...
-  }
-
-  /** The multiplexed `stderr`. */
-  val stderr: PrintStream = {
-    val stream: OutputStream = (b: Int) => getStreams().err.write(b)
-    new PrintStream(stream)
-  }
-
-  initialize()
+  /**
```
jvican: I would remove this scaladoc, doesn't seem useful.

frontend/src/main/scala/bloop/exec/MultiplexedStreams.scala
```diff
@@ -51,16 +27,14 @@ object MultiplexedStreams {
    * @return The result of `op` wrapped in an `Some`, or None if an error occurred.
    */
   def withLoggerAsStreams[T](logger: Logger)(op: => T): Option[T] = {
+    initialize()
```
jvican: :tada:

jvican:
LGTM bro, nice to see all thie code reuse :smile:.

PR #163, opened 2018-01-15
Handle configure in Cli pattern matching

jvican:
```
[W]  [E1] /src/main/scala/bloop/Cli.scala
[W]       match may not be exhaustive.
[W]       It would fail on the following input: Right(Configure(_, _))
[W]       L88:             command match {
[W]                        ^
[E] /src/main/scala/bloop/Cli.scala: L88 [E1]
```


jvican:
I think we didn't use it, and that's why we didn't catch this before.

PR #170, opened 2018-01-17
Use Monix tasks all over bloop

Duhemm:
Still needs more polishing and more tests.


Duhemm:
test performance please

bloopoid:

bloopoid:

Duhemm:
Wow that looks like a huge performance improvement! I can't explain it, but that's great news 🎉 

https://snapshot.raintank.io/dashboard/snapshot/taKH5bxe6uCUfFe4IljFmDekZnTLWyeQ

Duhemm:
Not really actually, it looks like it's performing worse when doing compilation.

project/Dependencies.scala
```diff
@@ -13,6 +13,7 @@ object Dependencies {
   val junitVersion = "0.11"
   val graphvizVersion = "0.2.2"
   val directoryWatcherVersion = "0.4.0"
+  val monixVersion = "2.3.0"
```
olafurpg: You might want to give 3.0.0-RC3 a try, see https://github.com/monix/monix/releases/tag/v3.0.0-M3

> One other highlight of the release are the performance improvements for Task and Coeval, an on-going process to make sure that Monix's Task remains the best implementation in Scala's ecosystem.

frontend/src/main/scala/bloop/engine/tasks/Result.scala
```diff
@@ -0,0 +1,59 @@
+package bloop.engine.tasks
+
+/** The result of the execution of a `Task`. */
+sealed trait Result[+T] {
```
olafurpg: What is the difference between this and `Try[T]`?

Monix Task catches exceptions in map/flatMap, `Task.materialize` can turn Task[T] into `Task[Try[T]]`. It also has combinators to recover from errors.

Duhemm: Good find, thanks!

project/Dependencies.scala
Duhemm: I'll benchmark the two, thanks for the suggestion.

Duhemm:
test performance please

bloopoid:

frontend/src/main/scala/bloop/engine/tasks/Task.scala
```diff
@@ -3,16 +3,17 @@ package bloop.engine.tasks
...
 
 import monix.eval.{Task => MTask}
 import monix.execution.{CancelableFuture, Scheduler}
 
 /**
  * An asynchronous computation, possibly depending on other computations.
  */
-class Task[+T] private (private val underlying: MTask[Result[T]],
-                        origTask: MTask[Result[T]],
+class Task[+T] private (private val underlying: MTask[Try[T]],
```
olafurpg: Something feels wrong about this data structure. This seems to be closer to a tree rather than `Task`

```scala
case class Tree[T](node: Task[T], children: Seq[Tree[_]])
```

`children` might even be ListBuffer. `gatherUnordered` can be run at the end when you want to build your final `Task`.

The methods `map`/`flatMap` are not behaving the way I would expect them to, because it only applies to the underlying task and not dependencies.

jvican: I recommend using `Dag[Task[T]` here, `Dag` was created for this purpose and I find it a very good structure to deal with.

bloopoid:

frontend/src/main/scala/bloop/engine/tasks/Task.scala
Duhemm: @olafurpg Thanks, I've changed the tasks to call `gatherUnordered` at the end when building the final task as you've suggested. It's indeed clearer, and it fixed another issue.

I don't agree about the dependencies being a mutable list or about `map` and `flatMap`. I don't think that having the dependencies mutable solves any problem, but it makes it harder to reason about what happens in this case:

```scala
val t0 = Task { 0 }
val t1 = Task { 1 }
val t2 = t1.map(_ + 1)
t1.dependsOn(t0)
t2.runAsync // Should that trigger t0?
```

I also disagree with your comment about `map` and `flatMap`. The dependencies are arbitrary tasks, I don't see how we could apply an operation on all of them? It's also how these operations are defined on sbt's tasks, for instance.

Duhemm: @jvican The problem with `Dag[T]` is that all its elements have type `T`, it's not the case here.

jvican:
test performance please

bloopoid:

bloopoid:

jvican:
Results in the benchmarks look great to me, but we need to stabilize that machine -- it's just too unstable.

jvican:
In a follow-up PR, I'd like to add the following things:
  
* Propagate task up to the `Interpreter` interface.
* Only have one `Await` in our application.
* Make file watching cancelable and use tasks appropriately.

Duhemm:
LGTM

Duhemm:
test performance please

bloopoid:

bloopoid:

bloopoid:
performance test failed:



Error line number: 94

[check  for more information]

Duhemm:
test performance please

bloopoid:
performance test scheduled: 1 job(s) in queue, 1 running.

bloopoid:

Duhemm:
This will test the performance on `master`:
test performance please: 7ff9993d9ebc0ab0ddc8c3058efd73dafc54a4ff

Duhemm:
test performance please: 7ff9993d9ebc0ab0ddc8c3058efd73dafc54a4ff

bloopoid:

bloopoid:

Duhemm:
test performance please

bloopoid:

bloopoid:

PR #174, opened 2018-01-18
Fix alias in installation instructions

Duhemm:


jvican:
This won't be required after #175, closing :smile:

PR #177, opened 2018-01-18
Remove test from .gitignore

jvican:
Now that the integration tests are somewhere else, we can remove this
line that was causing us pain :)


jvican:
Updated with feedback, CI seems to pass :tada:!

jvican:
BYE BYE `git add --force`

PR #180, opened 2018-01-19
Revert "Remove `Bloop` shell"

jvican:
This reverts commit 74ce7cd0df022a9533e95cb20170081b12261ca5.


jvican:
Damn, this will probably fail now that I think about it.

PR #183, opened 2018-01-19
Sync nailgun

### LINKED ISSUE #169: Upgrade to latest nailgun
There have been important changes upstream. Catch up.

Duhemm:
Fixes #169 

@jvican Is that the naming scheme that you had in mind for the branches in the submodules?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,7 +5,7 @@
 [submodule "nailgun"]
 	path = nailgun
 	url = https://github.com/scalacenter/nailgun.git
-	branch = bloop-nailgun-coursier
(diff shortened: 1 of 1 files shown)


Duhemm:
I'm still wondering if we shouldn't just merge the remote branches in our forks, instead of rebasing. This would avoid having to switch the branches in the submodules...

jvican:
This is so weird, I would have sworn I already did this.
  
I expect to be more changes in nailgun, and merging branches will only complicate our lives. Rebasing is the way to go, the flatter the git history is the better will be to avoid issues rebasing and wasting our time.

PR #188, opened 2018-01-20
Close `DirectoryWatcher` when watching is finished

Duhemm:
Otherwise, the underlying `WatchService` will never be closed.


frontend/src/main/scala/bloop/io/SourceWatcher.scala
```diff
@@ -49,6 +49,6 @@ final class SourceWatcher(dirs0: Seq[Path], logger: Logger) {
         logger.error("Unexpected error happened when file watching.")
         logger.trace(t)
         result.mergeStatus(ExitStatus.UnexpectedError)
-    }
+    } finally watcher.close()
```
jvican: This won't ever be hit because there's no current "good" way of stop file watching, I propose waiting until I come up with a sane way of catching `CTRL-C` calls to indeed stop file watching.

Duhemm: I don't understand. This is a method that's defined on `DirectoryWatcher`:
https://github.com/gmethvin/directory-watcher/blob/b825bab32512692a80fe27738c5f4a2158f8069b/core/src/main/java/io/methvin/watcher/DirectoryWatcher.java#L195-L197

The `finally` clause will be executed, regardless of how we exit the block.

PR #189, opened 2018-01-21
Use correct branch when downloading nailgun client

Duhemm:


jvican:
What do you think about generating this file via sbt so that every time the version of some of our dependencies change the installation script is regenerated?

Duhemm:
Superseded by #193

PR #193, opened 2018-01-24
Generate installation script and Homebrew formula

Duhemm:
I rewrote the installation script in Python and tested it on Windows. It appears to be working.

I also added a Homebrew formula that uses the Python script to do the installation. I haven't been able to completely test it at the moment because we don't publish zinc and nailgun, so they're not found during installation (homebrew builds packages in a sandbox).

The scripts are generated with a very simple thing. Running `frontend/makeTemplates` generates the scripts.

Finally, I added a script to create releases in `bin/make-tag.sh`. Example usage: `bin/make-tag.sh v0.5`, `bin/make-tag.sh v0.5 my-remote-name`.

It'll create tags in both the Homebrew formula and Bloop's repo and push all of that. The remote name can be specified.

I know that's a lot of heterogeneous looking stuff, but everything was kind of built together...


olafurpg:
Let me know when you want a dummy tester for the new installation setup on mac!

Duhemm:
I will, thanks @olafurpg !

build.sbt
```diff
@@ -129,18 +130,21 @@ val bloop = project
...
+    s"^${sbtBloop.id}/$publishLocalCmd",
+    s"${mavenBloop.id}/$publishLocalCmd"
+  ).mkString(";", ";", "")
+)
 
-addCommandAlias("runTests", List(
-  s"${sbtBloop.id}/${scriptedAddSbtBloop.key.label}",
-  s"${sbtBloop.id}/${scripted.key.label}"
-).mkString(";", ";", ""))
+addCommandAlias("runTests",
```
jvican: Can we format this one like `setupIntegrations` :smile:?

frontend/src/main/scala/bloop/Cli.scala
```diff
@@ -23,10 +23,16 @@ object Cli {
...
     val command = ngContext.getCommand
-    val args =
+    val args = {
       if (command == "bloop.Cli") ngContext.getArgs
       else command +: ngContext.getArgs
-    val cmd = parse(args, nailgunOptions)
+    }
+    val cmd = {
+      // If no command is given to bloop, we'll receive the script's name.
+      if (command == "bloop")
```
jvican: Why is this necessary?

project/BuildPlugin.scala
```diff
@@ -53,17 +55,26 @@ object BuildKeys {
...
-    integrationTestsLocation := (Keys.baseDirectory in sbt.ThisBuild).value / "integration-tests" / "integration-projects",
+    integrationTestsLocation := buildBase.value / "integration-tests" / "integration-projects",
     Keys.testOptions += Tests.Argument(TestFrameworks.JUnit, "-v", "-a"),
     Keys.libraryDependencies ++= List(
       Dependencies.junit % Test
     ),
   )
 
+  import ohnosequences.sbt.GithubRelease.{keys => GHReleaseKeys}
+  val releaseSettings = Seq(
```
jvican: So beautiful 💛

project/ReleaseUtils.scala
```diff
@@ -0,0 +1,131 @@
+package build
+
+import java.io.File
+
+import sbt.{Def, Keys, MessageOnlyException}
+import sbt.io.syntax.fileToRichFile
+import sbt.io.IO
+
+/** Utilities that are useful for releasing Bloop */
+object ReleaseUtils {
```
jvican: This looks awesome, great work 👍

project/ReleaseUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Creates a new installation script (based on the normal installation script) that has default
+   * values for the nailgun commit and version of Bloop to install.
+   *
+   * This lets us create an installation script that doesn't need any additional input to install
+   * the version of Bloop that we're releasing.
+   */
+  val versionedInstallScript = Def.task {
+    IO.readLines(installScript.value) match {
+      case shebang :: rest =>
```
jvican: Dynamic scripts FTW

project/ReleaseUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+
+/** Utilities that are useful for releasing Bloop */
+object ReleaseUtils {
+
+  /** The path to our installation script */
+  private val installScript = Def.setting { BuildKeys.buildBase.value / "bin" / "install.py" }
+
+  /** The latest commit in the `nailgun` submodule. */
+  private val nailgunCommit = Def.setting {
+    GitUtils.withGit(GitUtils.submoduleGitDir("nailgun").value) { git =>
```
jvican: A small question: why is this necessary? Can we not reuse sbt dynver's utils for nailgun? `version in NailgunBuild` should give us this commit hash 🤔

project/ReleaseUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * The content of the Homebrew Formula to install the version of Bloop that we're releasing.
+   *
+   * @param version The version of Bloop that we're releasing (usually `Keys.version.value`)
+   * @param tagName The name of the tag that we're releasing
+   * @param installSha The SHA-256 of the versioned installation script.
+   */
+  def formulaContent(version: String, tagName: String, installSha: String): String = {
```
jvican: I'm really happy with how everything has turned out to be

project/build.sbt
```diff
@@ -4,6 +4,8 @@ val root = project
   .in(file("."))
   .settings(
     scalaVersion := "2.12.4",
+    resolvers += "Jenkins repo" at "http://repo.jenkins-ci.org/public/",
```
jvican: Agrrrrr.

project/BuildPlugin.scala
```diff
@@ -53,17 +55,26 @@ object BuildKeys {
...
+    GHReleaseKeys.ghreleaseRepoOrg := "scalacenter",
+    GHReleaseKeys.ghreleaseRepoName := "bloop",
+    GHReleaseKeys.ghreleaseAssets += ReleaseUtils.versionedInstallScript.value,
```
jvican: This means we're running `versionedInstallScript` twice, one here and one in `updateHomebrewFormula`. Would it be possible to make `versionedInstallScript` a setting instead of a task?

build.sbt
Duhemm: scalafmt doesn't agree -- do you really want to go against it?

olafurpg: @Duhemm http://scalameta.org/scalafmt/#Configstyle

olafurpg:
```
align = none
danglingParentheses = true
```
gives you this by default, fwiw

project/ReleaseUtils.scala
Duhemm:
```scala
scala> (version in build.BuildKeys.NailgunBuild).eval
res0: String = 0.1-SNAPSHOT
```

I didn't manage to get the value I wanted with `gitHeadCommit` either...

frontend/src/main/scala/bloop/Cli.scala
Duhemm: It's not necessary for this PR, I didn't intend to keep this commit here...
The reason is that at the moment, calling `bloop` without any argument will print `bloop: Command not found` which is extraordinarily surprising for people trying bloop for the first time. I'm happy to remove it and put that in another PR (as I initially intended to) if you prefer.

build.sbt
Duhemm: Actually scalafmt is smart enough to understand what I want to do on its own 😱

project/BuildPlugin.scala
Duhemm: I changed it, but now it's a side-effecting setting which is not super cool either...

jvican: Oh damn... if it's side-effecting, it's better as a task. Can we have it as a cache that you only run once (with some kind of semaphore or sync utility) please? 🙏

project/ReleaseUtils.scala
jvican: We can keep it this way and I'll make sbt-dynver work and remove this piece of code in the future.

frontend/src/main/scala/bloop/Cli.scala
jvican: > It's not necessary for this PR, I didn't intend to keep this commit here...
The reason is that at the moment, calling bloop without any argument will print bloop: Command not found which is extraordinarily surprising for people trying bloop for the first time. I'm happy to remove it and put that in another PR (as I initially intended to) if you prefer.

jvican: Does this happen even with our latest changes in nailgun?

Issue #100, opened 2017-12-01
Unable to to pass option to bloop script


leszekgruchala:
MacOS 10.13.1
Zsh
Bloop version: `66018e0a`
```
➜  some-project git:(feature/some-branch) ✗ bloop projects
Projects loaded from '/Users/lgr/projects/some-project/.bloop-config':
 * core
 * core-test
 * root
 * root-test
 * web
 * web-test

➜  some-project git:(feature/some-branch) ✗ bloop compile -p root
Usage: bloop-ng.py [options] cmd arg1 arg2 ...

bloop-ng.py: error: no such option: -p

➜  some-project git:(feature/some-branch) ✗ bloop compile root
Required option --project / -p not specified

➜  some-project git:(feature/some-branch) ✗ bloop compile --project root
Usage: bloop-ng.py [options] cmd arg1 arg2 ...

bloop-ng.py: error: no such option: --project
```

Duhemm:
I think that `bloop -- compile -p root` should work as a workaround.

leszekgruchala:
That was quick! Thank you, helped.

But another thing, running tests does nothing
```
✗ bloop -- test -p web
✗ bloop -- test -p web-test
```
No output at all. Should I fill another issue?

Duhemm:
Yes please! There are some issues with messages being lost between the server and the client, that could also be the cause. In that case, restarting the server usually helps.

You can also add `--verbose` to test to have bloop show the tests that it discovered.

jvican:
What was the way we decided to solve this issue? Forwarding CLI arguments from the nailgun server to the underlying app?

olafurpg:
I think I fixed this with `alias bloop = ng bloop.Cli` @jvican

jvican:
I have the same alias and this problem still occurs in my machine.

Duhemm:
I have
```sh
alias bloop="~/.bloop/bloop-ng.py bloop.Cli --"
```

And commands such as 
```sh
bloop run -p blabla --args arg0 --args arg1
```

work fine.

jvican:
> alias bloop="~/.bloop/bloop-ng.py bloop.Cli --"
  
The `--` at the end does the trick. However, I wonder if we can fix this in a more fundamental way. I'd like to avoid the need of `--` altogether.

jvican:
[...]

PR #194, opened 2018-01-26
Change the `bloop` alias suggested in the installation instructions

propensive:
As suggested here,

  https://github.com/scalacenter/bloop/issues/100

Even if there's a better long-term solution, it would be good to merge
this so that nobody else hits the same problems I did.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ We suggest that you add the following to your shell configuration:
 
 ```sh
 export PATH="$PATH:~/.bloop"
-alias bloop="~/.bloop/bloop-ng.py bloop.Cli --"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,7 +55,7 @@ We suggest that you add the following to your shell configuration:
 
 ```sh
 export PATH="$PATH:~/.bloop"
-alias bloop="bloop-ng.py"
+alias bloop="~/.bloop/bloop-ng.py bloop.Cli --"
```
Duhemm: Could you please remove the prefixing `~/.bloop/`? Otherwise it'd break for people that install the client on their `$PATH` but not in `~/.bloop`.

PR #200, opened 2018-01-26
Correct and improve installation instructions

propensive:
Check what you think about the updated installation instructions, which should now be a bit shorter and require less human interaction.

This patch also makes the installation script executable, which it wasn't before.


README.md
```diff
@@ -35,14 +35,10 @@ sbt:
...
-> frontend/version # copy this version number
-> cd nailgun
-> git rev-parse HEAD # copy this commit SHA
-> cd ..
-# paste the version number and SHA obtained above in the following command:
-$ bin/install.py --dest $HOME/.bloop --nailgun <nailgun-commit-sha> --version <version>
+$ sbt install
+$ VERSION=$(sbt -no-colors frontend/version | tail -n1 | cut -d' ' -f2)
+$ NAILGUN=$(git --git-dir=nailgun/.git rev-parse HEAD)
+$ bin/install.py --dest $HOME/.bloop --nailgun $NAILGUN --version $VERSION
```
Duhemm: We actually have git integration inside sbt with sbt-git 😄, so it should have been fine.

Unfortunately, this will not work on Windows (we work had to support it, that's why we re-wrote the installation script in Python).

Maybe that we could keep the current instruction, and add another snippet saying something like:

> If your shell supports it (i.e. you don't use Windows), you can also use this one liner which does the same thing:
> ```sh
> $ sbt install && bin/install.py --dest $HOME/.bloop --nailgun $(git --git-dir=nailgun/.git rev-parse HEAD) --version $(sbt -no-colors frontend/version | tail -n1 | cut -d' ' -f2)
> ```

(I may have screwed up my copy-pasting)

propensive: It might not be worth the effort of making the installation copy/pastable if it makes it harder to get it working on Windows...

If the `sbt install` command could automate the variable extraction (save them in `.nailgun_version` and `.frontend_version`, maybe?) that might make the installation easier... but maybe it's not worth the trouble of optimizing this if most people are going to be installing a published version soon?

jvican: What do we do with this PR? Is there something from here we can merge and that works for all of our users? @Duhemm

propensive: I'm not bothered if it's easier to close it. :) But the README still has `>`s where it should have `$`s for the prompt. It's a trivial fix, so just change that, and make the script executable. :)

jvican: I filed your advice in #225.

PR #210, opened 2018-01-29
Build PRs in the CI

Duhemm:
It looks like we were no longer building PRs, only cloning.


Duhemm:
The PR doesn't pick up my changes to `.drone.yml`. Does it use the `.drone.yml` that's on master? Or maybe my fix doesn't work at all?

PR #215, opened 2018-01-30
Write config to .bloop-config in integration tests

Duhemm:
I don't remember why we don't write them to the default location. It just makes it annoying when testing with Nailgun afterwards.


Duhemm:
test performance please

bloopoid:

bloopoid:

PR #216, opened 2018-01-30
Trigger Jenkins job in Drone CI

Duhemm:


Duhemm:
@jvican OMG it's green 😮

backend/src/main/scala/bloop/logging/ProcessLogger.scala
```diff
@@ -41,15 +41,24 @@ object ProcessLogger {
     val outputStream = new OutputStream {
       private[this] var dirty = false
       private[this] val buffer = new ByteArrayOutputStream
+      private[this] var sawR = false
 
       override def write(x: Int): Unit = synchronized {
         dirty = true
-        if (x == '\n') flush()
-        else buffer.write(x)
+        if (x == '\r') sawR = true
```
jvican: How will this affect Windows users when they run and test in Windows machines? I wonder if it wouldn't be better to enable this only for our CI? I'm afraid about the implications of this change regarding correctness and performance (running a verbose application with tons of logs).

bin/start-jenkins-job.sh
```diff
@@ -0,0 +1,15 @@
...
+JENKINS_AUTH=$(echo -n "$1" | base64)
+LENGTH=$(echo -n "$1" | wc -c)
+BROKEN=$(echo -n "bloop:" | wc -c)
+
+if [ $LENGTH = $BROKEN ]; then
+    echo "No token..."
+    exit 1
+fi
+
+curl -X POST \
```
jvican: :tada: This is great work.

.drone.yml
```diff
@@ -56,6 +57,16 @@ pipeline:
       - ${SBT_RUN}
       - ./bin/ci-clean-cache.sh
 
+  build_windows:
```
jvican: This is currently a Drone CI step, I wonder if it would be better to run it in parallel. http://docs.drone.io/pipelines/

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
```diff
@@ -54,6 +54,7 @@ class BspProtocolSpec {
     test(10)
   }
 
+  @Ignore
```
jvican: Can you open ticket about this and assign it to me?

frontend/src/test/scala/bloop/tasks/ProjectHelpers.scala
```diff
@@ -216,7 +217,8 @@ object ProjectHelpers {
         }
 
         override def postVisitDirectory(dir: Path, exc: IOException): FileVisitResult = {
-          Files.delete(dir)
+          try Files.delete(dir)
+          catch { case _: DirectoryNotEmptyException => () } // Happens sometimes on Windows?
```
jvican: This makes me uncomfortable... I don't want bloop become sbt in being super defensive on IO because of Windows support. Do you know exactly why this happens and in which scenario?

project/BuildPlugin.scala
```diff
@@ -373,7 +375,8 @@ object BuildImplementation {
       val indexProperty = s"-Dbloop.integrations.index=${buildIndexFile.getAbsolutePath}"
       val properties = stagingProperty :: indexProperty :: pluginsProperty :: settingsProperty :: Nil
       val toRun = "installBloop" :: "buildIndex" :: Nil
-      val cmd = "sbt" :: (properties ++ toRun)
+      val cmdBase = if (isWindows) "cmd.exe" :: "/C" :: "sbt.bat" :: Nil else "sbt" :: Nil
```
jvican: Sigh, Windows.

.drone.yml
Duhemm: This runs concurrently with the `build` step, is that what you wanted?

frontend/src/test/scala/bloop/tasks/ProjectHelpers.scala
Duhemm: It happens in the file watching tests, but I don't know anything more...

project/BuildPlugin.scala
Duhemm: 💣 🔫 🔥 🦀

backend/src/main/scala/bloop/logging/ProcessLogger.scala
Duhemm: I'll experiment more today and try to remove it, but it seemed to me this was required for having the `ProcessLogger` work on Windows...

Duhemm: The question is, what message should the logger receive when someone does:
`System.out.println("foo")` ?
Should it contain `foo` (as it is currently in this PR), or
`foo\r` (as it is in `master`)

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
Duhemm: See #281

backend/src/main/scala/bloop/logging/ProcessLogger.scala
Duhemm: Fixed as discussed today. We just don't care about `\r` in the tests.

frontend/src/test/scala/bloop/tasks/ProjectHelpers.scala
Duhemm: I'll look a bit further tomorrow, but I don't really know what to do about that. I strongly suspect that Windows thinks that the directory is not empty because of the file watching, but I can't tell for sure.

Note that this is only in test code, when the test is done and everything has run smoothly.

jvican:
Seems reliable! Let's merge :smile:.

Duhemm:
@jvican Do you want to take a last look before ewe merge?

PR #222, opened 2018-01-31
Squash all history of our integration projects

Duhemm:
We really don't care about their history. It just makes it awfully slow
to clone our repo, which is bad for CI and for external contributors.

The orphaned commits have been created with `git checkout --orphan`.


Duhemm:
I've squashed them all! 🔨

PR #223, opened 2018-01-31
Add coursier PLEASE

jvican:
I invoke the god of coursier to make my CI faster.

Before we were blocked by
https://github.com/coursier/coursier/issues/388, but we work around that
issue by adding the coursier plugin to the global plugins directory so
that all the sbt builds get it.

### DIFF
--- a/bin/add-coursier.sh
+++ b/bin/add-coursier.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -eu
+set -o nounset
+
+PLUGINS_DIR="$DRONE_DIR/.sbt/1.0/plugins"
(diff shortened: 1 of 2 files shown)


bin/add-coursier.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -eu
+set -o nounset
+
+PLUGINS_FILE="/home/jvican/.sbt/1.0/plugins/plugins.sbt"
```
Duhemm: How will drone pick that up? Shouldn't it be in `/root/.sbt/...`?

jvican: JEEZ, forgot to update it.

PR #228, opened 2018-02-02
Add command to release all source modules

jvican:


.drone.yml
```diff
@@ -11,9 +11,10 @@ matrix:
           "frontend/test"
 
   SBT_PUBLISH:
+    # There is a bug in sbt that ignores these settings when defined in `.sbt` files and plugins :(
     - sbt "set pgpPublicRing in Global := file(\"/drone/.gnupg/pubring.asc\")" \
           "set pgpSecretRing in Global := file(\"/drone/.gnupg/secring.asc\")" \
-          ";backend/releaseEarly;frontend/releaseEarly;^sbtBloop/releaseEarly"
+          "releaseBloop"
```
Duhemm: That should probably be `publishBloop` rather than `releaseBloop` (it's part of the `publish` step, no `release`)

jvican: But it invokes `releaseEarly` in all the projects, so calling it `publish` is a misnomer :smile:. I should have called `sbt-release-early` differently, maybe `sbt-publish-early`. Are you happy if I call it `releaseBloopModules`?

Duhemm: We can also keep it that way, it's just a nitpick.

PR #232, opened 2018-02-02
Use bintray to release everything

jvican:
Bintray is so much nicer and faster. The slugishness of sbt-sonatype is
unbearable. We still publish to sonatype the sbt-bloop plugin because Bintray
does not allow to sync to Maven Central repositories that use the sbt plugin
repository layout.


jvican:
It goes without saying we're syncing up with Maven Central in the bloop, nailgun, bsp and zinc repositories in our organization: https://bintray.com/scalacenter. I've asked for these packages to be synced up with Maven Central, and I'll get response from Bintray soon.
  
When we do, we'll cut an official release. Until then, we can cut on merge releases :smile:.

PR #237, opened 2018-02-06
Update to latest zinc

Duhemm:
I've rebased on the 1.x branch of Zinc. The build is certainly a lot less elegant than it was before, but I've tried to minimise the diff with upstream Zinc.

### DIFF
--- a/project/BuildPlugin.scala
+++ b/project/BuildPlugin.scala
@@ -144,6 +144,9 @@ object BuildKeys {
       )
     }
   )
+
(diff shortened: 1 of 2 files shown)


Duhemm:
I've fixed a mistake I introduced during the rebase, I was not publishing `compiler-interface` when publishing the compiler bridge. It works now, and should be easy to rebase.

jvican:
Closing in favor of #242.

PR #238, opened 2018-02-06
Publish website on merge

Duhemm:


jvican:
Can you look into why the CI fails?

Duhemm:
It fails because hugo is not installed in the CI. We need to upgrade the docker image.

Duhemm:
Yay, it works, we just need to publish the real docker image (I've tried here with the image published under my name)

Duhemm:
This should be ready for review @jvican :)

jvican:
Feel free to merge when CI passes in last commit.

PR #239, opened 2018-02-06
Add scalameta bintray repo in installation script

Duhemm:
This is the repository where lsp4s lives. We need to tell coursier to
use it, otherwise we won't find this artifact.

Also, make the installation script executable.


olafurpg:
Whoops, we will likely cut a v0.1 release of metals soon. I'm tempted to do the release on Feb 13th at the Lausanne meetup

PR #241, opened 2018-02-08
Setup SSH key to push to github pages

Duhemm:


.drone.yml
```diff
@@ -75,6 +75,11 @@ pipeline:
       - export DRONE_DIR="/drone"
       - ./bin/add-coursier.sh
       - /usr/local/bin/setup_keys.sh
+      # Setup everything to be able to push to github pages
```
olafurpg: I recommend moving these sort of things into bash scripts, large .travis/.drone.yml files are a code smell IMO

Duhemm:
Thanks @olafurpg, that's a good point. I would have been faster if I hadn't run it on my machine 😭

PR #242, opened 2018-02-08
End the pain

jvican:


jvican:
A gist of what's happened here: I thought I would be able to outsmart sbt, but my sbt foo was useless, he won. I just give up, I want to be a normal sbt user again.
  
This pull request removes the source deps logic, moves some of that logic to the zinc and bsp repositories, and uses a binary dependency for them. It then moves the release logic to the nailgun repository too. It builds upon #237.

project/BuildPlugin.scala
```diff
@@ -186,47 +177,43 @@ object BuildImplementation {
...
 
   final val buildSettings: Seq[Def.Setting[_]] = Seq(
     Keys.organization := "ch.epfl.scala",
     Keys.updateOptions := Keys.updateOptions.value.withCachedResolution(true),
     Keys.scalaVersion := "2.12.4",
     Keys.triggeredMessage := Watched.clearWhenTriggered,
+    Keys.resolvers := {
+      val oldResolvers = Keys.resolvers.value
+      val sonatypeStaging = Resolver.sonatypeRepo("staging")
+      val scalametaResolver = Resolver.bintrayRepo("scalameta", "maven")
```
jvican: Let's wait for lsp4s to be published to Central before removing this.

Duhemm:
Why can't it find the compiler bridge?

Duhemm:
When this is green, we should also probably run the benchmarks to make sure nothing breaks here too.

Duhemm:
The benchmarks run fine on my machine, let's merge that!

PR #247, opened 2018-02-09
Add a random change to try the community build

jvican:


jvican:
The check for the community build now works, the only issue is that our bloop config file seems to be exporting the `test` project in the scalac build and compiling it, which obviously makes it fail. I'll fix this on Monday.

jvican:
By the way, when compiling the whole community build we get this beautiful error:
  
```
Java HotSpot(TM) 64-Bit Server VM warning: CodeCache is full. Compiler has been disabled.
Java HotSpot(TM) 64-Bit Server VM warning: Try increasing the code cache size using -XX:ReservedCodeCacheSize=
CodeCache: size=245760Kb used=243305Kb max_used=243346Kb free=2455Kb
 bounds [0x00007f2f41000000, 0x00007f2f50000000, 0x00007f2f50000000]
 total_blobs=61394 nmethods=60717 adapters=575
compilation: disabled (not enough contiguous free space left)
```
  
Solutions:
  
1. We can change the bloop instance compiling the community build from time to time.
2. We can increase the code cache by a lot so that C2 can compile all the projects with ease.

Duhemm:
@jvican I've cherry-picked your fix in #110. I think that the problem you mention with scala has been fixed in #272.

jvican:
Can we close this PR? It seems it's useless now!

Issue #254, opened 2018-02-12
Working directory for bloop run is root


olafurpg:
I installed bloop via homebrew and the working directory with `bloop run` seems to be root `/`

```
$ bloop run -p vork --args --in --args docs
error: File /docs does not exist.
```

Adding `$(pwd)/` fixes the problem
```
$ bloop run -p vork --args --in --args $(pwd)/docs
```

I would expect bloop to use the working directory from where the nailgun client is invoked, which might require forking. Not sure if setting `sys.props("user.dir")` could work

Duhemm:
Thanks for the report!

This is easy to fix when `fork = true`, but it's going to be a bit more involved when running in the same JVM. We'll probably do as we did with the standard streams.

PR #260, opened 2018-02-13
Remove in process run and test

### LINKED ISSUE #249: Fix spurious error in TestLoggingSpec
```
[error] Test bloop.tasks.TestLoggingSpec.concurrentInProcessTestRunsHaveDifferentStreams failed: java.lang.AssertionError: expected:<10> but was:<0>, took 3.117 sec
[error] at bloop.tasks.TestLoggingSpec.concurrentInProcessTestRunsHaveDifferentStreams(TestLoggingSpec.scala:50)
```
  
It seems this error happens quite often on the CI. @Duhemm Could you look into it? I think you're the author of this test.

Duhemm:
Supporting in process run and test adds a lot of unnecessary complexity. It is indeed slightly faster to be able to run without starting a JVM, but that means that we need to replace the standard streams and highjack the system properties (plus many other things that I may not have thought about).

I can't think of a situation where running in-process is a necessity. I think it just adds a lot of complexity without bringing anything to the table.

This PR removes in process run and test and always forks. Also, it sets the current working directory in the forked JVM to the user's CWD.

Fixes #249, #254


olafurpg:
> This PR removes in process run and test and always forks. Also, it sets the current working directory in the forked JVM to the user's CWD.

To strictly follow sbt semantics the cwd should match the project sourceDirectory if `fork:=true`. I personally dislike that behavior of sbt, but I suspect many builds are organized around it.

jvican:
@Duhemm Can you rebase and I review? I'd like to merge this one soon :smile:, it's been sitting on the queue for too long.

Duhemm:
Rebased and ready for review

PR #264, opened 2018-02-13
Reload projects when configurations change

Duhemm:
I've quickly run some benchmarks with hyperfine using scalameta/scalameta as example:

This is Bloop 1.0.0-M1:
```
Benchmark #1: bloop projects

  Time (mean ± σ):      74.0 ms ±   2.3 ms    [User: 35.7 ms, System: 31.9 ms]
 
  Range (min … max):    69.0 ms …  78.4 ms
 ```

This is this PR:
```
Benchmark #1: bloop projects

  Time (mean ± σ):      77.1 ms ±   2.0 ms    [User: 37.1 ms, System: 32.9 ms]
 
  Range (min … max):    73.0 ms …  80.5 ms
 
Benchmark #2: touch .bloop-config/benchmarks.config && bloop projects

  Time (mean ± σ):     173.0 ms ± 204.1 ms    [User: 38.0 ms, System: 33.5 ms]
 
  Range (min … max):    76.6 ms … 630.4 ms
 
Benchmark #3: echo '#foo' >> .bloop-config/benchmarks.config && bloop projects

  Time (mean ± σ):      1.210 s ±  0.064 s    [User: 38.0 ms, System: 34.6 ms]
 
  Range (min … max):    1.168 s …  1.385 s
 ```

### DIFF
--- a/frontend/src/main/scala/bloop/Project.scala
+++ b/frontend/src/main/scala/bloop/Project.scala
@@ -51,10 +51,21 @@ case class Project(name: String,
 }
 
 object Project {
+
+  /** The pattern used to find configuration files */
+  final val loadPattern: String = "glob:**.config"
+
+  /**
+   * Load all the projects from `config`.
+   *
+   * @param config The base directory from which to load the projects.
+   * @param logger The logger that collects messages about project loading.
+   * @return The list of loaded projects.
+   */
   def fromDir(config: AbsolutePath, logger: Logger): List[Project] = {
     timed(logger) {
       // TODO: We're not handling projects with duplicated names here.
-      val configFiles = Paths.getAll(config, "glob:**.config")
+      val configFiles = Paths.getAll(config, loadPattern)
       logger.debug(s"Loading ${configFiles.length} projects from '${config.syntax}'...")
       configFiles.par.map(configFile => fromFile(configFile, logger)).toList
     }
--- a/frontend/src/main/scala/bloop/engine/Build.scala
+++ b/frontend/src/main/scala/bloop/engine/Build.scala
@@ -3,9 +3,10 @@ package bloop.engine
 import bloop.Project
 import bloop.io.{AbsolutePath, DirChecksum}
 
-final case class Build(origin: AbsolutePath, projects: List[Project]) {
+final case class Build private (origin: AbsolutePath,
+                                projects: List[Project],
+                                originChecksum: DirChecksum) {
 
-  private val originChecksum = DirChecksum(origin, Project.loadPattern)
   private val stringToProjects: Map[String, Project] = projects.map(p => p.name -> p).toMap
   private[bloop] val dags: List[Dag[Project]] = Dag.fromMap(stringToProjects)
 
@@ -14,6 +15,13 @@ final case class Build(origin: AbsolutePath, projects: List[Project]) {
     Dag.dagFor(dags, project).getOrElse(sys.error(s"Project $project does not have a DAG!"))
 
   /** Has this build definition changed since it was loaded? */
-  def changed(): Boolean = originChecksum.changed()
+  def changed(): DirChecksum.DirStatus = originChecksum.changed()
 
+}
+
+object Build {
+  def apply(origin: AbsolutePath, projects: List[Project]): Build = {
+    val checksum = DirChecksum(origin, Project.loadPattern)
+    new Build(origin, projects, checksum)
+  }
 }
--- a/frontend/src/main/scala/bloop/engine/Build.scala
+++ b/frontend/src/main/scala/bloop/engine/Build.scala
@@ -1,13 +1,19 @@
 package bloop.engine
 
 import bloop.Project
-import bloop.io.AbsolutePath
+import bloop.io.{AbsolutePath, DirChecksum}
 
 final case class Build(origin: AbsolutePath, projects: List[Project]) {
+
+  private val originChecksum = DirChecksum(origin, Project.loadPattern)
   private val stringToProjects: Map[String, Project] = projects.map(p => p.name -> p).toMap
   private[bloop] val dags: List[Dag[Project]] = Dag.fromMap(stringToProjects)
 
   def getProjectFor(name: String): Option[Project] = stringToProjects.get(name)
   def getDagFor(project: Project): Dag[Project] =
     Dag.dagFor(dags, project).getOrElse(sys.error(s"Project $project does not have a DAG!"))
+
+  /** Has this build definition changed since it was loaded? */
+  def changed(): Boolean = originChecksum.changed()
+
 }
--- a/frontend/src/main/scala/bloop/engine/Build.scala
+++ b/frontend/src/main/scala/bloop/engine/Build.scala
@@ -2,6 +2,7 @@ package bloop.engine
 
 import bloop.Project
 import bloop.io.{AbsolutePath, FileTracker}
+import bloop.logging.Logger
 
 final case class Build private (origin: AbsolutePath,
                                 projects: List[Project],
@@ -14,8 +15,13 @@ final case class Build private (origin: AbsolutePath,
   def getDagFor(project: Project): Dag[Project] =
     Dag.dagFor(dags, project).getOrElse(sys.error(s"Project $project does not have a DAG!"))
 
-  /** Has this build definition changed since it was loaded? */
-  def changed(): FileTracker.Status = originChecksum.changed()
+  /**
+   * Has this build definition changed since it was loaded?
+   *
+   * @param logger A logger that receives errors, if any.
+   * @return The status of the directory from which the build was loaded.
+   */
+  def changed(logger: Logger): FileTracker.Status = originChecksum.changed(logger)
 
 }
 
--- a/frontend/src/main/scala/bloop/engine/Build.scala
+++ b/frontend/src/main/scala/bloop/engine/Build.scala
@@ -1,11 +1,11 @@
 package bloop.engine
 
 import bloop.Project
-import bloop.io.{AbsolutePath, DirChecksum}
+import bloop.io.{AbsolutePath, FileTracker}
 
 final case class Build private (origin: AbsolutePath,
                                 projects: List[Project],
-                                originChecksum: DirChecksum) {
+                                originChecksum: FileTracker) {
 
   private val stringToProjects: Map[String, Project] = projects.map(p => p.name -> p).toMap
   private[bloop] val dags: List[Dag[Project]] = Dag.fromMap(stringToProjects)
@@ -15,13 +15,13 @@ final case class Build private (origin: AbsolutePath,
     Dag.dagFor(dags, project).getOrElse(sys.error(s"Project $project does not have a DAG!"))
 
   /** Has this build definition changed since it was loaded? */
-  def changed(): DirChecksum.DirStatus = originChecksum.changed()
+  def changed(): FileTracker.Status = originChecksum.changed()
 
 }
 
 object Build {
   def apply(origin: AbsolutePath, projects: List[Project]): Build = {
-    val checksum = DirChecksum(origin, Project.loadPattern)
+    val checksum = FileTracker(origin, Project.loadPattern)
     new Build(origin, projects, checksum)
   }
 }
--- a/frontend/src/main/scala/bloop/io/DirChecksum.scala
+++ b/frontend/src/main/scala/bloop/io/DirChecksum.scala
@@ -16,7 +16,7 @@ import java.util.zip.{Adler32, CheckedInputStream}
  */
 final class DirChecksum(dir: AbsolutePath,
                         pattern: String,
-                        private[this] var modifiedTimes: List[(AbsolutePath, FileTime)],
+                        modifiedTimes: List[(AbsolutePath, FileTime)],
                         contentsChecksum: Long) {
 
   /**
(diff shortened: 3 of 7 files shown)


frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,92 @@
...
+ * Represents the state of a directory and some files in which we track changes.
+ *
+ * @param dir           The directory to track.
+ * @param pattern       The pattern matching the files that must be tracked inside `dir`.
+ * @param modifiedTimes The last modification time of the tracked files.
+ * @param contentsChecksum The checksum of all the contents of the directory.
+ */
+final class DirChecksum(dir: AbsolutePath,
+                        pattern: String,
+                        private[this] var modifiedTimes: List[(AbsolutePath, FileTime)],
```
jvican: Why is this `private[this]`?

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Inspects the directory for changes.
+   *
+   * @return true if the directory has changed, false otherwise.
+   */
+  def changed(): Boolean = {
+    val newModifiedTimes = DirChecksum.getFiles(dir, pattern)
+    if (newModifiedTimes == modifiedTimes) false
+    else
```
jvican: Curly braces around?

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,92 @@
...
+      synchronized {
```
jvican: Is synchronizing here working? We're only syncing on the instance of `DirChecksum`.

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,92 @@
...
+        val newChecksum = DirChecksum.filesChecksum(newModifiedTimes.map(_._1))
+        if (newChecksum != contentsChecksum) true
+        else {
+          modifiedTimes = newModifiedTimes
```
jvican: Can we not make all this thing immutable? Easier to reason about it.

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Returns all the tracked files inside this directory, associated with their last
+   * modification time.
+   *
+   * @param dir     The directory to track.
+   * @param pattern The pattern to find the files to track.
+   * @return A map associating each tracked file with its last modification time.
+   */
+  private def getFiles(dir: AbsolutePath, pattern: String): List[(AbsolutePath, FileTime)] =
```
jvican: Curly braces around?

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,92 @@
...
+  private def filesChecksum(files: List[AbsolutePath]): Long = {
+    val streams = new Enumeration[InputStream] {
+      val streams = files.map(f => Files.newInputStream(f.underlying)).toIterator
+      override def hasMoreElements(): Boolean = streams.hasNext
+      override def nextElement(): InputStream = streams.next()
+    }
+
+    val checkedStream = new CheckedInputStream(new SequenceInputStream(streams), new Adler32)
+    try {
+      val buf = new Array[Byte](4096)
```
jvican: Not used?

frontend/src/main/scala/bloop/Project.scala
```diff
@@ -52,10 +52,21 @@ case class Project(name: String,
 }
 
 object Project {
+
+  /** The pattern used to find configuration files */
+  final val loadPattern: String = "glob:**.config"
```
jvican: We should change the `.config` ending at some point and favor something more bloop specific.

frontend/src/main/scala/bloop/engine/Build.scala
```diff
@@ -1,13 +1,19 @@
 package bloop.engine
 
 import bloop.Project
-import bloop.io.AbsolutePath
+import bloop.io.{AbsolutePath, DirChecksum}
 
 final case class Build(origin: AbsolutePath, projects: List[Project]) {
+
+  private val originChecksum = DirChecksum(origin, Project.loadPattern)
```
jvican: Following up on the previous comment on mutability, I think it would be better to have the checksums live in the state itself and be independent of this data structure.

Duhemm: What are your arguments for that? `Build` is just a wrapper around the list of projects, so I think it makes more sense to keep here the information needed to tell whether the `Build` is outdated or not.

frontend/src/main/scala/bloop/engine/Build.scala
```diff
@@ -1,13 +1,27 @@
...
-final case class Build(origin: AbsolutePath, projects: List[Project]) {
   private val stringToProjects: Map[String, Project] = projects.map(p => p.name -> p).toMap
   private[bloop] val dags: List[Dag[Project]] = Dag.fromMap(stringToProjects)
 
   def getProjectFor(name: String): Option[Project] = stringToProjects.get(name)
   def getDagFor(project: Project): Dag[Project] =
     Dag.dagFor(dags, project).getOrElse(sys.error(s"Project $project does not have a DAG!"))
+
+  /** Has this build definition changed since it was loaded? */
+  def changed(): DirChecksum.DirStatus = originChecksum.changed()
```
jvican: No parens for `changed` in the method definition?

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,106 @@
...
+                        pattern: String,
+                        modifiedTimes: List[(AbsolutePath, FileTime)],
+                        contentsChecksum: Long) {
+
+  /**
+   * Inspects the directory for changes.
+   *
+   * @return true if the directory has changed, false otherwise.
+   */
+  def changed(): DirChecksum.DirStatus = {
```
jvican: No parens?

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,106 @@
...
+ * Represents the state of a directory and some files in which we track changes.
+ *
+ * @param dir           The directory to track.
+ * @param pattern       The pattern matching the files that must be tracked inside `dir`.
+ * @param modifiedTimes The last modification time of the tracked files.
+ * @param contentsChecksum The checksum of all the contents of the directory.
+ */
+final class DirChecksum(dir: AbsolutePath,
+                        pattern: String,
+                        modifiedTimes: List[(AbsolutePath, FileTime)],
```
jvican: `lastModifiedTimes`?

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Represents the state of a directory and some files in which we track changes.
+ *
+ * @param dir           The directory to track.
+ * @param pattern       The pattern matching the files that must be tracked inside `dir`.
+ * @param modifiedTimes The last modification time of the tracked files.
+ * @param contentsChecksum The checksum of all the contents of the directory.
+ */
+final class DirChecksum(dir: AbsolutePath,
```
jvican: What about calling this `BuildSignature` or `BuildChecksum` instead of `DirChecksum`?

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,106 @@
...
+                        pattern: String,
+                        modifiedTimes: List[(AbsolutePath, FileTime)],
+                        contentsChecksum: Long) {
```
jvican: `checksum` alone?

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Inspects the directory for changes.
+   *
+   * @return true if the directory has changed, false otherwise.
+   */
+  def changed(): DirChecksum.DirStatus = {
+    val newModifiedTimes = DirChecksum.getFiles(dir, pattern)
+    if (newModifiedTimes == modifiedTimes) DirChecksum.DirUnchanged(None)
```
jvican: Can we make `DirChecksum.DirUnchanged` a final val in `DirUnchanged`s companion?

frontend/src/main/scala/bloop/io/DirChecksum.scala
jvican: Also, we may want to abstract over `(AbsolutePath, FileTime)` with a `FileSignature` or `FileState` class.

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,106 @@
...
+   * modification time.
+   *
+   * @param dir     The directory to track.
+   * @param pattern The pattern to find the files to track.
+   * @return A map associating each tracked file with its last modification time.
+   */
+  private def getFiles(dir: AbsolutePath, pattern: String): List[(AbsolutePath, FileTime)] = {
+    Paths
+      .getAll(dir, pattern)
+      .map { path =>
```
jvican: Can we keep this in one line?

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,106 @@
...
+   * @return The checksum.
+   */
+  private def filesChecksum(files: List[AbsolutePath]): Long = {
+    val streams = new Enumeration[InputStream] {
+      val streams = files.map(f => Files.newInputStream(f.underlying)).toIterator
+      override def hasMoreElements(): Boolean = streams.hasNext
+      override def nextElement(): InputStream = streams.next()
+    }
+
+    val checkedStream = new CheckedInputStream(new SequenceInputStream(streams), new Adler32)
```
jvican: Can you elaborate on the contract of `CheckedInputStream` and `SequenceInputStream`? Do they immediately create a signature out of an input stream given a concrete algorithm? Is there any detail we should be particularly wary of?

frontend/src/main/scala/bloop/io/DirChecksum.scala
```diff
@@ -0,0 +1,106 @@
...
+    try {
```
jvican: If something here fails, can we report the error in a nice way and assume the worst (e.g. the build has changed)?

frontend/src/main/scala/bloop/engine/Build.scala
```diff
@@ -1,13 +1,27 @@
 package bloop.engine
 
 import bloop.Project
-import bloop.io.AbsolutePath
+import bloop.io.{AbsolutePath, DirChecksum}
+
+final case class Build private (origin: AbsolutePath,
+                                projects: List[Project],
+                                originChecksum: DirChecksum) {
```
jvican: `DirChecksum` is not a case class, contrary to what `Build` is. That means that comparing builds that may represent the same build will always yield false (because of object equality in `DirChecksum`). Should we make `DirChecksum` a case class too? What are the disadvantages of doing that?

frontend/src/main/scala/bloop/engine/Build.scala
jvican: As I see it, we have `Build` and then we have the state of `Build`. Their representation should be separate so that we can reason about it better, and so that we can update or check its state without necessarily updating the build.

frontend/src/main/scala/bloop/engine/Build.scala
Duhemm: I've put the parentheses because `changed` isn't pure

frontend/src/main/scala/bloop/io/DirChecksum.scala
Duhemm: It's not really related to `Build`, we just happen to use it for `Build`?

frontend/src/main/scala/bloop/engine/Build.scala
Duhemm: Making `DirChecksum` a case class sounds good.

frontend/src/main/scala/bloop/io/DirChecksum.scala
Duhemm: > Can you elaborate on the contract of CheckedInputStream and SequenceInputStream?

`CheckedInputStream` keeps an internal value that represents the checksum and that is updated as the wrapped stream is read through. `SequenceInputStream` reads several streams back to back. Closing the `CheckedInputStream` closes the underlying stream, and closing the `SequenceInputStream` closes all the underlying streams.

> Do they immediately create a signature out of an input stream given a concrete algorithm?

Exactly.

> Is there any detail we should be particularly wary of?

None that I know of.

frontend/src/main/scala/bloop/io/DirChecksum.scala
Duhemm: Done

frontend/src/main/scala/bloop/io/DirChecksum.scala
Duhemm: Not sure what you mean here? `DirUnchanged` is a case class. Did you mean `DirChanged`?

frontend/src/main/scala/bloop/io/DirChecksum.scala
Duhemm: We agreed on `FileTracker`

frontend/src/main/scala/bloop/io/DirChecksum.scala
Duhemm: I'm not sure it's worth it right now, maybe in the future.

frontend/src/main/scala/bloop/io/DirChecksum.scala
Duhemm: Sounds good

frontend/src/main/scala/bloop/io/DirChecksum.scala
Duhemm: Done

frontend/src/main/scala/bloop/io/FileTracker.scala
```diff
@@ -0,0 +1,119 @@
...
+        val newChecksum = FileTracker.filesChecksum(newModifiedTimes.map(_._1))
+        if (newChecksum != checksum) FileTracker.Changed
+        else {
+          val checksum = new FileTracker(base, pattern, newModifiedTimes, newChecksum)
+          FileTracker.Unchanged(Some(checksum))
+        }
+      } catch {
+        case NonFatal(e) =>
+          logger.error(
+            s"An error occurred while checking the build for changes; assuming it changed.")
```
jvican: :tada:

PR #265, opened 2018-02-13
Better shortening of path in compilation errors

### LINKED ISSUE #263: File paths include leading slash /
```
Obtained
[E] [E1] /src/test/scala/scala/meta/tests/semanticdb/XXX.scala

Expected
[E] [E1] src/test/scala/scala/meta/tests/semanticdb/XXX.scala
```
The leading slash makes it impossible to cmd+click on the path to open in the editor.

Duhemm:
Also, relativize the compilation error messages to the base of the build
instead of the base of the project.

Fixes #263

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
@@ -49,7 +49,8 @@ object Tasks {
       val target = project.tmp
       val scalacOptions = project.scalacOptions
       val javacOptions = project.javacOptions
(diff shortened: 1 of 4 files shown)


jvican:
This looks great. Can we do like http://ensime.github.io/editors/emacs/sbt-mode/#related-customizations and keep @fommil happy?

fommil:
i.e. output full paths, with file / column information exactly like scalac, and let text editors do the shortening.

Duhemm:
> i.e. output full paths, with file / column information exactly like scalac, and let text editors do the shortening.

Isn't that what `--reporter scalac` does?

PR #276, opened 2018-02-18
fix install commands (exit sbt shell)

rethab:


rethab:
whoopsy, duplicates #275

PR #277, opened 2018-02-19
Fix commit sha for guardian/frontend

Duhemm:
Currently, sbt fails to load the project because of a bug in either
coursier or sbt. This changes the sha to apply a workaround.

### DIFF
--- a/build-integrations/sbt-1.0/build.sbt
+++ b/build-integrations/sbt-1.0/build.sbt
@@ -1,7 +1,7 @@
 val SbtSbt = RootProject(
   uri("git://github.com/scalacenter/sbt.git#c84db3a3969e269aeed0e9d9f2f384a0b029b82c"))
 val GuardianFrontend = RootProject(
-  uri("git://github.com/guardian/frontend.git#c6611d93635133a8a0282c4ffc8d304effdca374"))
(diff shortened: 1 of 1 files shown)


Duhemm:
I'm running the benchmarks on my machine. I really hope that's the last fix...

build-integrations/sbt-1.0/build.sbt
```diff
@@ -1,7 +1,7 @@
 val SbtSbt = RootProject(
   uri("git://github.com/scalacenter/sbt.git#c84db3a3969e269aeed0e9d9f2f384a0b029b82c"))
 val GuardianFrontend = RootProject(
-  uri("git://github.com/guardian/frontend.git#c6611d93635133a8a0282c4ffc8d304effdca374"))
+  uri("git://github.com/scalacenter/frontend.git#fd8da1929d8a3bd39ca6027ffba6c0850e036ce3"))
```
Duhemm: This is not a revert to the previous commit that we had, I just added this workaround for the issue that we had: https://github.com/coursier/coursier/issues/450#issuecomment-302267082

jvican: Sigh. Yeah, sourcedeps in sbt should work.

PR #288, opened 2018-02-21
Add command reference

Duhemm:


frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -63,10 +63,10 @@ object Commands {
 
   @CommandName("bsp")
   case class Bsp(
-      @ExtraName("h")
```
jvican: Thanks for fixing this, it was an oversight.

PR #291, opened 2018-02-26
Add support for `testOptions`

Duhemm:
Ignore all the bad serialisation stuff for now. It'll be better when we have a saner format for configuration.


jvican:
Which issue is this fixing?

Duhemm:
For instance, JUnit is logging its stuff at the `debug` level, meaning that we won't see any output unless we set the good test arguments.

jvican:
So, this is all about the test options that we pass in to every test framework. What happens if these test options are framework specific in a project that has several frameworks enabled?

Duhemm:
Same as sbt. This is replicating sbt's behaviour. You can set options that must apply only to certain frameworks.

jvican:
I'm closing this pull request, I'll create a follow-up including these commits after the redesign of the configuration format is done.

PR #305, opened 2018-03-06
Fix regression in parallel compilation

jvican:
See first commit for more information.

### DIFF
--- a/bin/install.py
+++ b/bin/install.py
@@ -49,11 +49,9 @@ NAILGUN_COMMIT = args.nailgun
 NAILGUN_CLIENT_URL = "https://raw.githubusercontent.com/scalacenter/nailgun/%s/pynailgun/ng.py" % NAILGUN_COMMIT
 BLOOP_COURSIER_TARGET = join(BLOOP_INSTALLATION_TARGET, "coursier")
 BLOOP_SERVER_TARGET = join(BLOOP_INSTALLATION_TARGET, "blp-server")
-BLOOP_SHELL_TARGET = join(BLOOP_INSTALLATION_TARGET, "blp-shell")
 BLOOP_CLIENT_TARGET = join(BLOOP_INSTALLATION_TARGET, "bloop")
 
 BLOOP_ARTIFACT = "ch.epfl.scala:bloop-frontend_2.12:%s" % BLOOP_VERSION
-
 
 def download_and_install(url, target):
     try:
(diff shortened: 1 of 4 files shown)


bin/install.py
```diff
@@ -47,7 +47,7 @@
 NAILGUN_COMMIT = args.nailgun
 
 NAILGUN_CLIENT_URL = "https://raw.githubusercontent.com/scalacenter/nailgun/%s/pynailgun/ng.py" % NAILGUN_COMMIT
-BLOOP_COURSIER_TARGET = join(BLOOP_INSTALLATION_TARGET, "coursier")
+BLOOP_COURSIER_TARGET = join(BLOOP_INSTALLATION_TARGET, "blp-coursier")
```
Duhemm: This change will break the homebrew formula: 

https://github.com/scalacenter/bloop/blob/6fa5009c41a5d8e35f927f02685e0e0075bc7009/project/ReleaseUtils.scala#L75

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
```diff
@@ -66,21 +66,25 @@ object Tasks {
...
-        visited.add(dag)
-        dag match {
-          case Leaf(project) => Task(List(compile(project)))
-          case Parent(project, dependencies) =>
-            val downstream = Task.wanderUnordered(dependencies)(compileTree).map(_.flatten)
-            downstream.flatMap { results =>
-              if (results.exists(_._2 == FailedResult)) Task.now(results)
-              else Task(compile(project)).map(r => r :: results)
-            }
+      val visited = scala.collection.mutable.HashSet.empty[Dag[Project]]
```
Duhemm: Why is commit 4cc1ad4c39770be29879a8b2884879078f22371b necessary? Looking at the diff it looks to me like it's only nested one level deeper without changing anything?

bin/install.py
jvican: Nice catch :+1:

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
jvican: It's better to scope `visited` in `compileTree` rather than the `compile` function so that it's idempotent. Otherwise adding printlns before the actual use of the function makes return the latter one an empty list.

olafurpg:
Cool! Any chance to give this a try without publishLocal?

Duhemm:
@olafurpg I think this should do the trick once publishing is done:

```sh
bin/install.py -v 1.0.0-M4+15-a56919c6 -n 12448b9b2c074f8048960ce87240a0dfb646303a
```

PR #310, opened 2018-03-06
Add search support for docs [DOCS]

### LINKED ISSUE #306: Add search box to docs
This needs to be ready before 1.0.0. I have signed up for Algolia's free service https://community.algolia.com/docsearch/, and they will index our page for free. All we need to do is to reuse the html utils we have from Kube to add search boxes and add some javascript.
  
This is blocked until Algolia's team gets back to me.

jvican:
Via Algolia's docsearch.

Fixes #306.

/cc @olafurpg


jvican:
bloop-search

jvican:
The only inconvenience of having Algolia index our website every 24h is that the search results redirect to the official website instead of the local copy. But setting this up was too easy, and fixing this issue would require we send to algolia the index every time we push, and that's just too time-consuming.

s-pace:
👋,

We can change this behaviour by forcing some links / avoid scraping some URLs (regex)

We can also diable this crawling every 24h.

Please provide me a sitemap and it will be 💯

jvican:
@s-pace Thanks for chiming in! I'll look into a way of providing you with a sitemap, being able to have search results relative to the base url would be huge! :top:
  
I'll let you know when I have all ready on my side.

s-pace:
@jvican Perfect. My pleasure!

Issue #154, opened 2018-01-12
Avoid "Required option --project / -p not specified"


olafurpg:
I ran
```
$ bloop compile a
Required option --project / -p not specified
```

It would be nice if bloop inferred `-p` here.

propensive:
Relatedly,
`bloop clean -p foo`
works where
`bloop clean --project foo`
does not.

This isn't a problem with the `compile` subcommand, though it may be for others.

jvican:
@propensive I think that's a case-app bug :smile:. Would be great if you report it.

jvican:
@propensive It's not, it was due to the fact that this field was called `projects` instead of `project`. Fixed that as well as the issue described in this ticket.

PR #313, opened 2018-03-07
Infer projects from cli remaining args

jvican:
This is probably not the best patch this repository has seen, but I'm
quite happy with being able to do it at all so I won't complain.

The CLI handling in bloop has to fundamentally change because it's
introducing a lot of complexity, but for now we can deal with this
complexity since there are more important blockers on the road to
success.

The issue in this commit is that, because case-app derives the CLI
parsers from case classes, we have to turn projects into an optional
argument and set it in our CLI interpreter if the `-p` or `--project`
flags have not been passsed.

This introduces the problem that we, as maintainers, could be creating
instances of these case classes with no specified project at all, but
there is no short-term solution to this problem rather than rewriting
the whole CLI support and having two different ADTs for both parsing and
interpreting.

How bad the state of the art is doesn't matter much for now, this is
something that should affect core developers only and it solves a real
pain in the ass.

Fixes https://github.com/scalacenter/bloop/issues/154.

### DIFF
--- a/frontend/src/test/scala/bloop/nailgun/NailgunTest.scala
+++ b/frontend/src/test/scala/bloop/nailgun/NailgunTest.scala
@@ -28,11 +28,11 @@ abstract class NailgunTest {
     val serverThread =
       new Thread {
         override def run(): Unit = {
-          Server.main(Array(TEST_PORT.toString))
           val outStream = ProcessLogger.toOutputStream(log.info)
           val errStream = ProcessLogger.toOutputStream(log.error)
           Console.withOut(outStream) {
             Console.withErr(errStream) {
+              Server.main(Array(TEST_PORT.toString))
             }
           }
         }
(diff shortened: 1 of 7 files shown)


frontend/src/test/scala/bloop/nailgun/NailgunTest.scala
```diff
@@ -28,11 +28,11 @@ abstract class NailgunTest {
     val serverThread =
       new Thread {
         override def run(): Unit = {
+          Server.main(Array(TEST_PORT.toString))
           val outStream = ProcessLogger.toOutputStream(log.info)
           val errStream = ProcessLogger.toOutputStream(log.error)
           Console.withOut(outStream) {
             Console.withErr(errStream) {
-              Server.main(Array(TEST_PORT.toString))
```
Duhemm: Why is this change necessary ? `Console.withOut` and `Console.withErr` do nothing here after the change, should we remove them completely?

jvican: This was a leftover from debugging, will remove!

PR #314, opened 2018-03-08
Add release notes for 1.0.0-M5

jvican:


jvican:
I didn't tag this as `[DOCS]` or `[CI SKIP]` because this will be the commit we'll need to build when we push the release tag.

PR #319, opened 2018-03-08
Add curl command to install bloop [DOCS]

jvican:


website/content/docs/installation.md
```diff
@@ -39,20 +39,20 @@ Go to [Next steps]({{< ref "#next-steps" >}}) to configure Bloop and start using
...
+$ curl -L https://github.com/scalacenter/bloop/releases/download/v1.0.0-M5/install.py | python2
 ```
 
-You can then move the generated binaries somewhere else on your disk. Optionally, you can also pass
-a destination where the installation script should put the binaries:
+<span class="label warning">Note</span> that you may need to update the version
+in the URL. Check the last released version in [Bloop GitHub releases]({{<
+githubrepo >}}/releases).
+
+#### Installing in a different repository
```
Duhemm: `s/repository/directory/`

website/content/docs/installation.md
```diff
@@ -39,20 +39,20 @@ Go to [Next steps]({{< ref "#next-steps" >}}) to configure Bloop and start using
...
+
+By default, the script installs bloop binaries in a `.bloop` directory in your
+`$HOME` folder. You can change that with the `--dest` option:
```
Duhemm: I think that the default is to install in the current working directory, no?

jvican: `BLOOP_DEFAULT_INSTALLATION_TARGET = join(expanduser("~"), ".bloop")`

Duhemm: 😝 

https://github.com/scalacenter/bloop/blob/d44ecafd86414aa8a02462e91cc607ae9567cba2/bin/install.py#L26-L30

`BLOOP_DEFAULT_INSTALLATION_TARGET` is not used anywhere. What do you prefer we do?

PR #321, opened 2018-03-09
Make install script compatible with Python2 and 3

Duhemm:
Translated using `futurize`.

@valencik I don't know much about Python, is this the way to go?

### DIFF
--- a/bin/install.py
+++ b/bin/install.py
@@ -1,7 +1,10 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 
+from __future__ import print_function
+from future import standard_library
+standard_library.install_aliases()
 import argparse
-import urllib
+import urllib.request, urllib.parse, urllib.error
 import os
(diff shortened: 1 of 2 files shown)


bin/install.py
```diff
@@ -55,10 +58,10 @@
 
 def download_and_install(url, target):
     try:
-        urllib.urlretrieve(url, target)
-        os.chmod(target, 0755)
+        urllib.request.urlretrieve(url, target)
+        os.chmod(target, 0o755)
```
jvican: `0o`? You sure?

Duhemm: Yes (that's generated code, and it runs fine in the Python REPL)

jvican: Let the Python REPL guide us!

valencik:
Oh wow, I didn't know `futurize` could do this without having a new dependency. Very nice!

PR #326, opened 2018-03-09
Add a bunch of fixes to compilation

jvican:


jvican:
test performance please

jvican:
The community build passed, this can be reviewed.

.drone.yml
```diff
@@ -52,7 +52,7 @@ pipeline:
       - export DRONE_DIR="/drone"
       - git log | head -n 20
       - ./bin/add-coursier.sh
-      - ./bin/detect-community-build.sh
+      - ./bin/detect-community-build.sh && export RUN_COMMUNITY_BUILD=true
```
Duhemm: `./bin/detect-community-build.sh && export RUN_COMMUNITY_BUILD=true || echo "Not running the community build."`

Otherwise this will kill the build for non-community-builds

frontend/src/main/scala/bloop/Project.scala
```diff
@@ -22,6 +22,8 @@ case class Project(name: String,
                    javaEnv: JavaEnv,
                    tmp: AbsolutePath,
                    bloopConfigDir: AbsolutePath) {
+  override def toString: String = s"$name"
```
Duhemm: `name` ?

frontend/src/test/scala/bloop/nailgun/NailgunTest.scala
```diff
@@ -40,6 +40,7 @@ abstract class NailgunTest {
 
     serverThread.start()
 
+    Thread.sleep(500)
```
Duhemm: Leftover from debugging?

jvican: This is not a leftover from debugging, it's necessary avoid spurious errors in `NailgunTest`.

frontend/src/main/scala/bloop/Project.scala
jvican: Yes, otherwise the pretty-printed version of the project is terrible: it prints all the malformatted arrays to the users. This fixes some usability errors.
  
In the future, after 1.0.0, I want to use `Show` for practically everything. But now that we're under a deadline, I just prefer to do it this way.

.drone.yml
jvican: `"Not running the community build"` is already outputted by the script, but I'll make `|| 0` to avoid the failure.

PR #328, opened 2018-03-12
Remove log4j

Duhemm:


backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -1,89 +1,67 @@
...
-    val appender = OutputStreamAppender
-      .newBuilder()
-      .setName(s"$name-$nameSuffix")
-      .setFilter(filter)
-      .setLayout(layout)
-      .setTarget(stream)
-      .build()
-
-    appender
+  private def printTrace(line: String): Unit = {
```
jvican: Shall we add a method for `Throwable` here?

jvican: Oh no, just realized that the method targeting `Throwable` comes inherited.

backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -1,89 +1,71 @@
...
+      print(prefix + exception.toString(), printTrace)
+      exception.getStackTrace.foreach(ste => print("\t" + ste.toString, printTrace))
+      trace("Caused by: ", exception.getCause)
+    }
+  }
 
-    outAppender.start()
-    errAppender.start()
+  private def print(msg: String, fn: String => Unit): Unit = {
+    if (msg == null) ()
```
jvican: This implementation is pretty weird. why `.lines` and why the NPE guard? Is there a way we can remove that from this core logging API endpoint?

Duhemm: `.lines` is to have the prefix on every line (think `log.error("foo\nbar")`). The NPE guard is to make sure `.lines` won't crash. There are cases where `exception.getMessage` is `null`.

frontend/src/test/scala/bloop/nailgun/BasicNailgunSpec.scala
```diff
@@ -119,7 +123,54 @@ class BasicNailgunSpec extends NailgunTest {
...
+
+      assertTrue("`logger1` received messages of `logger2`",
+                 !msgs1.exists(_.startsWith("Projects loaded from")))
+      assertEquals("`logger` didn't receive verbose messages",
+                   3,
+                   msgs1.count(_.startsWith(s"${debugPrefix}Elapsed:")).toLong)
+      assertTrue("`logger2` received messages of `logger1`",
+                 !msgs2.exists(_.startsWith("Compiling")))
+      assertTrue("`logger2` received verbose messages of `logger1`",
+                 !msgs2.exists(_.startsWith(debugPrefix)))
```
jvican: :tada:

PR #332, opened 2018-03-12
Add temporary option to debug file watching

jvican:


jvican:
Closing this, since the root of the file watching issue has been found and I want to get some time to add a proper selective debug CLI option that allows us to filter out the sea of logs we see under `--debug`. For example, `--debug file-watching` would show only logs related to file watching, `--debug compilation` only those related to compilation, `--debug test` those related to test execution, etcetera.

Issue #257, opened 2018-02-12
--watch prints out a lot of noise and false compile errors


olafurpg:
At commit https://github.com/olafurpg/vork/commit/85e63f7dff59b2a6051f77fbea0c7ed3e17b4a80
```
$ bloop compile -p vork --watch
Compiling 19 Scala sources to /Users/ollie/dev/vork/vork/target/scala-2.12/classes ...
Done compiling. // All OK
// add single newline to Processor.scala
Compiling 2 Scala sources to /Users/ollie/dev/vork/runtime/target/scala-2.12/classes ...
Compiling 2 Scala sources to /Users/ollie/dev/vork/runtime/target/scala-2.12/classes ...
Compiling 2 Scala sources to /Users/ollie/dev/vork/runtime/target/scala-2.12/classes ...
Compiling 1 Scala source to /Users/ollie/dev/vork/vork/target/scala-2.12/classes ...
Compiling 2 Scala sources to /Users/ollie/dev/vork/runtime/target/scala-2.12/classes ...
Done compiling.
[W]  [E-1] there were three deprecation warnings (since 2.11.0); re-run with -deprecation for details
[W]  [E-1] there were three deprecation warnings (since 2.11.0); re-run with -deprecation for details
[W]  [E-1] there were three deprecation warnings (since 2.11.0); re-run with -deprecation for details
Done compiling.
Done compiling.
Done compiling.
[W]  [E-1] there were three deprecation warnings (since 2.11.0); re-run with -deprecation for details
Done compiling.
Compiling 2 Scala sources to /Users/ollie/dev/vork/runtime/target/scala-2.12/classes ...
Compiling 19 Scala sources to /Users/ollie/dev/vork/vork/target/scala-2.12/classes ...
Compiling 19 Scala sources to /Users/ollie/dev/vork/vork/target/scala-2.12/classes ...
Compiling 19 Scala sources to /Users/ollie/dev/vork/vork/target/scala-2.12/classes ...
Compiling 19 Scala sources to /Users/ollie/dev/vork/vork/target/scala-2.12/classes ...
[W]  [E-1] there were three deprecation warnings (since 2.11.0); re-run with -deprecation for details
Done compiling.
Compiling 19 Scala sources to /Users/ollie/dev/vork/vork/target/scala-2.12/classes ...
[E] [E14] /src/main/scala/vork/markdown/processors/MarkdownCompiler.scala
[E]       not found: type Section
[...]

PR #345, opened 2018-03-14
Redesign file watching and add client listeners

jvican:
Nailgun has this great concept of client listeners: closures that are called
whenever the nailgun server thinks that the client has disconnected. These
closures receive the kind of event that the nailgun server got, and executes
some logic.

This concept allows bloop to register callbacks that are run to clean
resources. Cleaning resources is paramount to keep the right behaviour of
long-running tasks like file watching and bsp servers.

This PR does three things:

1. Exposes the client listeners to the state, in a new abstraction called `ClientPool`.
2. It uses client listeners to cancel tasks, whenever the tasks accept cancellation.
3. Redesigns file watching in a Monix idiomatic style, with observables and
   consumers that help us reason about file watching events more reasonably.

Fixes https://github.com/scalacenter/bloop/issues/257 which was caused
because the cancellation of tasks for file watching did not work correctly
when `CTRL-C` was used.

### DIFF
--- a/backend/src/main/scala/bloop/logging/BloopLogger.scala
+++ b/backend/src/main/scala/bloop/logging/BloopLogger.scala
@@ -1,11 +1,10 @@
 package bloop.logging
 
-import java.io.{FileOutputStream, OutputStream, PrintStream}
+import java.io.PrintStream
 import java.util.concurrent.atomic.AtomicInteger
 
 import scala.Console.{CYAN, GREEN, RED, RESET, YELLOW}
-
-import bloop.io.AbsolutePath
+import com.martiansoftware.nailgun.NGCommunicator
 
 /**
  * Creates a logger that writes to the given streams.
--- a/frontend/src/main/scala/bloop/Server.scala
+++ b/frontend/src/main/scala/bloop/Server.scala
@@ -10,11 +10,15 @@ class Server
 object Server {
   private val defaultPort: Int = 8212 // 8100 + 'p'
   def main(args: Array[String]): Unit = {
+    run(mainTest(args))
+  }
+
+  private[bloop] def mainTest(args: Array[String]): NGServer = {
     val port = Try(args(0).toInt).getOrElse(Server.defaultPort)
     val addr = InetAddress.getLoopbackAddress
     val server = new NGServer(addr, port)
     registerAliases(server)
-    run(server)
+    server
   }
 
   def nailMain(ngContext: NGContext): Unit = {
--- a/frontend/src/main/scala/bloop/Server.scala
+++ b/frontend/src/main/scala/bloop/Server.scala
@@ -10,10 +10,10 @@ class Server
 object Server {
   private val defaultPort: Int = 8212 // 8100 + 'p'
   def main(args: Array[String]): Unit = {
-    run(mainTest(args))
+    run(instantiateServer(args))
   }
 
-  private[bloop] def mainTest(args: Array[String]): NGServer = {
+  private[bloop] def instantiateServer(args: Array[String]): NGServer = {
     val port = Try(args(0).toInt).getOrElse(Server.defaultPort)
     val addr = InetAddress.getLoopbackAddress
     val server = new NGServer(addr, port)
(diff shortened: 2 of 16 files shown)


backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -1,11 +1,10 @@
 package bloop.logging
 
-import java.io.{FileOutputStream, OutputStream, PrintStream}
+import java.io.PrintStream
 import java.util.concurrent.atomic.AtomicInteger
 
 import scala.Console.{CYAN, GREEN, RED, RESET, YELLOW}
-
-import bloop.io.AbsolutePath
+import com.martiansoftware.nailgun.NGCommunicator
```
Duhemm: Unused import?

frontend/src/main/scala/bloop/engine/ClientPool.scala
```diff
@@ -0,0 +1,38 @@
+package bloop.engine
+
+import com.martiansoftware.nailgun.{NGClientDisconnectReason, NGClientListener, NGContext}
+
+sealed trait CloseEvent
+case object Heartbeat extends CloseEvent
+case object SocketError extends CloseEvent
+case object SocketTimeout extends CloseEvent
+case object SessionShutdown extends CloseEvent
+case object InternalError extends CloseEvent
```
Duhemm: Why not using `NGClientDisconnectReason` directly?

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -252,10 +251,32 @@ object Interpreter {
...
-      Await.result(newState.runAsync(previousState.scheduler), Duration.Inf)
+      val handle = newState
+        .executeWithOptions(_.enableAutoCancelableRunLoops)
+        .runAsync(previousState.scheduler)
+
+      // Let's cancel tasks (if supported by the underlying implementation) when clients disconnect
+      pool.addListener {
+        case e: CloseEvent =>
+          if (!handle.isCompleted) {
+            System.out.println(
```
Duhemm: So that's writing to stdout of the server? Isn't that what `ngout` is for?

frontend/src/main/scala/bloop/engine/ClientPool.scala
jvican: Because I want not to depend on nailgun data structures as much as possible. Who knows what we'll use in the future. It didn't take a long time to do this, but in my opinion it's worth it :smile:

frontend/src/main/scala/bloop/engine/Interpreter.scala
jvican: Indeed, but in this case the stdout that we see here is already ngout! I'll change it though, for clarity :smile:

frontend/src/main/scala/bloop/Server.scala
```diff
@@ -10,11 +10,15 @@ class Server
 object Server {
   private val defaultPort: Int = 8212 // 8100 + 'p'
   def main(args: Array[String]): Unit = {
+    run(mainTest(args))
+  }
+
+  private[bloop] def mainTest(args: Array[String]): NGServer = {
```
Duhemm: Not sure that's the best name? `getServer`? `instantiateServer`?

frontend/src/test/scala/bloop/nailgun/BasicNailgunSpec.scala
```diff
@@ -80,6 +83,38 @@ class BasicNailgunSpec extends NailgunTest {
...
+      fileWatchingProcess2.waitFor(4, TimeUnit.SECONDS)
+      fileWatchingProcess2.destroyForcibly()
+
+      // Ugly, but necessary for the sake of testing.
+      rlogger = logger
+    }
+
+    // We check here the logs because until 'exit' the server doesn't flush everything
+    val serverInfos = rlogger.getMessages().filter(_._1 == "server-info").map(_._2)
+    val cancellingTasks = serverInfos.filter(_.contains("Cancelling tasks..."))
```
Duhemm: `serverInfos.count(_.contains("Cancelling tasks..."))` ?

jvican:
Done! Please check your feedback is properly addressed :smile:

Duhemm:
Great! Please merge when it's green 😄

Issue #195, opened 2018-01-26
Bloop returns nonzero exit status without explatation


propensive:
Sometimes Bloop will return a nonzero exit status, without printing any output, so it's not clear why it wasn't successful. I think that in the cases where it fails, it should always print *something*.

Here is an example, which I've not been able to diagnose:
```sh
# bloop compile -p core || echo failed
failed
#
```
while the bloop-server reports only the following:
```
Jan 26, 2018 8:24:00 AM com.martiansoftware.nailgun.NGSession run
INFO: NGSession shutting down
Jan 26, 2018 8:24:00 AM com.martiansoftware.nailgun.NGInputStream lambda$new$1
WARNING: Nailgun client read future was interrupted
java.lang.InterruptedException
        at java.util.concurrent.FutureTask.awaitDone(FutureTask.java:404)
        at java.util.concurrent.FutureTask.get(FutureTask.java:204)
        at com.martiansoftware.nailgun.NGInputStream.lambda$new$1(NGInputStream.java:117)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

Duhemm:
Thanks for the report!

Does it happen in cases where you the command to fail, or it doesn't matter? Do you get at least some output when running with `--verbose`?

propensive:
It happens in a case where I expect the command to succeed, so it's failing and I don't know why. `--verbose` does not change anything, unfortunately (though thanks for making me aware of it)!

propensive:
I say "where I expect the command to succeed", but there are quite possibly reasons why it wouldn't succeed.

propensive:
The cause for this particular failure was trying to do a `bloop run` with the `classpath` parameter in the configuration file empty, which is obviously a mistake: it could do nothing other than fail.

I haven't experimented much, but the complete lack of output may have been a consequence of not even the Scala library jars being on my classpath.

jvican:
The examples to reproduce this ticket now work. We can reopen if users catch more examples, but I'm fairly confident that the handling of status codes is now more reasonable and resilient.

propensive:
[...]

PR #347, opened 2018-03-14
Propagate error to avoid downstream execution

jvican:
As the compilation task does not throw an exception, but rather we model
failures in the DAG, we need to propagate the error down the task
dependency chain so that tasks that depend on compilation do not continue
if compilation has failed.

In this vein, this commit aligns the codebase to behave like this, and then
avoid testing a project that has not correctly compiled.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -101,9 +101,8 @@ object Interpreter {
 
     state.build.getProjectFor(cmd.project) match {
       case Some(project) =>
-        def doCompile(state: State): Task[State] = {
+        def doCompile(state: State): Task[State] =
           Tasks.compile(state, project, reporterConfig).map(_.mergeStatus(ExitStatus.Ok))
-        }
 
         val initialState = {
           if (cmd.incremental) Task(state)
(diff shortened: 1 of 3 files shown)


frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -139,15 +139,31 @@ object Interpreter {
...
   }
 
+  private def compileAnd(
+      state: State,
+      project: Project,
+      reporterConfig: ReporterConfig,
+      excludeRoot: Boolean
+  )(next: State => Task[State]): Task[State] = {
+    Tasks.compile(state, project, reporterConfig, excludeRoot).flatMap { compiled =>
+      if (compiled.status != ExitStatus.CompilationError) next(compiled)
```
Duhemm: Shouldn't we only continue if the status is `Ok`?

jvican: I'm a little bit afraid of doing that now because of https://github.com/scalacenter/bloop/issues/195 and other errors we've had in this area (which is probably the less tested overall in bloop), so I'd prefer to keep it this way for now.

Duhemm:
Drone didn't run...

Issue #344, opened 2018-03-14
Use a different classes directory than sbt


olafurpg:
Currently, `bloop compile` writes classfiles to the same directoy as `sbt compile`. This is annoying when working in a project where I mix `bloop compile` and `sbt compile` because it seems to trigger clean recompilation. In a perfect world, the two can share analysis files so that `sbt compile` can pick up the latest incremental updates from `bloop compile`, and vice versa. In the meantime however, I think it would be nice that `bloop compile` emitted files elsewhere, for example in `out/projectname/compile`

PR #348, opened 2018-03-14
Use different classes directory than sbt

jvican:
Fixes https://github.com/scalacenter/bloop/issues/344.

We don't do what the ticket describes yet because that's something left
for the future. So far, reusing the `.bloop` directory (which should
already be in .gitignore) for the target makes sense.

In the future, we'll get the sbt integration to talk to the bsp server
of bloop and compile to that directory, making the experience of
switching from sbt to bloop and back and forth smooth.

### DIFF
--- a/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
+++ b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
@@ -87,8 +87,9 @@ object PluginImplementation {
         Keys.crossPaths.value
       )
 
-      if (!bloopTarget.exists()) sbt.io.IO.createDirectory(bloopTarget)
       val classesDir = bloopTarget / (Defaults.prefix(configuration.name) + "classes")
+      if (!classesDir.exists()) sbt.io.IO.createDirectory(classesDir)
+
       val sourceDirs = Keys.sourceDirectories.value
       val testFrameworks = Keys.testFrameworks.value.map(_.implClassNames)
       val scalacOptions = Keys.scalacOptions.value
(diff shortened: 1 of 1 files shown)


Duhemm:
One second, what's going on with Drone?

jvican:
Now Drone is running for the last commit.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -73,12 +73,23 @@ object PluginImplementation {
...
-      val classesDir = Keys.classDirectory.value.getAbsoluteFile
+      val bloopTarget = Defaults.makeCrossTarget(
+        bloopConfigDir,
+        Keys.scalaBinaryVersion.value,
+        (Keys.sbtBinaryVersion in Keys.pluginCrossBuild).value,
+        Keys.sbtPlugin.value,
+        Keys.crossPaths.value
+      )
+
+      val classesDir = bloopTarget / (Defaults.prefix(configuration.name) + "classes")
```
olafurpg: Will the classes directories live in `.bloop/` or in some nested directory? I would like to be able to easily remove all `classes` directories without removing `.bloop/*.config` files

jvican: `.bloop/target...` etc. It will have the same path as it used to have in `target/` (sbt) but this time rebased to `.bloop`.

olafurpg: I would consider specifying the location per project, it's good to know when you want to `rm -rf` and I should be able to rely on this location not changing.

olafurpg: transparent cache locations are a nice and cheap feature to support

Issue #338, opened 2018-03-13
Make bloop projects output friendly for scripting


olafurpg:
```
$ bloop projects
Projects loaded from '/Users/ollie/dev/scalafix-scopes/.bloop-config':
 * cli
 * cli-test
 * cli211
 * cli211-test
 * core211JS
 * core211JS-test
 * core211JVM
 * core211JVM-test
 * coreJS
```

It would be nice if the output only contained the project names, that would make it easy to for example use it for tab completions.

```
$ bloop projects
cli
cli-test
cli211
cli211-test
core211JS
core211JS-test
core211JVM
coreJS
```

Duhemm:
I'm working on that as we speak :)

PR #349, opened 2018-03-14
Pretty print projects to be script friendly

jvican:
Now, the "Projects loaded from" is hidden behind the verbose flag.

Fixes https://github.com/scalacenter/bloop/issues/338.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -129,12 +129,9 @@ object Interpreter {
       val contents = Dag.toDotGraph(state.build.dags)
       logger.info(contents)
     } else {
-      // TODO: Pretty print output of show projects, please.
       val configDirectory = state.build.origin.syntax
-      logger.info(s"Projects loaded from '$configDirectory':")
-      state.build.projects.map(_.name).sorted.foreach { projectName =>
-        logger.info(s" * $projectName")
-      }
+      logger.debug(s"Projects loaded from '$configDirectory':")
+      state.build.projects.map(_.name).sorted.foreach(logger.info)
     }
 
     state.mergeStatus(ExitStatus.Ok)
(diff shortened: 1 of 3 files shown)


frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -129,12 +129,9 @@ object Interpreter {
...
       logger.info(contents)
     } else {
-      // TODO: Pretty print output of show projects, please.
       val configDirectory = state.build.origin.syntax
-      logger.info(s"Projects loaded from '$configDirectory':")
-      state.build.projects.map(_.name).sorted.foreach { projectName =>
-        logger.info(s" * $projectName")
-      }
+      logger.debug(s"Projects loaded from '$configDirectory':")
+      state.build.projects.map(_.name).sorted.foreach(logger.info)
```
olafurpg: What's the order of the projects?

jvican: Whatever the order of reading them is. It's non-deterministic, if you're asking for that.

Duhemm: They're printed in alphabetical order if that's what you're asking?

olafurpg: alphabetical order is good IMO, having a deterministic order is good IMO

jvican: Ah no! Sorry, it's alphabetical by default.

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -129,12 +129,9 @@ object Interpreter {
       val contents = Dag.toDotGraph(state.build.dags)
       logger.info(contents)
     } else {
-      // TODO: Pretty print output of show projects, please.
       val configDirectory = state.build.origin.syntax
-      logger.info(s"Projects loaded from '$configDirectory':")
-      state.build.projects.map(_.name).sorted.foreach { projectName =>
-        logger.info(s" * $projectName")
-      }
+      logger.debug(s"Projects loaded from '$configDirectory':")
```
Duhemm: https://github.com/scalacenter/bloop/blob/072b2fe908970658d112db844d0062eb4c0ee596/frontend/src/test/scala/bloop/engine/InterpreterSpec.scala#L34

This needs to be adapted

jvican: I've already done that in my last commit.

Issue #330, opened 2018-03-12
Clear the screen before each iteration in `--watch` mode


Duhemm:
When running Bloop with `--watch`, we should clear the screen before each iteration. Otherwise, it's hard to know to what iteration every bit of output belongs.

If we don't clear the screen, we should at least print some separator?

```
bloop+2… vork ❯ bloop compile -p vork --watch
Compiling 2 Scala sources to /Users/martin/Documents/Projects.nosync/olafurpg/vork/runtime/target/scala-2.12/classes ...
[W]  [E-1] there were three deprecation warnings (since 2.11.0); re-run with -deprecation for details
Done compiling.
Compiling 19 Scala sources to /Users/martin/Documents/Projects.nosync/olafurpg/vork/vork/target/scala-2.12/classes ...
Done compiling.
Compiling 1 Scala source to /Users/martin/Documents/Projects.nosync/olafurpg/vork/vork/target/scala-2.12/classes ...
Done compiling.
Compiling 1 Scala source to /Users/martin/Documents/Projects.nosync/olafurpg/vork/vork/target/scala-2.12/classes ...
Done compiling.
Compiling 1 Scala source to /Users/martin/Documents/Projects.nosync/olafurpg/vork/vork/target/scala-2.12/classes ...
Done compiling.
Compiling 1 Scala source to /Users/martin/Documents/Projects.nosync/olafurpg/vork/vork/target/scala-2.12/classes ...
Done compiling.
Compiling 1 Scala source to /Users/martin/Documents/Projects.nosync/olafurpg/vork/vork/target/scala-2.12/classes ...
Done compiling.
```

jvican:
Fixed in #350.

PR #350, opened 2018-03-14
Clear screen before each watch iteration

jvican:
From https://stackoverflow.com/questions/2979383/java-clear-the-console.

### DIFF
--- a/frontend/src/main/scala/bloop/io/SourceWatcher.scala
+++ b/frontend/src/main/scala/bloop/io/SourceWatcher.scala
@@ -3,6 +3,7 @@ package bloop.io
 import java.nio.file.Path
 
 import bloop.Project
+import bloop.bsp.BspServer
 import bloop.engine.{ExecutionContext, State}
 import bloop.logging.Logger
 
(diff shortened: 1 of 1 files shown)


jvican:
Fixes https://github.com/scalacenter/bloop/issues/330.

Duhemm:
What about Windows?

jvican:
Doesn't Windows support escape characters? I think it does. https://en.wikipedia.org/wiki/ANSI_escape_code I don't care about not supporting Windows versions < 10.

Duhemm:
This comment hints that it doesn't work: https://stackoverflow.com/questions/2979383/java-clear-the-console#comment79501906_32295974

I don't think it matters that much. I hope for people on Windows they use a better terminal.

jvican:
Let's try to deal with this misbehaviour in the future then.

olafurpg:
I propose you don't clear the screen on Windows if you suspect it will cause problems. It's better to have somebody more knowledgeable about Windows contribute window clearing.

jvican:
That makes total sense.

PR #352, opened 2018-03-14
Fix clearing of screen in SourceWatcher + drone insanity fix

jvican:
For some reason, Drone doesn't seem to update the drone commit
message... Let's do it manually.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -53,11 +53,11 @@ pipeline:
       - git log | head -n 20
       - ./bin/add-coursier.sh
       - . bin/detect-community-build.sh # We need to source it for the env variable to be exported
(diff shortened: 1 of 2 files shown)


jvican:
Don't merge yet.

Duhemm:
Actually, I'm thinking that we should NOT clear the screen if the logs are verbose, otherwise we'll never see why the file watch got triggered.

jvican:
Or, we can move the `debug` after we clear the screen.

Duhemm:
That should work too, we can always change that if it's annoying

jvican:
Let's merge this as soon as CI is green. I want a release out tonight :man_dancing:

PR #353, opened 2018-03-14
Add autocompletion with Bash and zsh

Duhemm:
This adds a new command `completion` to bloop that takes care of providing autocompletion suggestions. The output is configured with `--format`, so that it can be understood by different autocompletion engines.

The suggestions for autocompletion depend on the context where they are applied: For instance, after a command, Bloop will suggest project names. If the command is `run` and the flag is `--main`, Bloop will propose main classes.

Currently Bloop will not suggest test names in (`test`, `--filter`) because extracting the test names requires some refactoring that I've undertaken in my WIP to support Scala Native. I'll fix that later.

The homebrew formula installs both autocompletions for Bash and zsh.


valencik:
WOW! This is wicked

frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -34,6 +34,14 @@ object Commands {
       @Recurse cliOptions: CliOptions = CliOptions.default
   ) extends RawCommand
 
+  case class Completion(
```
jvican: Can we call this `autocompletion` so that `completion` doesn't conflict with `compile`?

Duhemm:
This is my first feature written 100% with Bloop 😄

frontend/src/main/scala/bloop/cli/Commands.scala
olafurpg: I was gonna suggest the same, ideally there should be no conflicting tasks with `test` and `compile` to make `c<TAB> => compile` and `t<TAB>` completion work

jvican:
I've just tried this out, it's awesome.
  
Someone, in the future, should check whether we abide by the rules described in the zsh completions bible or not. https://github.com/zsh-users/zsh/blob/master/Etc/completion-style-guide
  
This is not a blocker for now, it works.
  
I added the following to my `~/.zshrc`:
  
```
fpath=(/data/rw/code/scala/loop/etc/zsh $fpath)
autoload -U _bloop
compdef _bloop bloop
```

olafurpg:
JUST RELEASE ALREADY WILL YA 😸

jvican:
Waiting on a PR to update the docs and update the python script and we release M7 folks! :tada:

PR #357, opened 2018-03-14
Add the project to the classes dir path

jvican:
I mistakenly forgot to add the project to the bloop configuration
directory, meaning that all class files effectively go to the same
classes directory.

### DIFF
--- a/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
+++ b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
@@ -17,6 +17,8 @@ object AutoImported {
   import sbt.{SettingKey, TaskKey, settingKey, taskKey}
   val bloopConfigDir: SettingKey[File] =
     settingKey[File]("Directory where to write bloop configuration files")
+  val bloopProductDirectories: TaskKey[Seq[File]] =
+    taskKey[Seq[File]]("Directory where to write the class files")
   val bloopInstall: TaskKey[Unit] =
     taskKey[Unit]("Generate all bloop configuration files")
   val bloopGenerate: sbt.TaskKey[Unit] =
@@ -30,7 +32,10 @@ object PluginImplementation {
   )
 
   val configSettings: Seq[Def.Setting[_]] = {
-    val rawSettingsInConfigs = List(BloopKeys.bloopGenerate := PluginDefaults.bloopGenerate.value)
+    val rawSettingsInConfigs = List(
+      BloopKeys.bloopProductDirectories := PluginDefaults.generateBloopProductDirectories.value,
+      BloopKeys.bloopGenerate := PluginDefaults.bloopGenerate.value
+    )
     val all = rawSettingsInConfigs ++ DiscoveredSbtPlugins.settings
     all.flatMap(ss => sbt.inConfig(Compile)(ss) ++ sbt.inConfig(Test)(ss))
   }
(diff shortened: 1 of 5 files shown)


jvican:
I've added all we needed to effectively change our sbt target dir to our own. The changes were bigger than I expected, but this change should make bloop work for real.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -104,7 +107,20 @@ object PluginImplementation {
...
-      val scalacOptions = Keys.scalacOptions.value
+
+      val scalacOptions = {
+        val scalacOptions0 = Keys.scalacOptions.value
+        val internalClasspath = AutoImported.bloopInternalClasspath.value
+        internalClasspath.foldLeft(scalacOptions0) {
+          case (scalacOptions, (oldClassesDir, newClassesDir)) =>
+            scalacOptions.map { scalacOption =>
+              scalacOption
+                .replace(oldClassesDir.toString, newClassesDir.getAbsolutePath)
```
Duhemm: I imagine that it doesn't matter, but is it on purpose that you pass the absolute path here?

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -33,18 +41,22 @@ object PluginImplementation {
     BloopKeys.bloopInstall := PluginDefaults.bloopInstall.value
   )
 
+  import Compat._
   val configSettings: Seq[Def.Setting[_]] = {
     val rawSettingsInConfigs = List(
-      BloopKeys.bloopProductDirectories := PluginDefaults.generateBloopProductDirectories.value,
+      BloopKeys.bloopProductDirectories := List(BloopKeys.bloopClassDirectory.value),
+      BloopKeys.bloopManagedResourceDirectories := PluginDefaults.managedResourceDirs.value,
+      BloopKeys.bloopClassDirectory := PluginDefaults.generateBloopProductDirectories.value,
```
Duhemm: Maybe rename `generateBloopProductDirectories` to `generateBloopClassDirectory` ?

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -195,6 +211,40 @@ object PluginImplementation {
...
+            .value
+        val s = Keys.streams.value
+        val cacheStore = Compat.generateCacheFile(s, "copy-resources-bloop")
+        val mappings = (sbt.PathFinder(Keys.resources.value) --- dirs) pair (sbt.Path
+          .rebase(dirs, t) | sbt.Path.flat(t))
+        s.log.debug("Copy resource mappings: " + mappings.mkString("\n\t", "\n\t", ""))
+        sbt.Sync(cacheStore)(mappings)
+        mappings
+      }
+    }
```
Duhemm: Can you add a comment explaining the difference between this task and sbt's implementation, and why we need to do this?

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
jvican: Yes.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
jvican: No, because it's named after `productDirectories` in sbt.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
jvican: The implementation is the same, but we need to hijack all the use sites of `targetDir`, and that means reimplementing this logic.
  
There's still work to do (I think, even though right now it works in our community build and all the projects i've tried it with), so I prefer to add the comment in a follow-up PR.

PR #358, opened 2018-03-14
Install the completions in the installation script

Duhemm:


Duhemm:
I've tested the script with python2 and python3 :)

bin/install.py
```diff
@@ -7,6 +7,7 @@
 from os.path import expanduser, isdir, isfile, join
 from subprocess import CalledProcessError, check_call
 import sys
+import re
```
Duhemm: This comes from the python stdlib, so no worries

jvican:
:shipit:

bin/install.py
```diff
@@ -53,11 +54,30 @@
...
+ZSH_COMPLETION_DIR = join(BLOOP_INSTALLATION_TARGET, "zsh")
+BASH_COMPLETION_DIR = join(BLOOP_INSTALLATION_TARGET, "bash")
+
+# If this is not a released version of Bloop, we need to extract the commit SHA
+# to know how to download the completion scripts.
+# If we can't get the SHA, just download from master.
+if CUSTOMIZED_SCRIPT:
+    COMPLETION_VERSION = BLOOP_VERSION
+else:
+    pattern = '(?:.+?)-([0-9a-f]{8})(?:\+\d{8}-\d{4})?'
```
Duhemm: That's the same regex we use to do the same thing in the benchmarks

PR #362, opened 2018-03-14
Add one more core to the available processors

jvican:
Let's add one more core because the monix infrastructure steals us one
blocking. Blocking is something that computers can easily put on timers,
so this way we won't waste one core that could be used to compile.


jvican:
In the future I want to rework this area by having a scheduler per client. But for now this is good to go.

PR #363, opened 2018-03-15
Update the build in every watch iteration

jvican:
Nailgun shuts down the executor service running the "server code"
whenever an exception is catched. This usually takes around 1000s after
the execption has been thrown.

This exception is thrown after a SIGKILL or SIGINT, for example. The
problem is that in 1000s our bloop tasks do not have time to return, and
that means that the `Await.result` in `Interpreter` never finishes.

Therefore, the logic never reaches `Cli` back, and the state is not
saved. This code is a temporary workaround thought to overcome this
limitation. It fixes the current problem in which the compilations that
happen in file watching are thrown away. Now they are cached, and
subsequent compile commands won't trigger compile afresh.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/caches/StateCache.scala
+++ b/frontend/src/main/scala/bloop/engine/caches/StateCache.scala
@@ -24,7 +24,7 @@ final class StateCache(cache: ConcurrentHashMap[AbsolutePath, State]) {
    * @return The passed `State`, with a status code of `Ok`.
    */
   def updateBuild(state: State): State = {
-    cache.put(state.build.origin, state)
+    Option(cache.put(state.build.origin, state)).getOrElse(state)
   }
 
   /**
--- a/frontend/src/test/scala/bloop/engine/FileWatchingSpec.scala
+++ b/frontend/src/test/scala/bloop/engine/FileWatchingSpec.scala
@@ -17,7 +17,7 @@ import org.junit.experimental.categories.Category
 class FileWatchingSpec {
   @scala.annotation.tailrec
   final def readCompilingLines(target: Int, msg: String, out: ByteArrayOutputStream): Int = {
-    Thread.sleep(100) // Wait 100ms for the OS's file system
+    Thread.sleep(100) // Wait 150ms for the OS's file system
     val allContents = out.toString("UTF-8")
     val allLines = allContents.split(System.lineSeparator())
     val compiled = allLines.count(_.replaceAll("\u001B\\[[;\\d]*m", "").contains(msg))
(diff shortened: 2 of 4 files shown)


frontend/src/test/scala/bloop/engine/FileWatchingSpec.scala
```diff
@@ -17,7 +17,7 @@ import org.junit.experimental.categories.Category
 class FileWatchingSpec {
   @scala.annotation.tailrec
   final def readCompilingLines(target: Int, msg: String, out: ByteArrayOutputStream): Int = {
-    Thread.sleep(100) // Wait 100ms for the OS's file system
+    Thread.sleep(100) // Wait 150ms for the OS's file system
```
Duhemm: Comment are code are not in sync :)

project/BuildPlugin.scala
```diff
@@ -196,7 +196,7 @@ object BuildImplementation {
   import ch.epfl.scala.sbt.release.ReleaseEarlyPlugin.{autoImport => ReleaseEarlyKeys}
 
   final val globalSettings: Seq[Def.Setting[_]] = Seq(
-    BuildKeys.schemaVersion := "1.2",
+    BuildKeys.schemaVersion := "1.2-refresh",
```
Duhemm: Is that necessary ? We shouldn't need to re generate the integration configs in this PR right?

frontend/src/main/scala/bloop/engine/caches/StateCache.scala
```diff
@@ -24,7 +24,7 @@ final class StateCache(cache: ConcurrentHashMap[AbsolutePath, State]) {
    * @return The passed `State`, with a status code of `Ok`.
    */
   def updateBuild(state: State): State = {
-    cache.put(state.build.origin, state)
+    Option(cache.put(state.build.origin, state)).getOrElse(state)
```
Duhemm: Did you ever hit `null` here? Wouldn't that be a scary bug ???

Duhemm: Also it looks like the java doc is not in sync with the code, but it already wasn't...

project/BuildPlugin.scala
jvican: Yes, it's necessary

frontend/src/main/scala/bloop/engine/caches/StateCache.scala
Duhemm: Not scary in the hand, this happens in tests with a hand-rolled state

PR #366, opened 2018-03-15
Fix maximum depth when loading projects

Duhemm:


jvican:
Just rebased this so that we can quickly merge it.

PR #380, opened 2018-03-16
Change python2 to python in install instructions

valencik:
Changes the installation instructions for other platforms to use generic `python` now that install scripts support both python 2 and 3.
This appears to be the last reference to `python2` in the project.


valencik:
Rebased.

Issue #379, opened 2018-03-16
Even watch file iterations do not trigger test or compile


MasseGuillaume:
![out2](https://user-images.githubusercontent.com/921490/37526995-2f106882-2931-11e8-817e-2a2f506de17e.gif)

steps:

```
bloop test -w test-project-name
edit a file
save // 1
nothing happens
edit a file
save // 2
tests are run
```

expected:

tests should be triggered the first time I save.

PR #381, opened 2018-03-16
Remove clean terminal before file watching

jvican:
This is causing misbehaviours in the terminals, because it sometimes
cleans the screen right after the compilation logs are outputted, and
therefore they are lost.

Fixes https://github.com/scalacenter/bloop/issues/379.


jvican:
I really don't know how to fix this otherwise. I'm making this temporary fix for now. Whomever gets this back into the repo will need to check that it doesn't cause a race condition.

jvican:
(I double checked a race condition is what happens because the compiler does update the class files.)

PR #389, opened 2018-03-23
Ensure that file watching task exits

### LINKED ISSUE #379: Even watch file iterations do not trigger test or compile

jvican:
Previously, `cancel` would not complete the task. This refactoring makes
sure that the task is cancelled and that the completion returns the last
state processed by the consumer.

Such improvement makes sure that there's no hanging task listening to file
watching events.

Fixes #379.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/ExecutionContext.scala
+++ b/frontend/src/main/scala/bloop/engine/ExecutionContext.scala
@@ -1,7 +1,16 @@
 package bloop.engine
 
-import java.util.concurrent.{LinkedBlockingQueue, ThreadPoolExecutor, TimeUnit}
-import monix.execution.Scheduler
+import java.lang.Thread.UncaughtExceptionHandler
+import java.util.concurrent.{
+  LinkedBlockingQueue,
+  SynchronousQueue,
+  ThreadFactory,
+  ThreadPoolExecutor,
+  TimeUnit
+}
+
+import monix.execution.{ExecutionModel, UncaughtExceptionReporter}
+import monix.execution.schedulers.ExecutorScheduler
 
 object ExecutionContext {
   private[bloop] val nCPUs = Runtime.getRuntime.availableProcessors() + 1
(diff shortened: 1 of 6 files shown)


jvican:
This error was created because the file watching task was not correctly killed, and monix was somehow running it in the background. The compilation was always happening in #379, which explains the red herring in #354, but it was happening by an stale file watching process from a previous run.
  
Coincidentally, this watching process was being run on the background only in the even runs, stealing the job from the new file watching process and then making it return.

frontend/src/main/scala/bloop/io/SourceWatcher.scala
```diff
@@ -61,22 +63,38 @@ final class SourceWatcher(project: Project, dirs0: Seq[Path], logger: Logger) {
...
+            case EventType.OVERFLOW => runAction(state, event)
+            case EventType.DELETE => Task.now(state)
+          }
+      }
+    }
 
+    val watchHandle = watchController.runAsync(ExecutionContext.ioScheduler)
     observable
       .consumeWith(fileEventConsumer)
-      .doOnFinish(_ => Task(watchHandle.cancel()))
```
jvican: @alexandru If you add `doOnFinish` here, the cancellation logic of the consumer is not executed. Is this a bug?

frontend/src/main/scala/bloop/io/SourceWatcher.scala
```diff
@@ -61,22 +63,38 @@ final class SourceWatcher(project: Project, dirs0: Seq[Path], logger: Logger) {
...
-      try watcher.watch()
+      try watcherHandle.get()
       finally watcher.close()
-    }.doOnCancel(Task {
+      logger.debug("The file watcher was successfully closed.")
+    }
+
+    val watchCancellation = Cancelable { () =>
+      watchingEnabled = false
+      watcherHandle.complete(null)
```
Duhemm: Can the `null` find its way to the `fileEventConsumer`?

frontend/src/main/scala/bloop/monix/FoldLeftAsyncConsumer.scala
```diff
@@ -0,0 +1,83 @@
...
+import monix.execution.cancelables.{AssignableCancelable, CompositeCancelable}
+import monix.execution.misc.NonFatal
+import monix.reactive.{Consumer, Observable}
+import monix.reactive.observers.Subscriber
+
+import scala.collection.mutable.ListBuffer
+import scala.concurrent.Future
+
+// Fork of `FoldLeftAsyncConsumer` from Monix
+final class FoldLeftAsyncConsumer[A, R](
```
Duhemm: What are the differences with the reference implementation from Monix? Why can't we use it?

frontend/src/test/scala/bloop/engine/FileWatchingSpec.scala
```diff
@@ -160,4 +166,39 @@ class FileWatchingSpec {
...
+          Task(handle.cancel()).delayExecution(FiniteDuration(2, TimeUnit.SECONDS))
+        val waitWatchingHandle = waitForWatching.runAsync(s1)
+        val waitCancelHandle = waitAndCancelWatching.runAsync(s1)
+
+        try Await.result(waitCancelHandle, FiniteDuration(3, TimeUnit.SECONDS))
+        finally println(s"Cancellation was executed ${waitCancelHandle.isCompleted}")
+        try Await.result(waitWatchingHandle, FiniteDuration(5, TimeUnit.SECONDS))
+        finally println(s"Watching task was completed ${waitWatchingHandle.isCompleted}")
+    }
+    ()
```
Duhemm: It's not immediately clear to me how this test can fail, can you please explain?

jvican: This test can fail if the watching handle doesn't return (as it was happening before), which means that Bloop is blocked.

frontend/src/main/scala/bloop/monix/FoldLeftAsyncConsumer.scala
jvican: https://github.com/scalacenter/bloop/pull/389/files/3b07914214c27039658685c68355de169f083d3c#diff-935b4e27e7e2af3e3f5402c4e6b9bcc4R62 and the implementation of `apply` are the differences. Essentially, we add cancellation logic to this consumer.

frontend/src/main/scala/bloop/io/SourceWatcher.scala
jvican: Yeah, this works, Java expects `Void` and for that you need to use null in Scala.

frontend/src/main/scala/bloop/engine/ExecutionContext.scala
```diff
@@ -16,6 +25,42 @@ object ExecutionContext {
...
+      TimeUnit.SECONDS,
+      new SynchronousQueue[Runnable](false),
+      threadFactory
+    )
+  }
+
+  implicit lazy val ioScheduler: Scheduler =
+    ExecutorScheduler(ioExecutor, ioReporter, ExecutionModel.Default)
+
+  // Inlined from `monix.execution.schedulers.ThreadFactoryBuilder`
```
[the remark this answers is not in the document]
jvican: Because the executor needs to be exposed so that https://github.com/scalacenter/bloop/pull/389/files/3b07914214c27039658685c68355de169f083d3c#diff-f9ab286c6a27ce9fefdf1f0373d83486R67 works.

PR #390, opened 2018-03-23
Use nailgun that doesn't throw closed socket

jvican:
Uses the improvement merged in scalacenter/nailgun and referenced in
https://github.com/facebook/nailgun/issues/112.


jvican:
Merging this since it's an easy change.

PR #394, opened 2018-03-26
Add flag to support source dependencies

### LINKED ISSUE #375: Support source dependencies in sbt-bloop
Bloop doesn't support source dependencies in sbt because they are tricky to support and there are more important things to work on in the sbt integration. The main issue seems to be that doing `bloopInstall` applies the project filter in the *current* build, but not all the source builds loaded by sbt.
  
This is a ticket to keep track of support of source deps in sbt.

jvican:
The main issue with source deps in Bloop arises from the fact that, by
design, `bloopConfigDir` is build dependent, which means that it changes
across different source deps.

We don't want this to be the case for source dependencies though, since
Bloop needs all the project dependencies to be in the same directory.
For that reason, this flag aggregates the bloop config files to be
written to the same bloop configuration directory, and hence have it
work nicely with bloop.

Fixes #375.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,8 @@
 import build.BuildImplementation.BuildDefaults
 
+// Tell bloop to aggregate source deps (benchmark) config files in the same bloop config dir
+bloopAggregateSourceDependencies in Global := true
+
 /***************************************************************************************************/
 /*                      This is the build definition of the source deps                            */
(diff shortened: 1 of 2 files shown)


jvican:
I double checked manually this does not affect the bloop files in our community build because this setting is always false. :smile:

Issue #320, opened 2018-03-08
Rename blp-server to bloop-server


olafurpg:
blp-server pros:
* tab completion

blp-server cons:
* not self-explanatory. One more thing the user needs to learn, whereas `bloop-server` needs no explanation
* cryptic named long-running binary process that appears in htop/activity monitor

Most users won't invoke blp-server manually because it can be handled by something like `services start bloop`. And even if you invoke it manually, you do it once and let it run in the background so the tab completion is not that desirable.

Same applies to blp-coursier.

cc/ @propensive

olafurpg:
Alternatively, we could tweak the `bloop` python client to support `bloop startserver` and the user never needs to know the difference between bloop and bloop-server.

jvican:
I lean towards having `bloop server --start` so that if we, in the future, want to add more flags to control the server, we can.

propensive:
Aesthetically, @jvican's suggestion is the best I've seen. :)

PR #400, opened 2018-03-27
Add support for `bloop server`

jvican:
Adds improvements in nailgun that fix https://github.com/scalacenter/bloop/issues/320.

### DIFF
--- a/website/content/docs/how-to-use-bloop.md
+++ b/website/content/docs/how-to-use-bloop.md
@@ -10,36 +10,37 @@ bref = "Learn how to use bloop and get familiar with its client-server architect
 
 Bloop has two main components, a client and a server. If you're looking for
 ways to extend this model, head to the [Integration
-guide]({{< ref "integration-guide.md" >}}).
+guide]({{< ref "integration-guide.md" >}}). Otherwise, keep reading.
 
 ### `bloop` client
 
-The `bloop` client is the user-facing tool to run any bloop command. The
-client is a python script that implements Nailgun's protocol and communicates
-with the server. It gives you immediate feedback from the server, usually in
-the order of milliseconds.
+The `bloop` client is the user-facing tool to run any bloop command. The client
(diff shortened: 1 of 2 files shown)


jvican:
CI error in https://github.com/scalacenter/bloop/pull/400/commits/53de32c19e3d61bb301ee4e1ca43d3244e02fd46 is spurious.

website/content/docs/how-to-use-bloop.md
```diff
@@ -10,22 +10,57 @@ bref = "Learn how to use bloop and get familiar with its client-server architect
...
+
+<span class="label warning">Note</span> There can only be one server running
+per user session. If you want to run more than one server (strongly not
+recommended unless you know what you're doing), then specify the nailgun port
+in both the server and the client.
+
+The server is installed by default in the same directory than the `bloop`
+client, so `bloop server` will try to find the location of the server in the
+same directory where `bloop` is installed. If this is not the case, run `bloop
+---server-location=/foo/my-blp-server server` instead.
```
Duhemm: Are there 3 `-` on purpose?

website/content/docs/how-to-use-bloop.md
```diff
@@ -10,22 +10,57 @@ bref = "Learn how to use bloop and get familiar with its client-server architect
...
+
+Have you forgotten to start bloop's server? Run it with `bloop server`.
+Check our usage instructions in https://scalacenter.github.io/bloop/
+
+Type `--nailgun-help` for help on the Nailgun CLI tool.
+```
+
+The bloop server can serve responses of many concurrent clients.
+
+#### Example
```
Duhemm: I think that we should get rid of this paragraph. It doesn't bring much, and the given example is not a runnable command (no project is specified).

jvican:
> Maybe that we should also use bloop server to start the service in the homebrew formula?
  
`bloop server` is meant to be used by the users, so I don't see much value in doing that.
  
Feedback addressed.

PR #402, opened 2018-03-28
Disable spurious test

jvican:
This fails spuriously on Windows and it's quite annoying. I want to
disable it for now until we get perfect support for concurrent clients.


Duhemm:
What’s that second commit?

jvican:
Something I added because the stream logic was misbehaving. If it happens again, I'll see why it wasn't returning.

PR #411, opened 2018-03-29
Add improvements to sbt plugin and migrate to `.bloop`

jvican:
I noticed that our `integrationSetUpBloop` was displaying logs with the
paths shortened, when they shouldn't because all the configuration files
are written in the staging sbt directory (because they are source deps).
This commit fixes it.

### DIFF
--- a/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
+++ b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
@@ -60,6 +60,7 @@ object PluginImplementation {
   // We create build setting proxies to global settings so that we get autocompletion (sbt bug)
   val buildSettings: Seq[Def.Setting[_]] = List(
     BloopKeys.bloopInstall := BloopKeys.bloopInstall.in(Global).value,
+    // Bloop users: Do NEVER override this setting as a user if you want it to work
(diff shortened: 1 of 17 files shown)


integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -60,6 +60,7 @@ object PluginImplementation {
   // We create build setting proxies to global settings so that we get autocompletion (sbt bug)
   val buildSettings: Seq[Def.Setting[_]] = List(
     BloopKeys.bloopInstall := BloopKeys.bloopInstall.in(Global).value,
+    // Bloop users: Do NEVER override this setting as a user if you want it to work
```
Duhemm: This should probably be in the documentation, no?

jvican: We don't have docs for this yet. Do you want me to add them?

jvican: Just added them.

jvican:
I need this for #413 and the JSON configuration changes, so I'm merging for now. @Duhemm Please do raise questions or provide comments if something is not clear. I double checked all the code and the configuration files pass, as well as the community build.

Issue #335, opened 2018-03-13
Make sure that we get the environment variables from Nailgun


jvican:
The nailgun protocol transmits all the environment variables every time there is a connection between the client and the server so that the server can be up-to-date with changes in their developer environment.
  
I think nailgun does not update the environment variables on its own, but rather exposes them in the `NGContext` and waits for us to do something with them. To be on the safe side, we'd need bloop to take either of these strategies:
  
1. Update the mutable `sys.env` with the new environment variable changes.
2. Keep the environment variables map around in the state and change use sites to query that one instead of `sys.env`. This may be possible within bloop sources, but I doubt we'll be able to make Zinc work with it (because we'd need to change its sources and we only do that for critical changes).

jvican:
See my upstream proposal to add this feature in a generic way: https://github.com/facebook/nailgun/issues/129

jvican:
Reminder: ping @etaque who showed interest in this ticket.

jvican:
@etaque Are you using env variables for running and testing right? I guess you're not using them for compilation.

etaque:
@jvican exact, using them for app configuration

jvican:
This is merged now and published in version 1.0.0-M8+20-0dbe10a0: https://ci.scala-lang.org/scalacenter/bloop/2629/9! I'll be cutting M9 tomorrow afternoon, in case you want a better stable.

jvican:
M9 is available @etaque, just in case you want to give it a try :wink:.

PR #413, opened 2018-03-29
Use nailgun environment in forked processes

jvican:
This commit propagates the `remoteEnv` properties in the nailgun context
to the place where we run and test projects in forked projects. Thanks
to the fact that we always fork, we're able to use the environment
variables from the user without weird hacks to mutate `System.getenv`
and scope it per client.

Fixes https://github.com/scalacenter/bloop/issues/335.


build-integrations/sbt-1.0/project/Integrations.scala
```diff
@@ -9,8 +9,8 @@ object Integrations {
   val MiniBetterFiles = RootProject(uri(
     "git://github.com/scalacenter/mini-better-files.git#0ed848993a2fd5a36e4366b5efb9c68dce958fc2"))
   val WithResources = RootProject(
-    uri("git://github.com/scalacenter/with-resources.git#7529b2c3ac455cbb1889d4791c4e0d4957e29306"))
+    uri("git://github.com/scalacenter/with-resources.git#f0a46830cae7ef6282d9bba64b6da34bae18f339"))
```
jvican: These new hashes contain modifications to the dummy code to fail if the env variable `BLOOP_OWNER` is not accessible via `sys.env`.

jvican:
What this means is that now Bloop will use the environment variable of the places where the bloop client is involved (e.g. your terminal) to run and test code.

jvican:
This is ready for review :tada:.

frontend/src/main/scala/bloop/exec/ForkProcess.scala
```diff
@@ -47,6 +47,7 @@ final case class ForkProcess(javaEnv: JavaEnv, classpath: Array[AbsolutePath]) {
               className: String,
               args: Array[String],
               logger: Logger,
+              env: Properties,
```
Duhemm: Javadoc needs updating

frontend/src/main/scala/bloop/exec/ForkProcess.scala
```diff
@@ -68,6 +69,9 @@ final case class ForkProcess(javaEnv: JavaEnv, classpath: Array[AbsolutePath]) {
     } else {
       val processBuilder = new ProcessBuilder(cmd: _*)
       processBuilder.directory(cwd.toFile)
+      val processEnv = processBuilder.environment()
+      processEnv.clear()
+      env.forEach { case (k: String, v: String) => processEnv.put(k, v); () }
```
Duhemm: I think that you can use `putAll` here

frontend/src/main/scala/bloop/exec/ForkProcess.scala
```diff
@@ -68,6 +69,9 @@ final case class ForkProcess(javaEnv: JavaEnv, classpath: Array[AbsolutePath]) {
     } else {
       val processBuilder = new ProcessBuilder(cmd: _*)
       processBuilder.directory(cwd.toFile)
+      val processEnv = processBuilder.environment()
+      processEnv.clear()
```
Duhemm: Are we sure that we want to discard everything?

frontend/src/main/scala/bloop/testing/TestInternals.scala
```diff
@@ -79,7 +73,8 @@ object TestInternals {
                    fork: ForkProcess,
                    discoveredTests: DiscoveredTests,
                    eventHandler: EventHandler,
-                   logger: Logger): Unit = {
+                   logger: Logger,
+                   env: Properties): Unit = {
```
Duhemm: Javadoc needs updating

frontend/src/main/scala/bloop/cli/CommonOptions.scala
```diff
@@ -22,11 +23,18 @@ case class CommonOptions(
...
   def workingPath: AbsolutePath = AbsolutePath(workingDirectory)
 }
 
 object CommonOptions {
   final val default = CommonOptions()
+  final val currentProperties: Properties = {
+    import scala.collection.JavaConverters._
+    System.getenv().asScala.foldLeft(new Properties()) {
+      case (props, (key, value)) => props.setProperty(key, value); props
+    }
```
Duhemm: Just use `clone`?

jvican:
> Are we sure that we want to remove everything from a forked process' environment before adding more values? Could we be discarding values that users would expect to be present in the environment?
  
This PR will not expose any of the server's environment variables to the test and run tasks, so yeah, if users expect those to be in they won't. I think that doing this is the right behaviour, and it's very unlikely someone would want env variables from the server for these particular tasks.
  
However, the rest of the tasks will still use the environment variables of the server (as expected).

frontend/src/main/scala/bloop/exec/ForkProcess.scala
jvican: Yeah, I elaborated on my other comment.

Issue #58, opened 2017-11-23
Use a standard configuration format that other tools can reuse


jvican:
This is a ticket to keep track of collaboration with other tools in the Scala commuinty. Bloop should settle on a configuration format that can be read by any tool (from LSP servers, scalafix to other build tools).
  
As this format doesn't exist, we should probably pioneer one. Another possibility is to reuse `.eclipse` configuration files -- that are widely supported by Eclipse, Intellij and vscode (not natively, but some plugin extensions do) -- or directly generate `.idea` configuration files. However, we may want to encode more information in the file and it's not clear yet whether eclipse configuration files allow it. 
  
Supporting all this configuration files is easy, so we may also consider understanding all of these so that people that heavily use Intellij and rely on sbt-idea to import their projects do not need to run `install` in the build. This would mean there's a standard, preferred configuration file but bloop can also understand others. This is a nitpick and it's only UX work, so this won't happen anytime soon. But it's a thing to keep an eye on.
  
My ideal format is json because it can be edited and it's readable (unlike protobuf). This json format should have also a library that provides bincompat readers and writers so that other tools that want to read/write from it only need to depend on the library and use it.
  
When bloop is public, I'll start summoning people up in this ticket so that we converge to one solution and explore how useful this file would be, and how it could help make better tooling.

ShaneDelmore:
2 more cents, no more novelty.  The Clang community has settled on a format and quite a lot of tooling has sprung up around it.  I would love if we could do the same in the scala community.  It may make sense to re-use the same keys they use for our shared bits of information and then just enrich the file with additional information we need to help us interoperate with existing tooling.

https://sarcasm.github.io/notes/dev/compilation-database.html

My other thought, I would like for this file to contain enough information to be able to determine if artifacts built with these settings are still valid or if they need to be rebuilt.  I don't know all of the variables besides the scala source code, maybe it is just classpath, compiler flags, and scalac version?

propensive:
I know I brought this up in another ticket, but I'll reiterate here, as it's more relevant.

If the file format needs to be read by other tools, JSON is too heavyweight a dependency to require, as it becomes largely inaccessible to shell scripting (which is a particular use-case for Bloop). JSON can be read with `jq`, but it's not standard.

The existing key/value format works very well for shell scripting, as `grep`, `cut` and `tr` provide most of the functionality required for manipulating keys, even if they're rather blunt tools for the task.

[...]

Issue #329, opened 2018-03-12
The junit test framework doesn't output test logs by default


jvican:
If you do `bloop test frontend -f bloop.engine.DagSpec`, you'll see that the test execution doesn't output anything. However, the tests are *indeed* run! The only issue why it doesn't work is because the junit-interface framework needs `-v` (https://github.com/sbt/junit-interface) to show the logs, and that needs to be set manually via a test option. This test option is a default in sbt.
  
We'll fix this issue automatically when we have https://github.com/scalacenter/bloop/pull/291/ in, and when we set the default options for the junit test interface (which should be the same as in sbt).

PR #415, opened 2018-04-05
Add json file format and implement test options

jvican:
This pull request does the following things:

* Splits up `integrationCore` to an independent module.
* Uses the json config module in the integration infrastructure and build.
* Generates json files instead of java property files.
* Implements test options support.
* Adds a json schema and docs about how to generate a file.
* Improves the website with syntax highlighting and some extra style.
* Cleans up the internals to implement the aforementioned features.

Builds upon https://github.com/scalacenter/bloop/pull/291.

Fixes https://github.com/scalacenter/bloop/issues/58.


build-integrations/global/plugins.sbt
```diff
@@ -23,9 +32,19 @@ libraryDependencies := {
...
+      typesafeConfig,
+      metaconfigCore,
+      metaconfigDocs,
+      metaconfigConfig,
+      circeDerivation
     )
   } else {
-    List()
+    List(
+      typesafeConfig,
```
jvican: We need to do this for all the builds using 2.10 (sbt 0.13) because circe-derivation and metaconfig are not published to 2.10. As we only care about the encoders, we add circe core and circe generic and make the module cross-compile.

build-integrations/sbt-1.0/project/Integrations.scala
```diff
@@ -3,7 +3,7 @@ import sbt.{RootProject, uri}
 object Integrations {
 
   val SbtSbt = RootProject(
-    uri("git://github.com/scalacenter/sbt.git#c84db3a3969e269aeed0e9d9f2f384a0b029b82c"))
+    uri("git://github.com/scalacenter/sbt.git#fd66dfa1d40a01634f8ce65e299ce7a27cdb247c"))
```
jvican: This was updated to avoid compiling sbt when we generate the bloop config files.

frontend/src/main/scala/bloop/cli/CommonOptions.scala
```diff
@@ -23,17 +23,33 @@ case class CommonOptions(
...
+
+  // Our own version of properties in which we override `toString`
+  final class PrettyProperties extends Properties {
+    override def toString: String = synchronized {
+      super.keySet().toArray.map(_.toString).mkString(", ")
+    }
+  }
+
+  object PrettyProperties {
+    def from(p: Properties): PrettyProperties = {
```
jvican: We have `PrettyProperties` so that `toString` in `commonOptions` doesn't print the property values, only the keys, to avoid littering the screen.

jvican:
The docs use `docson` like bucklescript does to display the json schema interactively. The user can toggle fields to see more examples and navigate the whole structure. Sneak peek:

fd1689bb-df87-4ee5-92d7-aabeeb680324

bin/install.py
```diff
@@ -95,6 +95,7 @@ def coursier_bootstrap(target, main):
             "java", "-jar", BLOOP_COURSIER_TARGET, "bootstrap", BLOOP_ARTIFACT,
             "-r", "bintray:scalameta/maven",
             "-r", "bintray:scalacenter/releases",
+            "-r", "https://oss.sonatype.org/content/repositories/snapshots",
```
Duhemm: `"-r", "sonatype:snapshots",`

build.sbt
```diff
@@ -39,10 +38,41 @@ val backend = project
     )
   )
 
+// Needs to be called `jsonConfig` because of naming conflict with sbt universe...
```
Duhemm: You mean it'd conflict with `bloop-config`?

bin/install.py
jvican: Will this work? I copy pasted this line from the coursier README.

build.sbt
jvican: No, i mean that the name `config` is imported implicitly by sbt, so we cannot call it `config`, we need to prepend json.

jvican:
Fixes https://github.com/scalacenter/bloop/issues/329 too.

PR #421, opened 2018-04-07
Converting camel case to kebab case configuration flags under autocomplete

rberenguel:
Note: all the tests running `checkInvalidAddress` in `Validate.scala` are failing for me locally (even in a clean pull with no changes done). Not sure why it could be.

I got Case.scala from @olafurpg metaconfig, thanks, it's way neater than what I would have written.


rberenguel:
I think something on my fork is wrong (I'd blame the submodules, but can't see why), re-forking (also, the approach I took here won't work if autocompletions are autogenerated anyway)

jvican:
Yes, looks like the submodules are off. Did you do `git submodule update --init` right after the clone? That should make things work :wink:

rberenguel:
I knew I was forgetting something, thanks! Not very used to submodules. Will do for the next one, I'm trying another approach to avoid messing with zsh and bash separately in a new fork now :+1:

jvican:
Great, keep in mind that you can also use `hub` which wraps `git` to make sure all submodules are initialized when cloning, it's quite useful just in case you forget to do it yourself :smile:

PR #423, opened 2018-04-07
Fix/camel case autocompletion

rberenguel:
Thanks @olafurpg for the Kebabizer :)


frontend/src/main/scala/bloop/cli/completion/Case.scala
```diff
@@ -0,0 +1,22 @@
...
+    val m = Camel.pattern.matcher(camel)
+    val sb = new StringBuffer
+    while (m.find()) {
+      m.appendReplacement(sb, "-" + m.group().toLowerCase())
+    }
+    m.appendTail(sb)
+    sb.toString
+  }
+
+  def kebabizeArg(arg: Arg): Arg = arg.copy(extraNames =
```
jvican: Can we put the body of this method inside curly braces and move the implementation of `extraNames` to a new `val`? :smile:

rberenguel: Sure!

rberenguel:
@jvican actually I was thinking about that on Friday, I'm not sure how autocompletion could be tested completely (covering all possible shells). Any idea? I could create the issue if there is any possible approach

jvican:
@rberenguel Covering `bash` and `zsh` autocompletions is a good start. The good thing is that you can create a simple test suite for each one that uses the `Autocomplete` command defined in `Commands.scala`. You can have different input and simulate what would be the expected response.

PR #424, opened 2018-04-07
Fix/watch waiting for #422

### LINKED ISSUE #422: Bloop should indicate when --watch is waiting for source changes
I ran `bloop test --watch tests-test` in the metadoc repo and when the tests complete the last line that is printed out is

```
- .semanticdb.json
```
It would be nice if that line was followed with some message from bloop indicating that bloop is waiting for more source changes, something like

```
Waiting for source changes... (press ??? to interrupt)
```

rberenguel:
Log watching status after state change.
  
Fixes #422.


rberenguel:
@jvican but that log there makes no sense, since the screen is cleaned before the test is run, when it appears nothing is actually watched for change then. Also, the output from the tests would wipe it out so the situation would still be the same as what @olafurpg reports.

The only place that can show the message after the action has been executed and we are watching for the next changes (and not requiring some rewrite of the actual actions) is where I put it, unless there's some task chaining I'm not aware of with Monix (never used it, so may be the case)

jvican:
You're right Ruben, I thought we were clearing the screen right after the compilation is finished, but we only do it when there's a new event. :+1:

rberenguel:
Thanks @jvican awesome work you (I mean you and the rest of the contributors and maintainers) have with bloop. Not only because it's very fast (wish I'd known about it before!), but the code base is easy to follow and the documentation (for users and contributors) is very clear. Doesn't happen everywhere!

jvican:
@Duhemm and I are very happy to hear that. Thanks for the nice words!

PR #425, opened 2018-04-08
Generalize `sourceDirectories` to `sources`

jvican:
Our project abstraction only accepted source directories. This
limitation is unfounded and makes it difficult to support build tools
whose project definitions have no source directories. One example of
such a build tool is Bazel, whose `Target`s only know about source
files.

This patch:

1. Renames `sourceDirectories` in the source code.
2. Makes the rest of the code (`Tasks` and `SourceWatcher`) aware that
sources does not strictly mean directories.
3. Build upon @rberenguel's work to show the same "Watching ..." message
in every watch iteration.


jvican:
@Duhemm There's no new logic :wink:

PR #430, opened 2018-04-10
Simple improvement to zsh completion when server is not started

propensive:
This will hide the error if attempting a tab completion to bloop while
the bloop server has not been started, and will offer the single
completion of `bloop server` instead.

Someone might like to do the same thing for bash....

### DIFF
--- a/etc/bash/bloop
+++ b/etc/bash/bloop
@@ -5,13 +5,13 @@ _files() {
 }
 
 _projects() {
-    projects=$(bloop autocomplete --mode=projects --format=bash)
+    projects=$(bloop autocomplete --mode=projects --format=bash 2> /dev/null)
     COMPREPLY=( $(compgen -W "$projects" -- $cur) )
     return 0
 }
 
 _commands() {
-    commands=$(bloop autocomplete --mode=commands --format=bash)
+    commands=$(bloop autocomplete --mode=commands --format=bash 2> /dev/null || _notStarted)
     COMPREPLY=( $(compgen -W "$commands" -- $cur) )
     return 0
 }
@@ -60,6 +60,10 @@ _callCompletionForCommandAndFlag() {
 
     return 0
 
+}
+
+_notStarted() {
+    echo "server"
 }
 
 _bloop() {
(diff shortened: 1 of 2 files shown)


jvican:
@propensive Thanks for this! Could you try to do the same for `bash` or would that take you too much time?

jvican:
@rberenguel Can you take a peek at this PR and approve? I'm curious why the `2> /dev/null` is only needed in `_commands` and `_projects`, and not everywhere else too.

rberenguel:
Looks good @propensive nice approach

@jvican the `2> /dev/null` for commands in combination with the `notStarted` function ensure the first completion is for `server`... Which then behaves as a command. If you press tab twice without a server you'd get an error without the other `2> /dev/null` because it triggers `project` for command `server`. Since tags is optional, it does not trigger an error.

rberenguel:
I leave the merge for @Duhemm so he can also have a look (since he's also requested as reviewer)

rberenguel:
Btw I just checked and looks like the same approach works also for bash just as fine. @propensive feel like it?

jvican:
> Btw I just checked and looks like the same approach works also for bash just as fine. @propensive feel like it?
  
I think Jon is a little bit absent. Since you already tried @rberenguel, can you push your bash changes to this branch and then we merge :smile:? As a contributor, you should be able to checkout jon's repo and push to his branch. See https://github.com/github/hub.

rberenguel:
Yup, no problem (well, needs to be during lunch break)

rberenguel:
Could it be that "Allow edits from maintainers" is not enabled in this PR? I can't co the PR branch in Jon's fork, but can co it in a (fresh) clone (not a fork though?) of bloop. Sorry if this is evident, first time I have to do this :D

jvican:
It could be so @rberenguel, it's Jon the one that can control that. Let's merge this PR, you can follow up with another PR, what do you think? Push the merge button if you agree :wink:

Duhemm:
It looks like Jon pushed his branch directly to this repository.

rberenguel:
Oh, if I can push directly to the branch in this repository then I can submit the changes

jvican:
@rberenguel Thanks for completing this pull request Ruben! Merging.

propensive:
Sorry, was too busy today! Thanks for sorting it, everyone!

Issue #217, opened 2018-01-30
Implement good batching support for compilation in bsp


jvican:
See https://github.com/scalacenter/bloop/pull/153/files/74c8d7a3d08d5c0c34b994564f6b19303fb2c48d#diff-6a5568f260cb9c6ed0c7ff2ccc7e5f5e. I'll get to this soon, when I get back to #170.

jvican:
Problem statement: given a list of projects, we want Bloop to take into account the project graph to minimize the amount of compiler requests done.
  
For example, take the following structure:

* `A` -> `B`
* `B` -> `C`
* `C` -> `D`
  
Let's say we enable `compile` to accept a list of projects to compils like `bloop compile A B C D`. We want bloop to detect that the compilation of `B`, `C`, and `D` is subsumed by the one in `A`, and therefore only do 4 compiles.
  
The naive treatment of the previous command would be to compile `A` first, then `B`, then `C`, and lastly `D`, therefore trying to compile 10 times.

jvican:
Fixed in https://github.com/scalacenter/bloop/commit/c6f3f3c1011f6195e347901441a0710bc6d3d569.

PR #431, opened 2018-04-10
Continue the bsp implementation

jvican:
This pull request adds several compilation tests that check the correct
compilation of a target, as well as the potential errors showing up from an
invalid URI format.

It also improves the logging internals of BSP: now, logback is not used
anymore, and Bloop has its own slf4j logger adapter to wrap with scalalogging
and passes it in the lsp4s internals.

This pull request also fixes the handling of the root URI: before, it had to be
configuration directory `.bloop`, but that was wrong since the root URI must be
the base directory of the workspace. Now the detection of the configuration
directory is done automatically by Bloop.

/cc @jastice


jvican:
The latest additions to this PR also implement `buildTarget/dependencySources` and `buildTarget/scalacOptions`, which are added in this PR on scalacenter/bsp https://github.com/scalacenter/bsp/pull/5. It also adds tests for both new entrypoints.

backend/src/main/scala/bloop/io/AbsolutePath.scala
```diff
@@ -30,4 +32,8 @@ object AbsolutePath {
   def apply(path: String)(implicit cwd: AbsolutePath): AbsolutePath = apply(NioPaths.get(path))(cwd)
   def apply(path: Path)(implicit cwd: AbsolutePath): AbsolutePath =
     if (path.isAbsolute) new AbsolutePath(path) else cwd.resolve(path.toString)
+
+  // Necessary to test wrong paths in tests...
+  private[bloop] def completelyUnsafe(path: String): AbsolutePath =
```
rberenguel: Wouldn't `unsafe` or `toUnsafe` sound a bit less scary?

jvican: It's meant to be super scary, so that anyone but me uses it :stuck_out_tongue: This makes an absolute path out of something it's not an absolute path.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -73,42 +82,147 @@ final class BloopBspServices(callSiteState: State, client: JsonRpcClient, bspLog
...
     }
+  }
 
-    Interpreter.execute(action, Task.now(currentState)).map { state =>
-      currentState = state
-      val items = projectsToCompile.map(p => CompileReportItem(target = Some(p._1)))
-      Right(CompileReport(items))
+  def compile(params: CompileParams): BspResponse[CompileReport] = {
+    def compile(projects: Seq[ProjectMapping]): BspResponse[CompileReport] = {
+      // TODO(jvican): Naive approach, we need to implement batching here.
```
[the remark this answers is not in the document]
jvican: Batching in the sense that bloop needs to find out the smallest group of disjoint DAGs that includes a given list of nodes. This is for performance reasons, a very simple example is described in https://github.com/scalacenter/bloop/issues/217.

backend/src/main/scala/bloop/io/AbsolutePath.scala
rberenguel: Haha got it, didn't think it was _that_ scary

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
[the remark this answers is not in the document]
rberenguel: Oh I see. The problem somehow sounds familiar...

PR #432, opened 2018-04-10
Use Python3 in Homebrew formula

Duhemm:
Our homebrew formula download Coursier from the installation script.
Sometimes, this fails because it looks like some versions of MacOS have
an outdated version of Python 2 with a faulty SSL module (or something
along those lines).

This commit forces Python3 to be used for running the installation
script from the Homebrew formula, so that downloading Coursier no longer
fails.


jvican:
Fixes https://github.com/scalacenter/homebrew-bloop/issues/2#issuecomment-380053270.

PR #435, opened 2018-04-12
Report compilation logs via publishDiagnostics

jvican:
The previous implementation was sending the compilation infos as if they
were normal logs. This patch splits up the reporters in two: one used
for normal logging and the other one used for bsp. The bsp reporter will
`textDocument/publishDiagnostics` all the compilation infos instead of
forwarding them with `window/showMessage`.

### DIFF
--- a/backend/src/main/scala/bloop/reporter/DefaultReporterFormat.scala
+++ b/backend/src/main/scala/bloop/reporter/DefaultReporterFormat.scala
@@ -8,6 +8,11 @@ import scala.compat.Platform.EOL
 object DefaultReporterFormat extends (ConfigurableReporter => ReporterFormat) {
   override def apply(reporter: ConfigurableReporter): DefaultReporterFormat =
     new DefaultReporterFormat(reporter)
+
+  def toOption[T](m: java.util.Optional[T]): Option[T] = if (m.isPresent) Some(m.get) else None
+  def position(position: xsbti.Position): Option[(Int, Int)] = {
+    toOption(position.line).flatMap(l => toOption(position.pointer).map(c => (l, c)))
+  }
 }
 
 /**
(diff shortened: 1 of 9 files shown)


backend/src/main/scala/bloop/reporter/DefaultReporterFormat.scala
```diff
@@ -8,6 +8,11 @@ import scala.compat.Platform.EOL
 object DefaultReporterFormat extends (ConfigurableReporter => ReporterFormat) {
   override def apply(reporter: ConfigurableReporter): DefaultReporterFormat =
     new DefaultReporterFormat(reporter)
+
+  def toOption[T](m: java.util.Optional[T]): Option[T] = if (m.isPresent) Some(m.get) else None
+  def position(position: xsbti.Position): Option[(Int, Int)] = {
```
olafurpg: What is `(Int, Int)`? Is it start/end offsets? Or line/column?

olafurpg:
Is it possible to get range positions if -Yrangepos is enabled? Range positions are extra helpful in the editor.

backend/src/main/scala/bloop/reporter/DefaultReporterFormat.scala
jvican: line/column

olafurpg: I would introduce a case class for this instead of using tuple of ints. I would also make it clear if lines and columns start at 0 or 1

jvican: Yup, all this area needs to be improved when I’ve got some time, so it will be most likely introduced in a future PR

jvican:
@olafurpg I need to double check but IIRC the zinc api doesn’t expose this information. We’ll need to add it in a pr upstream sinc this is indeed critical

jastice:
with full ranges it should be possible to put red squiggles into IntelliJ's editor based on bsp build :)

PR #436, opened 2018-04-12
Use range positions in bsp if Zinc returns them

jvican:
The trick to get access to this info was in downcasting to the internal
position in Zinc's land that exposes `startPosition` and `endPosition`,
meaning the start and end of the column count.

With this information, Intellij is now able to show the red squigglies
in the editor.

I couldn't resist, thanks @olafurpg for the reminder and @jastice for getting
my excitement up with the comment on red squigglies.


jvican:
Merging so that tomorrow @jastice can use the version published by the release process, available in the CI release step.

PR #438, opened 2018-04-13
Improve autocompletion for zsh to be more context-aware for #356

### LINKED ISSUE #356: Add autocompletions for flags of commands when no project is required
Support for completions has just been merged in #353. However, it's not yet complete. One thing that is missing is that we don't autocomplete flags of commands *when no project is required*.
  
That is, `bloop compile foo -<TAB>` autocompletes, but `bloop configure -` or `bloop compile -` doesn't.

rberenguel:
Fixes #356.

### DIFF
--- a/etc/zsh/_bloop
+++ b/etc/zsh/_bloop
@@ -53,11 +53,11 @@ _flags_for_cmd() {
 }
 
 _project_or_flags() {
-    local project_bound_cmd=( clean compile run test console )
+    local project_cmd=($(bloop autocomplete --format zsh --mode project-commands))
     local cmd=${words[2]}
     # For project related commands, try to complete project. If not in a project
     # folder, switch to flags. Otherwise, complete flags
-    if  [[ $CURRENT == 3 ]] && [[ -n "${project_bound_cmd[(r)$cmd]}" ]]; then
+    if  [[ $CURRENT == 3 ]] && [[ -n "${project_cmd[(r)$cmd]}" ]]; then
         local projs=($(bloop autocomplete --format zsh --mode projects 2> /dev/null))
         if  [[ "$projs" != " " && "$projs" != "" ]]; then
             _projects
--- a/etc/zsh/_bloop
+++ b/etc/zsh/_bloop
@@ -37,6 +37,11 @@ _mainsfqcn() {
 }
 
 _flags() {
+    local cmd="$words[1]"
+    _flags_for_cmd $cmd
+}
+
+_flags_for_cmd() {
     local cmd=$1
     local saveIFS=$IFS
     IFS=$'\n'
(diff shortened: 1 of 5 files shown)


etc/zsh/_bloop
```diff
@@ -25,30 +27,53 @@ _boolean() {
...
+    local flags=($(bloop autocomplete --format zsh --mode flags --command $cmd))
+    flags=($flags ':approximate-4:::' )
     IFS=$saveIFS
+    _arguments $flags
+    return 0
+}
 
-    _arguments $results
+_project_or_flags() {
+    local project_bound_cmd=( clean compile run test console )
```
jvican: Would it be possible to get the list of commands that take a project from `bloop autocomplete` too? :smile:

rberenguel:
@jvican yep, done (much better now that Scala handles that). Not very happy with how the filter-map combo look, but seems to be the most seamless way to write this. Any suggestions to make it look better?

jvican:
Fixes #356.

jvican:
@rberenguel Btw, just a tip: you can use "Fixes #ticket-number" to tell GitHub to close the ticket when this PR is merge.

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -213,9 +214,15 @@ object Interpreter {
     }
   }
 
+  private val projectBound = CommandsMessages.messages.filter {
+    case (name, CommandMessages(args, _)) => args.exists(_.name == "project")
+  }.map(_._1).mkString(" ")
```
Duhemm: The filter-map combo can be replaced with `collect`:

```scala
private val projectBound = CommandsMessages.messages.filter {
  case (name, CommandMessages(args, _)) if args.exists(_.name == "projects") => name
}.mkString(" ")
```

rberenguel: Thanks @Duhemm ! I knew there was to be something in the library to do this in one go

rberenguel:
Any idea what might be up with the Windows test? `bloop.nailgun.BasicNailgunSpec` has failed twice in Windows (last two commits) but passes fine in Linux

jvican:
> Oh thanks @jvican (btw had a facepalm with the put in a val :D ) Do you think it would make sense to actually place it in Commands.scala?
  
Yeah :smile:, I think that's the perfect place.
  
> Any idea what might be up with the Windows test? bloop.nailgun.BasicNailgunSpec has failed twice in Windows (last two commits) but passes fine in Linux
  
It's an spurious error, we just restarted the build :+1:

rberenguel:
Moved the list of project commands to Commands. Next one to tackle will be adding (scala) tests for autocompletion :)

jvican:
Awesome, merging :smile:

PR #439, opened 2018-04-15
Notify client when the server starts listening

jvican:
The client needs to know when it can connect to the socket because the
socket file needs to be created by the server.

### DIFF
--- a/frontend/src/test/scala/bloop/bsp/BspClientTest.scala
+++ b/frontend/src/test/scala/bloop/bsp/BspClientTest.scala
@@ -1,6 +1,7 @@
 package bloop.bsp
 
 import java.nio.file.Files
+import java.util.concurrent.TimeUnit
 
 import bloop.cli.Commands
 import bloop.io.AbsolutePath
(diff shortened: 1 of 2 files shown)


frontend/src/test/scala/bloop/bsp/BspClientTest.scala
```diff
@@ -104,12 +106,23 @@ object BspClientTest {
...
-      val initializeServer = endpoints.Build.initialize.request(
-        InitializeBuildParams(
-          rootUri = cwd.toAbsolutePath.toString,
-          Some(BuildClientCapabilities(List("scala")))
-        )
-      )
+      val initializeServer = {
+        def messages = logger0.underlying.getMessages().filter(_._1 == "info")
+        def waitForLiveServer: Task[Unit] = Task(()).flatMap { _ =>
+          if (!messages.exists(_._2.startsWith("The server started to listen")))
```
rberenguel: This may be a personal preference, but I would create a string in BspServer with `The server started to listen` and use it here and when building the message string, to avoid a minor change of this string from breaking anything

jvican: I'd do the same if this was the main code, but it's just test sources :smile:, so I didn't really put too much thinking into it. I've found that inlining the expected messages in the test sources makes it easier to add more tests and fix them if they regress.

jvican:
I'm closing this, it doesn't seem the best strategy to notify the client that the server is up because of issues with system process buffering and how it varies across different operating systems.

Issue #337, opened 2018-03-13
Add docs to set up systemd to run bloop server


jvican:
It would be great if Linux/*BSD users have a template they can copy-paste to install a systemd service in their machine that would run the bloop server in the background. This would be quite a big improvement in terms of usability for those users (e.g. me) because it would mean they don't need to manually handle the creation of the server themselves, and they can keep track of their lifetime over time.
  
Relevant links:
  
1. https://wiki.archlinux.org/index.php/systemd
2. https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/sect-managing_services_with_systemd-unit_files

jvican:
Another avenue we may want to consider is using desktop entries: https://wiki.archlinux.org/index.php/desktop_entries. We can still consider having both ways of starting up a service, but the good thing is that this mechanism works across different Linux distributions.
  
A desktop entry looks like the following one, extracted from jetbrains-toolbox:
  
```
[Desktop Entry]
Type=Application
Name=JetBrains Toolbox
Exec=/opt/jetbrains-toolbox/jetbrains-toolbox --minimize
Icon=/home/jvican/.local/share/JetBrains/Toolbox/toolbox.svg
StartupNotify=false
Categories=Development;IDE;
Terminal=false
X-GNOME-Autostart-enabled=true
StartupWMClass=jetbrains-toolbox
MimeType=x-scheme-handler/jetbrains;
Hidden=true
```
  
The good thing about desktop entries is that they are automatically started by the desktop environment on startup if they are XDG compliant. The desktop entries need to be in `$XDG_CONFIG_DIR/autostart`.

jvican:
Kudos to @tues for picking this up :tada:.

tues:
Hi @jvican, thanks for merging #440 (and thanks to the reviewers) 🙂

What about https://github.com/tues/bloop/tree/ticket/337-xdg? Do you want me to open a PR for that one as well?
I guess it's not very important to have both a systemd service and a desktop entry, but in case anyone is interested it looks like this: https://imgur.com/CU9mz3L

jvican:
@tues it looks great! Didn’t know that Ubuntu shows desktop entries in the launcher. Can you PR the changes? 😄 it may be worth it explaining in which cases users should prefer desktop entries over systemd .

tues:
Created PR.

KDE does the same and other DEs should do as well :slightly_smiling_face: 

[...]

PR #440, opened 2018-04-15
Add systemd service file

tues:
Adds a user service file for systemd and makes `install.py` download it
automatically to `~/.bloop/systemd/bloop.service`.
Updates the documentation with steps needed to enable this service.

Fixes systemd part of #337. A Desktop Entry solution will follow (most likely).

### DIFF
--- a/bin/install.py
+++ b/bin/install.py
@@ -12,10 +12,8 @@ import re
 IS_PY2 = sys.version[0] == '2'
 if IS_PY2:
     from urllib import urlretrieve as urlretrieve
-    from urllib import urlopen as urlopen
 else:
     from urllib.request import urlretrieve as urlretrieve
-    from urllib.request import urlopen as urlopen
 
 # INSERT_INSTALL_VARIABLES
 BLOOP_DEFAULT_INSTALLATION_TARGET = join(expanduser("~"), ".bloop")
@@ -86,29 +84,29 @@ SYSTEMD_SERVICE_TARGET = join(SYSTEMD_SERVICE_DIR, "bloop.service")
 
 BLOOP_ARTIFACT = "ch.epfl.scala:bloop-frontend_2.12:%s" % BLOOP_VERSION
 
+def download(url, target):
+    try:
+        urlretrieve(url, target)
+    except IOError:
+        print("Couldn't download %s, please try again." % url)
+        sys.exit(1)
+
+def download_and_install(url, target, permissions=0o644):
+    download(url, target)
+    os.chmod(target, permissions)
+
 def replace_template_variables(template):
     return template.replace("__BLOOP_INSTALLATION_TARGET__", BLOOP_INSTALLATION_TARGET)
 
-def download_and_install(url, target):
-    try:
-        urlretrieve(url, target)
-        os.chmod(target, 0o755)
-    except IOError:
-        print("Couldn't download %s, please try again." % url)
-        sys.exit(1)
-
-def download_and_install_template(url, target):
-    try:
-        socket = urlopen(url)
-        template = socket.read()
-        with open(target, "w") as file:
-            output = replace_template_variables(template)
-            file.write(output)
-        socket.close()
-        os.chmod(target, 0o755)
-    except IOError:
-        print("Couldn't download %s, please try again." % url)
-        sys.exit(1)
+def download_and_install_template(url, target, permissions=0o644):
+    template_target = target + ".tmp"
+    download(url, template_target)
+    with open(template_target, "r") as template_file, open(target, "w") as output_file:
+        template = template_file.read()
+        output = replace_template_variables(template)
+        output_file.write(output)
+    os.remove(template_target)
+    os.chmod(target, permissions)
 
 def coursier_bootstrap(target, main):
     try:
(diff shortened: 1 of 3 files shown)


etc/systemd/bloop.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Bloop Scala build server
+
+[Service]
+ExecStart=%h/.bloop/bloop server
```
Duhemm: Should we assume that `bloop` is in the `$PATH` and just use `bloop server` here?

I'm asking because we support `--destination` in the installation script, which will override the directory where Bloop is installed.

bin/install.py
```diff
@@ -126,3 +130,6 @@ def makedir(directory):
 
 download_and_install(BASH_COMPLETION_URL, BASH_COMPLETION_TARGET)
 print("Installed Bash completion in '%s'" % BASH_COMPLETION_TARGET)
+
+download_and_install(SYSTEMD_SERVICE_URL, SYSTEMD_SERVICE_TARGET)
```
jvican: Can we do this if it's only Linux/BSD?

etc/systemd/bloop.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Bloop Scala build server
+
+[Service]
```
jvican: Can we redirect logs to a file? Maybe `/var/log` if we can or a log file called `server.log` in the bloop installation target?

etc/systemd/bloop.service
rberenguel: Seems better to expect it to be on the path, otherwise we need to get the installation script to template this file with the chosen location (which sounds good as a solution too)

tues: True, I haven't noticed `--destination` when I started working on this. OTOH I don't think we can expect to have `bloop` in `$PATH` because systemd services don't use `.bashrc` etc., so a template seems to be the only solution. We'll also need templates for Desktop Entries anyway because we can't use environment variables at all there.

bin/install.py
tues: Is there any reason to do this only for Linux/BSD? Bash and zsh completions are downloaded even if you don't have bash/zsh installed.

etc/systemd/bloop.service
tues: I think the standard way to do this in `systemd` is to redirect `stdout`/`stderr` to system log and let it be handled by `journalctl`. The issue with `/var/log` is that the user running Bloop most likely won't have permissions to write there.

etc/systemd/bloop.service
tues: Alternatively, we could try to add something like `BLOOP_PATH` to `~/.config/environment.d/`, which should be visible to the service and we could then use it in place of `%h/.bloop`. However, it's not that much different from using a template for `bloop.service`.

I don't think we should change `PATH` automatically because that's something that some users might not want to do.

etc/systemd/bloop.service
tues: Please see https://github.com/scalacenter/bloop/pull/440/commits/e2a4ff525da6a92e95598aa04d3f5eeb98592ee6

tues:
I've added `download_and_install_template` to `install.py` and used it for `bloop.service` in place of normal `download_and_install`. It resulted in some duplicated code, but we don't want to download `coursier` or `nailgun` using `urlopen`, so I've kept `urlretrieve` for these.

tues:
> Why not urllib here? urllib2 does not have retrieve but both have open, I'd guess they do what we want the same and is more consistent

`urllib2` doesn't silently ignore HTTP errors, but yes, I guess it's better to replace it with `urllib` for consistency and worry about error handling later.

> Maybe it's better to use {bloop_installation_target} in the template and then use template.format(bloop_installation_target=BLOOP_INSTALLATION_TARGET) (takes advantage of format to do the work for us)

This is an interesting idea but it would mean that we need to remember to double any literal `{` and `}` in the templates, which I think is a bit annoying and risky. Currently we're not using these characters in `bloop.service`, but we might use them quite often in other files in the future. What do you think?

rberenguel:
Dang, you are right @tues , we might have issues with braces later on. It's just that I inherited quite a lot of "non-quite standard templating systems" in Python and other languages and I've found that in the long run it pays to just use format. If we name arguments the script will fail outright on creating the template if there are unmatched braces, so it should be relatively safe. Do you have any preference @jvican @Duhemm ?

tues:
Created a new branch for the Desktop Entry stuff: https://github.com/tues/bloop/commit/0fd7655cc9b7896e4de303107286df8a0df68811
Please let me know if you want it included as well.

bin/install.py
```diff
@@ -89,6 +97,19 @@ def download_and_install(url, target):
         print("Couldn't download %s, please try again." % url)
         sys.exit(1)
 
+def download_and_install_template(url, target):
```
valencik: I think you could reuse the existing `download_and_install` and then modify the template once on disk (inside your `replace_template_variables`) instead of introducing socket handling and new urllib code

jvican:
Hey guys, I'll get back to this PR probably tomorrow in a break, I'm a little bit busy preparing our presentation for Scalasphere. Thanks for your work here @tues, it's excellent.

jvican:
I was fiddling with this PR, the documentation and instructions to get running are :100:. One thing I've noticed though is that I get different results if I run with python2 or python3:
  
```
jvican in /data/rw/code/scala/loop                                                                                                                                                                [15:13:41] 
> $ python2 /data/rw/code/scala/loop/bin/install.py --dest ~/.bloop --nailgun 615737a9408654744bc066fa6c3cb938f3f139c6 --version 1.0.0-M9                                     [±tues-ticket/337]
Installed bloop server in '/home/jvican/.bloop/blp-server'
Installed bloop client in '/home/jvican/.bloop/bloop'
Installed zsh completion in '/home/jvican/.bloop/zsh/_bloop'
Installed Bash completion in '/home/jvican/.bloop/bash/bloop'
Installed systemd service in '/home/jvican/.bloop/systemd/bloop.service'
                                                                                                                                                                                                             
jvican in /data/rw/code/scala/loop                                                                                                                                                                [15:15:16] 
> $ python3 /data/rw/code/scala/loop/bin/install.py --dest ~/.bloop --nailgun 615737a9408654744bc066fa6c3cb938f3f139c6 --version 1.0.0-M9                                    [±tues-ticket/337]
Installed bloop server in '/home/jvican/.bloop/blp-server'
Installed bloop client in '/home/jvican/.bloop/bloop'
Installed zsh completion in '/home/jvican/.bloop/zsh/_bloop'
Installed Bash completion in '/home/jvican/.bloop/bash/bloop'
Couldn't download https://raw.githubusercontent.com/scalacenter/bloop/2e43fa5a/etc/systemd/bloop.service, please try again.
```
  
@tues Do you know what could be happening? I was surprised this was working at all in python2 because the bloop.service is not yet available in the GitHub repo URLs.

bin/install.py
jvican: This is a good idea :+1:

tues:
Sorry for the delay, new version:

* Changed `download_and_install_template` so it uses the same HTTP code as `download_and_install`
* Fixed file permissions
* Replaced `urlretrieve` with `urlopen` to make sure we don't ignore HTTP errors
* Rebased on top of `master`

Unfortunately it seems it's still failing on Windows. I'm not even sure it makes any sense to run the tests, I don't think they do anything with `install.py`? Maybe I should add `[DOC]`?

PR #443, opened 2018-04-16
Speed up first bloop server load

jvican:
The first time the server starts up is a little bit slow (2s for the bloop
project, 3s for the spark project) because Bloop needs to load up the analysis
file from disk. This process was sequential. This patch makes it parallel, and
also runs the whole load process in the io executor which is thread-unbounded.


jvican:
The bloop build now takes "[D] Elapsed: 487.909977 ms", that's a 1.5s cut from the previous version.

PR #449, opened 2018-04-20
Revamp the test and run fork process infrastructure

jvican:
1. Migrate to `NuProcess` and avoid the annoying Java and Scala process APIs.
Nuprocess gives us a better foundation to deal with stdout and stderr, as well
as processing standard input from the host process (Bloop). `NuProcess` will
also be more efficient than our previous approach.

2. Migrate test server and the whole `runMain` machinery to `Task` so that we
can have a finer control over it. Note that now these actions can be cancelled!


jvican:
The logic in this pull request also makes Bloop able to cancel tests and applications run reliably.

jvican:
Looks like CI fails because of an annoying bug in Docker. `--privileged` doesn't enable one of the permissions we need to efficiently handle processes with nuprocess. I'm working around it temporarily.

jvican:
Modifying the drone test agents to allow SYS_ADMIN seccomp capabilities fixed this issue. CI is now passing.

frontend/src/main/scala/bloop/Cli.scala
```diff
@@ -257,7 +257,6 @@ object Cli {
         .flatMap(start => taskState.materialize.map(s => (s, start)))
         .map { case (state, start) => logElapsed(start); state }
         .dematerialize
-        .executeWithOptions(_.enableAutoCancelableRunLoops)
```
jvican: We don't need cancel to propagate through the current flatMap stack, so we disable this. I've also experienced some weird Monix behaviour when this is enabled. Since this is not a well-tested feature and it doesn't make much sense to have it enabled by default, I'm more comfortable with using the stock task options.

jvican:
Oh, and I forgot to say that now we're handling input streams correctly (which means that run applications that require stdin will work).

PR #451, opened 2018-04-20
Add freedesktop.org-compliant Desktop Entry

tues:
Second part of #337.

Adds a `bloop.desktop` file and an accompanying `bloop.png` icon which can be used in freedesktop.org-compliant desktop environments as either a simple shortcut or as an autostart item. The desktop entry is made executable to let users run it without issues.
Updates `install.py` and user documentation as well.

The icon was generated with:
`$ inkscape -z -e etc/xdg/bloop.png -w 256 -h 256 website/static/img/logo.svg`


etc/xdg/bloop.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
```
jvican: Do you know if having this desktop entry installed will conflict with a current system installation? It would also probably be worth it to make sure that users don't mistakenly install both systemd and desktop entry's files. Would it be possible to add some docs about that? :smile:

tues: No, just having them in `.config/systemd` and `.local/share/applications` respectively shouldn't cause any issues.

Enabling both to start automatically won't do too much harm, as the second one to start will simply throw an error and exit:
```
$ bloop server
java.net.BindException: Address already in use (Bind failed)
        at java.net.PlainSocketImpl.socketBind(Native Method)
        at java.net.AbstractPlainSocketImpl.bind(AbstractPlainSocketImpl.java:387)
        at java.net.ServerSocket.bind(ServerSocket.java:375)
        at java.net.ServerSocket.<init>(ServerSocket.java:237)
        at com.martiansoftware.nailgun.NGServer.run(NGServer.java:384)
        at bloop.Server$.run(Server.scala:30)
        at bloop.Server$.main(Server.scala:13)
        at bloop.Server.main(Server.scala)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at coursier.Bootstrap.main(Bootstrap.java:428)
Command not found: server
```

I will add a warning at the beginning of the "Automatic startup" section that users should pick only one method.

Of course we can also forget the Desktop Entry approach and just assume most people run systemd.

jvican: Systemd is not that popular these days anyway, I'd expect people to benefit from desktop entries too. So I incline towards merging this PR and just making it clear in the docs that one of the two methods should be chosen.

jvican:
Thanks @tues :smile:

PR #457, opened 2018-04-30
Support linking and running of Scala Native projects

Duhemm:
I have an implementation that also supports testing of Scala Native projects; however it needs the test runner to be split out of sbt-scala-native (see scala-native/scala-native#1234).

Also missing from this PR: extracting the other options relevant only for Scala Native (GC to use, release vs debug mode, etc.)

This is already large enough for one PR, I think it'd be better to open another PR for the rest.

### DIFF
--- a/config/src/main/scala/bloop/config/Config.scala
+++ b/config/src/main/scala/bloop/config/Config.scala
@@ -75,6 +75,30 @@ object Config {
     }
   }
 
+  case class NativeConfig(
+      toolchainClasspath: Array[Path],
+      gc: String,
+      clang: Path,
+      clangPP: Path,
+      linkingOptions: Array[String],
+      compileOptions: Array[String],
+      targetTriple: String,
+      nativelib: Path,
+      linkStubs: Boolean
+  )
+
+  object NativeConfig {
+    private[bloop] val empty = NativeConfig(Array.empty,
+                                            "",
+                                            emptyPath,
+                                            emptyPath,
+                                            Array.empty,
+                                            Array.empty,
+                                            "",
+                                            emptyPath,
+                                            false)
+  }
+
   case class Project(
       name: String,
       directory: Path,
@@ -91,7 +115,7 @@ object Config {
       java: Java,
       test: Test,
       platform: Platform,
-      nativeClasspath: Array[Path]
+      nativeConfig: Option[NativeConfig]
   )
 
   object Project {
@@ -99,7 +123,7 @@ object Config {
     private[bloop] val empty: Project =
       Project("", emptyPath, Array(), Array(), Array(), ClasspathOptions.empty,
         CompileOptions.empty, emptyPath, emptyPath, emptyPath, Scala.empty, Jvm.empty, Java.empty,
-        Test.empty, Platform.default, Array())
+        Test.empty, Platform.default, None)
     // FORMAT: ON
 
     def analysisFileName(projectName: String) = s"$projectName-analysis.bin"
@@ -153,7 +177,7 @@ object Config {
         Java(Array("-version")),
         Test(Array(), TestOptions(Nil, Nil)),
         Platform.default,
-        Array.empty
+        None
       )
 
       File(LatestVersion, project)
--- a/config/src/main/scala/bloop/config/Config.scala
+++ b/config/src/main/scala/bloop/config/Config.scala
@@ -59,6 +59,22 @@ object Config {
     private[bloop] val empty: Scala = Scala("", "", "", Array(), Array())
   }
 
+  sealed abstract class Platform(val name: String)
+  object Platform {
+    private[bloop] val default: Platform = JVM
+
+    case object JS extends Platform("JS")
+    case object JVM extends Platform("JVM")
+    case object Native extends Platform("Native")
+
+    def apply(platform: String): Platform = platform match {
+      case JS.name => JS
+      case JVM.name => JVM
+      case Native.name => Native
+      case _ => throw new IllegalArgumentException(s"Unknown platform: '$platform'")
+    }
+  }
+
   case class Project(
       name: String,
       directory: Path,
@@ -73,13 +89,16 @@ object Config {
       `scala`: Scala,
       jvm: Jvm,
       java: Java,
-      test: Test
+      test: Test,
+      platform: Platform
   )
 
   object Project {
     // FORMAT: OFF
     private[bloop] val empty: Project =
-      Project("", emptyPath, Array(), Array(), Array(), ClasspathOptions.empty, CompileOptions.empty, emptyPath, emptyPath, emptyPath, Scala.empty, Jvm.empty, Java.empty, Test.empty)
+      Project("", emptyPath, Array(), Array(), Array(), ClasspathOptions.empty,
+        CompileOptions.empty, emptyPath, emptyPath, emptyPath, Scala.empty, Jvm.empty, Java.empty,
+        Test.empty, Platform.default)
     // FORMAT: ON
 
     def analysisFileName(projectName: String) = s"$projectName-analysis.bin"
@@ -131,7 +150,8 @@ object Config {
         Scala("org.scala-lang", "scala-compiler", "2.12.4", Array("-warn"), Array()),
         Jvm(Some(Paths.get("/usr/lib/jvm/java-8-jdk")), Array()),
         Java(Array("-version")),
-        Test(Array(), TestOptions(Nil, Nil))
+        Test(Array(), TestOptions(Nil, Nil)),
+        Platform.default
       )
 
       File(LatestVersion, project)
--- a/config/src/main/scala/bloop/config/Config.scala
+++ b/config/src/main/scala/bloop/config/Config.scala
@@ -63,11 +63,11 @@ object Config {
   object Platform {
     private[bloop] val default: Platform = JVM
 
-    case object JS extends Platform("JS")
-    case object JVM extends Platform("JVM")
-    case object Native extends Platform("Native")
+    case object JS extends Platform("js")
+    case object JVM extends Platform("jvm")
+    case object Native extends Platform("native")
 
-    def apply(platform: String): Platform = platform match {
+    def apply(platform: String): Platform = platform.toLowerCase match {
       case JS.name => JS
       case JVM.name => JVM
       case Native.name => Native
--- a/config/src/main/scala/bloop/config/Config.scala
+++ b/config/src/main/scala/bloop/config/Config.scala
@@ -75,6 +75,11 @@ object Config {
     }
   }
 
+  /**
+   * Configures how to start and use the Scala Native toolchain, if needed.
+   * For the description of these fields, see:
+   * http://static.javadoc.io/org.scala-native/tools_2.10/0.3.7/index.html#scala.scalanative.build.Config
+   */
   case class NativeConfig(
       toolchainClasspath: Array[Path],
       gc: String,
@@ -88,15 +93,11 @@ object Config {
   )
 
   object NativeConfig {
-    private[bloop] val empty = NativeConfig(Array.empty,
-                                            "",
-                                            emptyPath,
-                                            emptyPath,
-                                            Array.empty,
-                                            Array.empty,
-                                            "",
-                                            emptyPath,
-                                            false)
+    // FORMAT: OFF
+    private[bloop] val empty: NativeConfig =
+      NativeConfig(Array.empty, "", emptyPath, emptyPath, Array.empty, Array.empty, "", emptyPath,
+        false)
+    // FORMAT: ON
   }
 
   case class Project(
--- a/config/src/main/scala/bloop/config/Config.scala
+++ b/config/src/main/scala/bloop/config/Config.scala
@@ -90,7 +90,8 @@ object Config {
       jvm: Jvm,
       java: Java,
       test: Test,
-      platform: Platform
+      platform: Platform,
+      nativeClasspath: Array[Path]
   )
 
   object Project {
@@ -98,7 +99,7 @@ object Config {
     private[bloop] val empty: Project =
       Project("", emptyPath, Array(), Array(), Array(), ClasspathOptions.empty,
         CompileOptions.empty, emptyPath, emptyPath, emptyPath, Scala.empty, Jvm.empty, Java.empty,
-        Test.empty, Platform.default)
+        Test.empty, Platform.default, Array())
     // FORMAT: ON
 
     def analysisFileName(projectName: String) = s"$projectName-analysis.bin"
@@ -151,7 +152,8 @@ object Config {
         Jvm(Some(Paths.get("/usr/lib/jvm/java-8-jdk")), Array()),
         Java(Array("-version")),
         Test(Array(), TestOptions(Nil, Nil)),
-        Platform.default
+        Platform.default,
+        Array.empty
       )
 
       File(LatestVersion, project)
--- a/frontend/src/main/scala/bloop/cli/Commands.scala
+++ b/frontend/src/main/scala/bloop/cli/Commands.scala
@@ -158,4 +158,19 @@ object Commands {
       watch: Boolean = false,
       @Recurse cliOptions: CliOptions = CliOptions.default
   ) extends CompilingCommand
+
+  case class NativeLink(
+      @ExtraName("p")
+      @HelpMessage("The project to run (will be inferred from remaining cli args).")
+      project: String = "",
+      @ExtraName("m")
+      @HelpMessage("The main class to link. Leave unset to let bloop select automatically.")
+      main: Option[String] = None,
+      @HelpMessage("Pick reporter to show compilation messages. By default, bloop's used.")
+      reporter: ReporterKind = BloopReporter,
+      @ExtraName("w")
+      @HelpMessage("If set, run the command whenever projects' source files change.")
+      watch: Boolean = false,
+      @Recurse cliOptions: CliOptions = CliOptions.default
+  ) extends CompilingCommand
 }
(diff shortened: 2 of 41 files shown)


frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -156,4 +158,19 @@ object Commands {
...
+
+  case class NativeLink(
+      @ExtraName("p")
+      @HelpMessage("The project to run (will be inferred from remaining cli args).")
+      project: String = "",
+      @ExtraName("m")
+      @HelpMessage("The main class to link. Leave unset to let bloop select automatically.")
+      main: Option[String] = None,
+      @HelpMessage("Pick reporter to show compilation messages. By default, bloop's used.")
+      reporter: ReporterKind = BloopReporter,
```
jvican: Why do we need the reporter here?

frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -156,4 +158,19 @@ object Commands {
...
+      @ExtraName("w")
+      @HelpMessage("If set, run the command whenever projects' source files change.")
+      watch: Boolean = false,
```
jvican: Do we really need `watch` here? If so, can we add tests for it?

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -270,6 +273,39 @@ object Interpreter {
...
+      case Some(project) =>
+        val reporter = ReporterKind.toReporterConfig(cmd.reporter)
+        def doRun(state: State): Task[State] = {
+          compileAnd(state, project, reporter, false, sequential, "`nativeLink`") { state =>
+            cmd.main.orElse(getMainClass(state, project)) match {
+              case None => Task(state.mergeStatus(ExitStatus.RunError))
+              case Some(main) =>
+                Task {
+                  val out = ScalaNative.nativeLink(project, main, state.logger)
+                  state.logger.info(s"Produced $out")
```
jvican: Can we have a more descriptive log?

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
```diff
@@ -0,0 +1,88 @@
...
+    initializeToolChain(logger)
+
+    val bridgeClazz = nativeClassLoader.loadClass("bloop.scalanative.NativeBridge")
+    val paramTypes = classOf[Project] :: classOf[String] :: classOf[Logger] :: Nil
+    val nativeLinkMeth = bridgeClazz.getMethod("nativeLink", paramTypes: _*)
+
+    // The Scala Native toolchain expects to receive the module class' name
+    val fullEntry = if (entry.endsWith("$")) entry else entry + "$"
+
+    AbsolutePath(nativeLinkMeth.invoke(null, project, fullEntry, logger).asInstanceOf[Path])
```
jvican: Can we protect us here from a potential null returned by `invoke` and only instantiate an absolute path if we're 100% sure the reflective execution didn't fail?

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
```diff
@@ -0,0 +1,88 @@
...
+   * @param main    The fully qualified main class name.
+   * @param args    The arguments to pass to the program.
+   * @return A task that links and run the project.
+   */
+  def run(state: State,
+          project: Project,
+          cwd: AbsolutePath,
+          main: String,
+          args: Array[String]): Task[State] = Task {
+    import scala.collection.JavaConverters.propertiesAsScalaMap
```
jvican: Could you rebase this on top of https://github.com/scalacenter/bloop/pull/449? Also, can we not reuse as much as possible the stock implementation of `run`? This seems to boilerplate-y :smile:

frontend/src/main/scala/bloop/cli/Commands.scala
tindzk: Yes, at least for Scala.js `--watch` is desirable because after every source change a new JavaScript file would be assembled. sbt supports the same with `~fastOptJS`.

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
tindzk: I think it is cleaner to pass the target path to `nativeLink()`. Then, we can ignore the result altogether.

frontend/src/main/scala/bloop/cli/Commands.scala
tindzk: It is needed because the link command may trigger a compilation pass.

tindzk:
@jvican I share your concerns about resolving the bridges at runtime. At the moment, bridges depend on the front-end and on the back-end JARs. I am creating a Docker image for Bloop with Scala.js support that can be easily used in a CI like Drone. With the current design, packaging the JARs in `.bloop` does not suffice. I also need to include `.ivy2`, so eventually end up with two JARs for the front-end and back-end each.

It would be appreciated if we could get rid of the Coursier dependency in Bloop as resolving dependencies should be ideally handled externally.

What do you think about making the bridges part of the front-end for now? We could modularise the code in a future pull request. As I understood, the reasoning behind modularising the bridges was that the dependencies on Scala Native and Scala.js should be opt-in, and that we need to support multiple versions like Scala.js 0.6 and 1.0. Is this correct?

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -59,6 +59,46 @@ object Config {
...
+      clangPP: Path,
+      linkingOptions: Array[String],
+      compileOptions: Array[String],
+      targetTriple: String,
+      nativelib: Path,
+      linkStubs: Boolean
+  )
+
+  object NativeConfig {
+    private[bloop] val empty = NativeConfig(Array.empty,
```
jvican: Nitpick: can we move this to only one line like we do in `Project.empty`?

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -271,6 +275,40 @@ object Interpreter {
     state
   }
 
+  private def link(cmd: Commands.Link, state: State, sequential: Boolean): Task[State] = {
+    state.build.getProjectFor(cmd.project) match {
+      case None =>
+        Task(reportMissing(cmd.project :: Nil, state))
+      case Some(project) if project.platform != Platform.Native =>
+        Task {
+          state.logger.error("This command can only be called on a Scala Native project.")
```
jvican: When you add Scala.js support, can we improve this message and have it be like: "Expected `$PROJECT_NAME` to be either a Scala.js or Scala Native project"?

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  private def toClassLoader(classpath: Array[AbsolutePath]): ClassLoader = {
+    val parent = this.getClass.getClassLoader
+    val entries = classpath.map(_.underlying.toUri.toURL)
+    new URLClassLoader(entries, parent)
+  }
+
+}
+
+class ScalaNative private (classLoader: ClassLoader) {
```
jvican: Can we move the class before the companion?

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
```diff
@@ -0,0 +1,130 @@
...
+          main: String,
+          args: Array[String],
+          optimize: OptimizerConfig): Task[State] = {
+    link(project, main, state.logger, optimize).flatMap {
+      case Success(nativeBinary) =>
+        val cmd = nativeBinary.syntax +: args
+        Forker.run(cwd, cmd, state.logger, state.commonOptions).map { exitCode =>
+          val exitStatus = {
+            if (exitCode == Forker.EXIT_OK) ExitStatus.Ok
+            else ExitStatus.UnexpectedError
```
jvican: Can we use our own status code, like `LinkingError` and use it in all the places?

frontend/src/main/scala/bloop/exec/Forker.scala
```diff
@@ -57,37 +57,64 @@ final case class Forker(javaEnv: JavaEnv, classpath: Array[AbsolutePath]) {
...
+      } else Task.unit
+    logTask.flatMap(_ => Forker.run(cwd, cmd, logger, opts))
+  }
+
+}
+
+object Forker {
+
+  /** The code returned after a successful execution. */
+  final val EXIT_OK = 0
```
jvican: Make both of them `private[Forker]`?

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -59,6 +59,46 @@ object Config {
...
+
+  case class NativeConfig(
+      toolchainClasspath: Array[Path],
+      gc: String,
+      clang: Path,
+      clangPP: Path,
+      linkingOptions: Array[String],
+      compileOptions: Array[String],
+      targetTriple: String,
+      nativelib: Path,
```
jvican: `nativeLib` for consistency?

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -59,6 +59,46 @@ object Config {
...
+      case Native.name => Native
+      case _ => throw new IllegalArgumentException(s"Unknown platform: '$platform'")
+    }
+  }
+
+  case class NativeConfig(
+      toolchainClasspath: Array[Path],
+      gc: String,
+      clang: Path,
+      clangPP: Path,
```
jvican: What does `clangPP` mean?

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -59,6 +59,46 @@ object Config {
...
+
+    def apply(platform: String): Platform = platform.toLowerCase match {
+      case JS.name => JS
+      case JVM.name => JVM
+      case Native.name => Native
+      case _ => throw new IllegalArgumentException(s"Unknown platform: '$platform'")
+    }
+  }
+
+  case class NativeConfig(
```
jvican: It would be great if there was a scaladoc explaining each of the fields of this case class, as well as the options that they can take (e.g. `gc`). If this is too cumbersome, a link to some kind of Scala Native documentation would also be helpful

config/src/main/scala/bloop/config/Config.scala
Duhemm: `clang++`

config/src/main/scala/bloop/config/Config.scala
Duhemm: That's how it's called in Scala Native's build API

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
```diff
@@ -0,0 +1,130 @@
...
+import bloop.{DependencyResolution, Project}
+import bloop.cli.{ExitStatus, OptimizerConfig}
+import bloop.engine.State
+import bloop.exec.Forker
+import bloop.io.AbsolutePath
+import bloop.logging.Logger
+
+import monix.eval.Task
+
+object ScalaNative {
```
jvican: Can we call this something else that clarifies more the meaning of the class? `ScalaNativeToolchain` for example?

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  private[this] var _ivyResolved: ScalaNative = _
+  private[this] val instancesCache: ConcurrentHashMap[Array[AbsolutePath], ScalaNative] =
+    new ConcurrentHashMap
+
+  def forProject(project: Project, logger: Logger): ScalaNative = {
+    project.nativeConfig match {
+      case None =>
+        ivyResolved(logger)
```
jvican: s/ivyResolved/resolveBridgeJars/g

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
```diff
@@ -0,0 +1,130 @@
...
+    if (_ivyResolved == null) {
+      val jars = bridgeJars(logger)
+      val nativeClassLoader = toClassLoader(jars)
+      _ivyResolved = new ScalaNative(nativeClassLoader)
+    }
+    _ivyResolved
+  }
+
+  def direct(classpath: Array[AbsolutePath]): ScalaNative = {
+    instancesCache.computeIfAbsent(classpath,
```
jvican: Hoist up the creation of `ScalaNative` up to its own variable?

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
```diff
@@ -0,0 +1,130 @@
...
+      case None =>
+        ivyResolved(logger)
+      case Some(config) =>
+        val classpath = config.toolchainClasspath.map(AbsolutePath.apply)
+        direct(classpath)
+    }
+  }
+
+  def ivyResolved(logger: Logger): ScalaNative = synchronized {
+    if (_ivyResolved == null) {
```
jvican: What is the purpose of `_ivyResolved` in the grand scheme of things? Don't we already cache with `instancesCache` and synchronize on the object here?

config/src/main/scala/bloop/config/Config.scala
jvican: Isn't it possible to have something like `clang++`?

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
Duhemm: This is only to completely bypass coursier (though I remember measuring it when we discussed about resolving the test agents, and it took only a few ms). I'm happy to get rid of it.

frontend/src/main/scala/bloop/exec/Forker.scala
Duhemm: That value is being used in other places.

config/src/main/scala/bloop/config/Config.scala
Duhemm: Personally, I prefer to keep it simple and name everything the way it is in Scala Native.

frontend/src/main/scala/bloop/engine/tasks/ScalaNativeToolchain.scala
```diff
@@ -0,0 +1,127 @@
...
+      case None =>
+        resolveNativeToolchain(logger)
+      case Some(config) =>
+        val classpath = config.toolchainClasspath.map(AbsolutePath.apply)
+        direct(classpath)
+    }
+  }
+
+  def resolveNativeToolchain(logger: Logger): ScalaNativeToolchain = {
+    val jars = bridgeJars(logger)
```
jvican: Last time I was touching `ScalaInstance` I noticed that the overhead of calling coursier is high, and skipping its invocation reduced starting up time significantly.

jvican: I think we should do the same we do with ScalaInstance here https://github.com/scalacenter/bloop/blob/topic/allow-only-java-projects/backend/src/main/scala/bloop/ScalaInstance.scala#L106-L107 so that we can share the resolution of the bridges across every project not defining the `toolchainClasspath`.

frontend/src/main/scala/bloop/engine/tasks/ScalaNative.scala
jvican: Just replied with another comment re caching.

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -59,6 +59,46 @@ object Config {
...
+      case JS.name => JS
+      case JVM.name => JVM
+      case Native.name => Native
+      case _ => throw new IllegalArgumentException(s"Unknown platform: '$platform'")
+    }
+  }
+
+  /**
+   * Configures how to start and use the Scala Native toolchain, if needed.
+   * See http://javadoc.io/doc/org.scala-native/tools_2.10/0.3.7 for the description of the fields.
```
jvican: This link points to the root of the scaladoc, would it be possible to link to the class containing all the fields? :smile:

frontend/src/main/scala/bloop/engine/tasks/ScalaNativeToolchain.scala
Duhemm: How high was it in your measurement? Resolving `bloop-native-bridge_2.12` is pretty fast on my machine, but I may very well be benchmarking it wrong:

```
[info] Benchmark                                                Mode   Cnt   Score   Error  Units
[info] Bench.resolveNativeBridge                              sample  4056  24.671 ± 0.086  ms/op
[info] Bench.resolveNativeBridge:resolveNativeBridge·p0.00    sample        22.315          ms/op
[info] Bench.resolveNativeBridge:resolveNativeBridge·p0.50    sample        24.052          ms/op
[info] Bench.resolveNativeBridge:resolveNativeBridge·p0.90    sample        27.197          ms/op
[info] Bench.resolveNativeBridge:resolveNativeBridge·p0.95    sample        27.787          ms/op
[info] Bench.resolveNativeBridge:resolveNativeBridge·p0.99    sample        29.014          ms/op
[info] Bench.resolveNativeBridge:resolveNativeBridge·p0.999   sample        30.079          ms/op
[info] Bench.resolveNativeBridge:resolveNativeBridge·p0.9999  sample        31.818          ms/op
[info] Bench.resolveNativeBridge:resolveNativeBridge·p1.00    sample        31.818          ms/op
```

The resolution of bridges was shared between projects that don't define `toolchainClasspath`. Should I revert https://github.com/scalacenter/bloop/pull/457/commits/27cd1a32bdd1d2d7c18acc7bce155bdbbe0fdf28 ?

config/src/main/scala/bloop/config/Config.scala
Duhemm: lol

PR #459, opened 2018-04-30
Make compilation of frontend faster

jvican:
by:

1. Upgrading to a custom version of 2.12.7 that includes https://github.com/scala/scala/pull/6582.
2. Removing the metals settings by default.


jvican:
I'll be opening a new PR to address this.

PR #464, opened 2018-05-02
Add scalameta to the community build

jvican:
The forked version of scalameta:

* Fix an issue with a relative path for the proto sources.
* Upgrade some plugin versions and disable coursier (the 0.13 community build
  doesn't like coursier).
* Removes the two out of three test projects because they require the
  compilation of the semanticdb module.

### DIFF
--- a/build-integrations/sbt-0.13/build.sbt
+++ b/build-integrations/sbt-0.13/build.sbt
@@ -2,7 +2,8 @@ val ApacheSpark = Integrations.ApacheSpark
 val LihaoyiUtest = Integrations.LihaoyiUtest
 val ScalaScala = Integrations.ScalaScala
 val ScalaCenterVersions = Integrations.ScalaCenterVersions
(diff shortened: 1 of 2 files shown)


olafurpg:
😭 anything I can help with?

jvican:
I wanted to hold off this change because I was getting some cryptic errors in the CI, probably related to the high intensity of compiling scalameta together with the rest of our projects.
  
My main objective was to use scalameta as a nice showcase for pipelined compilation because it has no Java sources; but unfortunately its design doesn't go along with pipelined compilation. I forgot that in Scalameta all the macros depend on other Scalameta modules. This breaks the invariant that "I can sidestep class file generation" because the macro classloader needs them to classload the macro execution. In the future, I want to do a best-effort in detecting build dependencies of macro projects and appyling pipelined compilation when possible, but this is out of scope for a prototype.

I think that maybe adding scalafix to our build would be better (note that macro dependencies that are binary and have no dependencies with build modules are ok).

PR #466, opened 2018-05-03
Add Ensime's config and cache files to .gitignore

tues:
It would make my life a little bit easier :smile: Pretty please :bowing_man:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ bin/.scalafmt*
 
 target/
 .bloop/
+
+# Ensime's config and cache
+.ensime
+.ensime_cache/


tues:
I guess the same can be said about `.idea/`? :wink:

I don't add them to my global `.gitignore` file on purpose, to remember to put them in project-specific `.gitignore`s seeing that everyone does so for other tools. If it's a problem then of course I'll ignore it locally :slightly_smiling_face:

jvican:
> I guess the same can be said about .idea/? wink

yeah, my gitignore is a template that i copy paste on every new project. Most of my dirs, like `.bloop` are ignored in `~/.gitignore`.

It's not an issue, let's merge this.

tues:
Thank you very much! :smile:

PR #468, opened 2018-05-03
Use -Xnojline option when starting console

### LINKED ISSUE #385: REPL appears not to start
A command invocation such as `bloop console kaleidoscope/core` appears to work right up until it starts the REPL. The last lines I see are (including a blank one),
```
Starting scala interpreter...

```
The invocation does not appear to terminate, and the cursor is left "hanging".

There is no output at all on the server until I hit `Ctrl+C`, then:
```
Client in <path> has disconnected with a 'SocketError' event. Cancelling tasks...
```

This is the beaviour for the first five invocations of `bloop console`. On the sixth attempt, no output is shown on the frontend at all.

Bloop-frontend: 1.0.0-M7
Zinc: 1.1.0+9-57ca9e0f
Scala: 2.12.4

tues:
Since we are using Nailgun, the REPL is being run in a different
process than the one connected to the user's terminal, so we cannot
run Scala's ILoop with InteractiveReader - we must use SimpleReader
instead. By passing -Xnojline argument to AnalyzingCompiler.console we
make sure that it will use SimpleReader.

Fixes #385.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
@@ -212,7 +212,8 @@ object Tasks {
     val loader = ClasspathUtilities.makeLoader(classpathFiles, scalaInstance)
     val compiler = state.compilerCache.get(scalaInstance).scalac.asInstanceOf[AnalyzingCompiler]
     val classpathOptions = ClasspathOptionsUtil.repl
-    compiler.console(classpathFiles, project.scalacOptions, classpathOptions, "", "", state.logger)(
+    val options = project.scalacOptions :+ "-Xnojline"
+    compiler.console(classpathFiles, options, classpathOptions, "", "", state.logger)(
       Some(loader))
     state
   }


frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
```diff
@@ -212,7 +212,8 @@ object Tasks {
     val loader = ClasspathUtilities.makeLoader(classpathFiles, scalaInstance)
     val compiler = state.compilerCache.get(scalaInstance).scalac.asInstanceOf[AnalyzingCompiler]
     val classpathOptions = ClasspathOptionsUtil.repl
-    compiler.console(classpathFiles, project.scalacOptions, classpathOptions, "", "", state.logger)(
+    val options = project.scalacOptions :+ "-Xnojline"
```
tues: I'd love to use a constant here, but `scala.tools.nsc.settings.ScalaSettings` has no companion object and I don't think that creating our own instance just to extract this value makes much sense.

jvican: The flag is unlikely to change, so I don't think it's a big deal :+1:

PR #470, opened 2018-05-03
Make some fixes before v1.0.0-M10 release

jvican:


jvican:
Enabling the community build on this PR to check that the error in the eighth integration test is gone.

PR #471, opened 2018-05-04
Make sure we always create source dirs in tests

jvican:
I have no idea why this hasn't been caught sooner, but
https://ci.scala-lang.org/scalacenter/bloop/3062/11 seems to point out
that as 6 directories don't exist within `with-tests`, we only print
that we're watching 2.

So let's see what happens with two.


jvican:
Aha, I see what's happening. The previous bloop would create non-existing directories, but the new bloop doesn't, so when we run without cache in the release process we only watch 2 folders, but here we watch 8 because the created source directories are cached.

Issue #62, opened 2017-11-24
Add lichess to our community build


jvican:
Current projects:
  
- sbt/sbt
- lihaoyi/utest
- scalacenter/versions
  
We need more. I propose to add heavy projects, possibly from the community build. Extra ideas:
  
* [ornicar/lila](https://github.com/ornicar/lila)
* [quasar-analytics/quasar](https://github.com/quasar-analytics/quasar)
* And similar projects... but we need a big testing suite to catch performance regressions.

jvican:
I've renamed this ticket into "Add lichess to our community build".

PR #474, opened 2018-05-06
Add lichess to our community build and optimize loading

jvican:
Fixes https://github.com/scalacenter/bloop/issues/62.

Lichess needs to go into an independent sbt 0.13 build because the dependencies of the play plugin conflict fatally with `sbt-maven-plugin` used by apache/spark.

### DIFF
--- a/backend/src/main/scala/bloop/io/Timer.scala
+++ b/backend/src/main/scala/bloop/io/Timer.scala
@@ -3,14 +3,12 @@ package bloop.io
 import bloop.logging.Logger
 
 object Timer {
-
-  @inline def timed[T](logger: Logger)(op: => T): T = {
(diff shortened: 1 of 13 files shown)


jvican:
CI is passing, but it's actually because the community build did not run, it's only set up. What seems to happen here is that ornicar/lila is so gigantic that 6gb of memory is not enough for it. The underlying reason for the high memory consumption is the huge size of the analysis file coming from Zinc. We may be talking about more than 1.5GB in in-memory analysis files.
  
I need to think about the right solution here. This may be a nice excuse to try to pull off the necessary changes in Zinc so that we can avoid storing the whole API in Zinc.

jvican:
Yayyyyyyyyyyyyyyyyyy.

PR #487, opened 2018-05-15
Kebabize arg names

Duhemm:
We were not kebabizing the name of the arguments, only the extra names,
which lead to wrong completion (like `--configDir` for instance).

This commit changes that behavior, so that we kebabize the name and the
extra names of the arguments of commands.


rberenguel:
I'll try to get the autocompletion tests working this week and then this should have been caught. Thanks @Duhemm :)

PR #489, opened 2018-05-15
Fix Maven plugin and add user docs

tues:
This is mostly Jorge's work from https://github.com/scalacenter/bloop/tree/topic/add-docs-maven-plugin, but rebased on top of newest `master` and with user documentation added.

I have tested it on a very large Maven project with weird configuration and it seems to work fine (at least much better than Maven itself). Considering that the current `master` doesn't work with Maven at all, I think this should be merged as is so that we can start collecting users' feedback.

By the way, when writing the documentation I assumed this would be released as `1.0.0-M10`.


jvican:
Thank you for putting this together and testing it in your project :heart:. I agree there may be some rough edges, but when I ran it with spark it was working great. @Duhemm can you have a look at it?

website/content/docs/installation.md
```diff
@@ -170,6 +176,14 @@ bloopAggregateSourceDependencies in Global := true
 
 <span class="label warning">Note</span> that `bloopAggregateSourceDependencies` has to be scoped globally.
 
+### Maven
+
+To generate Bloop's configuration using Maven, run (from your project directory):
+
+```sh
+$ mvn ch.epfl.scala:maven-bloop_2.12:1.0.0-M10:bloopInstall
```
Duhemm: You can use the same trick we did here to get the latest version automatically:
https://github.com/scalacenter/bloop/pull/489/files#diff-85567f8b3a7a1207fae1d03eb8722217R153

jvican: I'll fix this myself!

jvican:
@tues I had to force push to your branch because I rebased the changes with latest master :wink: I think this is good to go, I'll merge when CI passes.

website/content/docs/installation.md
tues: Thanks, I did look briefly at that snippet but thought it was just formatting.

BTW, the `:1.0.0-M10` part can be skipped altogether, but I'm not sure how Maven resolves the version then: if I try this now it downloads `1.0.0-M10+8-b6694b98`.

jvican: I'll add that detail to the docs. It may be useful to those that want to use latest master release.

Issue #496, opened 2018-05-19
Test failure returns always success exit code


tindzk:
The exit code is always 0 regardless of whether the tests failed or succeeded. Unfortunately, this makes it difficult to use `bloop test` in CI jobs.

jvican:
Thanks for the bug report, I'm happy to accept a PR to fix this! By the way Tim, I'll try to have a look at your open PR soon, I've been quite busy these last days at Scaladays.

tindzk:
No problem. I submitted a pull request to implement the required functionality.

tindzk:
Implemented in #499.

PR #499, opened 2018-05-21
Print report for test results and return correct exit code

### LINKED ISSUE #340: Print out summary after bloop test completes
I have a project using scalatest and currently `bloop test` does not include how many tests failed/succeeded. It would be great if bloop printed out a summary of the test run.

The lack of a summary is particularly confusing with `bloop test --watch` because you don't return to the shell after the tests completes so it's not clear whether the tests are hanging (infinite loop) or if bloop is waiting for more file events.

tindzk:
These changes print a report after the execution of each test suite and after all test suites. It also corrects the behaviour to return the correct exit code.

I tried to make the output more readable than sbt's. We might further improve the UX by using colours and formatted text (bold/italic).

Fixes #340 and #496.


jvican:
Let me know when this one passes CI and I'll review @tindzk. BTW, could you format the code with scalafmt? The project has a scalafmt config file that will be picked up whenever you format either from the CLI or IntelliJ.

jvican:
The errors are real, I'm fixing them tonight.

frontend/src/main/scala/bloop/testing/TestSuiteEvent.scala
```diff
@@ -63,38 +59,40 @@ class LoggingEventHandler(logger: Logger) extends TestSuiteEventHandler {
...
       suitesTotal += 1
       suitesDuration += duration
 
-    case TestSuiteEvent.Done =>
+    case TestSuiteEvent.Done => ()
   }
 
-  def report(): Unit = {
+  override def report(): Unit = {
+    // TODO: Shall we think of a better way to format this delimiter based on screen length?
```
tindzk: Good idea, however determining the terminal width may not always be possible. We could count the number of characters of the longest row instead.

frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -45,11 +45,11 @@ object Commands {
   ) extends RawCommand
 
   /** List of commands accepting a project as argument, used for autopomcletion */
-  val projectBound = CommandsMessages.messages
-    .collect {
-      case (name, CommandMessages(args, _)) if args.exists(_.name == "project") => name
-    }
-    .mkString(" ")
+  val projectBound: String = {
```
tindzk: Did you use a different scalafmt configuration to format the sources? In my previous commit, I used the `scalafmt` command within sbt.

tindzk:
Thanks for your changes! Feel free to squash and merge the pull request once you are done.

frontend/src/main/scala/bloop/cli/Commands.scala
jvican: In this case, scalafmt formats it differently if the collect is only in one line. I happen to know these tricks and that's why I reformatted so that it reads better :smile:

jvican:
@tindzk An squash won't be necessary. In my last commit I've removed the human readable format by a concise numeric one because I don't find the first one gives valuable information to the user and I find it verbose and difficult to parse. Let me know what you think. I'll be merging this change as soon as CI passes, you can submit more changes against my changes in another PR :wink:

tindzk:
Looks great. Thanks for merging!

frontend/src/main/scala/bloop/cli/Commands.scala
tindzk: I see. I was under the impression that scalafmt's formatters were deterministic.

jvican: Oh they are, but in this case I moved the collect to a new line and that wraps it’s body to the same line if it fits.

PR #502, opened 2018-05-22
Add option to configure the analysis file location

msvaljek:
Added support for configuring location of the compilation analysis file.


frontend/src/test/scala/bloop/tasks/TestUtil.scala
```diff
@@ -228,7 +228,8 @@ object TestUtil {
       testFrameworks = Array.empty,
       testOptions = Config.TestOptions.empty,
       javaEnv = javaEnv,
-      out = AbsolutePath(baseDirectory) // This means nothing in tests
+      out = AbsolutePath(baseDirectory), // This means nothing in tests
+      analysisOut = AbsolutePath(baseDirectory.resolve(s"$name-analysis.bin"))
```
jvican: Can we have something like the now removed `ResultsCache.pathToAnalysis` but move it to the `jsonConfig` module and reuse it both here and in all the integration code?

msvaljek: I didn't add dependency of config to `bloop.io` for `AbsolutePath` so I just created a method to create analysis file name there. Let me know what you think.

jvican:
This is great @msvaljek, good contribution! I added some improvements and fixed an issue in one of the areas you touched that revealed the reason why an unrelated bug was happening. I'm merging this, thanks again! :100:

Issue #429, opened 2018-04-09
Reference to undefined settings error during bloopInstall with Lagom plugin


matthewpflueger:
This seems like a conflict with the Lagom plugin and its internal projects. I have a fairly simple project using Lagom 1.4.4 and receive this error when running bloopInstall (I'm using bloop M9):

```
[error] sbt.internal.util.Init$RuntimeUndefined: References to undefined settings at runtime.
[error] setting(ScopedKey(Global,bloopInstall)) at LinePosition((bloop.integrations.sbt.PluginImplementation.globalSettings) SbtBloop.scala,56) referenced from ScopedKey(Scope(Select(ProjectRef(file:/opt/vp/scm/,lagom-internal-meta-project-kafka)), Select(ConfigKey(test)), Zero, Zero),bloopGenerate)
[error] setting(ScopedKey(Global,bloopInstall)) at LinePosition((bloop.integrations.sbt.PluginImplementation.globalSettings) SbtBloop.scala,56) referenced from ScopedKey(Scope(Select(ProjectRef(file:/opt/vp/scm/,lagom-internal-meta-project-cassandra)), Select(ConfigKey(compile)), Zero, Zero),bloopGenerate)
[error] setting(ScopedKey(Global,bloopInstall)) at LinePosition((bloop.integrations.sbt.PluginImplementation.globalSettings) SbtBloop.scala,56) referenced from ScopedKey(Scope(Select(ProjectRef(file:/opt/vp/scm/,lagom-internal-meta-project-kafka)), Select(ConfigKey(compile)), Zero, Zero),bloopGenerate)
[error] setting(ScopedKey(Global,bloopInstall)) at LinePosition((bloop.integrations.sbt.PluginImplementation.globalSettings) SbtBloop.scala,56) referenced from ScopedKey(Scope(Select(ProjectRef(file:/opt/vp/scm/,lagom-internal-meta-project-service-locator)), Select(ConfigKey(compile)), Zero, Zero),bloopGenerate)
[error] setting(ScopedKey(Global,bloopInstall)) at LinePosition((bloop.integrations.sbt.PluginImplementation.globalSettings) SbtBloop.scala,56) referenced from ScopedKey(Scope(Select(ProjectRef(file:/opt/vp/scm/,lagom-internal-meta-project-service-locator)), Select(ConfigKey(test)), Zero, Zero),bloopGenerate)
[...]

PR #512, opened 2018-06-01
Sbt: Export custom configurations

### LINKED ISSUE #505: Export custom sbt configurations (e.g. IntegrationTest)
Currently the sbt plugin exports only `Main` (e.g. `root`) and `Test` (e.g. `root-test`) configurations. It's quite common for projects to also have `IntegrationTest` configuration: https://www.scala-sbt.org/1.x/docs/Testing.html#Integration+Tests

It would be nice to support at least common configurations like `IntegrationTest`. Exporting all custom configurations would be even better, but obviously Bloop is not a build tool so it's unlikely it would support all weird configurations people may come up with.

Being able to enable/disable export of these additional configurations would probably be a desirable feature for people who don't want to clutter their project namespace.

tues:
To make use of this, one needs to add something along these lines:

```
import bloop.integrations.sbt.BloopDefaults

inConfig(IntegrationTest)(BloopDefaults.configSettings)
```

to project-local settings (preferably a .gitignored `local.sbt`).

TODO:

- [x] Inter-project dependencies (e.g. if `IntegrationTest` depends on `Test`)
- [x] Tests
- [ ] Documentation
- [ ] Probably resolve conflicts with https://github.com/scalacenter/bloop/compare/topic/fix-errors
- [x] Fix CI test failures
- [ ] Update https://github.com/scalacenter/bloop/blob/ticket%2F505-sbt-custom-configs/website/content/faq/_index.md#L44 when the commit ID is known (or change it to `master`?)

Closes #505.

### DIFF
--- a/build-integrations/global/src/main/scala/bloop/build/integrations/IntegrationPlugin.scala
+++ b/build-integrations/global/src/main/scala/bloop/build/integrations/IntegrationPlugin.scala
@@ -2,13 +2,13 @@ package bloop.build.integrations
 
 import java.io.File
 
-import bloop.integrations.sbt.{AutoImported => BloopKeys, SbtBloop}
+import bloop.integrations.sbt.{BloopKeys, BloopPlugin => SbtBloopPlugin}
 import sbt.{AutoPlugin, Def, PluginTrigger, Plugins, Keys, ThisBuild, Global}
 
 object IntegrationPlugin extends AutoPlugin {
   import sbt.plugins.JvmPlugin
   override def trigger: PluginTrigger = allRequirements
-  override def requires: Plugins = JvmPlugin && SbtBloop
+  override def requires: Plugins = JvmPlugin && SbtBloopPlugin
   val autoImport = PluginKeys
 
   override def globalSettings: Seq[Def.Setting[_]] =
--- a/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
+++ b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
@@ -153,8 +153,8 @@ object BloopDefaults {
   lazy val bloopGenerate: Def.Initialize[Task[File]] = Def.task {
     val logger = Keys.streams.value.log
     val project = Keys.thisProject.value
-    def nameFromString(name: String, configuration: Configuration): String =
-      if (configuration == Compile) name else name + "-test"
+      def nameFromString(name: String, configuration: Configuration): String =
+        if (configuration == Compile) name else s"$name-${configuration.name}"
 
     def nameFromRef(dep: ClasspathDep[ProjectRef], configuration: Configuration): String = {
       val ref = dep.project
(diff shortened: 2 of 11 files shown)


tues:
The test failure is my fault, I'll fix that.

tues:
I feel I'm stuck, so I have committed my work-in-progress changes adding `bloopDependencies` setting to work around sbt's limitations. It seems to be working well, but I don't like this. I should probably merge `nameFromRef` and `nameFromDependency` into one function, but I'm not sure what `sbt.Classpaths.mapped` is meant to do and the whole `ClasspathDep` vs `ClasspathDependency` and `ProjectRef` vs `ProjectReference` thing is confusing, to say the least.

This could be merged without the last commit (the one marked as `WIP`) and it would still be quite useful, the only disadvantage being the necessity of setting `dependencies` manually in JSON configs for custom configurations (at least if they depend on `compile` or `test`).

Example usage can be found in:
https://github.com/tues/bloop/blob/8516140b246d4c802d476da9f91f4b8aa842e57a/integrations/sbt-bloop/src/sbt-test/sbt-bloop/custom-configurations/build.sbt

jvican:
> I feel I'm stuck, so I have committed my work-in-progress changes adding bloopDependencies setting to work around sbt's limitations. It seems to be working well, but I don't like this. I should probably merge nameFromRef and nameFromDependency into one function, but I'm not sure what sbt.Classpaths.mapped is meant to do and the whole ClasspathDep vs ClasspathDependency and ProjectRef vs ProjectReference thing is confusing, to say the least.
  
Oh yeah, this part of sbt is particularly confusing and magic. It takes a lot of reading to fully understand how it works.
  
I added a commit that solves the problems with configurations automatically so that the user doesn't need to do anything. The key realization in my solution is that dependencies between configurations do not appear in normal classpath deps, but are embedded in the configurations themselves.
  
I've added some code that detects the dependent configurations from which we have to derive inter-project dependencies. These configurations are filtered out so that only the configurations that will generate a bloop configuration file are used.
  
> BTW, the second commit (Rename PluginImplementation to BloopDefaults) is best viewed with white-space changes hidden.
  
Yup, it's not a scary change when you see it only adds 42 lines :smile:
  
I'll be merging this as soon as CI passes since it already gives value. Do you think you can submit another PR adding information to the docs on how to generate configuration files for integrations? I was thinking it may be better to have a new docs page called "Advanced sbt configuration" or something similar in which we put nitty-gritty details on how to do things that are out of scope for the basic installation process.
  
An appereance in the FAQ would also be helpful, redirecting to the advanced sbt configuration page.

jvican:
I'm running the community build to make sure that the generation of bloop config files works for all the projects. In the last commit, I added a test case for #429 which you accidentally fixed by using the well-chosen `inTasks(bloopGenerate)` scope filter :wink:

tues:
Sure, I'll open another PR for the docs later :slightly_smiling_face:

I'm not sure about `config.extendsConfigs` because that was my first approach and it didn't work for the project on which I tested these changes, but maybe there was some other issue involved. We'll see, but I'm sure this is already very useful at least for the majority of sbt projects using custom configs :slightly_smiling_face:

jvican:
> I'm not sure about config.extendsConfigs because that was my first approach and it didn't work for the project on which I tested these changes, but maybe there was some other issue involved. We'll see, but I'm sure this is already very useful at least for the majority of sbt projects using custom configs
  
If it doesn't work for your project, then the issue may be that you're adding a custom classpath dependency that we don't detect in `nameFromDependency`, concretely in the `mapping`. Adding `it` there in the `masterConfs` parameters should work.

PR #517, opened 2018-06-05
fix mvn docs. dodgy dot and missing target

stephennancekivell:


jvican:
Thanks @stephennancekivell :smile:

PR #527, opened 2018-06-08
Run `updateClassifiers` when exporting the project

jvican:
`updateClassifiers` is necessary so that IDEs can index source jars and
provide a nice language experience when browsing code of dependencies.


jvican:
Interesting data about `updateClassifiers`:
  
1. The sbt implementation doesn't allow you to modify whether you want to resolve sources or javadocs, it does both. I've tried several big workarounds and none worked.
2. Running `updateClassifiers` in `bloopInstall` is considerably slower in the csae it downloads sources or javadocs. If it doesn't, then the overhead against the normal `bloopInstall` is negligible (at least with `coursier`).
3. Builds that don't use `coursier` will get really nasty logs (e.g. errors of artifacts that failed to be resolved even though this is valid under `updateClassifiers`). There is no way we can fix this in our end.
  
Suggestion: if you use this feature, make everything possible to use `coursier` instead of `ivy` in your build. You won't regret it.

PR #529, opened 2018-06-09
Redesign configuration file format from first principles

jvican:
This is a complete redesign of the configuration file format to
reduce its size and improve its overall shape for the upcoming
1.0.


integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -352,6 +333,15 @@ object BloopDefaults {
     }.distinct
   }
 
+  def onlyCompilationModules(ms: Seq[Config.Module], classpath: Array[Path]): Seq[Config.Module] = {
+    val classpathFiles = classpath.filter(p => Files.exists(p) && !Files.isDirectory(p))
```
tues: I think `Files.isRegularFile` should be at least as good as `Files.exists && !Files.isDirectory` in this case, but I can see the latter being easier to understand (one doesn't need to check the docs to make sure that symbolic links are followed by default).

jvican: I'll add `Files.isRegularFile` :wink:

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -63,45 +59,68 @@ object Config {
...
    */
   case class NativeConfig(
-      toolchainClasspath: Array[Path],
       gc: String,
-      clang: Path,
-      clangPP: Path,
-      linkingOptions: Array[String],
-      compileOptions: Array[String],
-      targetTriple: String,
+      platform: String,
```
tues: Are you sure about renaming `targetTriple` to `platform`? We already have `platform` as in `JVM` vs `JS` vs `Native` so this is a bit confusing. `target` is the standard way to refer to architecture/OS/ABI when compiling native code. OTOH `target` may also refer to the path where compiled binaries are written, so maybe something like `targetPlatform`?

jvican: You're right, I didn't notice we also had another `platform`. I just pushed a commit with the fix :wink:

tues:
Yea, probably best to keep it called `targetTriple` if that's what Scala Native uses.

LGTM :+1:

jvican:
Alright... what started as a small PR has grown into a hell of changes because they were all kinda required to get the tests passing.
  
Things that this PR does:
  
1. Redesign caching of toolchains for Native and Scalajs.
1. Add support for version-dependent toolchains (scalajs 0.6 and 1.0).
1. Fully extract scala native and scalajs information from builds with several sbt tricks.
1. Add checksums when possible since sbt doesn't populate them.
1. Reduce file size by removing unnecessary information and only persisting resolved modules that are visible in the classpath.
1. Rename some fields in the specification.
  
Changes to the docs (in all aspects) will be done pre 1.0 release.

PR #530, opened 2018-06-09
Add debugging instructions to developer documentation [DOCS]

tues:
It's not that obvious how to run Bloop server directly from sbt and how to enable debugging for forked processes. Hopefully these instructions will save people some time figuring things out.

### DIFF
--- a/website/content/docs/developer-documentation.md
+++ b/website/content/docs/developer-documentation.md
@@ -128,3 +128,52 @@ will then kick in and confirm your request.
 the benchmarks for it. Run it to execute the benchmark suite locally, but keep
 in mind that you must make sure your machine is stable and has been tweaked to
 get stable results.
+
+## Debugging
+
+When analysing a bug, it is often infeasible to rely only on tests. It might be
+more convenient to run Bloop and attach a debugger instead.
+
+### Running Bloop server from sbt
+
+To run the server directly from sbt:
+
+```sh
+$ sbt
+> frontend/runMain bloop.Server
+```
+
+If you want to be able to stop the server with <kbd>Ctrl</kbd><kbd>C</kbd>
+without killing the sbt shell, you need to set `cancelable` to `true` first:
+
+```
+> set cancelable in Global := true
+```
+
+#### Attaching debugger
+
+In order to attach a debugger you need to run Bloop with some additional JVM
+options to enable debugging in the first place. Here we will use the standard
+JDWP (Java Debug Wire Protocol) agent. Type the following in your sbt shell:
+
+```sh
+$ sbt
+> set javaOptions in (frontend, run) += "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005"
+> frontend/runMain bloop.Server
+```
+
+Now attach your favourite debugger using port `5005`. The JVM will wait with
+starting Bloop until a debugger is attached -- change `suspend` to `n` if you
+don't want this behaviour.
+
+#### Ensime
+
+If you are using [Ensime](https://ensime.github.io/), there is `ensimeRunDebug`
+task defined by `sbt-ensime` plugin which lets you simply do:
+
+```sh
+$ sbt
+> frontend/ensimeRunDebug bloop.Server
+```
+
+which is equivalent to setting `javaOptions` like above.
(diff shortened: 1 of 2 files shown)


website/content/docs/developer-documentation.md
```diff
@@ -128,3 +128,52 @@ will then kick in and confirm your request.
...
+### Running Bloop server from sbt
+
+To run the server directly from sbt:
+
+```sh
+$ sbt
+> frontend/runMain bloop.Server
+```
+
+If you want to be able to stop the server with <kbd>Ctrl</kbd><kbd>C</kbd>
```
jvican: What about we add it to the build directly and just tell people they can do `CTRL-C` to kill it?

jvican:
Did you eventually find out whether our nailgun script supports this? Would be even better to allow JDWP via the installed bloop server.

website/content/docs/developer-documentation.md
tues: I was actually thinking about that, I guess we might. Usually I add it to my global `~/.sbt/` because I want to be able to cancel any sbt task in any project. Will do.

tues:
> Did you eventually find out whether our nailgun script supports this?

No, unfortunately it doesn't seem to work. `ng.py` calls JVM in the following way:
```
            check_call(["java", "-jar", server_location] + cmd_args)
```
so it doesn't support adding any options between `java` and `-jar`.

> Would be even better to allow JDWP via the installed bloop server.

IMO it doesn't make much sense because we want to have the sources in sync with what's running in JVM. I was using it only temporarily because I had some issues running the server from sbt which I solved later. If you do see any value in this though, I can add support for additional JVM options to the Nailgun client :slightly_smiling_face:

jvican:
> IMO it doesn't make much sense because we want to have the sources in sync with what's running in JVM.
  
Do you mean that you don't wanna nuke the server in the background? The way i would do it is that i would spawn another bloop server in another port, and then connect to it from the bloop client so that it doesn't interrrup the other, active and hot bloop server.
  
> If you do see any value in this though, I can add support for additional JVM options to the Nailgun client :slightly_smiling_face: 
  
I'm thinking more about normal users trying to debug what's happening. Asking them to clone the repo and do it from the sbt build (which may require importing from intellij or updating the whole build) may be too demanding.
  
I think it would be great if we make the nailgun script work :smile:. We need it anyway to control things like `-Xmx` and `-Xms` etc

jvican:
Ah, I see! Alright, fair point. But I agree that having a ticket for this would be great. I wanna be able to change the jvm options of the bloop server so that I can shove my favorite `-XX:MaxInlineLevel=20` into the server :smile:
  
Are you going to update this to include the cancelable setup? Let me know when you want another review!

jvican:
I added some commits here to add some improvements to the docs and modify the nailgun script, what do you think @tues?

website/content/docs/developer-documentation.md
```diff
@@ -131,49 +131,82 @@ get stable results.
...
 
-If you are using [Ensime](https://ensime.github.io/), there is `ensimeRunDebug`
-task defined by `sbt-ensime` plugin which lets you simply do:
+##### sbt
 
 ```sh
 $ sbt
-> frontend/ensimeRunDebug bloop.Server
+> set javaOptions in (frontend, run) += "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005"
+> frontend/runMain bloop.Server --config-dir /foo/bar
```
tues: Is config dir needed for `bloop.Server`? It is for `bloop.Cli`, but the server seems to run fine without it.

website/content/docs/developer-documentation.md
```diff
@@ -131,49 +131,82 @@ get stable results.
 
 ## Debugging
 
-When analysing a bug, it is often infeasible to rely only on tests. It might be
-more convenient to run Bloop and attach a debugger instead.
+If you're trying to catch a bug, it might ben convenient to attach a debugger
```
tues: s/ben/be/

website/content/docs/developer-documentation.md
jvican: True! Forgot that we were doing server here.

PR #532, opened 2018-06-10
Add basic mill integration

Baccata:
Adds a `bloop.integrations.mill.BloopModule` mill external module that can be used to generate bloop config from mill. The support is limited to JVM modules at the moment, and does not fill the "resolution" configuration element.

See https://www.lihaoyi.com/mill/page/modules.html#external-modules

## Usage :

1. add ```import $ivy.`ch.epfl.scala::mill-bloop:$bloopVersion` ``` to the build.sc of a project (it's probably possible to add it to a `~/.mill/ammonite/predef.sc` or a similar file instead)

2. run `mill bloop.integrations.mill.BloopModule/install` from the directory of the mill project.

## NB : 

Due to some [obscure bug](https://gitter.im/scalacenter/bloop?at=5b1ba015144c8c6fea8223c9) that prevents me from safely running non-brewed bloop, I'm only able to run `1.0.0-M10`, which fails to run the tests on my machine due to a bug in nuprocess.

### DIFF
--- a/integrations/mill-bloop/src/main/scala/bloop/integrations/mill/MillBloop.scala
+++ b/integrations/mill-bloop/src/main/scala/bloop/integrations/mill/MillBloop.scala
@@ -52,10 +52,12 @@ object Bloop extends ExternalModule {
       case _ => T.task(Config.Scala.empty)
     }
 
-    val jvmConfig = T.task {
-      Config.Jvm(
-        home = T.ctx().env.get("JAVA_HOME").map(s => Path(s).toNIO),
-        options = module.forkArgs().toArray
+    val platform = T.task {
+      Config.Platform.Jvm(
+        Config.JvmConfig(
+          home = T.ctx().env.get("JAVA_HOME").map(s => Path(s).toNIO),
+          options = module.forkArgs().toList
+        )
       )
     }
 
@@ -89,6 +91,14 @@ object Bloop extends ExternalModule {
     }
 
     val classpath = T.task(transitiveClasspath(module)() ++ ivyDepsClasspath())
+    val compileSetup = Config.CompileSetup(
+      Config.Mixed,
+      addLibraryToBootClasspath = true,
+      addCompilerToClasspath = false,
+      addExtraJarsToClasspath = false,
+      manageBootClasspath = true,
+      filterLibraryFromClasspath = true
+    )
 
     val project = T.task {
       Config.Project(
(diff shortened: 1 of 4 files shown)


project/Dependencies.scala
```diff
@@ -66,4 +67,5 @@ object Dependencies {
 
   val scalaNativeTools = "org.scala-native" %% "tools" % scalaNativeVersion
   val scalaJsTools = "org.scala-js" %% "scalajs-tools" % scalaJsVersion
+  val mill = "com.lihaoyi" %% "mill-scalalib"	% millVersion
```
Baccata: probably should tag this as "provided"

integrations/mill-bloop/src/main/scala/bloop/integrations/mill/MillBloop.scala
```diff
@@ -0,0 +1,138 @@
...
+        ),
+        out = out(module).toNIO,
+        analysisOut = analysisOut(module).toNIO,
+        classesDir = classes(module).toNIO,
+        scala = scalaConfig(),
+        jvm = jvmConfig(),
+        java = javaConfig(),
+        test = testConfig(),
+        platform = Config.Platform.JVM,
+        nativeConfig = None,
```
Baccata: filling this will depend on https://github.com/lihaoyi/mill/pull/206

project/Dependencies.scala
jvican: Yes, I think so. It depends on how mill implements plugin classloading, but if the mill classloader is the parent of your plugin classloader, it should be `Provided` :+1:

Baccata: Addressed in 4943ed5 (tested locally)

Baccata:
> Do you have any more changes you want to include before I rebase?

Just pushed the "provided" 

> By the way, I'm curious, how long does it take to generate the bloop configuration file? 

I nuked my mill cache, and running the `Bloop/install` task took 19 seconds. That included compilation of all my (non-trivial, 16 modules) build , and running all the upstream tasks. From then, if I remove the `.bloop` folder and regenerate, it seems to take ~3-5 seconds.

jvican:
Does `bloopInstall` require the compilation of all your modules to generate bloop configuration files? It seems that should be improved upon. The sbt build, by default, does this too when you ask for the `dependencyClasspath`, but the plugin creates its own implementation of `dependencyClasspath` that doesn't require compilation.
  
Would you mind creating a ticket to keep track of that?

jvican:
(I'll be pushing to this branch in half an hour or so to rebase everything.)

Baccata:
> Does bloopInstall require the compilation of all your modules to generate bloop configuration files

Nope, I made sure that's not the case. When I said "compilation of the build", I meant compilation of the build file itself and all the ammonite script it depends on, not the modules. 

@Duhemm, Mill does provide a `generatedSources` task you use override with whatever behaviour. You can also write a resources generators very simply overriding the `resources` task to point to potential resource folders you might want to add. It's really super easy. Both tasks are run by the `install` task provided in this PR.

integrations/mill-bloop/src/main/scala/bloop/integrations/mill/MillBloop.scala
```diff
@@ -0,0 +1,138 @@
...
+        }
+      case _ => T.task(Config.Test.empty)
+    }
+
+    val ivyDepsClasspath =
+      module
+        .resolveDeps(T.task { module.compileIvyDeps() ++ module.transitiveIvyDeps() })
+        .map(_.map(_.path).toSeq)
+
+    def transitiveClasspath(m: JavaModule): Task[Seq[Path]] = T.task {
```
Baccata: @jvican this is what I had to re-implement (rather than calling the mill one directly) to make sure the `install` task does not call upon the `compile` task

jvican: Great! I cannot get back to the rebase tonight but wil do tomorrow morning.

jvican:
Done! Thanks again for the contribution @Baccata :+1:

PR #538, opened 2018-06-12
Persist sbt metadata in project model

jvican:
Add the sbt metadata field in the project configuration file so that we
can present it to IDEs via BSP. A nice property of the approach taken in
this PR is that we automatically create a project file for the build
under the meta build `.bloop` directory. This is helpful for two
reasons:

1. Bloop can compile your build just as it can compile your project.
2. Bloop knows immediately any change that happens to your build.

The implementation here proposed is quite tricky because of some sbt
implementation details. We cannot get the `autoImports` from the build
of the build, so we need to roll out our own piece of code to do so with
the `pluginData` of the meta build. This repeats some work because it
needs to classload the build data for our project's build, but it's
negligible in terms of time spent (<500 ms in the cases I've tested).
And it does reuse the parent classloader from the meta build.

Aside from implementing this feature, this commit also removes any cross
published module and instead creates a static module for it. This helps
development because we can compile the projects for 0.13 and 1.0
immediately without needing to use `+` and `^`, which are completely
broken in our build.


jvican:
The last commit adds a fancier `discoveredSbtPlugins` @Duhemm. Importing our build now takes 11 seconds (which is quite an achievement considering there are 38 modules).

integrations/sbt-bloop/src/sbt-test/sbt-bloop/cross-compile-test-configurations/build.sbt
```diff
@@ -1,5 +1,6 @@
 import java.nio.file.{Files, Path}
 
+val BloopClassifiers = Set("sources", "javadoc")
```
Duhemm: Is this used somewhere?

jvican: Nope, just removed it.

PR #541, opened 2018-06-13
Style improvements

tindzk:
- Improve wording of messages
- Do not use full stops in log messages unless they describe processes in which case the sentences should end with "..."
- Use correct spellings of "Scala.js" and "JavaScript" in messages and their camel case counterparts in code
- Do not use abbreviations
- Do not use contractions (didn't, shouldn't etc.)
- Implement IntelliJ's code suggestions


frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -280,6 +279,26 @@ object Interpreter {
...
+    }
+
+    val msg = s"Artifact $artifactName for Scala Native toolchain could not be resolved in project '$project'"
+    state.withError(msg, ExitStatus.LinkingError)
+  }
+
+  private def missingScalaJsToolchain(state: State, project: Project): Task[State] = {
+    val artifactName = project.platform match {
+      case Config.Platform.Js(config) => ScalaJsToolchain.artifactNameFrom(config.version)
+      case _ => throw new IllegalArgumentException
```
jvican: Let's remove these `IllegalArgumentException`. If we hit this path, the project isn't directly a javascript project so we'd better report that explicitly to users.

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -347,7 +363,7 @@ object Interpreter {
             else doJsRun(project, config)(state)
 
           case Platform.Jvm(_) =>
-            val msg = "Expected Scala Native or Scala.js project to run `link`."
+            val msg = "Only Scala.js and Scala Native projects can be linked"
```
jvican: I prefer to use well-structured and easy to parse error messages that start with "Expected" or "Missing". What about "Link failed. ${project.name} is not a Scala Native or Scala.js project."?

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -625,7 +623,7 @@ object BloopDefaults {
             case (options, sbt.Tests.Exclude(tests)) =>
               options.copy(excludes = tests.toList ++ options.excludes)
             case (options, other: sbt.TestOption) =>
-              logger.info(s"Skipped test option '${other}' as it can only be used within sbt.")
```
jvican: FTR, i find it better to wrap them in curly braces so that you don't need to add them whenever expressions are added.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -746,19 +744,19 @@ object BloopDefaults {
 
       import scala.collection.JavaConverters._
       val visited = Classpaths.interSort(currentProject, conf, data, deps)
-      val productDirs = (new java.util.LinkedHashSet[Task[Seq[File]]]).asScala
-      val bloopProductDirs = (new java.util.LinkedHashSet[Task[Seq[File]]]).asScala
+      val productDirs = new java.util.LinkedHashSet[Task[Seq[File]]].asScala
```
jvican: This is more difficult to parse, revert?

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -89,7 +89,7 @@ object Interpreter {
     logger.info(s"$header$line")
     logger.info(s"$t$bloopName is made with love at the Scala Center <3$line")
     logger.info(s"$versions$line$line") // I have no idea why two are required, one is not enough
-    logger.info(s"${t}It is maintained by $developers.")
+    logger.info(s"${t}It is maintained by $developers")
```
Duhemm: Just curious, what's wrong with having a dot at the end of sentences?

frontend/src/main/scala/bloop/engine/Interpreter.scala
tindzk: Should I write `sys.error("Scala.js project expected")` instead?

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
tindzk: It is a matter of taste. I prefer the more concise notation and this is also the style IntelliJ recommends.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
tindzk: Okay. IntelliJ was pointing it out too.

frontend/src/main/scala/bloop/engine/Interpreter.scala
tindzk: That's better. How about the following:

> Cannot link JVM project ${project.name}. `link` is only available for Scala Native and Scala.js projects.

frontend/src/main/scala/bloop/engine/Interpreter.scala
tindzk: Either variant is fine as long as we use punctuation consistently. Right now, periods appear in some log messages, but not all.

Based on your remark, I did some more research and there appears to be a [concensus](https://ux.stackexchange.com/questions/18671/periods-at-the-end-of-a-sentence-in-alert-message) to favour punctuation in log messages, but not in titles. I will add them back.

frontend/src/main/scala/bloop/engine/Interpreter.scala
jvican: Perfect!

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
jvican: Let's keep these changes as they are, but let's not do more of these stylistic subjective changes in the future. We have too many things to improve in the codebase and we don't wanna waste time discussing about syntax or coding style :wink:

frontend/src/main/scala/bloop/engine/Interpreter.scala
jvican: Thanks for the link, that was interesting! We should indeed be consistent here, so you did a good job at trying to consolidate one approach. Do you by any chance know how `npm` and `cargo` do this? I think they don't use dots there. It may be interesting dropping the dots in favor of whatever the majority of build tools in other language communities do!

frontend/src/main/scala/bloop/engine/Interpreter.scala
jvican: What about "Fatal error: Scala.js project ${project.name} doesn't have a js config. Report upstream."

tindzk: Sounds good, but I will change "js config" to "JavaScript configuration".

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
tindzk: Agreed. If we all agree on using a particular code linter configuration (such as IntelliJ's defaults), we could avoid these discussions.

jvican: I don't use such defaults, so it's better that everyone makes the necessary changes with their own style, only touching the code required for the change. That helps keep git history clean.

frontend/src/main/scala/bloop/engine/Interpreter.scala
tindzk: npm does not have the best UI and they do not use periods consistently either. Cargo does not use periods in messages ([screenshot](https://user-images.githubusercontent.com/15943089/36560466-a4e31226-17c5-11e8-807c-b9dd62cbcc39.png)).

tindzk:
I rebased the changes and incorporated your feedback.

jvican:
@tindzk I have a last request, sorry for the delay. Let's do what cargo does and not use full stops. Let's only full stops for paragraphs explaining things rather than just statements of a situation/problem.

tindzk:
@jvican Sure. I updated the commit.

jvican:
The CI is failing, possibly because you didn't remove the full stops from the tests :smile:

PR #545, opened 2018-06-14
Link: Add ability to specify custom output path

tindzk:


bridges/scala-native/src/main/scala/bloop/scalanative/NativeBridge.scala
```diff
@@ -9,12 +11,11 @@ import java.nio.file.{Files, Path}
 import scala.scalanative.build.{Build, Config, Discover, GC, Mode, Logger => NativeLogger}
 
 object NativeBridge {
-  def nativeLink(config0: NativeConfig, project: Project, entry: String, logger: Logger): Path = {
+  def nativeLink(config0: NativeConfig, project: Project, entry: String, target: File, logger: Logger): Path = {
```
Duhemm: Should be `java.nio.file.Path`

bridges/scalajs/src/main/scala/bloop/scalajs/JsBridge.scala
```diff
@@ -38,40 +40,31 @@ object JsBridge {
...
     IRContainer.Jar(new FileVirtualBinaryFile(jar.toFile) with VirtualJarFile)
 
   def link(
       config: JsConfig,
       project: Project,
       mainClass: String,
+      target: File,
       logger: BloopLogger
-  ): Path = {
+  ): Unit = {
```
Duhemm: Why? I think it's better to return the `target`, even if it's already one of the inputs.

bridges/scalajs/src/main/scala/bloop/scalajs/JsBridge.scala
```diff
@@ -38,40 +40,31 @@ object JsBridge {
...
 
-  private def toIRJar(jar: Path) =
+  private def toIrJar(jar: Path) =
     IRContainer.Jar(new FileVirtualBinaryFile(jar.toFile) with VirtualJarFile)
 
   def link(
       config: JsConfig,
       project: Project,
       mainClass: String,
+      target: File,
```
Duhemm: `Path`

bridges/scalajs/src/main/scala/bloop/scalajs/JsBridge.scala
```diff
@@ -38,40 +40,31 @@ object JsBridge {
...
-    val outputPath = project.out.underlying
-    val target = project.out.resolve("out.js")
-
-    val enableOptimizer = config.mode match {
-      case LinkerMode.Debug => false
-      case LinkerMode.Release => true
-    }
-
-    val jarFiles = classpath.filter(isJarFile).map(toIRJar)
+    val optimizer = config.mode == LinkerMode.Release
```
Duhemm: Do you feel strongly about this rename? I think it's better to have the name of a Boolean variable describe precisely what it controls.

bridges/scalajs/src/main/scala/bloop/scalajs/JsBridge.scala
```diff
@@ -38,40 +40,31 @@ object JsBridge {
...
+    val jarFiles = classpath.filter(isJarFile).map(toIrJar)
```
Duhemm: I think this is unused?

frontend/src/main/scala/bloop/engine/tasks/ScalaJsToolchain.scala
```diff
@@ -23,22 +23,23 @@ class ScalaJsToolchain private (classLoader: ClassLoader) {
...
    * @return The absolute path to the generated JS source.
    */
   def link(
       config: JsConfig,
       project: Project,
       mainClass: String,
+      target: AbsolutePath,
       logger: Logger
-  ): Task[Try[AbsolutePath]] = {
+  ): Task[Try[Unit]] = {
```
Duhemm: Can we keep the return type here too?

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -280,7 +285,23 @@ object Interpreter {
...
+      state.logger.error(s"The output path ${output.syntax} does not point to a file. Using default...")
+      None
+    } else if (!Files.isWritable(output.getParent.underlying)) {
+      state.logger.error(s"The output path ${output.getParent.syntax} cannot be created. Using default...")
+      None
+    } else if (!output.syntax.endsWith(extension)) {
+      // This is required for the Scala.js linker, otherwise it will throw an exception
+      state.logger.warn(s"The output path ${output.syntax} must have the extension '$extension'. Appending...")
+      Some(output.getParent.resolve(output.toFile.getName + extension))
+    } else Some(output)
```
Duhemm: I would completely fail the execution if the validation fails, instead of writing the output somewhere the user doesn't expect it.

bridges/scalajs/src/main/scala/bloop/scalajs/JsBridge.scala
tindzk: Because we are passing in the path to the function now. Otherwise, we would make the assumption that the linker may save the file to a different path, which goes counter to the idea of having a parameter like `--output`.

bridges/scalajs/src/main/scala/bloop/scalajs/JsBridge.scala
tindzk: No, I will change it back.

bridges/scalajs/src/main/scala/bloop/scalajs/JsBridge.scala
tindzk: Do you mean `jarFiles`? It is accessed in the line below.

frontend/src/main/scala/bloop/engine/Interpreter.scala
tindzk: You are right. The behaviour should be more deterministic. I will change it.

tindzk:
Thanks for your feedback! Please let me know if there is anything else I should change.

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -179,7 +184,7 @@ object Interpreter {
           Tasks.console(state, project, config, cmd.excludeRoot)
         }
       case None =>
-        Task(reportMissing(cmd.project :: Nil, state))
+        Task.pure(reportMissing(cmd.project :: Nil, state))
```
jvican: Can you use `now` instead? It's used all across the codebase

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -280,24 +285,45 @@ object Interpreter {
     }
   }
 
+  private def validOutputPath(output: AbsolutePath, state: State, extension: String): Boolean =
```
jvican: Let's model this as a method `reasonOfInvalidPath(output: AbsolutePath, state: State, extension: String): Option[String]` and then log the errors at the call-sites instead of side effecting here.
  
Also, can you add curly braces around the expression?

tindzk:
I will make the remaining changes tomorrow. Should I also remove `out` from `Config.Project`?

jvican:
Good question. Let's keep it as is for two reasons:
  
1. `out` in `Config.Project` points to the root directory where all the compilation products are generated (imagine the `out` generated by C compilers).
2. `out` in `JvmConfig` could be used for an hypothetical target file if we add support for `jlink` in the future.

tindzk:
I made the requested change of removing the `-o` parameter. Note that `run` now uses the same output path as `link`.

For consistency, the main class should also be moved to the Bloop configuration file. The main class for Scala.js projects can be configured via sbt, so we could export it.

jvican:
> For consistency, the main class should also be moved to the Bloop configuration file. The main class for Scala.js projects can be configured via sbt, so we could export it.
  
This is a good idea. We could add a field called `mainClasses` that is an array of fully qualified names as strings in every link command (because it's platform dependent). In case of several main classes, we should keep the following semantics: the first class of the array is the default one (and therefore we should correctly follow this contract when exporting from builds).
  
Do you want to do this in this PR or in another one?

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -86,9 +86,13 @@ object Config {
...
   }
 
-  sealed trait PlatformConfig
+  sealed trait PlatformConfig {
+    def output: Option[Path]
+  }
 
-  case class JvmConfig(home: Option[Path], options: List[String]) extends PlatformConfig
+  case class JvmConfig(home: Option[Path], options: List[String]) extends PlatformConfig {
+    override def output: Option[Path] = None
```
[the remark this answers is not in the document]
tindzk: I am not sure either, but you were mentioning that we may run `jlink` in the future on JVM projects.

frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -176,9 +176,9 @@ object Commands {
       @ExtraName("w")
       @HelpMessage("If set, run the command whenever projects' source files change.")
       watch: Boolean = false,
-      @ExtraName("o")
+      @ExtraName("opt")
```
tindzk: I will also change this back to `o`.

tindzk:
> the first class of the array is the default one (and therefore we should correctly follow this contract when exporting from builds)

Does it mean that the user will only be able to specify classes that were encoded in the Bloop configuration?

> Do you want to do this in this PR or in another one?

I will open a separate pull request such that the changes can be reviewed better.

tindzk:
@jvican I implemented your suggestions. Let me know if anything else is off.

bridges/scalajs/src/main/scala/bloop/scalajs/JsBridge.scala
Duhemm: Ugh sorry

frontend/src/main/scala/bloop/engine/tasks/ScalaJsToolchain.scala
Duhemm: I don't really feel strongly about that, but I don't like checking that's the result is a `Success(_)` and do nothing with that. If we change it, we should do the same in the `ScalaNativeToolchain` though!

Duhemm:
Thanks a lot @tindzk ! If you want to change the return types to `Unit` in the Scala.js toolchain, go ahead, but please also change in the `ScalaNativeToolchain` too then.

tindzk:
@Duhemm Do you refer to the `link` method? I already changed it in both toolchains to return `Try[Unit]`. I could change the type to `Option[Throwable]`, but `Try[Unit]` gets the meaning across better.

Duhemm:
Oh I missed the changes in the ScalaNativeToolchain, sorry about that!

There are just a few conflicts; let's resolve them and get this in.

Thanks a lot for the contribution 🎉

tindzk:
Thanks, @Duhemm! I rebased the changes.

PR #551, opened 2018-06-18
Add ability to set default main class

tindzk:
@jvican I think it would be more intuitive to allow users to specify up to one main class in the configuration file (i.e. `Option[String]` instead of `List[String]`).

### DIFF
--- a/config/src/main/scala-2.10/bloop/config/ConfigEncoderDecoders.scala
+++ b/config/src/main/scala-2.10/bloop/config/ConfigEncoderDecoders.scala
@@ -109,40 +109,39 @@ object ConfigEncoderDecoders {
       case Platform.Jvm(config, mainClass) =>
         val configJson = jvmEncoder(config)
         val mainClassJson = implicitly[RootEncoder[Option[String]]].apply(mainClass)
-        Json.fromFields(List(
-          (N, Json.fromString(Platform.Jvm.name)),
-          (C, configJson),
-          (M, mainClassJson)))
+        Json.fromFields(
+          List((N, Json.fromString(Platform.Jvm.name)), (C, configJson), (M, mainClassJson)))
       case Platform.Js(config, mainClass) =>
         val configJson = jsEncoder(config)
         val mainClassJson = implicitly[RootEncoder[Option[String]]].apply(mainClass)
-        Json.fromFields(List(
-          (N, Json.fromString(Platform.Js.name)),
-          (C, configJson),
-          (M, mainClassJson)))
+        Json.fromFields(
+          List((N, Json.fromString(Platform.Js.name)), (C, configJson), (M, mainClassJson)))
       case Platform.Native(config, mainClass) =>
         val configJson = nativeEncoder(config)
         val mainClassJson = implicitly[RootEncoder[Option[String]]].apply(mainClass)
-        Json.fromFields(List(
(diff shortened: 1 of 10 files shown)


jvican:
Initially, I thought that having a list of main classes would be better because build tools could persist this list if they have it, and just one main class if they don't. But I've just realized that the manifest file only allows a jar to define a main class and this restriction is kind of natural, so let's just add a field `mainClass: Option[String]` instead :+1: 
  
Also, since in Bloop every build target represents a platform (there cannot be a build target that is both jvm and js, or js and native, or native and jvm), let's move the definition from `PlatformConfig` to the `Platform` as a field `def mainClass: Option[String]`. WDYT?

jvican:
I think the style improvements PR has created some conflicts in the PR, could you fix them so that we can merge this soon ? :wink:

tindzk:
I rebased it and fixed the failing test case. It would be good to also export the default main class from the sbt plug-in, but I could not figure it out yet. Any ideas?

jvican:
It seems there are still some conflicts because of the other PR and CI doesn't pass.

tindzk:
I rebased again. Strangely, only the Windows build is failing: https://ci.scala-lang.org/scalacenter/bloop/3450/6 Is there anything special about the Windows tests I should be aware of?

Duhemm:
I think it's failing because this PR changes the format of the configuration file, but the version of the schema is not updated. It should be as simple as bumping `schemaVersion`: 
https://github.com/scalacenter/bloop/blob/07bf16e7731457ca48e32df47e22e9e0b8b89f0a/project/BuildPlugin.scala#L243

tindzk:
It failed again, but this time there was not sufficient memory:

```
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (malloc) failed to allocate 880976 bytes for Chunk::new
# An error report file with more information is saved as:
# C:\Program Files (x86)\Jenkins\workspace\bloop\build-integrations\sbt-1.0\hs_err_pid3672.log
#
# Compiler replay data is saved as:
# C:\Program Files (x86)\Jenkins\workspace\bloop\build-integrations\sbt-1.0\replay_pid3672.log
```

config/src/main/scala-2.10/bloop/config/ConfigEncoderDecoders.scala
```diff
@@ -109,40 +109,39 @@ object ConfigEncoderDecoders {
...
-          c.get[List[String]](M).map(mainClass =>
-            Platform.Jvm(config, mainClass.headOption)))
-        case Platform.Js.name => c.get[JsConfig](C).flatMap(config =>
-          c.get[List[String]](M).map(mainClass =>
-            Platform.Js(config, mainClass.headOption)))
-        case Platform.Native.name => c.get[NativeConfig](C).flatMap(config =>
-          c.get[List[String]](M).map(mainClass =>
-            Platform.Native(config, mainClass.headOption)))
+        case Platform.Jvm.name =>
+          for {
```
jvican: I had to write this with a for-comprehension because scalafmt was making the previous statement too big, and this way was simpler and easier to parse to my eye.

tindzk: This looks much better. Thanks.

Issue #333, opened 2018-03-12
Allow bloop to filter out debug output


jvican:
The sea of logs we get under `--debug` is difficult to read because it contains too much unrelated information. Particularly, it shows a lot of unnecessary info from Zinc's side, and annoying gigantic classpaths that clutter up our users' screens.
  
I propose to add a CLI option in `CliOptions` that is called `--debug` and takes an optional selector (`file-watching`, `compilation`, `test`, `bsp` et cetera) so that bloop only shows the debug information related to those tasks. It would be similar to what the reporter does in scalac.

ioleo:
ScalaSpree - started work.

PR #554, opened 2018-06-19
Support filtering of the debug messages

shamsak58:


frontend/src/main/scala/bloop/cli/CliOptions.scala
```diff
@@ -14,6 +14,9 @@ case class CliOptions(
     version: Boolean = false,
     @HelpMessage("If set, print out debugging information to stderr.")
     verbose: Boolean = false,
+    @ExtraName("f")
+    @HelpMessage("If set, print debug messages that contain given string.")
+    filter: String = "",
```
olafurpg: Would it be possible to use `bloop compile foo | grep filter` instead?

jvican: Good question, it got me thinking.
  
@Duhemm made the point that this is less than ideal when the log messages use more than one line (which not only happens for compiler messages but for other bloop logs). I think this is a good argument to merit a filtering option on its own.

jvican:
@shamsak58 The work here is great.
  
I think this approach is too general and pollutes the logger API unnecessarily with a filter that we will most likely not need. In #333 I sketched out a superior approach that would simplify the design. We'd create a child logger every time we are doing a core task (compilation, test, run, link) and we would then check if that logger has the label that the user has passed in via the CLI. Here you use a label as a `filter` in `CliOptions`, I think that we should just make `debug: List[String]` instead of `debug:Boolean` where `Nil` accounts for all of the debug logs. If that doesn't work, we should probably add a new flag called `debugFilters: List[String]`. We also need to create the child loggers in every use site of the tasks, and remove the filter from `BloopLogger`.
  
Given that there's a lot of work to be done to match the API I describe, let's close this pull request. For the moment, I won't have time to come back to this since I'll be quite busy in the next weeks. Anyone is invited to give #333 another try! :smile:

PR #555, opened 2018-06-19
Add the possibility to disable color 

shamsak58:
No Color users add support

### DIFF
--- a/frontend/src/main/scala/bloop/cli/CliOptions.scala
+++ b/frontend/src/main/scala/bloop/cli/CliOptions.scala
@@ -14,6 +14,8 @@ case class CliOptions(
     version: Boolean = false,
     @HelpMessage("If set, print out debugging information to stderr.")
     verbose: Boolean = false,
+    @HelpMessage("If set, do not color output.")
+    noColor: Boolean = false,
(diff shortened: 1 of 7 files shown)


frontend/src/main/scala/bloop/cli/CliOptions.scala
```diff
@@ -14,6 +14,8 @@ case class CliOptions(
     version: Boolean = false,
     @HelpMessage("If set, print out debugging information to stderr.")
     verbose: Boolean = false,
+    @HelpMessage("If set, do not color output.")
+    noColor: Boolean = false,
```
olafurpg: Would it be better to respect the environment variable `NO_COLOR`? http://no-color.org/

jvican: Indeed, this would be a great default. I'll commit a message adding it so that we can unblock this pull request next week.

jvican:
Fixes #568

jvican:
Added bloop to the list of software that now supports `NO_COLOR` https://github.com/jcs/no_color/pull/30, thanks for the suggestion @olafurpg

jvican:
Great job @shamsak58 :tada: Thank you for this pull request.

PR #560, opened 2018-06-22
Update systemd instructions

japgolly:


japgolly:
No problem. Patch updated :)

tues:
Thanks! I've finally managed to test it on a system with an XDG-compliant DE (normally I use Bloop on a headless VM) and it works fine :slightly_smiling_face:

Issue #152, opened 2018-01-11
Implement the gradle integration


jvican:
This is work in progress. It's an umbrella ticket so that interested people can subscribe to this and receive updates when the integrations is merged. /cc @time2die

jvican:
> I've taken a look at the Maven integration and it seems that all that the plugin is doing is writing a configuration file for Bloop, is that correct? I guess I'm missing something because then how is Maven aware of the compilation result?
  
Yes, writing a bloop configuration file https://scalacenter.github.io/bloop/docs/configuration-format/ is all it's required.
  
In the sbt integration, bloop does not share the classes directories with the stock build tool by default to avoid misinteractions. This can be changed by redefining an sbt setting, though. In Maven's case, it currently shares the classes directories. We can change that as we see fit.
  
I was adding some fixes to the maven integration two days ago, they are available in this branch https://github.com/scalacenter/bloop/tree/topic/add-docs-maven-plugin. I still need to work a little bit more on it, but I was trying it out in Spark's mvn build and was working :smile:.
  
@eyalroth If you wanna help, starting to put together a Gradle integration would be invaluable -- Maven integration was extremely tricky to get right because I used some tricks to recycle all the settings from `scala-maven-plugin`, but in Gradle's case all that complexity goes away because it has official Scala support. If you need live chatting, contact me at https://gitter.im/scalacenter/bloop, happy to answer any questions.
  
> Also, is there a plan to integrate Bloop with IntelliJ?
  
Yes, we're working with the build tool team at IntelliJ to use the Build Server Protocol http://github.com/scalacenter/bsp/ to communicate with bloop to import projects and compile in Bloop's build server. We're moving fast in this area, so we'll have exciting progress to show soon.

eyalroth:
I'm a little confused about this.

From the original PR of the maven integration (#148) I understand that the integration is in fact an additional maven plugin which extends from [davidB's Scala Maven Plugin](https://github.com/davidB/scala-maven-plugin), and according to issue [#208](https://github.com/davidB/scala-maven-plugin/issues/208) on said repository I gather that the maven plugin integrates with the [old (and deprecated) Zinc](https://github.com/typesafehub/zinc).

Now, as said earlier, the bloop maven integration only writes configuration files to bloop (this is done via the `installBloop` maven task, correct?). I'm assuming it is also possible to integrate this task / goal in the `pom.xml` so that Maven would update bloop with project changes (though this does not appear to be explained in Bloop's official guide).

[...]

PR #566, opened 2018-06-29
Add gradle integration

vigoo:
This is a first stage gradle implementation (https://github.com/scalacenter/bloop/issues/152).

Featuring:
- Works with Gradle 3.+ (current version is 4.8.1)
- Adds a `bloopInstall` gradle task that, similarly to the `sbt` integration, generates Bloop's configuration files
- Registers a `bloop` project extension where some customizations can be done from the Gradle build script
- Multiple source sets are supported, `main` is treated special, others always depend on it
- Project dependencies are generated correctly (so they become dependencies in Bloop level too)
- Scala and Java compiler options

NOTE: A custom sbt task downloads Gradle and runs a dummy Gradle build compile time to generate the gradle-api artifact

I was not sure about:
- Where should I put user documentation for it
- What kind of tests I should provide. It would be possible include some small test Gradle projects, compile them both with Gradle and with Bloop and compare something (successful compilation + output of the generated application? or even binary equivalence?)
so these are not included (yet).


jvican:
Thank you @vigoo, this looks great. I'll have a closer look on Monday (on vacation right now), and will fix the CI. It seems we're having some trouble with the Windows CI.

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
```diff
@@ -0,0 +1,247 @@
...
+    ))
+
+  private val scalaTestFramework = Config.TestFramework(
+    List(
+      "org.scalatest.tools.Framework",
+      "org.scalatest.tools.ScalaTestFramework"
+    )
+  )
+
+  private val specsFramework = Config.TestFramework(
```
jvican: What do you think of sharing these default frameworks with `MojoImplementation` by adding them to the `jsonConfig` module? The idea would be to have a value called `DefaultTestFrameworks` that can be invoked by both integrations.
  
Another thing that we better do now here is to have a function `def detectTestFrameworks(artifacts: List[Config.Artifact]): List[Config.TestFramework]` that would add these only if the classpath contains the artifact associated with every framework (scalatest, specs, junit, etc).
  
(This last point is minor, if you don't feel like taking on it we can add a ticket so that another contributor does it :smile:)

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/tasks/BloopInstallTask.scala
```diff
@@ -0,0 +1,63 @@
...
+
+    // The 'main' source set maps to the raw project name (as all integrations do)
+    val mainProjectName = converter.getProjectName(project, mainSourceSet)
+    generateBloopConfiguration(mainProjectName, Set.empty, mainSourceSet, targetDir)
+
+    // Generate the bloop configuration files for the rest of the source sets
+    for (sourceSet <- otherSourceSets) {
+      val projectName = converter.getProjectName(project, sourceSet)
+      // Hardcore an implicit dependency for every source set to the main source set (compile)
+      generateBloopConfiguration(projectName, Set(mainProjectName), sourceSet, targetDir)
```
jvican: I'm surprised that this is not present in the project configuration of gradle. Would you mind confirming that adding this dependency is necessary?

project/BuildPlugin.scala
```diff
@@ -357,6 +371,31 @@ object BuildImplementation {
...
+      Keys.unmanagedJars.in(Compile) := unmanagedJarsWithGradleApi.value,
+      BuildKeys.fetchGradleApi := {
+        val logger = Keys.streams.value.log
+        // TODO: we may want to fetch it to a custom unmanaged lib directory under build
+        val targetDir = (Keys.baseDirectory in Compile).value / "lib"
+        GradleIntegration.fetchGradleApi(Dependencies.gradleVersion, targetDir, logger)
+      },
+    )
+
+    lazy val unmanagedJarsWithGradleApi: Def.Initialize[Task[Keys.Classpath]] = Def.taskDyn {
```
jvican: This is a trick I did so that it `fetchGradleApi` is done exactly before it's required (this is important so that IntelliJ and other tools using the output of library dependencies of a project can see the lib jar if it's not yet downlaoded).

project/Dependencies.scala
```diff
@@ -18,6 +18,8 @@ object Dependencies {
   val mavenApiVersion = "3.5.2"
   val mavenAnnotationsVersion = "3.5"
   val mavenScalaPluginVersion = "3.2.2"
+  val gradleVersion = "4.8.1"
```
jvican: I took the liberty of updating this to 4.8.1, 3.0.0 is very outdated.

vigoo: I chose 3.0 because that is the oldest version compatible. By compiling it with `3.0` it is compatible with all versions above 3  including the latest `4.8.1` (I tried it). I also asked a friend who works for Gradle about this and he recommended to choose the oldest one we would like to support.

vigoo:
Thanks for the review, I will go through all your comments tomorrow.

project/Dependencies.scala
jvican: Oh interesting! So did your friend recommend you 3.0 for testing the gradle API or for testing the plugin itself? I'm fine leaving this for 3.0, but I'd like to test the plugin with the latest gradle version so that we are closer to the versions that our users will be at.

vigoo: I think the best would be to compile with the oldest version possible to cover as much versions as we can, but always test it with the latest (or with both). 
My fear of people not having the latest Gradle version may be a bit subjective as we at Prezi are still stuck with 2.13 :) But covering more versions without extra effort would no hurt, I think.

jvican:
The last support adds the required sbt magic to be able to access the project base directory for `apache/kafka` from either unit or integration tests. Kafka will be cloned automatically by the build tool (and cached). This general-purpose mechanism will allow us to do the same for our future maven and mill infrastructure.

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/tasks/BloopInstallTask.scala
vigoo: What I found so far is:
- The `test` source set's compile classpath is extended with the `main` source set's compilation output: https://github.com/gradle/gradle/blob/master/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaPlugin.java#L277
- The `test` configuration "extends from" the `main` configuration: https://github.com/gradle/gradle/blob/master/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaPlugin.java#L427

So it's not as high level as I would expect but hopefully can be somehow used to figure out the depenencies. Will experiment with it.

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
vigoo: Good idea, will do it

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/tasks/BloopInstallTask.scala
jvican: After thinking more about this, we do need to add this implicit dependency as a project dependency. As you have figured out, Gradle adds the main output dir to the test classpath, but at the project level we still need to add this dependency so that Bloop can have properly schedule compilation sequentially (if there was no dependency, Bloop would run the compilation of main and test separately, which would fail because test depends on main).

jvican:
Whenever you add changes to the pull request, remember to rebase so that CI is faster :wink:

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/tasks/BloopInstallTask.scala
vigoo: Yes that was my orgiginal intention too. But it would be nice figuring out the dependencies between the Gradle source sets instead of hardcoding that all non-main depends on main.

jvican: Yeah. That’s most likely not supported by Gradle, though... Maybe exposing a way to configure that in the gradle config would make sense.

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/BloopParameters.scala
```diff
@@ -6,64 +6,47 @@ import org.gradle.api.Project
...
+ *
+ * {{{
+ * bloop {
+ *   targetDir "$projectDir/.bloop"
+ *   mainSourceSet "main"
+ *   compilerName "scala-compiler"
+ * }
+ * }}}
+ */
+final class BloopParameters(project: Project) {
```
vigoo: Seems like it cannot be final, because of some unexpected Gradle magic :)

```
Caused by: org.gradle.api.GradleException: Could not generate a proxy class for class bloop.integrations.gradle.BloopParameters.
        at org.gradle.api.internal.AbstractClassGenerator.generateUnderLock(AbstractClassGenerator.java:201)
        at org.gradle.api.internal.AbstractClassGenerator.generate(AbstractClassGenerator.java:64)
        at org.gradle.api.internal.ClassGeneratorBackedInstantiator.newInstance(ClassGeneratorBackedInstantiator.java:36)
        at org.gradle.api.internal.plugins.DefaultConvention.create(DefaultConvention.java:103)
        at bloop.integrations.gradle.syntax$ProjectExtension.createExtension(syntax.scala:42)
        at bloop.integrations.gradle.BloopPlugin.apply(BloopPlugin.scala:20)
        at bloop.integrations.gradle.BloopPlugin.apply(BloopPlugin.scala:17)
        at org.gradle.api.internal.plugins.ImperativeOnlyPluginApplicator.applyImperative(ImperativeOnlyPluginApplicator.java:35)
        at org.gradle.api.internal.plugins.RuleBasedPluginApplicator.applyImperative(RuleBasedPluginApplicator.java:43)
        at org.gradle.api.internal.plugins.DefaultPluginManager.doApply(DefaultPluginManager.java:137)
        ... 124 more
Caused by: java.lang.VerifyError: Cannot inherit from final class
        at java.lang.ClassLoader.defineClass1(Native Method)
        at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
        at java.lang.ClassLoader.defineClass(ClassLoader.java:642)
        at sun.reflect.GeneratedMethodAccessor8.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:75)
        at org.gradle.api.internal.AsmBackedClassGenerator$ClassBuilderImpl.generate(AsmBackedClassGenerator.java:908)
        at org.gradle.api.internal.AbstractClassGenerator.generateUnderLock(AbstractClassGenerator.java:199)```

**Comment by vigoo** [comment]
FYI: I was playing with manually trying out the plugin with Kafka, does not work yet, will continue investigating sometime in the next days

**Comment by vigoo** [comment]
Added `gradle-test-kit` too from the downloaded Gradle build as a test dependency and wrote an initial test which only checks that the `bloopInstall` task is added to the project. Looks like the test kit downloads the given Gradle version for itself, so we can test with a different version (for example the latest 4.8.1) than the version used for the deps.

File: .drone.yml [hunk 6]
```diff
@@ -3,9 +3,11 @@
 matrix:
   SBT_RUN:
     - sbt "benchmarks/compile" \
-          "millBloop/compile" \
           "jsonConfig210/test" \
           "jsonConfig212/test" \
+          "millBloop/compile" \
+          "mavenBloop/compile" \
+          "gradleBloop/compile" \
```
jvican: Can we change this to `gradleBloop/test` to test the new gradle plugins in the CI?

jvican:
@vigoo Is this working on your gradle builds? If it is, maybe we can hold off from adding the kafka integration test and merge this in 1.1.0 and label it as beta. What do you think?

vigoo:
@jvican sorry I did not have time to work on this since my last commit but would like to continue this week. Last time I checked it worked for some example projects which are not added yet as tests, and did not work with Kafka yet. What is the planned date for 1.1.0?

jvican:
@vigoo :+1: We're flexible on the date, but I have the hope of releasing it in two/three weeks. We release on merge though, so as soon as we merge this Gradle users have a version they can use to test the integration and report bugs. That can be useful to improve the quality of the integration for 1.1.0+, and catch things that our tests won't.

vigoo:
Made some progress in applying the plugin for Kafka, still not finished though. 
I had to create a 2.11 version to be able to apply the Gradle plugin in the first time. I'm not completely sure how the plugin's dependencies are conflicting with the target project's dependencies though (the plugin failed to load without any error/debug messages). 
There were some smaller issues now fixed. 
The current biggest problem I see is the Java-only projects in Kafka where the Scala configuration cannot be determined by looking at the Gradle (sub)project. By leaving the scala config empty for bloop it defaults for 2.12.6 in `/Users/martin/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.12.6/scala-compiler-2.12.6.jar` :) 
I'm not sure how to handle this:
- find a valid scala configuration in other subprojects of the big project and use that (that would work in this case but sounds a bit random)
- require additional config in the gradle build definition for java-only projects to support building them with bloop
- do something in bloop to handle this case correctly
- or what is probably the most reasonable is to not create bloop projects for these Java-only subprojects at all but compile them with gradle and add as simple classpath dependencies.

jvican:
> I had to create a 2.11 version to be able to apply the Gradle plugin in the first time. I'm not completely sure how the plugin's dependencies are conflicting with the target project's dependencies though (the plugin failed to load without any error/debug messages).
  
This most likely has to do with Gradle class loading Scala 2.11, and then it's only normal that class loading the plugin fails, good catch.
  
> The current biggest problem I see is the Java-only projects in Kafka where the Scala configuration cannot be determined by looking at the Gradle (sub)project. By leaving the scala config empty for bloop it defaults for 2.12.6
  
Yes, and with the latest commit that I've added and had staged in another branch, the scala instance for an empty scala config will always be the instance for the scala version that Bloop server uses. The previous approach wasn't quite working and I did it because I kind of predicted this would be a problem for the authors of integrations :smile:.
  
Note that this is not problematic with the rest of the scala versions of your project because the scala compiler will never be run (there are no scala sources in the project). Unfortunately, because of design limitations of Zinc, we're forced to always pass in an object of `ScalaInstance` even though we only want to compile Java sources.
  
> or what is probably the most reasonable is to not create bloop projects for these Java-only subprojects at all but compile them with gradle and add as simple classpath dependencies.
  
In all the other integrations, we can compile Java sources without a problem. This is, actually, mandatory to some extent because Java sources and Scala sources can have both way dependencies.
  
Let me know if I've clarified everything! I just rebased your PR and added the commit I'm talking you about. I had to force push to rebase, so make sure you fully update your branch. Most likely, `git reset --hard HEAD~15 && git pull` will fix it.

jvican:
@vigoo Feel free to have a look at my latest change in the config API and the bloop configuration, this should make the CI pass again. Note that I've done a change important for semantics: the classpath and all the artifact-related variables are lists instead of sets. The reason for this is that we need to respect the order of the classpath that Gradle gives us, because the order of the classpath modifies the semantics of compilation (and Bloop wants to compile projects as similar as possible to Gradle).

jvican:
I rebased everything on top of latest master that adds much faster CI turnarounds (the caching wasn't working because of a regression and was killing me).
  
I just realized that I may have misunderstood the reason why you and I think that a 2.11 version of gradle is required. You seem to think that this is the case for Gradle projects that use Scala 2.11, and that's why you left Gradle for 2.12, and I think this is due to gradle's scala support being written in Scala 2.11 and class loaded (this would make sense because they use a very old version of think from three/four years ago). Based on this, I intuited that gradle 2.12 wasn't required and therefore removed it.
  
@vigoo What do you think? Please let me know if gradle 2.12 is indeed required, or feel free to just add it back. By the way, try to have a look at my two commits and review them whenever you've got some time :wink:

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
```diff
@@ -76,18 +75,16 @@ final class BloopConverter(parameters: BloopParameters) {
         name = getProjectName(project, sourceSet),
         directory = project.getProjectDir.toPath,
         sources = getSources(sourceSet),
-        dependencies = allDependencies.toArray,
+        dependencies = allDependencies.toList,
         classpath = classpath,
         out = project.getBuildDir.toPath,
-        analysisOut = getAnalysisOut(project, sourceSet, targetDir).toPath,
```
vigoo: Not providing this property causes the following error for me with the generated bloop project:
```[E] Attempt to decode value on failed cursor: DownField(analysisOut),DownField(project)```

why did you remove it?

vigoo:
@jvican I finally had some time to see what fails in Kafka, and it is still the Java-only projects, it now fails with "`Problem(-1,Error,Failed to compile 'clients's Java sources because the default Scala instance couldn't be created.,NoPosition,), Problem(-1,Error,Add Scala configuration to the project. If that doesn't fix it, report it upstream,NoPosition,)` (had to temporarily change the show instance of `Task` to be able to see it). Would you like me to investigate this further or is it something you will solve?

jvican:
> I finally had some time to see what fails in Kafka, and it is still the Java-only projects
  
Thanks for confirming what the bug is Daniel. I'll have a look on Monday, this error may be related to a change I introduced in this branch to improve the reliability of the scala class loaders instantiation.

jvican:
By the way, I didn't say anything but I got `bloopInstall` to work with Kafka. I didn't have time to run bloop on the project and see if all the modules compiled, so I leave that task for the future.

jvican:
I couldn't find your handle in Twitter @vigoo, so I linked to your GH profile:
https://twitter.com/jvican/status/1026875681987325958

PR #572, opened 2018-07-03
Add initial support for running Scala.js tests

tindzk:
This pull request adds support for running Scala.js tests.

**To do:**
- [x] Add Scala.js 1.0 support
- [x] Eliminate warning triggered by calling `dispose()` too early
- [x] Introduce a configuration parameter to use `JSDOMNodeJSEnv` only if the project has DOM tests, otherwise it should default to `NodeJSEnv`
- [x] Spawn Node.js process using NuProcess

**Changes:**
- `TestSuiteEvent`: Remove `logger.trace()` as exceptions and their
      traces are already printed by the test runner
- `Tasks`: Only link projects and look up tests if test frameworks are
      defined
- `ScalaJsToolchain`: Make main class optional since test projects do
      not necessarily have a main class
- `ScalaJsToolchain`: Add function to retrieve test frameworks
- `DiscoveredTests`: Rename to `DiscoveredTestFrameworks` and make it
      platform-agnostic


jvican:
Excited to see this! I won't be able to review/merge this until I'm done with pipelined compilation (this pull request will need to be updated on top of my latest changes there). Coincidentally, I've changed some of the same code that handles link in Scala Native and Scala.js, you can see a diff [here](https://github.com/scalacenter/bloop/commit/98ab9a843482092c27b49019d24871dfe25f60eb).

jvican:
I just merged the changes that I linked in the diff here, so you should be good to rebase this PR + finish the TO-DOs 😍 I would love to have cross-platform testing in the next release. @tindzk

tindzk:
Great. I will look into it later this week.

tindzk:
@jvican Please feel free to review! Using NuProcess for spawning Node is not that straightforward because we would have to implement a custom `JSEnv` and duplicate much of the wiring Scala.js provides us with.

jvican:
> Using NuProcess for spawning Node is not that straightforward because we would have to implement a custom JSEnv and duplicate much of the wiring Scala.js provides us with.
  
To have a consistent story for test execution, we need to run tests in a forked process. Could you have a look at how Scala.js does this in its sbt plugin? I'm pretty sure they respect the `fork::test` setting and act accordingly.

olafurpg:
Scala.js prohibits `fork := true` https://github.com/scala-js/scala-js/issues/1590 iiuc, tests run in a forked environment either way so it doesn't matter in the case for bloop

tindzk:
@jvican It uses Scala's `ProcessBuilder`: https://github.com/scala-js/scala-js/blob/0ff1fbe3f7a003482c3b785b6a374202f44f47c1/js-envs/src/main/scala/org/scalajs/jsenv/ExternalJSRun.scala#L125-L141

I saw that my tests failed. It seems to be related to `lamppc23.epfl.ch:8080` not being accessible. Could you look into that?

jvican:
> I would also prefer if we could handle the process creation ourselves, but it would require a significant amount of work to support all JavaScript environments and we would have to keep the code in-sync. See also this discussion.
  
Thanks for the link to the discussion. I think duplicating those environments is a worthy price given what's at stake here. This PR also duplicates some of the unnecessary logic to run scalajs environment bits and the test adapters, so I feel that doing as I propose will give us not only better control but we'll end up with easier to maintain and understand bloop code.
  
I've been having a look at the code and it doesn't look like we would need to duplicate much of the code. The environments in which Scala.js runs are a few only, and they are not likely to change in the future, and the test adapter logic is mostly in its entirety already available in Bloop.
  
I agree with you in the future we should talk to Sébastien to change this upstream and abstract over the details of the adapter execution and the process spawn. However, I don't see much value to this for now because we still have the dependency on Scala.js 0.6.x.
  
> Where do I have to set the Bloop configuration version exactly? project/Build.scala does not exist unfortunately.
  
I'm sorry, I meant `project/BuildPlugin.scala`.

tindzk:
While there is not too much effort for Node.js, I am more concerned about alternative environments. For example, jsdom's API changed a few times in recent years and the Scala.js developers went to great lengths to keep their code backward-compatible (see [JSDOMNodeJSEnv.scala](https://github.com/scala-js/scala-js-env-jsdom-nodejs/blob/master/jsdom-nodejs-env/src/main/scala/org/scalajs/jsenv/jsdomnodejs/JSDOMNodeJSEnv.scala#L86)).

There is also an [environment](https://github.com/scala-js/scala-js-env-selenium) for Selenium which uses a browser to run the linked JavaScript file. For this it uses an external dependency, so I am doubtful whether we will have much control over the process creation here.

I do not mind if we duplicate all these environments, but we have to be conscious of the additional maintenance overhead.

jvican:
I propose we fork Scala.js and modify the access of their APIs (and maybe add some interfaces) so that we can extend it in bloop (by overriding the method that spawns tasks for example). This way we can support both 0.6.x and 1.x, while still letting Scala.js do all the logic and book-keeping to support all ecosystems. In the future, we can ask Sebastien to merge changes in the 1.x branch so that we can extend it in a nicer way in the 1.x series. What do you think?

tindzk:
This is a reasonable solution, especially since our changes could also be used by other build tools. I will look into it next week.

jvican:
@tindzk Hey Tim, I discussed this with @sjrd in person and he proposed we just duplicate all the required logic to replace the process implementation by ours. From this code, we can abstract and try to see which entrypoints we would need to add to scalajs 1.0 to support this use case.

sjrd:
> Selenium is deprecated in 1.x

No. You confused with PhantomJS ;)

tindzk:
I changed the Scala.js 1.0 bridge to use NuProcess. It required significantly less code duplication than I thought as I could re-use `ComRun`. Unfortunately, I could not get it fully to work.

The JavaScript binary files are correctly sent to the Node process. Node connects to the `ComRun`'s socket server. Looking at VisualVM, the thread is stuck at reading data from the Node client (at [this line](https://github.com/scala-js/scala-js/blob/master/nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/ComSupport.scala#L66)).

I added `console.log` calls before and after the test execution and these are processed fine. Here is the shortened log:

```
Generated JavaScript file '/tmp/build-pine/bloop/pine-test/out.js'
[D] Resolving test frameworks: List(List(utest.runner.Framework))
[D] Starting process node...
[D] Current working directory: Some(/home/tim/dev/pine)
[D] Process started at PID 10219
[D] Sending file MemVirtualBinaryFile(sourceMapSupport.js)...
[D] Writing 73 bytes...
[D] Closed stream
[D] Writing 1 bytes...
[D] Sending file MemVirtualBinaryFile(scalaJSEnvInfo.js)...
[D] Writing 86 bytes...
[D] Closed stream
[D] Writing 1 bytes...
[D] Sending file MemVirtualBinaryFile(nodeConsoleHack.js)...
[D] Writing 683 bytes...
[D] Closed stream
[D] Writing 1 bytes...
[D] Sending file MemVirtualBinaryFile(comSetup.js)...
[D] Writing 1507 bytes...
[D] Closed stream
[D] Writing 1 bytes...
[D] Sending file MemVirtualBinaryFile(before-tests.js)...
[D] Writing 28 bytes...
[D] Closed stream
[D] Writing 1 bytes...
[D] Sending file MemVirtualBinaryFile(/tmp/build-pine/bloop/pine-test/out.js)...
[D] Writing 65536 bytes...
[...]
[D] Writing 65536 bytes...
[D] Writing 47506 bytes...
[D] Closed stream
[D] Writing 1 bytes...
[D] Sending file MemVirtualBinaryFile(after-tests.js)...
[D] Writing 27 bytes...
[D] Closed stream
[D] Writing 1 bytes...
[D] Closing stdin stream...
[D] [stdout] before tests
[D] [stdout] after tests
```

jvican:
@tindzk Do you want me to eventually have a closer look to this issue or do you plan to? I suspect we will need this feature in one or two weeks to support #682 reliably in BSP.

tindzk:
Sorry, @jvican. I saw your comment, but did not have a chance to try your suggestions. I will do it this weekend.

tindzk:
The bug was related to a missing module initialiser for Scala.js' test interface. It now works as expected for Scala.js 1.0, including jsdom support.

Unfortunately, sharing the code across Scala.js 0.6 and 1.0 will not be possible since the testing interface has evolved significantly. As Scala.js 0.6 is being phased out, I propose we keep the spawning logic as-is for 0.6 and only use NuProcess for 1.0.

Which functions should I move out of `Tasks`? I put `linkJs` into the object because it also contains `runJVM` and `runNativeOrJs`.

jvican:
Excellent, could you rebase all your changes onto latest master? I'll then review the PR in depth and push whatever improvements I can find before merging the PR.

jvican:
Excellent, could you rebase all your changes onto latest master? I'll then review the PR in depth and push whatever improvements I can find before merging the PR.

jvican:
Excellent, could you rebase all your changes onto latest master? I'll then review the PR in depth and push whatever improvements I can find before merging the PR.

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -198,7 +200,7 @@ object Config {
 
   case class File(version: String, project: Project)
   object File {
-    final val LatestVersion = "1.0.0"
+    final val LatestVersion = "1.1.0"
```
jvican: Let's use 1.1.0-M1 here instead, as we'll do more changes before 1.1.0

tindzk:
Sure, I rebased the pull request. Feel free to make further changes.

jvican:
Thank you. Can we add more tests to the codebase to check that this works? I propose to modify `TestTaskTest` to discover tests and run them correctly in the JS modules of utest (just as we do for its counterpart in the JVM).

tindzk:
Sure. I noticed that the tests failed under Linux (https://ci.scala-lang.org/scalacenter/bloop/4072/14). I will look into it over the weekend and write an additional test case.

tindzk:
It appears that the test failure is due to a lack of memory. From the CI log:

```
Java HotSpot(TM) 64-Bit Server VM warning: CodeCache is full. Compiler has been disabled.
Java HotSpot(TM) 64-Bit Server VM warning: Try increasing the code cache size using -XX:ReservedCodeCacheSize=
CodeCache: size=262144Kb used=260120Kb max_used=260132Kb free=2023Kb
 bounds [0x00007fe2e4000000, 0x00007fe2f4000000, 0x00007fe2f4000000]
 total_blobs=69621 nmethods=68684 adapters=832
 compilation: disabled (not enough contiguous free space left)
```

Unfortunately, I cannot run the tests locally as I had to interrupt `frontend/integrationSetUpBloop` after 1.5 hours. Is it possible to run `bloop.scalajs.ScalaJsToolchainSpec` without cloning all those external code repositories? I am getting the following error:

```
[error] Test bloop.scalajs.ScalaJsToolchainSpec.canLinkScalaJsProject failed: java.lang.RuntimeException: Project scalajs-1 does not exist at /home/tim/.sbt/1.0/staging/bloop-integrations-4.0.csv, took 0.003 sec
[error]     at scala.sys.package$.error(package.scala:26)
[error]     at bloop.tasks.TestUtil$.$anonfun$getBloopConfigDir$1(TestUtil.scala:147)
[error]     at scala.Option.getOrElse(Option.scala:121)
[error]     at bloop.tasks.TestUtil$.getBloopConfigDir(TestUtil.scala:147)
[error]     at bloop.tasks.TestUtil$.loadTestProject(TestUtil.scala:153)
[error]     at bloop.scalajs.ScalaJsToolchainSpec.<init>(ScalaJsToolchainSpec.scala:19)
[error]     ...
```

jvican:
I'll look into these errors, thanks for reporting.
  
I don't know why `integrationSetUpBloop` was so slow... Once all the repos are cloned in the sbt staging directory, it runs on about 5 minutes in my machine. If you want to try it out without running it, I recommend cloning the version of utest specified in `build-integrations` somewhere in your computer, and then adding bloop manually, running `bloopInstall`, and pass the bloop config directory that was generated as the input of the test (so that the state loads the project definition from there).

tindzk:
I modified all `build.sbt` files in `build-integrations/` and disabled `twitterDodo` in `BuildPlugin.scala` (see [commit](https://github.com/tindzk/bloop/commit/78bdc6fa0fa83db005a04dcfdc0b19e8adc921ab)), but it still takes a significant amount of resources. Also, I noticed that each time I run `frontend/integrationSetUpBloop`, it re-creates all Bloop configurations.

It would be useful to be able to test Bloop's core functionality without having to clone any third-party repositories and download their dependencies.

tindzk:
> I would need to know what you mean by "significant amount of resources". The biggest issue is cloning guardian/frontend IIRC because sbt is incredibly slow doing so.

In my case, it produced 1.4 GiB in `$HOME/.sbt/1.0/staging/`, not counting the library dependencies. Spark, Lichess and Scala are the largest repositories here.

The process is slow mainly because for every project, it has to clone the Git repository, load the sbt build file, download its dependencies and then generate a Bloop configuration. I doubt that we can speed this up, but we should rethink whether Bloop's core functionality should be tested using integration tests. It might be better to inline a minimal testing framework like [minitest](https://github.com/monix/minitest) and modify it to our needs, such that we can perform more fine-grained testing. It would be useful to test scenarios like: Are all registered unit tests found and run? Are test failure counts correctly propagated to Bloop? etc.

Let's continue this conversation on a separate issue.

jvican:
I'm closing this pull request, I'll be submitting another one based on this one with significant changes to the whole test infrastructure + tests + reliable cancellation.

PR #579, opened 2018-07-05
Prepare configuration file for final 1.0.0

jvican:


jvican:
Once CI passes, I'll rework the JSON schema for the finished configuration file format.

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -188,31 +172,25 @@ object Config {
...
-
   case class Project(
       name: String,
       directory: Path,
-      sources: Array[Path],
-      dependencies: Array[String],
-      classpath: Array[Path],
+      sources: List[Path],
+      dependencies: List[String],
+      classpath: List[Path],
```
tindzk: Should it be spelled `classPath`?

frontend/src/main/scala/bloop/Project.scala
```diff
@@ -96,19 +95,19 @@ object Project {
   def fromConfig(file: Config.File, logger: Logger): Project = {
     val project = file.project
     val scala = project.`scala`
-    val isEmpty = scala.organization.isEmpty && scala.name.isEmpty && scala.version.isEmpty && scala.jars.isEmpty
```
tindzk: It did not work for me before because it should have been `||` instead of `&&`. The new solution is much cleaner though.

config/src/main/scala/bloop/config/Config.scala
jvican: The cli for Scala keeps it lowercase

PR #592, opened 2018-07-18
Add release notes for v1.0.0 and log plist service

jvican:


jvican:
I finally added some code that automatically generates html to document commands. At the moment a command changes, the `bloop.util.CommandsDocGenerator` can be run and its contents copy-pasted to `commands-reference.md`.

PR #594, opened 2018-07-23
Fix link to ruben's github profile

rethab:


rberenguel:
Haha @rethab @jvican I can only hope to be as productive as @Duhemm

PR #595, opened 2018-07-23
Add -sbt-dir in the drone yaml file

jvican:
There has been an unknown regression in caching in the CI that was
causing us CI times of about 16 seconds in linux. This is unbearable.
Hopefully this commit fixes it.


jvican:
I also changed the Windows script so that sbt is not invoked if DOCS is found in the commit message, we're done here, this is good to go.

PR #597, opened 2018-07-29
Fix grammar in the "Snappy" part of the homepage

kubukoz:


kubukoz:
Even better, updated. @propensive mind taking another look?

Issue #599, opened 2018-08-07
Remove dependency between test sources by default in sbt


jvican:
A lucky event yesterday made me realize that we were being too conservative when extracting configurations from sbt. A symptom of this is that if a project `A` depends on `B` (with no extra configuration), then the test sources of `B` would depend on the test sources of `A`. These are not the default semantics of sbt and they are effectively constraining opportunities to compile any builds faster (in most of the cases, the compilation of tests sources would be parallel).

PR #600, opened 2018-08-07
Remove default dependency between test sources

jvican:
Fixes https://github.com/scalacenter/bloop/issues/599.

The fix is just obvious: it doesn't use `configuration` when no
configuration is used, and instead relies on the default `Compile`.


jvican:
Let's run the sbt community build on this to be 100% sure that this works and doesn't break compilation.

PR #608, opened 2018-08-09
Ignore non-Scala and non-Java project dependencies

dsilvasc:
This avoids an exception when an extension method in syntax.scala tries to look up the JavaPluginConvention for a project that doesn't have one.

### DIFF
--- a/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
+++ b/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
@@ -51,7 +51,9 @@ final class BloopConverter(parameters: BloopParameters) {
 
     // We cannot turn this into a set directly because we need the topological order for correctness
     val projectDependencies: List[ProjectDependency] =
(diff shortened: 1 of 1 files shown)


dsilvasc:
The PR build failed because the CI instance was out of disk space.  How can I retry the build?

jvican:
@dsilvasc I've just given you access to the maintainer list so that you can retry the build as you want. Your eight PRs are more than worth it :wink: If you want to restore it, you need to go to ci.scala-lang.org, log in with GitHub, and then you'll be able to go to the CI run you want and click restart.

dsilvasc:
@jvican Thank you :) Can you please include @dsilva as well?  I shouldn't log into that CI system with my employer's account (this one).

jvican:
I'm gonna give you permission only to @dsilva then, and remove the permission for @dsilvac :wink:

jvican:
Instead of pushing all these commits, can you rebase all your PRs on top of latest master?

dsilvasc:
Just noticed that the gradle API has `findPlugin` which returns `null` on not found.  Switching to that in https://github.com/scalacenter/bloop/pull/631 and adding a test in https://github.com/scalacenter/bloop/pull/630

PR #611, opened 2018-08-09
Disambiguate Gradle project names with project paths

dsilvasc:
…projects with the same name.


dsilvasc:
We could do that, but it might result in odd or inconsistent names.  For example, suppose you have `:foo-service:client`, `:foo-service:server`, `:bar-service:server`, and `:bar-service:storage`.  You'd have `foo-service-server` and `bar-service-server` in bloop.  But you would also end up with `client` and `storage` in bloop and would have to remember that service foo has a client and service bar has a storage layer -- kind of confusing.  Then when you write a client for service bar, `client` in bloop becomes `foo-service-client`.

dsilvasc:
Yeah that seems reasonable.  I'll change it.

jvican:
Hey @dsilvasc, will you be coming back to this one? I would like to merge it soon so that we wrap up all the good work around the gradle plugin and we release it in a stable version.

PR #617, opened 2018-08-10
Fix homebrew formula regarding blp-server logs

daddykotex:


daddykotex:
I'm not sure how I can test a formula, so I'm not entirely sure that it works.

jvican:
Actually the fix should have also `FileUtils.chmod_R 0777, "log/bloop"` judging from the current formula in 1.0.0 https://github.com/scalacenter/homebrew-bloop/blob/v1.0.0/bloop.rb :smile:

daddykotex:
Ahhh I looked at master

PR #618, opened 2018-08-10
Fix permissions in Brew formula

jvican:
Alright, I think that the issue that was causing so many problems is that I was doing chmod 0777 on `log/bloop` but not on `log`. This should fix it.

### DIFF
--- a/project/ReleaseUtils.scala
+++ b/project/ReleaseUtils.scala
@@ -79,6 +79,7 @@ object ReleaseUtils {
        |      bash_completion.install "bin/bash/bloop"
        |      File.delete("bin/blp-coursier")
        |      FileUtils.mkdir_p("log/bloop/")
+       |      FileUtils.chmod_R 0777, "log"
        |
        |      prefix.install "bin"
        |      prefix.install "log"


jvican:
Can you quickly review this @daddykotex ?

jvican:
Compare with the status quo: https://github.com/scalacenter/homebrew-bloop/blob/v1.0.0/bloop.rb#L19

daddykotex:
Personnaly, on my setup, the log folder has the following permission:
`drwxr-xr-x  3 dfranco1  admin  102 10 Aug 08:37 log`
that would translate to 755 if i'm not wrong. But the initial owner was `root` and the `blp-server` process is started by `dfranco1` that's why t was not working.

So there are two solutions:
* allow write by anyone (777)
* allow write by user 755 but make sure owner is the user that starts the service


This MR goes with one, I'm ok with that personally

jvican:
The second option could be better, but I didn't figure out in 5 minutes what was the ruby function I should use for that, so let's just call it a day with the first one :stuck_out_tongue: If somebody takes issue with this approach, they can send a PR. Thanks for the review David!

PR #625, opened 2018-08-14
Update Mill

nightscape:
Newer Mill versions have a slightly different macro for expanding a Discovery,
so the current mill-bloop.jar doesn't work with newer Mill versions anymore

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -28,7 +28,7 @@ object Dependencies {
   val scalaNativeVersion = "0.3.7"
   val scalaJs06Version = "0.6.23"
   val scalaJs10Version = "1.0.0-M3"
-  val millVersion = "0.2.3"
+  val millVersion = "0.2.6"
 
   import sbt.librarymanagement.syntax.stringToOrganization
   val zinc = "ch.epfl.scala" %% "zinc" % zincVersion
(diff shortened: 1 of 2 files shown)


jvican:
Thanks @nightscape, is the latest mill release always recommend to be used? I'm thinking in terms of the impact of this change for our current mill users. Should we ask them to upgrade if they want to use the latest version? Is that unrealistically difficult?

nightscape:
I've been pondering the same thing.
I see several options:
1. Always build a new version for the latest Mill version (like in this PR). If people want updated Bloop functionality, they need to make sure they use the latest Mill version.
2. Cross-build for the last Mill versions.
3. Recommend a source dependency to the Bloop plugin:
```scala
import $url.{`https://raw.githubusercontent.com/scalacenter/bloop/master/integrations/mill-bloop/src/main/scala/bloop/integrations/mill/MillBloop.scala` => bloop}
```
This would not completely get rid of the problem, but make it less frequent (the macro problem would not have occurred with a source dependency).

I guess at the moment any of these would be fine, as Mill is still rather experimental and people probably don't rely on it in production scenarios.

jvican:
Yes, though that’s not a problem since we release on merge, so you’ll be able to use the mill plugin as soon as this makes it to master :) we can wait for source deps for now

nightscape:
:+1:
I added a small note in the docs regarding the fact that the plugin is always built for the latest Mill version.
So from my side this is good to merge.

PR #631, opened 2018-08-28
Filter project deps with JavaPluginConvention in Gradle

dsilvasc:
…instead of 'java' key


jvican:
Oh, you already did that in #631, great 👍

PR #632, opened 2018-08-28
configure jenv for JDK 1.8

dsilvasc:
Looks like this project needs to be built with JDK 1.8.  This is a config for people who use jenv to manage their jdk versions.


jvican:
Does it need to be build with JDK 1.8? What are the issues you get otherwise?

PR #640, opened 2018-09-11
Add note about running the cli from the project directory.

jan0sch:


jan0sch:
Good point, I moved it up there.

Issue #646, opened 2018-09-13
Homebrew formula is using chmod 777


tues:
A workaround proposed in #629 consisted of setting log directory's permissions to `0777`, which is a security risk Bloop's users might be uncomfortable with. @falconepl has a solution which avoids setting so liberal permissions, by changing the owner of the directory instead.

tues:
Well, I'm sorry to say that it's not Bloop's formula or Homebrew at fault here: it's MacOS itself. More specifically, it's `launchctl` which creates the `log` directory with wrong ownership. In our `plist`, we have:
```
    <key>StandardOutPath</key>
    <string>#{prefix}/log/bloop/bloop.out.log</string>
    <key>StandardErrorPath</key>
    <string>#{prefix}/log/bloop/bloop.err.log</string>
```
If you `brew install bloop`, the `log` directory won't exist yet. It's created when you `brew services start bloop`, which is just a more convenient way of doing:
```
ln -sf /path/to/Cellar/bloop/1.0.0/homebrew.mxcl.bloop.plist ~/Library/LaunchAgents
```
and then:
```
launchctl load  ~/Library/LaunchAgents/homebrew.mxcl.bloop.plist
```
This is when the directory gets created with `root` as the owner. This is completely broken and should be fixed upstream, but maybe we can work around it somehow. We could try creating these files during the `install` phase and setting correct owner there.

Oh, and one more thing: the current workaround consisting of `FileUtils.chmod_R 0777, "log"` isn't working for me. So I'm afraid #629 isn't fixed yet :disappointed:

tues:
Closing this one as the `chmod 0777` didn't work anyway.

PR #647, opened 2018-09-15
Homebrew: Touch log files to ensure correct owner

### LINKED ISSUE #629: Bloop server won't be found as brew service anymore
I used to have a working installation of bloop up-to milestone `1.0.0-M10` of the plugin.

I didn't use it for about 1 month and I tried today and to my dismay found that it doesn't work as it used to.

I installed it using `brew` and kept it up-to-date and the server worked as a registered service using `brew services start bloop`.

Now, even though brew seem to confirm the server is running, if I run the client (I upgraded the sbt version to latest) I get the usual error message for no server running.

Everything works if I manually run `bloop server` from the shell, so it must be something related to the brew service.

I tried reinstalling with brew from scratch with no avail. I also found something unusual: I have no `$HOME/.bloop` folder, though I would expect one to exist from reading the docs.

brew installed the `bloop` script under `/usr/local/bin` if it might help

tues:
Fixes #629, #646. At least unless someone removes these files manually
and then restarts Bloop service.

### DIFF
--- a/project/ReleaseUtils.scala
+++ b/project/ReleaseUtils.scala
@@ -77,9 +77,14 @@ object ReleaseUtils {
        |      system "python3", "install.py", "--dest", "bin", "--version", version
        |      zsh_completion.install "bin/zsh/_bloop"
        |      bash_completion.install "bin/bash/bloop"
+       |
        |      File.delete("bin/blp-coursier")
+       |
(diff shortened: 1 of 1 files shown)


jvican:
Thank you Pawel :smile:

PR #653, opened 2018-09-21
Speed up build loading and minimize its IO

jvican:
When bloop loads up a build from the configuration files, bloop performs
all the required book-keeping to be able to detect changes on it in the
future and correctly load up the build again if a user runs a command.

We used to implement this feature by checking last modified times and the
checksums produced by zip JDK utilities. However, this approach had several
inconvenients:

1. It was slow (reloading a build like lichess took 5s in our benchmarks)
2. It was repeating IO utilities. 3. The checksum taken at the end
   represented the whole build, but we are
   interested in updating the build incrementally only for the
   configuration files that changed instead of updating all the build.

The new approach introduced by this pull request fixes these issues. First,
every configuration file has access to its last modified time and its hash.
Second, it's fast. And third, we minimize IO (get modified time from the
file visitors and read only the configuration files once when we both check
for updates and update the build accordingly).

To prove it, here's the benchmarks after the change:

```
[info] Benchmark
Mode     Cnt   Score    Error  Units
[info] ProjectBenchmark.loadAkkaProject
sample  107924   0.463 ±  0.002  ms/op
[info] ProjectBenchmark.loadAkkaProject:loadAkkaProject·p0.00
sample           0.408           ms/op
[info] ProjectBenchmark.loadAkkaProject:loadAkkaProject·p0.50
sample           0.416           ms/op
[info] ProjectBenchmark.loadAkkaProject:loadAkkaProject·p0.90
sample           0.625           ms/op
[info] ProjectBenchmark.loadAkkaProject:loadAkkaProject·p0.95
sample           0.718           ms/op
[info] ProjectBenchmark.loadAkkaProject:loadAkkaProject·p0.99
sample           1.091           ms/op
[info] ProjectBenchmark.loadAkkaProject:loadAkkaProject·p0.999
sample           1.676           ms/op
[info] ProjectBenchmark.loadAkkaProject:loadAkkaProject·p0.9999
sample           4.652           ms/op
[info] ProjectBenchmark.loadAkkaProject:loadAkkaProject·p1.00
sample          12.976           ms/op
[info] ProjectBenchmark.loadLichessProject
sample   60764   0.822 ±  0.003  ms/op
[info] ProjectBenchmark.loadLichessProject:loadLichessProject·p0.00
sample           0.750           ms/op
[info] ProjectBenchmark.loadLichessProject:loadLichessProject·p0.50
sample           0.770           ms/op
[info] ProjectBenchmark.loadLichessProject:loadLichessProject·p0.90
[...]

PR #655, opened 2018-09-22
Clean all projects by default

### LINKED ISSUE #593: Redefine the semantics of clean without arguments
If you run `bloop clean` without specifying a project, nothing happen and it return a 0 (success) return code. I would expect that either:

1)  bloop prints the same error message as for the `compile` or `test` sub command (that would be `Required project name not specified for 'clean'.`) and return the same return code
2)  it cleans all the projects. It's not uncommon for clean commands to clean basically everything.

I think that 1 would the more consistent and less surprising.  

Likely related to that issue, the help message for the `-p` argument of `bloop clean` is different that the one for the `compile` and `test` sub command (it doesn't specify that the remaining arguments are used). It's also not very clear from the help message that clean can work with multiple projects as arguments.

tues:
Also add `CleanTaskSpec`, fix documentation and remove some unused variables/parameters.

Closes #593.

### DIFF
--- a/frontend/src/main/scala/bloop/cli/Commands.scala
+++ b/frontend/src/main/scala/bloop/cli/Commands.scala
@@ -79,7 +79,7 @@ object Commands {
 
   case class Clean(
       @ExtraName("p")
-      @HelpMessage("The projects to clean.")
+      @HelpMessage("The projects to clean (you can specify multiple). By default, all.")
       project: List[String] = Nil,
       @ExtraName("propagate")
       @HelpMessage("Run clean for the project's dependencies. By default, false.")
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -293,10 +293,14 @@ object Interpreter {
   }
 
   private def clean(cmd: Commands.Clean, state: State): Task[State] = {
-    val (projects, missing) = lookupProjects(cmd.project, state)
-    if (missing.isEmpty)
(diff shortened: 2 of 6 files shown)


tues:
Yeah, I'm not sure what's wrong with that. I can clearly see `frontend/src/main/scala/bloop/Project.scala` file in the repo, there's a `final case class Project` inside, and it compiles and works fine on my machine. Also passed some of the CI builds. Really weird :thinking:

jvican:
The issue is likely that you have not rebased on top of the latest master? I merged recently https://github.com/scalacenter/bloop/pull/653 which moves `Project` around.

jvican:
Yes, the CI will only really build either the last commit of your PR or any future commit in an open PR. Previous commits are not run because they were submitted to the repo before a PR is open. We do this to save resources.

tues:
I mean `windows extra` and `windows community-build` always succeed because they don't seem to do anything - I didn't notice it at first and thought that some builds failed and some passed, so it must be random failure.

Anyway, `windows basic` failed again, but this time it doesn't look like my fault :wink:

jvican:
We're experiencing some spurious failures in the Windows CI lately. Feel free to restart the build if it happens again (I just did, looking forward to merging this).

jvican:
Looks like I've fixed the root of our spurious CI failures, so they shouldn't happen again in the future :wink: I'll be merging this as soon as I've got green light.

tues:
Thanks :slightly_smiling_face: I'm afraid I lack permissions to restart builds, though, unless you're talking about artificially amending some commit to trick CI into rebuilding automatically.

PR #661, opened 2018-09-30
Make scalac and javac independent under pipelining

jvican:
To make scalac and javac independent, we pass the transitive java
sources of a project to the scalac compiler so that it populates the
symbol table from source instead than from class files.

As a plus, thanks to this independence between scala and java
compilation in the same target, when javac does not depend on the scala
sources (java->scala), we can run scalac and javac in parallel.

The approach taken in this PR is still less than optimal because we're
passing all the transitive java sources to scalac. A better approach
that I'll try later is to pass only the transitive java sources of those
javac compilers that have not yet finished, removing any kind of
overhead.

This change has only required two changes in our community build: one in
akka and the other one in apache/spark, and those changes are
one-liners.

In the default case, where users have mixed compilation, the scala
compiler is no longer depending on the transitive javac compilers, which
means we benefit from build pipelining as much as possible.

### DIFF
--- a/backend/src/main/scala/bloop/CompileMode.scala
+++ b/backend/src/main/scala/bloop/CompileMode.scala
@@ -1,11 +1,10 @@
-package sbt.internal.inc.bloop
+package bloop
 
-import java.io.File
+import _root_.monix.eval.Task
+
 import java.net.URI
 import java.util.Optional
 import java.util.concurrent.CompletableFuture
-
-import monix.eval.Task
 
 /**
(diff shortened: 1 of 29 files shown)


jvican:
The community build is passing. test performance please

bloopoid:

bloopoid:

jvican:
test performance please

bloopoid:

bloopoid:

Issue #662, opened 2018-10-01
Support command line args when there's more than 1 test framework


Qi77Qi:
Currently, I get the following warning when I tried to run tests with `bloop test root --args “directory.url=xxx”`
 
```
[W] Ignored CLI test options ‘directory.url=xxx' can only be applied to one framework, found: ScalaCheck, ScalaTest
```
 
Things will work if I hard code variables in `.bloop/root/scala-2.12/test-classes/reference.conf`.

PR #664, opened 2018-10-01
Apply system properties to all test framworks

Qi77Qi:
Fixes https://github.com/scalacenter/bloop/issues/662

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
@@ -133,26 +133,22 @@ object Tasks {
     def considerFrameworkArgs(frameworks: List[Framework]): List[Config.TestArgument] = {
       if (frameworkSpecificRawArgs.isEmpty) Nil
       else {
+        val cls = frameworks.map(f => f.getClass.getName)
         frameworks match {
           case Nil => Nil
-          case List(oneFramework) =>
+          case oneFramework :: Nil =>
             val rawArgs = frameworkSpecificRawArgs
-            val cls = oneFramework.getClass.getName()
             logger.debug(s"Test options '$rawArgs' assigned to the only found framework $cls'.")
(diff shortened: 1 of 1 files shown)


frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
```diff
@@ -142,10 +142,18 @@ object Tasks {
...
+            val (sysProperties, ignoredArgs) = frameworkSpecificRawArgs.partition(s => s.startsWith("-D"))
+
+            if(ignoredArgs.nonEmpty)
+              logger.warn(
               s"Ignored CLI test options '${ignoredArgs}' can only be applied to one framework, found: $frameworkNames")
-            Nil
+
+            val cls = frameworks.map(f => f.getClass.getName)
+
+            if(sysProperties.isEmpty)
```
jvican: I propose we add the `logger.warn` from above in this part of the `if` expression so that the logic is clearer to follow (we can assume `frameworkSpecificRawArgs.nonEmpty == true` always, which makes `ignoredArgs.nonEmpty == sysProperties.isEmpty`. Also, let's move `cls` to the block of `else` to avoid computing it the arguments are only system properties.

Qi77Qi: if user pass in both system properties and test framework specific args, `ignoredArgs.nonEmpty` and `sysProperties.isEmpty` won't be the same tho

jvican:
LGTM, I'll merge as soon as CI is green, thank you again!

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
jvican: Ohhh, I see what you wanted to see here. I apologize for my confusing comment. I'll submit another PR with this change plus an error message if the user passes both system properties and ignored args.

PR #667, opened 2018-10-04
Allow to pass several project arguments to compile

BardurArantsson:


frontend/src/main/scala/bloop/cli/Commands.scala
```diff
@@ -28,7 +28,7 @@ object Commands {
   sealed trait RawCommand extends Command
 
   sealed trait CompilingCommand extends RawCommand {
-    def project: String
+    def project: List[String] // FIXME: name change to "projects"?
```
jvican: I think it's fine to change the name here so long as we change it in the subclasses. When you do, you will also be changing the name of the command! Something we don't want to definitely do, because that would be a breaking change in our CLI. So let's add an extra `ExtraName("project")` before the `projects` definition so that people can continue using their current commands.
  
You will need to change also the part of the logic in `Cli` where we infer the projects from remaining cli args (right now we only infer one project, but now we should infer multiple, and note that we should not infer them for commands like `run`).

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -94,15 +94,18 @@ object Interpreter {
     state.mergeStatus(ExitStatus.Ok)
   }
 
+  private def getProjectsDag(projects: List[Project], state: State): Dag[Project] =
+    Aggregate[Project](projects.map(p => state.build.getDagFor(p)))
```
jvican: Fixing the type argument to `[Project]` is not necessary.

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -157,12 +161,16 @@ object Interpreter {
...
       deduplicateFailures: Boolean
   ): Task[State] = {
-    state.build.getProjectFor(cmd.project) match {
-      case Some(project) =>
-        if (!cmd.watch) runCompile(cmd, state, project, deduplicateFailures)
-        else watch(project, state)(runCompile(cmd, _, project, deduplicateFailures))
-      case None => Task.now(reportMissing(cmd.project :: Nil, state))
-    }
+    // FIXME: HERE
+    val (projects, missingProjects) = lookupProjects(cmd.project, state)
```
jvican: I reworked this part on a commit in the branch linked. The idea here is that if there's a project that is missing, we report an error immediately and bail out from compiling. Also, I undid the change you made to the return type of `reportMissing`, it's better returning a plain state than a `Task[State]`.

frontend/src/main/scala/bloop/engine/tasks/CompilationTask.scala
```diff
@@ -159,8 +158,9 @@ object CompilationTask {
         dependentResults.collect { case (p, Compiler.Result.NotOk(_)) => p }
       if (!failedDependentProjects.isEmpty) {
         val failedProjects = failedDependentProjects.map(p => s"'${p.name}'").mkString(", ")
-        logger.warn(
-          s"Skipping compilation of project '$project'; dependent $failedProjects failed to compile.")
+        // FIXME: What to print?!?
```
jvican: What you can do here is `Dag.directDependencies(dag)` which will return you a list of projects, and you can print the list.
  
I might rework this whenever this PR is ready in an independent commit and remove this part of the code -- I need to give it more thought. But for now, doing this is just fine.

frontend/src/main/scala/bloop/cli/Commands.scala
tues: Also please consider updating documentation if you change argument name :wink:

jvican:
Hey @BardurArantsson, do you think you'll be able to address the feedback here any time soon? I would like to have this merged into master soon.

BardurArantsson:
Unfortunately, I've been off sick. I'll probably not be able to pick this up again until Friday or Saturday at the earliest :/.

Btw, feel free to just close this and "take over" since it'll probably be a lot easier for you to implement this. I mostly just care about the feature and not so much actually doing the coding for it :).

jvican:
@BardurArantsson I hope you've recovered well. I'll try to pick this up before Friday or Saturday, as it's quite an urgent feature I'd like to have ready soon. Thanks for giving this a try Bardur, and hope to see you around implementing other improvements in the codebase.

PR #671, opened 2018-10-10
Replace picklepath by IR store

jvican:
Following the changes to the public Zinc API, that I will upstream to
sbt/zinc soon: https://github.com/scalacenter/zinc/commit/4b6a951335052bc49cfe6fff441348938de08d36


bloopoid:

jvican:
It turns out the reason why guardian/frontend was failing in the community build has to do with a symbol clash, where scalac was populating the symbol table from the wrong symbol in the IRs.
  
This is due to guardian/frontend defining the same symbol (same FQN) in different projects and not scoping them in different packages. This is generally regarded as an anti-pattern.
  
However, as it can happen, the fix to this problem is to make sure that the IRs are added in the same order as the classpath, making symbol loading hence reproducible, which is a great thing. In a build like frontend/guardian, we pay a negligible cost to run the computation required for it (16ms in total).

bloopoid:

jvican:
test performance please

bloopoid:

bloopoid:

PR #680, opened 2018-10-17
Use bsp 1.0.0 and implements its new methods

jvican:
This pull request upgrades to 1.0.0 to implements its new methods, which we didn't implement because Intellij didn't implement us yet. However, as they start to be implemented in IntelliJ and other clients, we add the implementations so that the existing clients, like IntelliJ or [Metals](https://github.com/scalameta/metals/), can start benefiting from them.
  
We miss the following items to be fully implementing BSP 1.0.0:
  
- [x] `buildTarget/test`
- [x] `buildTarget/run`
- [ ] `buildTarget/scalaTestClasses` 
- [ ] `buildTarget/scalaMainClasses`
- [ ] Notifications on build target modification
- [ ] Change of namespace to be compliant with 1.0.0
  
Alongside these missing items, I'd like to implement request buffering in bloop so that the clients can indiscriminately send request without taking into account whether they finished or not. I suspect this is not going to be easy.
  
I'm not sure yet how I'm going to organize my work to implement the missing entrypoints, but it's likely I will not do everything in this PR. I'm mostly trying to prioritize for the most important things.

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -6,14 +6,15 @@ import java.util.concurrent.atomic.AtomicInteger
 
 import bloop.{Compiler, ScalaInstance}
 import bloop.cli.{Commands, ExitStatus}
+import bloop.config.Config.Platform
 import bloop.data.Project
-import bloop.engine.tasks.Tasks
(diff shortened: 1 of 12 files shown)


olafurpg:
It would be great to implement `texDocument/buildTargets` if possible, I might be tempted to use that only instead of `workspace/buildTargets`

jvican:
I think I'm merging this to keep master synced with the latest changes. I'll follow up on the rest of missing items in other PRs.

PR #685, opened 2018-10-18
Allow reflective bloop invocation with cancellation

jvican:
1. Add a `reflectMain` method that exposes a Java-only API to run
   Bloop's CLI (usually for executing a long-running process like `bloop
   bsp`).
2. Add support for cancellation. Whenever the user feels like quitting,
   it can complete the `cancel: CompletableFuture[Boolean]` with a `true`
   to force the cancellation. If it completes it with a `false`, the
   cancellation will not happen.

### DIFF
--- a/frontend/src/main/scala/bloop/Cli.scala
+++ b/frontend/src/main/scala/bloop/Cli.scala
@@ -1,6 +1,8 @@
 package bloop
 
+import java.io.{InputStream, PrintStream}
 import java.nio.file.{Files, Path}
+import java.util.concurrent.CompletableFuture
 
 import bloop.bsp.BspServer
 import bloop.cli.validation.Validate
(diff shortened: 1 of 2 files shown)


frontend/src/main/scala/bloop/Cli.scala
```diff
@@ -22,6 +24,31 @@ object Cli {
...
   }
 
+  def reflectMain(
+      args: Array[String],
+      cwd: Path,
+      in: InputStream,
+      out: PrintStream,
+      err: PrintStream,
+      props: java.util.Properties,
+      cancel: CompletableFuture[Boolean]
```
olafurpg: Can we make this `java.lang.Boolean`?

olafurpg: Boxed `scala.Boolean` has a different signature
```java
➜  target git:(bsp) ✗ scalac foo.scala && javap a$  // with List[scala.Boolean]
Compiled from "foo.scala"
public final class a$ {
  public static a$ MODULE$;
  public static {};
  public void b(scala.collection.immutable.List<java.lang.Object>);
}
➜  target git:(bsp) ✗ scalac foo.scala && javap a$ // with List[java.lang.Boolean]
Compiled from "foo.scala"
public final class a$ {
  public static a$ MODULE$;
  public static {};
  public void b(scala.collection.immutable.List<java.lang.Boolean>);
```

PR #697, opened 2018-11-02
[DOCS] Fix version flag default when calling about

jlprat:
When calling `bloop about` the default value for the flag `--version` is true (by default it prints the header with the version).

I would even go as long as saying that the whole `--version` flag should be removed from the `about` command. WDYT?


jvican:
Hey Josep, thanks for the PR. The `--version` is present for every command (you can do `bloop compile --version` and it will print the version when compiling). For this reason, the default is `false` It just happens that `bloop about` also prints the version in its output.
  
We cannot remove `--version` without breaking existing clients. It's useful in scripting environments where you want to run a command but also print the version so that you know you're using the right one.

jlprat:
No worries. I will come back with a more or less satisfactory tab completion for Fish :smiley:

Issue #698, opened 2018-11-06
Support proxy in coursier instalation


crakjie:
`http_proxy` and `https_proxy` are not passed to Java when starting the coursier installation.
[Indeed](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html) java is not pulling `http_proxy` and `https_proxy` from environment vars.
I'm preparing a PR that correct this.

crakjie:
I thinks there is another problem deeper.
When running `bloop compile aproject` I got this error:

```
[E] Server access Error: Failed to connect to repo1.maven.org/151.101.120.209:443 url=https://repo1.maven.org/maven2/ch/epfl/scala/compiler-bridge_2.11/1.1.7%2B62-0f4ad9d5/compiler-bridge_2.11-1.1.7%2B62-0f4ad9d5.pom
[E] Server access Error: Failed to connect to oss.sonatype.org/52.6.213.150:443 url=https://oss.sonatype.org/content/repositories/staging/ch/epfl/scala/compiler-bridge_2.11/1.1.7%2B62-0f4ad9d5/compiler-bridge_2.11-1.1.7%2B62-0f4ad9d5.pom
[W] 	module not found: ch.epfl.scala#compiler-bridge_2.11;1.1.7+62-0f4ad9d5
[W] ==== local: tried
[W]   /home/etienne/.ivy2/local/ch.epfl.scala/compiler-bridge_2.11/1.1.7+62-0f4ad9d5/ivys/ivy.xml
[W] ==== public: tried
[W]   https://repo1.maven.org/maven2/ch/epfl/scala/compiler-bridge_2.11/1.1.7+62-0f4ad9d5/compiler-bridge_2.11-1.1.7+62-0f4ad9d5.pom
[W] ==== sonatype-staging: tried
[W]   https://oss.sonatype.org/content/repositories/staging/ch/epfl/scala/compiler-bridge_2.11/1.1.7+62-0f4ad9d5/compiler-bridge_2.11-1.1.7+62-0f4ad9d5.pom
[W] 	::::::::::::::::::::::::::::::::::::::::::::::
[W] 	::          UNRESOLVED DEPENDENCIES         ::
[W] 	::::::::::::::::::::::::::::::::::::::::::::::
[W] 	:: ch.epfl.scala#compiler-bridge_2.11;1.1.7+62-0f4ad9d5: not found
[W] 	::::::::::::::::::::::::::::::::::::::::::::::
[W] 
```

[...]

PR #699, opened 2018-11-06
#698 - Support proxy in coursier instalation

crakjie:
I'm not a python developer, I hope it's not too much outside of pythons standards.
The openurl lib from python can already read `http_proxy` and `https_proxy` so nothing is needed there.


crakjie:
I put the proxy in both case `is_local` and `!is_local`

PR #706, opened 2018-11-08
Add support for running Scala.js tests

jvican:
There's a new big commit stabilizing the test infrastructure and
modifying key things in Bloop to get test support for Scala.js working
reliably.

Follow-up of https://github.com/scalacenter/bloop/pull/572


jvican:
Things that need to be done in the test infrastructure before a stable release (non-blocking, for now):
  
- [ ] Print good test exceptions for testing/running anything.
- [ ] Populate the jsenv option from the sbt build and error if the jsenv is not supported
- [ ] Wire this infrastructure up with BSP to support the test classes endpoint

Issue #696, opened 2018-11-02
Re-publish diagnostics for warnings in sources that are not incrementally re-compiled 


olafurpg:
Currently, it's impossible for BSP clients to know when a warning diagnostic has been fixed.

To reproduce: 

- A.scala and B.scala have one warning each
- fix warning A.scala
- recompile, 0 warnings reported in compile report even if B.scala still has active warnings

It would be good if bloop re-published the warnings for B.scala in the second compilation.

PR #712, opened 2018-11-13
Re-publish diagnostics for warnings

jvican:
Introduces the ability to diff the changed source files from one analysis to the other one and then use this information to know which warnings in previous runs should still be logged because new compilations have not compiled their origin source files.
  
See https://github.com/scalacenter/bloop/issues/696


jvican:
Closing because #722 includes this change.

PR #713, opened 2018-11-13
Set jvm proxy from client environment variables

crakjie:
The test `detectsMacrosInClasspath` wich call directly `ch.epfl.scala` currently not work under a proxy.
Except that all other `frontend/test` is working.

I had initially called the proxy setting function in the `run` function but this function is bypass by the tests which by the way need proxy too.


project/BuildPlugin.scala
```diff
@@ -264,7 +264,7 @@ object BuildImplementation {
     BuildKeys.schemaVersion := "3.2",
     Keys.testOptions in Test += sbt.Tests.Argument("-oD"),
     Keys.onLoadMessage := Header.intro,
-    Keys.onLoad := BuildDefaults.bloopOnLoad.value,
+    //Keys.onLoad := BuildDefaults.bloopOnLoad.value,
```
jvican: It looks like this should be enabled back :wink:

frontend/src/main/scala/bloop/cli/Proxy.scala
```diff
@@ -0,0 +1,90 @@
+package bloop.cli
```
jvican: Can we move it to the `bloop.util` package?

frontend/src/main/scala/bloop/cli/Proxy.scala
```diff
@@ -0,0 +1,90 @@
...
+
+import scala.util.control.NonFatal
+
+import bloop.engine.State
+
+/**
+  * Proxy helper module
+  * Information came from  https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html
+  */
+object Proxy {
```
jvican: I would call this `ProxyEnvironment`

frontend/src/main/scala/bloop/cli/Proxy.scala
```diff
@@ -0,0 +1,90 @@
...
+import java.net.URL
+import java.net.MalformedURLException
+
+import scala.util.control.NonFatal
+
+import bloop.engine.State
+
+/**
+  * Proxy helper module
+  * Information came from  https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html
```
jvican: Can we add this link to the method, since it's the one implementing the spec, and remove this scaladoc?

frontend/src/main/scala/bloop/cli/Proxy.scala
```diff
@@ -0,0 +1,90 @@
...
+  */
+object Proxy {
+  /**
+    * Extract environments variables from the given state to set the proxy properties of the jvm.
+    * If environement variable are not found in the state, the property corresponding is cleared,
+    * to let user add and remove env var.
+    * */
+  def setProxyJvmSettings(state : State) : Unit = {
+    //no_proxy environment variable seems not be binded to a protocol.
```
jvican: Remove space between this and `maybeNoProxy` and add empty space before the start of the comment?

frontend/src/main/scala/bloop/cli/Proxy.scala
```diff
@@ -0,0 +1,90 @@
...
+
+    val maybeNoProxy = Option(state.commonOptions.env.getProperty("no_proxy")).map(_.replace(',','|'))
+
+    val maybeHttpProxy = state.commonOptions.env.getProperty("http_proxy")
```
jvican: All of this code is asking for a way to abstract over setting the property globally :wink: Can we create a method `setGlobalProperty` that will abstract over the code duplication when setting `http_proxy`, `https_proxy`, `ftp_proxy` and `socks_proxy`?

frontend/src/main/scala/bloop/cli/Proxy.scala
```diff
@@ -0,0 +1,90 @@
...
+    if(maybeHttpProxy != null) {
+      try {
+        val url =  new URL(maybeHttpProxy)
+        System.setProperty("http.proxyHost",url.getHost())
+        System.setProperty("http.proxyPort",url.getPort().toString)
+        maybeNoProxy.foreach(System.setProperty("http.nonProxyHosts", _))
+      } catch {
+        case e : MalformedURLException => //Ignore, but should maybe log a warning if the URL is not parsable.
```
jvican: Let's debug a message if this exception happens, such as "Unexpected non-URI format in proxy value of `$key`"?

frontend/src/main/scala/bloop/cli/Proxy.scala
```diff
@@ -0,0 +1,90 @@
...
+      }
+    } else {
+      System.clearProperty("http.proxyHost")
+      System.clearProperty("http.proxyPort")
+      if(maybeNoProxy.isEmpty)
+        System.clearProperty("http.nonProxyHosts")
+    }
+
+    val maybeHttpsProxy = state.commonOptions.env.getProperty("https_proxy")
+    if(maybeHttpsProxy != null) {
```
jvican: Can we reformat all of this file with `scalafmt`? The style scalafmt file is in the repository so it will pick up the changes.

frontend/src/main/scala/bloop/cli/Proxy.scala
```diff
@@ -0,0 +1,90 @@
...
+    if(maybeSocksProxy != null) {
+      try {
+        val url =  new URL(maybeSocksProxy)
+        System.setProperty("socksProxyHost",url.getHost())
+        System.setProperty("socksProxyPort",url.getPort().toString)
+      } catch {
+        case e : MalformedURLException => //Ignore, but should maybe log a warning if the URL is not parsable.
+      }
+    } else {
+      System.clearProperty("socksProxyHost")
```
jvican: I think we should only clear the properties *if* a client environment variable was set before. If, for example, the server was fired up with these properties, and the client doesn't have them, I don't think it's a good behavior to override the proxy properties a server was launched with.
  
This approach should be complementary to the server environment variables, rather than exclusionary. We should also not set the client environment variables if the server proxy settings are also set. We should only set globally client environment variables if the server has not set values for those properties before.
  
In order to achieve this, we need two steps: discovery of these environment variables when we boot up the server and setting the client environment variables based on that information. I would add an `AtomicAny[Set[String]]` field in `Cli` that is initialized the first time we start running `nailMain` with the contents of the server environment variables. Then, I would pass in this set to `setProxyJvmSettings`.

frontend/src/main/scala/bloop/cli/Proxy.scala
```diff
@@ -0,0 +1,90 @@
...
+
+import bloop.engine.State
+
+/**
+  * Proxy helper module
+  * Information came from  https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html
+  */
+object Proxy {
+  /**
+    * Extract environments variables from the given state to set the proxy properties of the jvm.
```
jvican: Can we explain the new policy I asked for in the comments and also mention that we're mutating system properties globally here and that we don't support the use case where two clients in the same machine have different settings?

frontend/src/main/scala/bloop/cli/Proxy.scala
crakjie: Yes of course

frontend/src/main/scala/bloop/cli/Proxy.scala
crakjie: That not down automatically when compiling?

project/BuildPlugin.scala
crakjie: Yes I remember that yester when I was in the subway. I had already amended that this morning. It's strange you still see that in the diff.

frontend/src/main/scala/bloop/cli/Proxy.scala
jvican: No, I think it's better to keep both separate, check https://scalameta.org/scalafmt/ if you need more info to get set up

frontend/src/main/scala/bloop/cli/Proxy.scala
crakjie: I agree for the first part. Fall back on server environement when it's not set by the client.

But I'm not sure about the second. If the user decides to change is proxy environement variable, he would have to restart the server? So settings the proxy per query would be just useful when the server was started without proxy environment variable? In this case the use won't have to restart the server. It seems to me a little bit to "conditional".

jvican: > But I'm not sure about the second. If the user decides to change is proxy environement variable, he would have to restart the server?
  
I didn't mean that :smile: I think that we should still be able to set client environment variables as many times as they change in the user environment, it's just that we should fast-fail if the server already had them in the first place.

crakjie: A val in Cli object wouldn't be enought to store boot environement variable since it's a singleton object ?

crakjie: And we should load theses environement variable in proxy settings on bootstrap. Because java don't load them automaticaly. ( execept if they are given via -Dhttp.proxyHost ... )

jvican: > A val in Cli object wouldn't be enought to store boot environement variable since it's a singleton object ?
  
The fact that it's a singleton object is what helps us now that it will contain the environment variables/system properties when the server is initialized. If they are not there when `Cli` was initialized, then we can safely know that the server was not configured to use a proxy.
  
> And we should load theses environement variable in JVM proxy settings on bootstrap. Because Java doesn't load them automatically. ( except if they are given via -Dhttp.proxyHost ... )
  
Sounds good :+1: I was only thinking of system properties but environment variables should be explicitly set during the first step for the rest of the logic to work.

jvican:
@crakjie I'm waiting for your changes here :smile: Whenever you have them ready, push them (make sure you rebase on latest master too) and I merge them.

jvican:
@crakjie Thanks for your work Etienne. I had to change quite a big part of your PR after I had a closer look at the best way of deal with proxy settings. I've added scaladocs and improved the FAQ entry to reflect what the current configuration requirements are.
  
Please have a look and let me know if this works for you (this approach is more conservative but i think covers most of the use cases reasonably well while reducing the danger that things go wrong in non-reproducible ways).

PR #717, opened 2018-11-14
Interpreter: Allow to run externally-defined main classes

tindzk:
Currently, the following error is triggered when specifying a valid
class that is not defined in the project:

```shell
$ bloop run root -m play.core.server.ProdServerStart
[E] Missing main class  in project 'root'
```

Simplify the logic of `findMainClasses()`. The validation logic is
redundant since specifying an invalid main class will be caught at a
later point:

```shell
 $ bloop run root -m test
[E] Error: Could not find or load main class test
```


jvican:
I merged this too quick, @tindzk would you mind following up this bug fix with a test? The logic is straightforward but every bug fix should have a test to ensure that we don't regress.

PR #720, opened 2018-11-15
Terminate processes gracefully

### LINKED ISSUE #719: Processes are not killed gracefully
When I run Play, kill the process with Ctrl-C and start it again, the output looks as follows:

```shell
~/dev/bloop-play/play-scala-seed $ bloop run root -m play.core.server.ProdServerStart
[info] play.api.Play - Application started (Prod)
[info] p.c.s.AkkaHttpServer - Listening for HTTP on /0:0:0:0:0:0:0:0:9000
^C%
~/dev/bloop-play/play-scala-seed $ bloop run root -m play.core.server.ProdServerStart
[E] This application is already running (Or delete /home/tim/dev/bloop-play/play-scala-seed/RUNNING_PID file).
```

However, when I kill the process manually (`kill $PID`), the the `RUNNING_PID` file disappeared and the server shut down gracefully:

```
[info] p.c.s.AkkaHttpServer - Stopping server...
````

As pointed out by @jvican, this could be due to forcibly closing the process via NuProcess.

Originally reported here: https://github.com/scalacenter/bloop/issues/227#event-1967188606

### PR DESCRIPTION
**Pull request by tindzk**
To-do:
- [ ] `stdout` is disconnected immediately when pressing Ctrl-C. The user does not see any log messages triggered by termination.
- [ ] Write test case

Closes #719.

### DIFF
--- a/frontend/src/main/scala/bloop/exec/Forker.scala
+++ b/frontend/src/main/scala/bloop/exec/Forker.scala
@@ -195,8 +195,8 @@ object Forker {
           gobbleInput.cancel()
           try process.closeStdin(true)
           finally {
-            process.destroy(true)
-            process.waitFor(200, _root_.java.util.concurrent.TimeUnit.MILLISECONDS)
+            process.destroy(false)
+            process.waitFor(400, _root_.java.util.concurrent.TimeUnit.MILLISECONDS)
             if (process.isRunning) {
               val msg = s"The cancellation could not destroy process ${process.getPID}"
               opts.ngout.println(msg)
--- a/frontend/src/main/scala/bloop/exec/Forker.scala
+++ b/frontend/src/main/scala/bloop/exec/Forker.scala
@@ -197,6 +197,7 @@ object Forker {
           finally {
             process.destroy(false)
             process.waitFor(400, _root_.java.util.concurrent.TimeUnit.MILLISECONDS)
+            process.destroy(true)
             if (process.isRunning) {
               val msg = s"The cancellation could not destroy process ${process.getPID}"
               opts.ngout.println(msg)
--- a/bridges/scalajs-1.0/src/main/scala/bloop/scalajs/jsenv/NodeJsEnv.scala
+++ b/bridges/scalajs-1.0/src/main/scala/bloop/scalajs/jsenv/NodeJsEnv.scala
@@ -177,7 +177,8 @@ object NodeJSEnv {
         // Make sure we only destroy the process once, the test adapter can call this several times!
         if (!isClosed.getAndSet(true)) {
           logger.debug(s"Destroying process...")
-          process.destroy(true)
+          process.destroy(false)
+          process.waitFor(400, _root_.java.util.concurrent.TimeUnit.MILLISECONDS)
           // Make sure that no matter what happens the `onExit` callback is invoked
           handler.cancel()
         }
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: frontend/src/main/scala/bloop/exec/Forker.scala [hunk 1]
```diff
@@ -195,7 +195,7 @@ object Forker {
           gobbleInput.cancel()
           try process.closeStdin(true)
           finally {
-            process.destroy(true)
+            process.destroy(false)
             process.waitFor(200, _root_.java.util.concurrent.TimeUnit.MILLISECONDS)
```
tindzk: I also propose we increase the timeout to 1s as 200 ms is rather short.

jvican: 1s is too much, 400ms is a good time. Also note that the JVM shutdown hooks give no guarantee to the programmer about how much time they are executed or if they are executed at all, so I believe 400ms is a good compromise.

jvican:
```suggestion
            process.waitFor(200, _root_.java.util.concurrent.TimeUnit.MILLISECONDS)
            process.destroy(true)
```

jvican:
ping @tindzk :smile:

frontend/src/main/scala/bloop/exec/Forker.scala
tindzk: Okay, I will change it to 400ms.

tindzk:
> This should not change. If you want to log something, use commonOptions.ngout.println instead.

A process can still write to stdout after receiving the `SIGINT` signal. Currently, no output is shown after pressing Ctrl-C, but stdout is propagated when I terminate the process using `kill`. This appears to be a shortcoming of the Nailgun script.

> I believe it will be difficult to write a test case for this. How are you planning to do it?

I presume that Play is using [addShutdownHook](https://docs.oracle.com/javase/6/docs/api/java/lang/Runtime.html#addShutdownHook%28java.lang.Thread%29). We could create a minimal project that waits 2s after the `SIGINT` signal. If the process terminated in < 2s, the test should fail. Similarly, any stdout sent by the process after `SIGINT` must be propagated.

jvican:
> A process can still write to stdout after receiving the SIGINT signal. Currently, no output is shown after pressing Ctrl-C, but stdout is propagated when I terminate the process using kill. This appears to be a shortcoming of the Nailgun script.
  
Yes, to summarize, at the moment the nailgun client receives a CTRL-C from the client it kills the client and the communication with the server it's broken so if a process sends things to stdout that output will be discarded/thrown away. I believe this is the correct behavior.
  
> I presume that Play is using addShutdownHook. We could create a minimal project that waits 2s after the SIGINT signal. If the process terminated in < 2s, the test should fail. Similarly, any stdout sent by the process after SIGINT must be propagated.
  
I usually don't advise this, but I propose we don't write a test case for this issue. It's complicated to reproduce and will be a drag in our CI more than anything else.
  
The only thing I think it's missing in this PR is to rebase with master and also propagate the changes to `process.destroy` to other use sites in the codebase (for example, we use it in the scalajs 1.0 too).

tindzk:
> Yes, to summarize, at the moment the nailgun client receives a CTRL-C from the client it kills the client and the communication with the server it's broken so if a process sends things to stdout that output will be discarded/thrown away. I believe this is the correct behavior.

I just checked how sbt behaves in this situation. When I press Ctrl-C in sbt, it waits until the child process has been killed, so any output sent to stdout is still visible and only after the child's termination, sbt is stopped. I am not sure whether it is a deliberate design decision, but this behaviour makes sense to me.

> I usually don't advise this, but I propose we don't write a test case for this issue. It's complicated to reproduce and will be a drag in our CI more than anything else.

Agreed.

> The only thing I think it's missing in this PR is to rebase with master and also propagate the changes to process.destroy to other use sites in the codebase (for example, we use it in the scalajs 1.0 too).

Okay, I will do it.

Issue #695, opened 2018-10-29
Add table completions for Fish shell


jlprat:
The same way there is support for autocompletion (tab completion) for Bash and Zsh, it would be cool to have the same functionality for users who use the Fish shell.

jvican:
Fixed by https://github.com/jlprat/bloop/commit/4d528b50ba6756f9c08cc88b8854c133ba76cd57

PR #721, opened 2018-11-15
Add fish tab completion capabilities

jlprat:
Refs: #695
Adds documentation on how to install
Adds documentation for developers when changing commands
Provides a file to generate tab completions for `fish`

Disclaimer on the approach: `fish` tab completions take a more descriptive approach than the one `Bash` or `zsh` have. The common way to write such tab completion files for `fish` is usually by listing all possible sub-commands and flags in such files.

I'm not sure if some other areas would need to be touched, like the installation script for example


jlprat:
This test (`Test bloop.tasks.JsTestSpec`) failed on CI, but I don't know how this is related to my changes.

jlprat:
You're welcome @jvican ! What about the installation scripts? Will the new file under `etc/fish/bloop.fish` be "magically" copied to `.bloop/fish/bloop.fish`?

jvican:
> You're welcome @jvican ! What about the installation scripts? Will the new file under etc/fish/bloop.fish be "magically" copied to .bloop/fish/bloop.fish?
  
They won't be magically copied, you need to modify two files:
  
1. The python script in `bin/install.py` for multi-OS installation
2. The contents of the `ReleaseUtils` brew installation script which we use to install bloop in OSX (https://github.com/scalacenter/bloop/blob/master/project/ReleaseUtils.scala#L98)
  
One question that comes to mind: can we avoid generating this file and instead let the server reply to the fish's autocompletion engine? The following SO question has two relevant answers that hint that this is possible. One reason why I prefer this approach is because it's likely that this file will go outdated.

jlprat:
Without being an expert in fish, letting the server respond will only work if the bloop server is running when the shell starts.
If we don't consider this, then it is possible to have everything scripted and never be outdated.

The main probles comes with the flag completion, where we need to have 1 `complete` per flag. I the server is down, the loop won't generate any entry.

I also didn't line this solution, but as it's being used already to keep the docs up, I guessed it won't be outdated that easily.


PS: you forgot the SO link 😉

jvican:
> you forgot the SO link 

https://stackoverflow.com/questions/20838284/how-can-i-provide-tab-completions-to-fish-shell-from-my-own-script
  
> Without being an expert in fish, letting the server respond will only work if the bloop server is running when the shell starts.
   
I would favor talking to the server to get the autocompletions directly since that's the way bash and zsh autocompletions work in bloop now. It's true that autocompletions won't work if the server is not running, but that's OK since you can do nothing when the server is down (the best way to ensure that this doesn't happen is that you have brew services/systemd enabled in your system).

jlprat:
If this is an acceptable case, then I can rewrite it to use the autocompletions directly from server.
The only problem I can image is, if somebody starts `bloop` manually, they would need to restart `fish` to get the completions working. But I guess, is an acceptable compromise.

jvican:
> if somebody starts bloop manually, they would need to restart fish to get the completions working.
  
Is this a fish-specific limitation? I've never had such issues with zsh/bash.
  
> PR update will come during next days.
  
Thanks Josep!

jlprat:
> Is this a fish-specific limitation? I've never had such issues with zsh/bash.

If I understand `zsh` completion correctly, one defines a syntax for the command with the functions that will provide the completions. This is way more flexible and fits better to `bloop`'s way of working. On the other hand, as far as my understanding of fish goes, yes, it is a fish limitation. Fish tab completions are designed to be generated from man pages, additionally and all the examples I saw around are upfront definitions. Nothing like the `zsh` style.

There are conditionals on the `complete` command and some interesting things can be done this way, like:
`complete -c bloop -f -n "__assert_args_count 0" -a '(_commands)'`
But you would need a `complete` command for each flag and subcommand, and this is where the things get trickier.

The main problem is that in `fish`, one can't define multiple flags (or options in `fish` naming convention) in one single `complete` command. This, together with the fact that `bloop autocomplete` only works when the server is started (obviously) make it more challenging. I have a way to overcome this I think.

bin/install.py
```diff
@@ -62,7 +62,7 @@
     help="Tell the location of the ivy home.")
 parser.add_argument(
     '--bloop-home',
-    help="Tell the location of the ivy home.")
```
jlprat: Fixed a typo here

etc/fish/bloop.fish
```diff
@@ -0,0 +1,102 @@
...
+function _bloop_cmpl_ -d 'Make a completion for a subcommand' --no-scope-shadowing -a minTokens cmd
+    set -e argv[1]
+    set -e argv[1]
+    echo ---$argv
+    complete -c bloop -f -n "__assert_args_at_least_count $minTokens; and __fish_seen_subcommand_from $cmd" $argv
+end
+
+
+# Nothing has been provided yet. Complete with commands
+complete -c bloop -f -n "__assert_args_count 0" -a '(_commands)'
```
jlprat: This will recover if `bloop server` is not running and then started.

etc/fish/bloop.fish
```diff
@@ -0,0 +1,102 @@
...
+
+# Only a project command has been provided
+complete -c bloop -f -n "__assert_args_count 1; and __fish_seen_subcommand_from_project_commands" -a '(_projects)'
```
jlprat: This will recover if `bloop server` is not running and then started.

etc/fish/bloop.fish
```diff
@@ -0,0 +1,102 @@
...
+
+
+for cmd in (_commands)
+    set -l minTokens (_is_project_command $cmd)
+    set -l flags (_flags_for_command $cmd)
+    for flag in $flags
+        set -l parsed (string split '#' $flag)
+        if test -n $parsed[1] -a -n $parsed[2] -a -n $parsed[3]
```
jlprat: These will not recover if `bloop server` is not running and then started.

website/content/docs/installation.md
```diff
@@ -87,6 +87,15 @@ To get the command line completions with Bash, add the following to your `~/.bas
 . $HOME/.bloop/bash/bloop
 ```
 
+#### Fish completions
+
+To get the command line completions with fish shell, navigate to your local completion location (e.g.: `~/.config/fish/completions`)
+and create a symlink to the file provided in the Bloop distribution:
```
jvican: Shall we mention here what to do if the bloop server is not running and how to get the fish completions to work after the user manually starts the server?

jlprat: Agree, makes sense to do so.

jvican: Great, whenever you make the change, please rebase the PR so that i can merge it right away.

jlprat: Change done, and rebased against master.

jvican:
🙇

PR #722, opened 2018-11-16
Migrate to bsp v2 partially

jvican:
We disable tests for `test` and `run` for now, only focusing on the most
important BSP tasks to ease the migration in the BSP clients,
concretely diagnostics, `sources` and compile. There is a lot of
information in the commit messages.

PR #725, opened 2018-11-19
Update schema docs for `sources`

jvican:
See review in https://github.com/scalacenter/bloop/pull/722

### DIFF
--- a/website/static/bloop-schema.json
+++ b/website/static/bloop-schema.json
@@ -306,12 +306,13 @@
           "items": {
             "$id": "/properties/project/properties/sources/items",
             "type": "string",
-            "title": "An example of a source directory",
+            "title": "An example of a source file or directory",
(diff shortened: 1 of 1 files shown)


website/static/bloop-schema.json
```diff
@@ -306,12 +306,13 @@
           "items": {
             "$id": "/properties/project/properties/sources/items",
             "type": "string",
-            "title": "An example of a source directory",
+            "title": "An example of a source file or directory",
```
olafurpg: Should directory values end with slash?

jvican: In the JSON files it's ok if they don't end with slash, distinguishing them from single source files is simple as collecting all the files that don't end with `*.scala` and `*.java`.

olafurpg:
LGTM 👍

PR #729, opened 2018-11-21
Re-publish all warnings on BSP server startup

jvican:
... and use the correct file URI schema for the files in the diagnostics notifications. More info in the commit messages.

### DIFF
--- a/frontend/src/main/scala/bloop/logging/BspServerLogger.scala
+++ b/frontend/src/main/scala/bloop/logging/BspServerLogger.scala
@@ -83,7 +83,7 @@ final class BspServerLogger private (
           case Severity.Info => bsp.DiagnosticSeverity.Information
         }
 
-        val uri = bsp.Uri(file.toURI)
+        val uri = bsp.Uri(file.toPath.toUri)
         val diagnostic = bsp.Diagnostic(pos, Some(severity), None, None, message, None)
         val textDocument = bsp.TextDocumentIdentifier(uri)
         val buildTargetId = bsp.BuildTargetIdentifier(project.bspUri)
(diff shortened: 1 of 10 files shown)


frontend/src/main/scala/bloop/logging/BspServerLogger.scala
```diff
@@ -101,7 +101,7 @@ final class BspServerLogger private (
   }
 
   def noDiagnostic(project: Project, file: File): Unit = {
-    val uri = bsp.Uri(file.toURI)
+    val uri = bsp.Uri(file.toPath.toUri)
```
olafurpg: 👍

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
```diff
@@ -395,6 +395,74 @@ class BspProtocolSpec {
...
+            Assert.assertTrue(
+              s"URI ${tid.uri} does not start with valid 'file:///'",
+              tid.uri.value.startsWith("file:///")
+            )
+            val relativePath = AbsolutePath(tid.uri.toPath).toRelative(baseDir)
+            val canonical = relativePath.toString.replace(File.separatorChar, '/')
+            val report = diagnostics
+              .map(_.toString.replace("\n", " ").replace(System.lineSeparator, " "))
+            builder
+              .++=(s"#${cycleIndex + 1}: $canonical\n")
```
olafurpg: nit I personally prefer `.append` over `++=` when using non-infix syntax (also makes it easier to refactor to `java.lang.StringBuilder`

jvican: I'll keep it in mind whenever I touch this code again :+1:

jvican:
> I am a bit concerned by the growing test case in BspProtocolSpec
  
Agreed, I would like to take some time down the road to deduplicate some of the logic here.

Issue #708, opened 2018-11-12
Handle build load errors gracefully in the server


jvican:
When `bloopInstall` fails to complete (for example, because of an error in the definition of a module in an sbt build), `bloop` usually fails to load the build. The load errors usually happen because there's a project reference in `dependencies` whose configuration file was not written to the config dir. In the past, `bloop` has been emitting `NoSuchElementException`s for such missing dependencies.
  
@olafurpg proposes that instead of failing early at such errors, we gracefully handle them and still allow the user to use bloop for all the modules that were successfully exported. Instead of an error, we can `warn` the user that a given project "touched" by a bloop invocation has a missing dependency.
  
On top of that, we should also add special logic such that if a command like `compile` fails with an error and one of its dependencies was missing a dependency, we log a postmortem telling users to double-check their build definitions and dependency misses.
  
All these changes must be done in the bloop server, not the build plugins.

PR #731, opened 2018-11-22
Handle build load errors gracefully in the server

jvican:
Imagine that `bloopInstall` fails to successfully complete. The process
did complete for a subset of the build graph, but not all of it. And now
imagine that the errors are not relevant and the user is willing to
ignore them temporarily. Isn't it frustrating that bloop will fail to do
anything with a `NoSuchElementException` because the build load fails?

This commit tries to improve on this situation by not making it an
all-or-nothing situation. Bloop can now handle missing dependencies in a
configuration file gracefully. To let the user know that there's
something off, though, we emit a warning every time we compile a module
that has a missing dependency (in case, for example, there's a failure
in compilation). However, we do not emit such warning if the user
compiles other modules in the build and therefore give more freedom to
the user.

Fixes https://github.com/scalacenter/bloop/issues/708

### DIFF
--- a/frontend/src/main/scala/bloop/engine/Dag.scala
+++ b/frontend/src/main/scala/bloop/engine/Dag.scala
@@ -13,7 +13,13 @@ object Dag {
   class RecursiveCycle(path: List[Project])
       extends Exception(s"Not a DAG, cycle detected in ${path.map(_.name).mkString(" -> ")} ")
 
-  def fromMap(projectsMap: Map[String, Project]): List[Dag[Project]] = {
+  case class DagResult(
+      dags: List[Dag[Project]],
+      missingDependencies: Map[Project, List[String]]
+  )
+
+  def fromMap(projectsMap: Map[String, Project]): DagResult = {
+    val missingDeps = new scala.collection.mutable.HashMap[Project, List[String]]()
     val visited = new scala.collection.mutable.HashMap[Project, Dag[Project]]()
     val visiting = new scala.collection.mutable.LinkedHashSet[Project]()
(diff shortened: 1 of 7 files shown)


frontend/src/main/scala/bloop/engine/Dag.scala
```diff
@@ -35,7 +41,19 @@ object Dag {
             project match {
               case leaf if project.dependencies.isEmpty => Leaf(leaf)
               case parent =>
-                val childrenNodes = project.dependencies.iterator.map(name => projectsMap(name))
```
olafurpg: Is this where the previous `NoSuchElementException` happened?

jvican: Yes, in the `apply` of `Map`

frontend/src/main/scala/bloop/engine/Feedback.scala
```diff
@@ -15,6 +15,16 @@ object Feedback {
     s"Failed to compile project '${project.name}': found Scala sources but project is missing Scala configuration."
   def missingInstanceForJavaCompilation(project: Project): String =
     s"Failed to compile Java sources in ${project.name}: default Zinc Scala instance couldn't be created!"
+  def detectMissingDependencies(project: Project, missing: List[String]): Option[String] = {
+    missing match {
+      case Nil => None
+      case missing :: Nil =>
+        Some(s"Detected missing dependency '$missing' in project '${project.name}' may cause compilation issues")
```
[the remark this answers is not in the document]
jvican: Great suggestion :smile:

PR #739, opened 2018-11-24
Fix test resource behavior for source-based projects

### LINKED ISSUE #738: Source dependencies' resources are not available
For source dependencies it's normal to have `application.conf` be assembled from (automatically merged) `application.conf`s of all subprojects. Right now this is not the case which results in test failing with 

```
  Cause: com.typesafe.config.ConfigException$Missing: No configuration setting found for key 'foo'
```

if that key is specified in dependency's config.

Note that the right entry is present in classpath in `toplevel-test.json`, it's just that there are no resources copied for subprojects.

tkroman:
Fixes #738.

**Issue** 
having the following project structure:

```
-toplevel
  |-src/test/resources/application.conf
-modules/foo
  |-src/test/resources/application.conf
```

where `toplevel` depends on `foo`, default (sbt/intellij/...) behavior is to have both `toplevel` and `foo` resources on classpath. Bloop only uses `toplevel` which leads to eg test errors when `modules` projects declare some shared/default configuration keys in configs, which are normally expected to be merged with `toplevel` configs.

**Fix**
include resources for all aggregated projects.

### DIFF
--- a/config/src/main/scala/bloop/config/util/ConfigUtil.scala
+++ b/config/src/main/scala/bloop/config/util/ConfigUtil.scala
@@ -1,26 +1,14 @@
 package bloop.config.util
 
-import java.nio.file.{Path, Files}
+import java.nio.file.Path
 
 object ConfigUtil {
   def pathsOutsideRoots(roots: Seq[Path], paths: Seq[Path]): Seq[Path] = {
-    paths.filterNot { path =>
-      roots.exists { root =>
-        var found: Boolean = false
-        val rootDirSize = root.toString.size
-        var currentTarget = (if (Files.isRegularFile(path)) path.getParent else path).toAbsolutePath
-        while (!found &&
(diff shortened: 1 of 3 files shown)


config/src/main/scala/bloop/config/util/ConfigUtil.scala
```diff
@@ -1,26 +1,14 @@
...
-               currentTarget.toString.size >= rootDirSize) {
-          if (currentTarget == root) {
-            found = true
-          }
-
-          currentTarget = currentTarget.getParent
-        }
-        found
-      }
+    for {
```
tkroman: this change doesn't necessarily have to be here, but this implementation is way more straightforward :)

tkroman:
not sure why linux test is failing, does drone ci use some caching? is there any way to run a completely clean test?

PR #740, opened 2018-11-25
Propagate main class in sbt plugin

### LINKED ISSUE #718: Propagate main class defined from the sbt build
As per #716, the detection of main classes is restricted to classes defined within the project. In the case of Play, the sbt project already sets the main class correctly:

```shell
[play-scala-seed] $ show mainClass
[info] Some(play.core.server.ProdServerStart)
```

It would be useful to use this value and set it in the Bloop configuration file such that it does not have to be passed as a command-line parameter.

tindzk:
Closes #718.


jvican:
@tindzk It looks like this change is making the CI hang on Windows, so I had to kill the process manually.

PR #741, opened 2018-11-25
Set cwd to project's base directory

### LINKED ISSUE #724: Set current working directory to the project base directory
When a user runs an application in bloop, the current working directory of the spawned system process is set to the workspace directory. However, this goes against the tradition of using instead the project base directory of the executed application in, for example, sbt projects that have enabled forked tests or program execution (which is the default in bloop).
  
Bloop should have the same behavior.

tindzk:
Closes #724.


tindzk:
The CI failure on Windows seems unrelated: https://ci.scala-lang.org/scalacenter/bloop/4291/8

Issue #743, opened 2018-11-28
Unexpected `NoClassDefFoundError` when compiling the bridge


ouchxp:
I'm getting NoClassDefFoundError: javax/tools/DiagnosticListener when running 2.11 project. However if I switch to 2.12, everything works just fine.
```
 ~/I/scalatest> bloop about
    Bloop-frontend version    `1.0.0`
    Zinc version     `1.1.7+62-0f4ad9d5`
    Scala version    `2.12.6`
```

```
~> scala
Welcome to Scala 2.11.12 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_192).
```

```
 ~/I/scalatest> bloop run garyscalatest
Compiling 1 Scala source to /Users/drwho/garyscalatest/.bloop/scalatest/scala-2.11/classes ...
Non-compiled module 'compiler-bridge_2.11' for Scala 2.11.12. Compiling...
error: java.lang.NoClassDefFoundError: javax/tools/DiagnosticListener
        at scala.tools.util.PathResolverBase$Calculated$.javaBootClasspath(PathResolver.scala:277)
        at scala.tools.util.PathResolverBase$Calculated$.basis(PathResolver.scala:283)
        at scala.tools.util.PathResolverBase$Calculated$.containers$lzycompute(PathResolver.scala:293)
        at scala.tools.util.PathResolverBase$Calculated$.containers(PathResolver.scala:293)
        at scala.tools.util.PathResolverBase.containers(PathResolver.scala:309)
        at scala.tools.util.PathResolver.computeResult(PathResolver.scala:341)
        at scala.tools.util.PathResolver.computeResult(PathResolver.scala:332)
        at scala.tools.util.PathResolverBase.result(PathResolver.scala:314)
        at scala.tools.nsc.backend.JavaPlatform$class.classPath(JavaPlatform.scala:28)
        at scala.tools.nsc.Global$GlobalPlatform.classPath(Global.scala:115)
        at scala.tools.nsc.Global.scala$tools$nsc$Global$$recursiveClassPath(Global.scala:131)
        at scala.tools.nsc.Global$GlobalMirror.rootLoader(Global.scala:64)
        at scala.reflect.internal.Mirrors$Roots$RootClass.<init>(Mirrors.scala:307)
        at scala.reflect.internal.Mirrors$Roots.RootClass$lzycompute(Mirrors.scala:321)
[...]

PR #747, opened 2018-12-03
Don't fail bloop if Java 11 installed in Mac OS

jvican:
This fix uses the godsend `/usr/libexec/java_home` to detect the java
home for 1.8 and use that one to run `coursier launch` (failing with a
helpful error message if it doesn't find it). The main idea here comes
from @olafurpg and some of the discussion of the bug

In reality, this means that the server must be run with Java 1.8 or it
will fail. This is a good compromise for now until we can ensure that
we are as efficient and correct in other Java versions as we're in Java
1.8. I've tested this in my machine where Java 8 is installed via jenv
(which doesn't set the java home variable directly) and Java 11 was
installed via homebrew casks. I believe this scenario stresses this fix
enough to make it valid.

For now, we only use this approach in Mac OS because libexec/java_home
only exist in OSX.

Fixes https://github.com/scalacenter/bloop/issues/743

### DIFF
--- a/bin/install.py
+++ b/bin/install.py
@@ -2,10 +2,12 @@
 
 from __future__ import print_function
 
+import sys
 import argparse
 import os
 from os.path import expanduser, isdir, isfile, join, dirname, abspath
 from subprocess import CalledProcessError, check_call
+import platform
 import sys
 import re
 from shutil import copyfileobj, copyfile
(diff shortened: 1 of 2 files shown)


bin/install.py
```diff
@@ -112,6 +115,20 @@
...
 BUFFER_SIZE = 4096
 
+MACOS_LAUNCH_SCRIPT_CONTENTS = """
+#!/usr/bin/env sh
+
+BASE_BIN_DIR=$(dirname "$0")
+COURSIER_BIN="$BASE_BIN_DIR/%s"
+/usr/libexec/java_home -v 1.8 -F -R --exec java \
+     -Divy.home=%s -jar "$COURSIER_BIN" launch %s \
+     -r bintray:scalameta/maven \
```
olafurpg: Why is this line needed?

jvican: Because of lsp4s

olafurpg: Isn't it released to maven central?

PR #756, opened 2018-12-11
Show release table and fix versions in docs [DOCS]

jvican:


docs/src/main/scala/bloop/Docs.scala
```diff
@@ -1,12 +1,27 @@
...
+    val cwd = if (!cwd0.resolve("docs").isDirectory) cwd0.toNIO.getParent else cwd0.toNIO
+
+    val settings = MainSettings()
+      .withSiteVariables(Map(
+        "VERSION" -> Sonatype.release.version,
+        "LATEST_VERSION" -> bloop.internal.build.BuildInfo.version
+      ))
       .withArgs(args.toList)
+      // it should work with mdoc when run inside bloop but it doesn't, let's wait until it's fixed
+      .withWorkingDirectory(cwd)
```
olafurpg: It will work if you update `withOut` instead of `withWorkingDirectory`.

I opened https://github.com/olafurpg/mdoc/pull/111 updating `withWorkingDirectory` to also update `docs` and `out`.

jvican:
Windows CI is flaky again, let's merge.

PR #758, opened 2018-12-12
Handle recursive dependencies gracefully

jvican:
See https://github.com/scalameta/metals/issues/396 for the issue that
triggered work on this area. Even if recursive dependencies is a fatal
error for bloop, we still handle gracefully and connect successfully
during a bsp connection. Builds that have problems will get:

1. An empty response to `workspace/buildTargets`.
2. A `build/showMessage` per every error found.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/Dag.scala
+++ b/frontend/src/main/scala/bloop/engine/Dag.scala
@@ -3,6 +3,8 @@ import bloop.data.Project
 import bloop.util.CacheHashCode
 import scalaz.Show
 
+import scala.collection.mutable.ListBuffer
+
 sealed trait Dag[T]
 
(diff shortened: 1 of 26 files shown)


frontend/src/main/scala/bloop/engine/Dag.scala
```diff
@@ -24,51 +25,59 @@ object Dag {
...
+        if (visiting.contains(project)) {
+          // We register the recursive cycle and return none
+          traces.+=(RecursiveTrace(visiting.toList :+ project))
+          None
+        } else {
+          visiting.+=(project)
+
+          val dag = visited.get(project).getOrElse {
+            markVisited {
+              project match {
```
olafurpg: nit: the pattern match could be replaced with `if (project.dependencies.isEmpty) .. else ..`

Issue #575, opened 2018-07-03
Compile, link and test on several projects at once


tindzk:
It should be possible to specify multiple projects when compiling, linking or testing:

```shell
bloop compile project1 project2
bloop link project1 project2
bloop test project1 project2
```

We need to make sure that this functionality also extends to the `--watch` parameter:

```shell
$ bloop compile --watch project1 project2
$ bloop link --watch project1 project2
$ bloop test --watch project1 project2
```

**Discussion:** https://gitter.im/scalacenter/bloop?at=5b2ceda1479ca2668975803b

jvican:
The logic for this is already implemented since we support that in BSP. The only change required for this is in the CLI handling logic.

jvican:
As said by @Duhemm in #578, we should probably allow `bloop test` to mean test all the projects in the build graph.

BardurArantsson:
Hi, so I started to have a look at this...

I'm totally new to the code base, but I'm assuming it's reasonable to start from `Commands.scala` and to go "inwards"?

Two immediate questions:

- CompilingCommand demands a "project: String". This could probably easily be changed to List[String] and then doing the follow-up changes. However, the field gets used in e.g. "run" and "console" which perhaps doesn't make much sense. (I guess it would be possible to just do them sequentially? Or maybe special-case the command handling for them to just accept a list and then bail out if multiple projects are specified on the command line?)

- Wrt. the mechanism for actually running the "compile" (or whatever): Is the idea to just iterate sequentially through the projects in e.g. Interpreter#test() and get through all the projects that way? Would there be a risk of the ordering being wrong? I'm also assuming that this might lead to some duplicate work... or is the rest of the machinery clever enough that this won't be problem?

jvican:
Hey @BardurArantsson, this is actually quite good timing to tackle this one. Yesterday I added the most critical blocker for this feature in https://github.com/scalacenter/bloop/commit/9cda6c7689e4a34d557f0eecd701fb8f9b12f40e, an aggregate node in our DAG abstraction (cherry-pick it in your branch before you start working on the ticket).
  
More to the point, let's answer your questions:

> CompilingCommand demands a "project: String". This could probably easily be changed to List[String] and then doing the follow-up changes. However, the field gets used in e.g. "run" and "console" which perhaps doesn't make much sense. (I guess it would be possible to just do them sequentially? Or maybe special-case the command handling for them to just accept a list and then bail out if multiple projects are specified on the command line?)
  
[...]

PR #772, opened 2018-12-26
Compile, test and run several projects in one command

jvican:
Superseeds https://github.com/scalacenter/bloop/pull/667


jvican:
Fixes https://github.com/scalacenter/bloop/issues/575

Issue #788, opened 2019-01-09
Report compiler progress notifications every 5% increments


jvican:
Bloop now publishes all compiler progress notifications that we get from scalac. These notifications can, however, be overwhelming for clients that compiles targets in big builds. In order to reduce network traffic and avoid users to process all this data, we send processed progress notifications that clients can directly report in their UIs.

Issue #785, opened 2019-01-08
Fixed compile errors are not cleared with reset diagnostics


olafurpg:
To reproduce

- compile the following source file, no compilation errors as expected
```scala
object A {
  val x = 2
}
```
- compile the following source, diagnostics are published as expected. However, note that two diagnostics are published with the same start offset but the second error has different end position.
```scala
object A {
  val x = 1
  val x = 2 // error: x is already defined
}
```
- compile the original source file, fixing the error, but bloop does not publish any diagnostics. I expected it to publish `reset: true, diagnostics: []`
```scala
object A {
  val x = 2
}
```

This bug results in errors remaining in the editor even if they have been fixed.
![screenshot 2019-01-08 at 14 45 40](https://user-images.githubusercontent.com/1408093/50834341-16127600-1354-11e9-9499-1a97990bb8ad.png)

Possibly related https://github.com/scalameta/metals/issues/426

<details><summary>JSON-RPC trace</summary>
{
  "jsonrpc": "2.0",
  "method": "build/initialized",
  "params": null
}
{
  "jsonrpc": "2.0",
  "id": "2",
  "method": "workspace/buildTargets",
...skipping...
  "method": "build/publishDiagnostics",
  "params": {
    "textDocument": {
      "uri": "file:///Users/olafurpg/dev/metals/test-workspace/src/main/scala/example/User.scala"
    },
    "buildTarget": {
      "uri": "file:/Users/olafurpg/dev/metals/test-workspace/?id\u003dtest-workspace"
    },
    "diagnostics": [
      {
        "range": {
          "start": {
            "line": 4,
            "character": 6
          },
          "end": {
            "line": 4,
            "character": 6
          }
        },
        "severity": 1,
        "message": "x is already defined as value x"
      }
    ],
    "reset": true
  }
}
{
  "jsonrpc": "2.0",
  "method": "build/publishDiagnostics",
  "params": {
    "textDocument": {
      "uri": "file:///Users/olafurpg/dev/metals/test-workspace/src/main/scala/example/User.scala"
    },
    "buildTarget": {
      "uri": "file:/Users/olafurpg/dev/metals/test-workspace/?id\u003dtest-workspace"
    },
    "diagnostics": [
      {
        "range": {
          "start": {
            "line": 4,
            "character": 6
          },
          "end": {
            "line": 4,
            "character": 11
          }
        },
        "severity": 1,
        "message": "x  is already defined as value x"
      }
    ],
    "reset": false
  }
[...]

PR #792, opened 2019-01-14
Assort fixes to diagnostics and progress notifications

jvican:
Changes:

d607a57c (Jorge Vicente Cantero, 87 minutes ago)
   Fix test case for #785 and improve reporting if no-op

   + adding a bunch more tests that battle-tests the current changes.

e2c1818d (Jorge Vicente Cantero, 2 days ago)
   Improve bsp test infra and add diagnostics test case

   Fixes https://github.com/scalacenter/bloop/issues/785

49131300 (Jorge Vicente Cantero, 3 days ago)
   Refactor and simplify bsp compile test suite

   A first step towards having better tests to check the correct behavior of
   all the compilation engine through the lens of BSP.

   We could simplify further, but tests already look pretty, so we'll do that
   in independent future pull requests if anything.

dacec8d2 (Jorge Vicente Cantero, 4 days ago)
   Change order of logs because it's non-deterministic

3dfedcc6 (jvican, 4 days ago)
   Clear previous problems as early as possible

   BSP requires that bloop clears problems that were present in previous
   compiler iterations manually. Previously, Bloop would meet this contract by
   clearing problems from previous compiler runs at the end of a successful
   incremental cycle. For every incremental cycle, bloop would go through the
   problems and clear those that were affected by changes occurred during that
   cycle. You can see more information in this pull request:
   https://github.com/scalacenter/bloop/pull/782

   However, this approach was slower than desired and required that we
   successfully finished an incremental compiler cycle before *starting* to
   clean some diagnostics in the recompiled files.

   Instead, this pull request takes a different approach to this problem. We
   clear diagnostics as soon as we know they are safe to be cleared. For
   example, if in a previous compiler run typer emitted an error, we have
   bloop record that a diagnostic happened during that phase and, in
   subsequent compiler runs, we clear the diagnostic as soon as we've passed
   typer for the source file that contained the error.

   That is, we clear diagnostics right after the phase that generated them
   succeeded.

   If a diagnostic could not be mapped to a phase, we fallback to the previous
   behavior: we will clear it in `reportRemainingProblems` right after every
   incremental cycle.

4d2f9f3e (Jorge Vicente Cantero, 5 days ago)
   Report progress notifications every 5% increments

   Fixes https://github.com/scalacenter/bloop/issues/788

a50095eb (jvican, 5 days ago)
   Fix name in cancelled variable

c8552e38 (jvican, 5 days ago)
   Log rich compilation msg in bsp compile task start

   Imitates the same logging behavior when a compilation is triggered in the
   CLI `Compiling foo (1 Scala source)`.

037759cb (jvican, 5 days ago)
   Report start and end of compilation per incremental cycle

   Previously, the start and end of the compilation would be reported once per
   build target compilation. With this change, bloop reports the start and end
   of compilation at the granularity of the incremental cycle.

   This helps:

   1. Reporting precise progress notifications to clients (e.g. Metals), as
     we know all the total units upfront. 2. Exposes information about how
   many incremental cycles are used to
     compile a project. The more transparent we are with this information,
     the better.

df8a8fe0 (jvican, 5 days ago)
   Fix order of bsp task progress params

c951b7f2 (jvican, 5 days ago)
   Send progress notifications after the phase runs


olafurpg:
Great job! The Metals test suite passes with these changes and I'm no longer able to reproduce #785. I can also confirm that diagnostics are cleared much faster now 👍

PR #811, opened 2019-02-04
Implement compile deduplication

jvican:
This is just a prototype for now. Expect big changes.


jvican:
The approach in this implementation has changed so radically that it's not worth keeping this PR open. Closing, will open another one soon.

PR #816, opened 2019-02-14
Enable source dependencies by default

olafurpg:


jvican:
Heads up this change will be the latest to make it for 1.3.0 because we will enable it and then test our community build.

PR #818, opened 2019-02-17
Include Gradle's transitive dependencies in classpath

### LINKED ISSUE #800: Gradle project dependency issues
I don't think the Gradle plugin handles some types of project dependencies. e.g X depends on Y depends on Z.

The Gradle test: `ConfigGenerationSuite.worksWithSeveralScalaProjects` has Project B which depends on Project A and C.
If I add a Project D which depends on Project B then Project A's _jar_ rather than its _classes_ directory is included in Bloop's classpath and Project C is not on the classpath at all.

I think the first issue (Project A included as a jar rather than included as another classes dir) means that project dependencies on other projects have to be recursively queried to find their classes dir.

I think the second issue (Project C not being on the classpath at all) is down to Gradle having a separate compile classpath and runtime classpath and Bloop not.  For now I'd suggest adding Gradle's runtime & compile classpath to Bloop's classpath.

### LINKED ISSUE #808: Test dependencies are not registered in gradle
I have a 3-project Gradle build with 3 projects, each with their own main and test:

* common
* api
* web

`api` and `web` depend on `common` and gradlew bloopInstall handles this great, but for the test projects it incorrectly depends on `common` twice instead of `common` and `common-test`, e.g. this is from the generated `.bloop/api-test.json`:

```
  "name" : "api-test",
  ...
  "dependencies" : [
    "api",
    "common",
    "common"
  ],
```

Arthurm1:
Changes to the Gradle plugin...
Includes transitive dependencies in classpath
Includes runtime dependencies in classpath (if they're not already in the classpath)
No longer includes "resources" paths in "sources" section
No longer includes "test" section for main source sets (non-test projects)
No longer creates test projects that are empty (i.e. Gradle's compulsory test source sets)
Adds default platform + main class if present
Adds source and target to java section

Fixes #800 and #808


jvican:
Thank you for the contribution @Arthurm1, I'll have a look at this soon :smile:

jvican:
@Arthurm1 What could we do here to fix #808? From the logic in this PR, I'd assume we're almost there...

Arthurm1:
#808 is about a project depending specifically on the "test" source set of another project.  This is a direct dependency, not transitive.

Gradle treats main and test source sets as the same project but Bloop treats them as separate projects.  All project dependencies on other projects reference their main source sets only.

AFAIK Gradle doesn't have a way of specifying a dependency on a specific source set.  Hence the roundabout way of creating a jar artifact containing the test source and referencing the configuration that owns that artifact from another project.

I think I'll be able to fix #808, hopefully in the next few weeks.

One workaround for now is to factor out the project's common test code into the main source set of a new project e.g. TestCommon and reference that main source set from other project's tests.
It shouldn't be a problem having common test code in the "main" sourceset if there aren't any actually tests in it (just code the tests use) and this project will not be included in any deployment if it's only referenced using testCompile.

ijuma:
> AFAIK Gradle doesn't have a way of specifying a dependency on a specific source set. Hence the roundabout way of creating a jar artifact containing the test source and referencing the configuration that owns that artifact from another project.

We do the following in the kafka project:

https://github.com/apache/kafka/blob/trunk/build.gradle#L660

Is that the workaround you mean?

Arthurm1:
To me there are issues with both your workaround and the workaround in #808.

The `testCompile project(':clients').sourceSets.test.output` example tells Gradle to include a specific output directory.  Knowledge of depending on project "clients" is lost.

#808 example tells Gradle to build a jar containing the the test code classes dir which Bloop won't do as it doesn't assemble jars AFAIK.  Also the knowledge of the source of the jar data (the test classes dir) doesn't seem to be accessible in Gradle's api.

A workaround (that would work with Bloop now) would be to factor out common test code into a new project e.g. "TestClients", put your code in the main source set (not test source set) and depend on that using testCompile (":TestClients").

Saying all that - I hope to have a PR to fix yours and the example in #808 this weekend.

Arthurm1:
@jvican  Not sure what I've done here - I was intending to add 1 more commit to fix #808 and cope with Scala compiler plugins.

Is this OK?  I'm not too familiar with additional commits + PRs on GitHub

jvican:
@Arthurm1 totally OK to add more commits to this PR, no worries :+1: it'll simplify reviewing. I intend to look at this shortly, sorry for the delay, I appreciate a lot your work!

Arthurm1:
Thanks for merging @jvican.

All the reflection stuff is down to Gradle changing method names in newer versions.  The newer method names don't exist in the Gradle API version that Bloop uses and the old method names have now been removed in newer Gradle versions.

I've no idea if the Gradle team have stats on how many users are on which versions so that we could update the minimum version and get rid of the reflection calls, but as the work is done I don't see an issue supporting older versions.

jvican:
Thanks for the explanation @Arthurm1 :+1:

PR #819, opened 2019-02-18
Remove command prompt for one-line commands

EnverOsmanov:
The dollar sign should not be copied when I click a "copy" button. At least for one-line commands.
__
Example page with "copy" button" - [link](https://scalacenter.github.io/bloop/setup#scoop).

### DIFF
--- a/docs/tools/homebrew/install.md
+++ b/docs/tools/homebrew/install.md
@@ -1,5 +1,5 @@
 ```bash
-$ brew install scalacenter/bloop/bloop
+brew install scalacenter/bloop/bloop
 ```
 
 ### Requirements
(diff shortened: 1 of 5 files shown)


olafurpg:
fwiw, I personally avoid "copy" buttons and prefer to select text and cmd+c so the approach in this PR would work better with that workflow

jvican:
> fwiw, I personally avoid "copy" buttons and prefer to select text and cmd+c so the approach in this PR would work better with that workflow
  
if you select the text manually, then having a `$` at the beginning of the line is not a major issue
  
The benefit about modifying the copy js script is that when there are multi-line commands they can all be copied no matter if there are comments in the middle explaining what the commands do

EnverOsmanov:
Let's try another way :)

- The code snippet is marked as ` ```bash `, but it is not valid bash script with dollar sign at the beginning.
- Code snippets for other languages (like Scala) don't have additional formatting.

Also I noticed some code snippets are without dollar sign ([link](https://scalacenter.github.io/bloop/setup#zsh-completions)).

PR #826, opened 2019-03-06
Allow customization of cores via system property

jvican:
Note that we can only customize the number of cores dedicated to the
computation thread pool, which is the main thread pool where all
compilation takes place. We usually want to be this pool to have roughly
the same amount of threads as cores supported in the running machine.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/ExecutionContext.scala
+++ b/frontend/src/main/scala/bloop/engine/ExecutionContext.scala
@@ -13,7 +13,15 @@ import monix.execution.{ExecutionModel, UncaughtExceptionReporter}
 import monix.execution.schedulers.ExecutorScheduler
 
 object ExecutionContext {
-  private[bloop] val nCPUs = Runtime.getRuntime.availableProcessors()
+  private[bloop] val nCPUs = {
+    val default = Runtime.getRuntime.availableProcessors()
+    Option(System.getProperty("bloop.computation.cores")) match {
+      case Some(value) =>
+        try Integer.parseInt(value)
+        catch { case scala.util.control.NonFatal(_) => default }
+      case None => default
+    }
+  }
 
   import monix.execution.Scheduler
   implicit lazy val bspScheduler: Scheduler = Scheduler {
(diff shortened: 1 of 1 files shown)


frontend/src/main/scala/bloop/engine/ExecutionContext.scala
```diff
@@ -13,7 +13,15 @@ import monix.execution.{ExecutionModel, UncaughtExceptionReporter}
 import monix.execution.schedulers.ExecutorScheduler
 
 object ExecutionContext {
-  private[bloop] val nCPUs = Runtime.getRuntime.availableProcessors()
+  private[bloop] val nCPUs = {
+    val default = Runtime.getRuntime.availableProcessors()
+    Option(System.getProperty("bloop.computation.cores")) match {
```
olafurpg: Can we also try an environment variable? For example, I’m not sure how I would pass a system property to the bloop launcher.

jvican: You would pass it as a normal argument prepended with `-J`. Given that you will have control over the launcher in your CI, do you think you can work with that? I don’t like exposing an environment variable because:

a) the server is spawned in places where environment variables are scarce and sometimes not modifiable (think launchd)

b) it gives the impression that changing the environment variable would dynamically change the thread pool

c) As a matter of style, I’m avoiding environment variables to configure bloop. System properties are more adequate because they are passed exactly when the server is started

olafurpg: Sure I can work with a system property

frontend/src/main/scala/bloop/engine/ExecutionContext.scala
```diff
@@ -13,7 +13,30 @@ import monix.execution.{ExecutionModel, UncaughtExceptionReporter}
...
+
+    Option(System.getProperty("bloop.computation.cores")) match {
+      case Some(value) => parseIntOrDefault(value)
+      case None =>
+        Option(System.getenv("BLOOP_COMPUTATION_CORES")) match {
+          case Some(value) => parseIntOrDefault(value)
+          case None =>
+            Option(System.getenv("TRAVIS")) match {
+              case Some(value) =>
+                try if (JBoolean.parseBoolean(value)) DefaultTravisCores else default
```
olafurpg: Thank you 🙏

PR #837, opened 2019-04-07
Duplicate compilation if project inputs change

jvican:


jvican:
Thank you God CI for making my life less miserable and making this PR finally pass in Windows. I'll be eternally grateful and will continue to worship you every day with the same passion as you require me.

PR #839, opened 2019-04-09
Fix java.lang.NullPointerException in resources

artyomboyko:
```
[ERROR] Failed to execute goal ch.epfl.scala:maven-bloop_2.10:1.2.5:bloopInstall (default-cli) on project project: Execution default-cli of goal ch.epfl.scala:maven-bloop_2.10:1.2.5:bloopInstall failed.: NullPointerException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal ch.epfl.scala:maven-bloop_2.10:1.2.5:bloopInstall (default-cli) on project project: Execution default-cli of goal ch.epfl.scala:maven-bloop_2.10:1.2.5:bloopInstall failed.
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:956)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:288)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:192)
    at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:498)
[...]

PR #840, opened 2019-04-09
Follow up to #839

artyomboyko:
@jvican 

I've made mistake here https://github.com/scalacenter/bloop/pull/839

```
val resources = Some(resources0.asScala.toList.map(Option(_.getTargetPath)).flatMap(classesDir.resolve))
```
should be
```
val resources = Some(resources0.asScala.toList.map(Option(_.getTargetPath)).flatten.map(classesDir.resolve))
```


jvican:
This code was not compiling either, I'm pushing to master to fix it.

PR #847, opened 2019-04-12
Updated mill version 0.3.5 -> 0.3.6

robby-phd:
With this version change, mill-bloop seems to work with mill 0.3.6-42-d1f1b6 also (latest published "unstable" mill version).


robby-phd:
Not sure why the DeduplicationSpec test failed (it worked in my machine); also, it does not seem related to the change in this PR.

PR #849, opened 2019-04-15
Export source dependencies from Gradle builds

tgodzik:
This applies the default Eclipse plugin from gradle that downloads the dependencies locally and links them in the bloop configuration.


integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
```diff
@@ -417,20 +469,45 @@ final class BloopConverter(parameters: BloopParameters) {
...
+              path = artifact.getFile.toPath
+            )
+        }
+        .toList
+
     Config.Module(
-      organization = artifact.getModuleVersion.getId.getGroup,
-      name = artifact.getName,
-      version = artifact.getModuleVersion.getId.getVersion,
+      organization = artifact.getModuleVersion().getId().getGroup(),
```
odisseus: Why?

tgodzik: I think it looks better that you actually know it's a function

Issue #848, opened 2019-04-15
Replace `UUID.toString` with Base64 representations


odisseus:
Random UUIDs are used to create a unique name for temporary directories, BSP client instances etc. Using a more compact representation of UUID will lead to shorter names and shorter debug logs.

Among different [widely known](https://blog.codinghorror.com/equipping-our-ascii-armor/) encodings, Base64 provides reasonable compression (32 hex characters to 22 Base64 characters) and is [safe](https://en.wikipedia.org/wiki/Base64#Filenames) to use in filenames.

jvican:
> Using a more compact representation of UUID will lead to shorter names and shorter debug logs.
  
I like the end goal. Mind making a PR?

odisseus:
Okay. Here's the implementation I'm going to use: https://scalafiddle.io/sf/cmu1FcG/0

jvican:
Fixed in #851

PR #851, opened 2019-04-16
Encode UUIDs with Base 64

odisseus:


propensive:
What's the motivation for this, @odisseus?

jvican:
@propensive The motivation is shorter names, which I kind of like too https://github.com/scalacenter/bloop/issues/848

backend/src/main/scala/bloop/util/Base64Uuid.scala
```diff
@@ -0,0 +1,27 @@
+package bloop.util
+import java.nio.ByteBuffer
+import java.util.{Base64, UUID}
+
+object Base64Uuid {
```
jvican: Let's rename to `UUIDUtil`

backend/src/main/scala/bloop/util/Base64Uuid.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  // Copied from https://stackoverflow.com/a/29836273
+  implicit class UuidWithBase64(val uuid: UUID) extends AnyVal{
```
jvican: I propose to turn this into a method, there's no need for a value class

backend/src/main/scala/bloop/util/Base64Uuid.scala
```diff
@@ -0,0 +1,27 @@
...
+    def toBase64String: String = {
+      val bb = ByteBuffer.wrap(new Array[Byte](16))
+      bb.putLong(uuid.getMostSignificantBits)
+      bb.putLong(uuid.getLeastSignificantBits)
+      Base64.getUrlEncoder.encodeToString(bb.array)
+    }
+  }
+
+  def apply(base64: String): UUID = {
```
jvican: Is this being used? If not, let's remove it, we should be never decoding Base 64 to UUID, all we want is a shorter UUID for a random id :smile:

backend/src/main/scala/bloop/util/Base64Uuid.scala
```diff
@@ -0,0 +1,27 @@
...
+    val bytes = Base64.getUrlDecoder.decode(base64)
+    val bb = ByteBuffer.wrap(bytes)
+    val high = bb.getLong
+    val low = bb.getLong
+    new UUID(high, low)
+  }
+
+  def randomUUID(): String = UUID.randomUUID.toBase64String
```
jvican: `def randomUUID: String` without `()` is more idiomatic

propensive:
Yes, I like it too. Base64 will use two or three characters in addition to the alphanumeric characters. These will be `+`, `/` and potentially `=` at the end, which aren't filesystem-safe. I didn't check what this version produces, but there should be a URL-safe version of Base64 which uses `-` and `_` instead.

jvican:
Good point @propensive , even though @odisseus linked to https://en.wikipedia.org/wiki/Base64#Filenames in his ticket, it doesn't look like this PR is using the filename safe base64 variant.

odisseus:
The method [`getUrlEncoder`](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html#getUrlEncoder--) of class `java.util.Base64` promises to return an encoder that produces values [safe](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html#url) for use in filenames and URLs, but I have to admit I haven't verified that.

propensive:
```
scala> java.util.Base64.getUrlEncoder.encodeToString((0 to 255).to[Array].map(_.toByte))
res1: String = AAECAwQFBgcICQoLDA0ODxAREhMUFRYXGBkaGxwdHh8gISIjJCUmJygpKissLS4vMDEyMzQ1Njc4OTo7PD0-P0BBQkNERUZHSElKS0xNTk9QUVJTVFVWV1hZWltcXV5fYGFiY2RlZmdoaWprbG1ub3BxcnN0dXZ3eHl6e3x9fn-AgYKDhIWGh4iJiouMjY6PkJGSk5SVlpeYmZqbnJ2en6ChoqOkpaanqKmqq6ytrq-wsbKztLW2t7i5uru8vb6_wMHCw8TFxsfIycrLzM3Oz9DR0tPU1dbX2Nna29zd3t_g4eLj5OXm5-jp6uvs7e7v8PHy8_T19vf4-fr7_P3-_w==
```

Looks good to me.

backend/src/main/scala/bloop/util/Base64Uuid.scala
odisseus: Ok

backend/src/main/scala/bloop/util/Base64Uuid.scala
odisseus: Removed

backend/src/main/scala/bloop/util/Base64Uuid.scala
odisseus: Done

backend/src/main/scala/bloop/util/Base64Uuid.scala
odisseus: Done

odisseus:
@jvican, can you please check why the build has failed? The logs suggest it's an infrastructure issue.

jvican:
There was an outage at EPFL that brought our Windows CI down. Whenever it's back up, I'll merge this.

PR #861, opened 2019-04-24
Log to the user a deduplication is happening

jvican:
Making more visible the cases where bloop performs a deduplication is better for debugging purposes and will inform users better about what work is the build server doing.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/compilation/CompileGraph.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/compilation/CompileGraph.scala
@@ -124,12 +124,12 @@ object CompileGraph {
     }
   }
 
-  case class RunningCompilation(
+  private[bloop] final case class RunningCompilation(
       traversal: CompileTraversal,
       previousLastSuccessful: LastSuccessfulResult,
-      cancelPromise: Promise[Unit],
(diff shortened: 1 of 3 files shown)


frontend/src/main/scala/bloop/engine/tasks/compilation/CompileGraph.scala
```diff
@@ -185,15 +185,17 @@ object CompileGraph {
...
+          scheduleCompilation(inputs, bundle, client, compile)
         }
       )
 
       if (!deduplicate) {
         ongoingCompilation.traversal
       } else {
         val rawLogger = logger.underlying
-        rawLogger.debug(s"Deduplicating compilation for ${bundle.project.name} (${client}")
+        rawLogger.info(
```
olafurpg: I'm not sure a regular user is interested in this log. I think this would be noise for me at least.

jvican: We can consider disabling it down the road in the next one or two stable releases. For now, I find it important that users know when a deduplication is happening in case there's some kind of misbehavior and want to report it, or in case a compilation is not rightfully deduplicated. I prefer trading log noise over easiness of reporting...

PR #803, opened 2019-01-21
Add ArchLinux package generation

TheElectronWill:
This PR adds two sbt tasks:
- `createLocalArchPackage`, which generates the package files locally
-  `updateArchPackage`, which updates the [AUR bloop package](https://aur.archlinux.org/packages/bloop/). Git authentication is done with an SSH key stored in a file. The file's path is set by the env variable `BLOOPOID_AUR_KEY_PATH`.

One last step to make this work: create a Bloopoid/ScalaCenter AUR account, and I'll make it a maintainer of the bloop package.

_Ping @jvican who I've discussed with about this._

PR #874, opened 2019-04-29
Array indices in fish start from 1

2m:
Was getting the following fish error:

```
/usr/share/fish/vendor_completions.d/bloop.fish (line 57): array indices start at 1, not 0.
        if string match -q -- $tokens[-1] $argv[0]
                                                ^
in function “__assert_prev_arg_in”
        called on standard input
        with parameter list “--only -o”

in command substitution
        called on standard input
```

when trying to autocomplete project name in `bloop test`.

### DIFF
--- a/etc/fish/bloop.fish
+++ b/etc/fish/bloop.fish
@@ -54,7 +54,7 @@ function __assert_prev_arg_in
         return 0
     end
     for arg in $argv
-        if string match -q -- $tokens[-1] $argv[0]
+        if string match -q -- $tokens[-1] $argv[1]
             return 1
         end
     end


jvican:
@2m Checking out master and running `sbt install` should give you a local version that you can then install via `frontend/target/install.py`. That will install the completions for fish too.
  
If you're using bloop via another installation method (e.g. brew), you'll need to wait until a new release is out to benefit from this completions fix

2m:
Thanks for the tips. I am using Archlinux package from AUR. I'll wait for a release.

jvican:
:+1: @2m FYI in the next release https://github.com/scalacenter/bloop/pull/803 will be merged too

PR #875, opened 2019-04-29
Do not include project jar dependencies in Gradle

tgodzik:
Seems that previously we added all artifacts as inputs which cause project jars to be added also. This caused issues when projects were not compiled, since the jar would not exists.

This PR makes sure that we don't add artifact dependencies from other projects.


jvican:
I'm merging despite CI, the error is expected for now until I fix latest master, I'm changing some things in the test infrastructure

PR #888, opened 2019-05-10
Add source dependencies to maven plugin

tgodzik:
This adds download of source dependencies for project and then appends them to the bloop config.

I have tried the official Maven Api, but that is way too complex to use. Here I use an existing plugin that downloads all source dependencies to local repo and then I resolve all source deps by hand from normal paths.

### DIFF
--- a/integrations/maven-bloop/src/main/java/bloop/integrations/maven/BloopMojo.java
+++ b/integrations/maven-bloop/src/main/java/bloop/integrations/maven/BloopMojo.java
@@ -37,6 +37,9 @@ public class BloopMojo extends ExtendedScalaContinuousCompileMojo {
     @Parameter(property = "bloop.executionFork", defaultValue = "false")
     private boolean bloopExecutionFork;
 
+    @Parameter(property = "downloadSources", defaultValue = "false")
+    private boolean downloadSources;
+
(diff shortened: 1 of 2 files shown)


jvican:
@tgodzik Thank you for this PR!
  
1. Do we want to enable this by default? Would be good to have an option to disable it if downloading source dependencies fails.
2. Is there any Maven user that could give this a try? We don't have tests for the maven integration so it would be great if someone can try this out in their builds.

tgodzik:
>> Do we want to enable this by default? Would be good to have an option to disable it if downloading source dependencies fails.

I think I will add the option to enable it, it can take a while to download.

>> Is there any Maven user that could give this a try? We don't have tests for the maven integration so it would be great if someone can try this out in their builds.

I will try to get some info on gitter.

jvican:
@tgodzik I propose we get this merged in 1.3.1 if it's ok for you :smile: I cannot find the time to review and add this change to the release notes in 1.3.0, we already have too many new features in the release and I have the impression adding more will create more confusion than anything else...

olafurpg:
I would appreciate to have this merged to unblock the Maven integration in Metals. The diff is small, the changes look reasonable to me (although I admit I don’t know Maven we’ll) and they can be reverted in the worst case if it turns out to be a regression for Maven Bloop users.

PR #897, opened 2019-05-16
Fix not found type errors during incremental compilation

jvican:
Implements a similar fix as
https://github.com/scalacenter/bloop/pull/881 but this time making sure
we remove generated class files in dependent projects from the
transitive invalidations.


jvican:
Error in CI is spurious, @tgodzik  I'm merging since this was producing some weird errors during compilation and the bug fix is critical for correctness! feel free to review this code and ask any questions

PR #902, opened 2019-05-22
Update BSP version  to 2.0.0-M4

tgodzik:


frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -599,8 +599,9 @@ final class BloopBspServices(
         projects.iterator.map {
           case (target, project) =>
             val sources = project.sources.map { s =>
+              val kind = if(s.isDirectory) bsp.SourceItemKind.Directory else bsp.SourceItemKind.File
```
jvican: `sources` can contain both source dirs and source files. if a path doesn't exist, `isDirectory` will be false and the kind will be file (even though it could be a non-existing directory). To fix this issue, we should imitate what we do in `toBspSourceUri`: if a path doesn't exist, use file item kind only if the file name looks like a source file name, otherwise default on a directory item kind. https://github.com/scalacenter/bloop/blob/de71440426b75e50106053d354acda6993d4299e/backend/src/main/scala/bloop/io/AbsolutePath.scala#L27-L36

tgodzik: I added an optional parameter to `isFile` - does this make sense?

tgodzik:
Sure, I will take a look at whether metals work with the new bloop with updated version.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -599,8 +594,12 @@ final class BloopBspServices(
         projects.iterator.map {
           case (target, project) =>
             val sources = project.sources.map { s =>
+              val uri = s.toBspSourceUri
```
tgodzik: We could also do it like:
```
                val kind = if (s.exists) {
                  if (s.isFile) bsp.SourceItemKind.File else bsp.SourceItemKind.Directory
                } else {
                  if (AbsolutePath.sourceFileNameMatcher.matches(s.underlying.getFileName))
                    bsp.SourceItemKind.File
                  else bsp.SourceItemKind.Directory
                }
```

But whatever you think is best.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -599,8 +594,12 @@ final class BloopBspServices(
...
+              val kind =
+                if (uri.toString().endsWith("/")) bsp.SourceItemKind.Directory
+                else bsp.SourceItemKind.File
               // TODO(jvican): Don't default on false for generated, add this info to JSON fields
-              bsp.SourceItem(bsp.Uri(s.toBspSourceUri), false)
+              bsp.SourceItem(bsp.Uri(s.toBspSourceUri), kind, false)
```
tgodzik: IF we do it like that then we can :
```scala
bsp.SourceItem(bsp.Uri(uri), kind, false)
```

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: Didn't go for this approach because it's basically inlining `toBspSourceUri`, but given that we only use `toBspSourceUri` once in all the codebase (we used to use it more) the most sensible thing might be to just inline it here and remove `toBspSourceUri` altogether. Can you push  that change :smile: ? 
  
I'll investigate what's up with Windows later on the day

tgodzik: Sure, I will take care of it :)

PR #904, opened 2019-05-24
Describe the DAP integration

marek1840:


docs/debug-adapter.md
```diff
@@ -0,0 +1,37 @@
...
+Bloop allows the client to run and debug both the main classes and the tests defined in the project. 
+This is possible thanks to the Debug Adapter Protocol, which bloop exposes to the client 
+in a run/test task start notification. This adapter expects either a 
+`ScalaTestClassesItem` or `ScalaMainClass` as a configuration parameter for the launch request.
+
+### Client-Server communication
+![communication sequence diagram](assets/dap-sequence.png)
+
+1. BSP Client chooses which main class to run. To do so, it requests the list of all runnable main classes from the BSP Server.
+2. BSP Client request the chosen class to be run.
```
marek1840: @jvican written:
> I really like that we use `build/taskStart` as the notification to tell the BSP client the connection details. Another alternative would be to use a custom Bloop-specific notification.

@olafurpg written:
> I like the idea of keeping `buildTarget/run` and establish the debug adapter session via `build/taskStart` notifications

marek1840: Since the `run` request responds with the report for the whole run, using a notification it is the only viable option. And I think that creating another, custom notification would only add to the complexity. 

I have one concern: I have proposed to use the `RunTask` as a data field in`build/taskStart` - but what about tests? I am not sure, but aren't these notifications sent for each suite/build target? 
That's why, I would now like to propose sending `build/taskStart` with `DebugAdapterTask` as data instead. Then both the `RunTask` and `TestTask` would remain unchanged.

docs/debug-adapter.md
```diff
@@ -0,0 +1,37 @@
...
+### Example: [metals](https://github.com/scalameta/metals) integration
+
+Metals is an interesting case of BSP client. 
+It is acting as a language server therefore it must allow the language client to handle the Debug Client. 
+
+![metals communication sequence diagram](assets/dap-example-metals.png)
+1. Metals caches the available main classes.
+2. User opens a source file. As a result, Language Client asks metals for code lenses, which are resolved using the cached main classes metadata.
+3. User clicks on the code lens with the intention to run the specific main class. Metals requests a Debug Adapter uri from the BSP Server and then passes it back to the Language Client.
+4. Language Client instruments the Debug Client to connect to this Debug Adapter
```
marek1840: @olafurpg  asked:
> What will be the arguments of the `launch` request via DAP? I think it would be great if we can build on top of the launch.json configuration in VS Code so that users can trigger run/test via keyboard shortcuts instead of clicking on "run/test" buttons.
> 
> The nice thing about launch.json configuration is that it could committed into git and shared across team members.

marek1840: Let me rephrase the problem and suggest a solution. Then we can discuss whether it solves your concerns and if it is viable at all.

Problem: 
1. Configurations must be visible from the outside of BSP/metals (in Language Client)
2. Configurations created/updated bythe Lanugage Client must remain usable

Propositions:
### configuration storage in metals

This storage could be an in-memory cache (populated on startup from client's `launch.json`) or a directory maintained by metals (but that is harder, I think). 

Whenever Language Client configuration store is updated (onDidSave for `launch.json`?), metals parses this file and updates its internal storage.

Whenever a configuration is run and is not present in this storage, metals appends the configuration to the `launch.json`. This configuration is a `scalaMainClass` or `scalaTestClassesItem` and is saved without any changes (meaning: as it was sent by the BSP Server earlier).

How will we know it is being run for the first time? We can compare checksums or something. That way we will persist every configuration that has been launched, which will allow the user to repeatedly run same configurations without using code lenses. Also, main classes never launched will not be stored. Another bonus is that now we can store test configuration which spans across many build targets (let's say in metals `slow` and `unit` but not `cross`).
	
The whole process would work like that (sorry for the lack of diagrams ;P):
	1. metals fetches main classes
	2. vs code requests code lenses
	3. user clicks on the code lens
	4. vs code sends a `metals-main-run` request with the configuration
	5. metals checks-up its internal cache of configurations (magic hashing techniques). If the configuration is not persisted in Language Client configuration cache, metals appends it. Then, regardless of whether it was saved or not, metals requests `buildTarget/run`
	6. user makes some changes to this configuration and runs it.
	7. vs code resolves the configuration by name and sends it to metals via `metals-main-run`. Metals checks if it has such configuration stored. It doesn't, since the user created it himself by modifying. So, metals just runs `buildTarget/run`.
	8. user clicks code lens once more. metals does not recognise this configuration and therefore appends it. Now there are two configurations in `launch.json`: modified by user and "bare" from clicking the code lens.

### configuration store in bloop

I have not explored this approach a lot but also I am not sure storing configurations fits into the range of responsibilities for the BSP Server. Thoughts?

marek1840:
I would also like to propose dropping the `buildTarget/test`. With the suggested approach it can all be handled by the single endpoint. `buildTarget/run` has the `dataKind` field and I believe it could (and should) be used to drive the particular mode of execution. 

If this field indicates that we are running tests, then the appropriate notifications and `TestReport` will be send to the client.
If it shows that the user requested us to run a `ScalaMainClass` then we return simple `RunReport` and issue notifications appropriate for this mode of execution. 

I think it allows the Build Server implementation a great deal more freedom.

Also, my proposition with configurations would be a lot harder (if not impossible without additional hacks) when still maintaning `buildTarget/test`

docs/debug-adapter.md
olafurpg: First, let me try to define a few objectives that I think are desirable for any solution around managing launch configurations:

- the user should be able to trigger run/test via `launch.json` using keyboard shortcuts, it shouldn't be necessary to use a mouse to click on a code lens
- the user should be able to commit `launch.json` to version control so that other team members can checkout a clean repo to reuse custom launch configurations

Here's a proposal for how we could manage launch configurations:

- no configuration storage in Bloop, Bloop only defines the spec for what launch configurations it supports (build target identifier, main class, test class, link options, etc.) and executes the launch configurations it receives via DAP `launch`.
- Metals stores all executed launch configuration in the `.metals/metals.db` SQL database

```sql
create table launch_configuration(
  when_launched timestamp,
  launch_data: varchar not null, -- JSON encoded string.
  -- Up for discussion, human readable description of the launch configuration.
  -- For example "Test UserSuite in core-test".
  display_name varchar not null
);
```
- Metals auto-generates a single launch configuration in VS Code `launch.json` with value `{"name": "Run Last Task"}`, which executes the persisted launch configuration with the most recent timestamp. No hashing or advanced logic is needed to generate this config because the JSON is so simple.
- Metals registers the LSP server command "Metals: Save Recent Task", which opens a select list showing the last 5-10 executed launch configurations. The configuration selected by the user will be persisted in the VS Code `launch.json` file, with full details including build target identifier and test/main class.
- Metals registers the LSP server command "Metals: Run Recent Task", which works the same as "Save Recent Task" except it starts a debug session to launch that configuration instead of saving it to launch.json

docs/debug-adapter.md
marek1840: > I am no longer sure if it's a good idea to establish DAP via buildTarget/run
I think it it would be best to introduce a custom JSON-RPC endpoint to start a debug session

Why? What are your concerns?

docs/debug-adapter.md
marek1840: > - the user should be able to trigger run/test via launch.json using keyboard shortcuts, it shouldn't be necessary to use a mouse to click on a code lens

That is the responsibility of the editor - to make configurations available to the user under the specified key binding. Especially since the editor has the full knowledge of the configurations. Currently, I could not make it work with the built-in `run last task` command - @jvican you mentioned something about it working, if I remember correctly?

Anyway, if it is in fact impossible to run a specific configuration in a convenient way (from the config file, using commands), then it probably merits an issue in the vs code. 

> - the user should be able to commit `launch.json` to version control so that other team members can checkout a clean repo and still reuse custom launch configurations

Agreed.

All in all, seems doable. But it is for sure more convenient for the developer - but is it more convenient for the user?

docs/debug-adapter.md
olafurpg: I think it complicates the pipeline to keep `buildTarget/run` and `buildTarget/test` in the mix. The two endpoints can be replaced with `debugLaunchConfiguration/resolve` and it would be clearer for a new person to read the spec and understand what's going on.

docs/debug-adapter.md
olafurpg: > is it more convenient for the user?

I think so, I tried to design for the user experience first and architecture (implementation details) second.

> the user should be able to trigger run/test via launch.json using keyboard shortcuts, it shouldn't be necessary to use a mouse to click on a code lens

Code lenses

docs/debug-adapter.md
marek1840: In that case:
> The two endpoints can be replaced with

Fully agree

> `debugLaunchConfiguration/resolve`

we could work on the name, since the it is a bit misleading (and a mouthful): DAP is just an implementation detail. 
First part could be something related to launching, starting, evaluating? 
Second could be any of: `startSession`, `openConnection`, and probably more?

docs/debug-adapter.md
olafurpg: > the user should be able to trigger run/test via launch.json using keyboard shortcuts, it shouldn't be necessary to use a mouse to click on a code lens

I agree the keyboard shortcut part is a responsibility of the editor. Let me try to reformulate the point: it should be possible to launch configurations through `launch.json` configurations in addition to code lenses because launch configurations are the standard interface to run Python/Node.js/Java/C# applications in VS Code.

marek1840: Well it surely depends on ones habits, so let's wait for the opinion of others

docs/debug-adapter.md
olafurpg: To clarify, I'm not proposing to remove `buildTarget/run` and `buildTarget/test` from BSP. Those two endpoints can be dealt with separately. My primary suggestion is to not bake the Debug Adapter Protocol into the semantics of `buildTarget/run` and `buildTarget/test`. I think it's clearer to have a custom endpoint `debugLaunchConfiguration/resolve` that deals only with how to establish a DAP session via LSP/BSP.

olafurpg: My thinking behind the name of `debugLaunchConfiguration/resolve`:

- DAP is not an implementation detail since the contract of the `port` field is that the client will be able to open a socket to that port number to speak DAP.
- it resembles `completionItem/resolve` from LSP, which is used in a similar way: client sends a `DebugLaunchConfiguration` JSON request and the response contains a `DebugLaunchConfiguration` JSON response with all fields populated.

marek1840: So, as I understand it: you propose that we:
- introduce a new endpoint for dealing with the DAP
- leave the `buildTarget/run/test` as they were

So now, when the user wants to run the main class, it uses the `buildTarget/run` endpoint but when he wants to debug it, he has to spin up a Debug Client and run `debugLaunch.../resolve`.
If that is not correct, please elaborate on your solution, because I am clearly missing something :(

olafurpg: > So now, when the user wants to run the main class, it uses the buildTarget/run endpoint

This is incorrect, the user always runs `debugLaunchConfiguration/resolve` (both for main/test) and always needs to spin up a Debug Client. We can rename the endpoint to `launchConfiguration/resolve` if that makes it clearer. The `buildTarget/run` and `buildTarget/test` can be left unchanged in the current BSP spec but Metals won't use them.

docs/debug-adapter.md
olafurpg: The primary interface that VS Code uses for launching applications is `launch.json`, which can be selected like this


Me and @jvican looked at how it works in the Python client and they use the same interface for debugging. For regular running without breakpoint debugging they use a separate custom interface, which is confusing in my opinion.

Screenshot 2019-05-24 at 21 34 46

docs/debug-adapter.md
marek1840: I am asking from the viewpoint of a BSP Client. If we leave "dangling endpoints", someone is bound to use them. If you consider this a non-issue, please say so.

Also, because I might have the wrong idea the whole time - are we brainstorming integrating the `DAP` with `bloop`, or are we spiking a marriage of `DAP` and `BSP`?

olafurpg: I think it's a non-issue to leave `buildTarget/run` and `buildTarget/test` unchanged for now. There may exist BSP clients which appreciate a simpler interface to run/test without the complexity of the DAP.

> marriage of DAP and BSP

We're integrating DAP into BSP by introducing a new request endpoint `launchConfiguration/resolve` that enables the build server to power a rich user interface for launching/evaluating source code with optional debugging functionality. I wouldn't call it a marriage since the two protocols can live fine without each other.

olafurpg: DAP has a lot of nice functionality and rich client/server implementations that we have no desire to merge in BSP, and BSP has a lot of functionality and rich client/server implementations that we have no desire to merge into DAP.

olafurpg: As discussed offline, the new JSON-RPC endpoint would be

```ts
// request: 'dap/connection' or 'debugAdapterProtocol/connection'
// params: empty object {}
// response: DebugAdapterProtocolConnection
interface DebugAdapterProtocolConnection {
  // Port number for client to establish a DAP session.
  port: number;
}
```

The `launchData` field has been removed, it will be the responsibility of Metals to figure out how to resolve the launch arguments.

jvican:
Closing in favor of #944

PR #905, opened 2019-05-26
Add build settings hashing to Buildpress

tkroman:
As discussed with @jvican.

Opening the PR early to get some feedback on certain points that I don't have much confidence in, especially:

- ~~added `compile` dependency on `frontend` to get access to `AbsolutePath`, hashing etc.~~ solved
- ~~configs: since serialization (more importantly, 2.10/2.11/2.12 workarounds) is encapsulated in the `config` module, I went with putting repo cache codecs into `config` too, is that OK?~~ solved
- no source-based plugin code change detection - I assume I would have to use semanticdb for that?
- ~~there is some duplication between community build code and buildpress code (config parsing), but i can't put the parsing into `config` modules because sbt refuses to load then with a `io.circe.parser` not found error. I saw main config's parser is also outside of the `config` module, is that for the same reason?~~ **solved** by adding a `circe-parser` dependency to all the build.sbt files that are involved in meta-build. ref https://github.com/scalacenter/compiler-benchmark/pull/7 

- couldn't get rid of `buildpressConfig` due to otherwise circular dependency launcher<->frontend<->buidlpress. now `frontend` depends purely on `buildpressConfig` to be able to get repo list from the cache.

Also, I'm not doing build file structure parsing for hashing yet.


tkroman:
@jvican do you mind addressing a first couple of questions in the PR description? Specifically interested to know how OK/not OK was it to add a dependency on backend and did I handle configs correctly?

jvican:
Hey @tkroman! I apologize for my slow response to this PR, these last three days have been quite busy and yesterday I was all day in Zurich attending a Scala Center spree.

> * added compile dependency on frontend to get access to AbsolutePath, hashing etc.
> * configs: since serialization (more importantly, 2.10/2.11/2.12 workarounds) is encapsulated in the config module, I went with putting repo cache codecs into config too, is that OK?

I think it's important we keep buildpress independent of bloop itself, so I propose the following:

* Move the bits you need carefully from frontend to an independent module called `io` (module name when publishing should be `bloop-io`, take a look at how we do this in the `frontend` project --published as `bloop-frontend`). That module can contain the hashing bits you need and `AbsolutePath` and `RelativePath` (let's only move the things we need in both buildpress and frontend).
* Move the serialization logic back to `buildpress`, the circe encoder/decoders in the config modules are cross-compiled because we use them across all our build integration projects (which run on different Scala versions), but that's not the case with the buildpress codecs which only need to work in 2.12
  
I hope I have clarified all your questions! I'll be active today so feel free to shoot any questions you have left.

tkroman:
Thanks for the feedback, makes sense, will do.

jvican:
@tkroman Having a look at this soon :wink:

tkroman:
Awesome! In the best case - let me know before the merge, I’d like to squash and write up a proper commit message.

tkroman:
Oh and I’d like to leave src-based sbt plugin change detection out of this if that’s ok with you.

jvican:
> Oh and I’d like to leave src-based sbt plugin change detection out of this if that’s ok with you.

Sounds good to me, we're doing the Scala Center spree today, I'm trying to review this PR as soon as I have some time available :smile:

buildpress-config/src/main/scala/buildpress/config/Config.scala
```diff
@@ -0,0 +1,79 @@
...
+    Encoder.encodeString.contramap[Path](_.toString)
+  implicit val pathDec: Decoder[Path] =
+    Decoder.decodeString.emapTry(s => Try(Paths.get(s)))
+
+  implicit val uriEnc: Encoder[URI] =
+    Encoder.encodeString.contramap[URI](_.toString)
+  implicit val uriDec: Decoder[URI] =
+    Decoder.decodeString.emapTry(s => Try(URI.create(s)))
+
+  @JsonCodec
```
jvican: We would usually avoid using these macro annotations at all costs and instead use semi-automatic derivation manually. However, i don't think it's a big deal so we can live with this code for now :smile:

buildpress-config/src/main/scala/buildpress/config/Config.scala
```diff
@@ -0,0 +1,79 @@
...
+      )
+  }
+
+  def write(all: RepoCacheFile, target: Path): Unit = {
+    Files.write(target, toStr(all).getBytes(StandardCharsets.UTF_8))
+    ()
+  }
+
+  def readBuildpressConfig(file: Path): Either[String, RepoCacheFile] = {
+    val cfgBytes: Array[Byte] = Files.readAllBytes(file)
```
jvican: Don't act on this, but using simple names such as `bytes` and `contents` (for the variable below) is usually better for such small bodies of functions where the semantics are so clear.

frontend/src/it/scala/bloop/CommunityBuild.scala
```diff
@@ -1,34 +1,21 @@
...
-import bloop.cli.Commands
-import bloop.engine.Action
+import scala.concurrent.Await
 import scala.concurrent.duration.Duration
-import bloop.engine.ExecutionContext
-import bloop.engine.Interpreter
+import bloop.cli.{Commands, ExitStatus}
+import bloop.config.Config
+import bloop.data.{Origin, Project}
+import bloop.engine._
```
jvican: Let's try to avoid wildcard imports as much as possible and instead prefer explicit import :smile: I suppose it was your IntelliJ configuration the one that did this change, so you might want to disable it when you work in the bloop codebase.

buildpress/src/main/scala/buildpress/io/SbtProjectHasher.scala
```diff
@@ -0,0 +1,78 @@
+package buildpress.io
+
+import java.io.IOException
+import java.nio.file._
```
jvican: Same here with regards to wildcard imports :smile:

buildpress/src/main/scala/buildpress/io/SbtProjectHasher.scala
```diff
@@ -0,0 +1,78 @@
...
+          exception: IOException
+      ): FileVisitResult = {
+        FileVisitResult.CONTINUE
+      }
+    }
+
+    Files.walkFileTree(
+      path,
+      java.util.EnumSet.of(FileVisitOption.FOLLOW_LINKS),
+      Int.MaxValue,
```
jvican: Shouldn't the depth level be here 1?

buildpress/src/main/scala/buildpress/io/SbtProjectHasher.scala
```diff
@@ -0,0 +1,78 @@
...
+          directory: Path,
+          exception: IOException
+      ): FileVisitResult = {
+        FileVisitResult.CONTINUE
+      }
+    }
+
+    Files.walkFileTree(
+      path,
+      java.util.EnumSet.of(FileVisitOption.FOLLOW_LINKS),
```
jvican: Is sbt following symlinks in their sbt files?

jvican:
Actually I didn't push to your branch because I saw you changed it, so I leave you here a patch you can apply on top of your work https://gist.github.com/jvican/efd0bc62453a088b993caa81f398585a

buildpress/src/main/scala/buildpress/io/SbtProjectHasher.scala
tkroman: practically yes, i haven't seen people going more than 1 `project/` level nesting, maybe I'll put 2 just to be safe, but this is a good catch performance-wise!

buildpress/src/main/scala/buildpress/io/SbtProjectHasher.scala
tkroman: actually no, i tried to symlink a shared sbt file once, loading failed so this is also a good catch!

buildpress/src/main/scala/buildpress/io/SbtProjectHasher.scala
jvican: I'm not sure sbt can read sbt files deeper than 1 level, especially in these meta builds, so I'd probably default on 1 :smile:

jvican:
I have killed the CI process because I'm making some changes in the CI, I'll restart it as soon as possible and review this PR then :smile:

jvican:
@tkroman I've push forced to rebase with latest master and add two changes. I misunderstood the way you were detecting all files belonging to a build and therefore my previous comment regarding the nesting was incorrect, sorry about that! I've pushed an improvement and document it so that we have a more precise detection of sbt/scala files that affect a build.
  
Would you mind having a look?

tkroman:
This is definitely more thoughtful, especially in the directory traversal part! Thank you!

jvican:
Please try it out, didn't test it manually! :smile:

jvican:
@tkroman Can we merge this already? I'd like to get this merged soon-ish because we're making several changes to the build definition and I don't like running "Import build" after checking out different branches :see_no_evil:

tkroman:
@jvican hi,just got back from vacation today, will get back to you with an answer and fixes (if any will be needed) by EOD.

tkroman:
Found 1 issue, fixed & verified manually.

PR #908, opened 2019-05-29
Display Java version and home in bloop about

cquiroz:
As suggested in gitter, this PR changes `bloop about` to show the JVM version and java home bloop is running on:

image

### DIFF
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -103,9 +103,9 @@ object Interpreter {
 
     logger.info(s"$bloopName v$bloopVersion")
     logger.info("")
-    logger.info(s"Running on Scala v$scalaVersion and Zinc v$zincVersion")
+    logger.info(s"Using Scala v$scalaVersion and Zinc v$zincVersion")
     (javaVersion, javaHome) match {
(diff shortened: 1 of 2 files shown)


frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -98,10 +98,16 @@ object Interpreter {
     val scalaVersion = bloop.internal.build.BuildInfo.scalaVersion
     val zincVersion = bloop.internal.build.BuildInfo.zincVersion
     val developers = bloop.internal.build.BuildInfo.developers.mkString(", ")
+    val javaVersion = Option(System.getProperty("java.version"))
```
tgodzik:
```scala
sys.props.get("java.version")
sys.props.get("java.home")
```
Should get an option and noe need to pack it in one here.

jvican: I actually prefer this over using the Scala API :smile:

tgodzik: Haha, fine with me. Just thought it might be a bit shorter.

cquiroz:
Thanks. I didn't think you'd check this on test but great that you do. I'll try to fix the test today

Issue #894, opened 2019-05-14
Don't print warnings when autocompleting tests in zsh


kubukoz:
I understand what went wrong, but the way the error message is displayed is weird (although hilarious!):

![image](https://user-images.githubusercontent.com/894884/57706115-69c7ff80-7665-11e9-992b-ea0be44dc926.png)

jvican:
Thanks for reporting! It looks like we should have bloop completely disable logger errors and warnings when `bloop autocomplete` (the command run by the terminal completions) is triggered. @kubukoz Fixing this should be straightforward, the implementation for autocompletion is short and self-contained https://github.com/scalacenter/bloop/blob/0d97ce30048ac2a8e14f0d837b7058cc1d3df7f3/frontend/src/main/scala/bloop/engine/Interpreter.scala#L366-L444
  
Adding a `state.copy(logger = NoopLogger)` to every use of state within that method would work, an example of where we use state is for completing test suites here https://github.com/scalacenter/bloop/blob/0d97ce30048ac2a8e14f0d837b7058cc1d3df7f3/frontend/src/main/scala/bloop/engine/Interpreter.scala#L431-L433

CucumisSativus:
I'll work on this one

jvican:
Fixed by https://github.com/scalacenter/bloop/pull/910

PR #910, opened 2019-05-29
Don't print warnings when autocompleting in zsh

CucumisSativus:
Since there are warning messages in findMainClasses and
findFullyQualifiedTestNames which will look very ugly in autocomplete
(check issue #894 to see the result) we are passing NoopLogger
to make sure that no warnings are printed in autocomplete.


jvican:
Fixes https://github.com/scalacenter/bloop/issues/894

PR #911, opened 2019-05-29
Update contributing guide

CucumisSativus:
Since user should not be worried about this error its nice to have it
mentioned


docs/contributing-guide.md
```diff
@@ -139,6 +139,18 @@ local Python script to override your installation in `$HOME/.bloop`:
 1. Give executable permission to the install script: `chmod +x $BLOOP_REPO/frontend/target/install.py`.
 1. Run script with `./$BLOOP_REPO/frontend/target/install.py`.
 
+You will something like this:
```
tkroman:
```suggestion
You will see something like this:
```

CucumisSativus: I must have been pretty hungry while writing this. Thanks for spotting it!

PR #914, opened 2019-06-01
Update gradle for java11 support

er1c:
I'm still getting errors locally, want to see the CI result


er1c:
I can't tell if the CI errors are related to this PR or not, they seem unrelated.

project/Dependencies.scala
```diff
@@ -26,8 +26,8 @@ object Dependencies {
   val mavenApiVersion = "3.6.1"
   val mavenAnnotationsVersion = "3.5"
   val mavenScalaPluginVersion = "3.2.2"
-  val gradleVersion = "3.0"
-  val groovyVersion = "2.5.0"
+  val gradleVersion = "5.4.1"
```
tgodzik: Do we need to change the version here? I think the idea is to keep it as low as possible and we have separate tests for different version. I woul dkeep it at `3.0`

jvican: Unfortunately I think we do because previous versions cannot build with JDK 11. I don't have any strong opinion on this, so we can merge it and see who complains and how strong their arguments are to avoid a build version bump.

tgodzik: But do the older versions need to build with JDK 11? They should still work with jdk 8, so they wouldn't upgrade jdk unless they upgraded gradle version

er1c: The older gradle versions don't build with jdk11, but if I understand you correctly, there might be a way to force the gradle project to build with jdk8 and keep the older versions?

tgodzik: Do we want to build the plugin with JDK11? I might be totally missing the point here. What exactly is the purpose of the PR? 

What is the failure when trying to run gradle 5.4.1 with this plugin and JDK 11?

er1c: I have been compiling the plugin (and whole project) with `JDK11`  It's up to the maintainers to decide if that is desirable or not (and you are raising a good question) I've had a myopic focus.  By updating gradle, this becomes possible.  I haven't explicitly tried to run 5.4.1 using the master branch - I've simply been focused on getting bloop to compile/test cleanly against jdk11 - not actually using the gradle plugin.  I'm fine with closing this PR until a more holistic plan is needed.

er1c:
I'm going to close this for now, post a conversation with @tgodzik - I have a couple research items on my list

PR #917, opened 2019-06-04
Catch exceptions when computing the compile bundle

jvican:
The compile bundle can throw if there is some IO error when, for
example, hashing sources or classpath entries. To protect us from that,
we make sure that we catch any exceptions in its computation and we
translate it to our compile result representation so that the user can
now about it.


jvican:
I checked manually CI errors are spurious.

PR #931, opened 2019-06-08
Fix memory leak in listing directories

jvican:
Two crashes due to excessive open file pointers in my macOS machine have
fired a red alarm that have led me to discover a memory leak in the way
we were listing directories. This memory leak was caused because
`Files.list` returns an stream that must be closed, but I have not seen
that before because I was converting that stream to a Scala iterator
via the automatic Java converters and at the end of that iteration
resources were not correctly cleaned up.

Now we make sure that we always close these resources.

### DIFF
--- a/backend/src/main/scala/bloop/io/ClasspathHasher.scala
+++ b/backend/src/main/scala/bloop/io/ClasspathHasher.scala
@@ -92,7 +92,7 @@ object ClasspathHasher {
             ()
           }
 
-          /* 
+          /*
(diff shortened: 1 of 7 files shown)


jvican:
I can confirm this PR fixes the resource leak with Sloth https://sveinbjorn.org/sloth
  
Screen Shot 2019-06-08 at 10 18 35 PM

I've checked manually the list of open file pointers in the bloop process (the second one) and all of them make sense. The first process shows the number of open file pointers owned by the Metals process.

Issue #933, opened 2019-06-11
Don't require Java 8 on macOS


fredfp:
Java 8 is still required when running on macOS (installed via brew)

jvican:
Adding here an explanation of why this is not yet enabled in macOS from the release notes:

> Contributed by er1c, bloop v1.3.0 supports JDK 9 and JDK 11. If you run
bloop with these Java runtimes, it is recommended to enable the full parallel
GC with -XX:+UseParallelGC as compilation is faster.

> For this reason, Java 8 still remains the recommended runtime for the time
being and our installers might install it by default. In the future, we might
consider upgrading to a newer Java runtime.

jvican:
Related https://github.com/scalacenter/bloop/issues/624

jvican:
I'm closing this ticket because we will be removing this limitation soon, thanks for reporting @fredfp and following up with the PR.

PR #934, opened 2019-06-11
Use java 8 or newer on macOS

fredfp:
fix for https://github.com/scalacenter/bloop/issues/933

### DIFF
--- a/bin/install.py
+++ b/bin/install.py
@@ -139,7 +139,7 @@ def macos_launch_script_contents(is_local):
     echo "Detected java options are '${java_opts[@]}'"
     BASE_BIN_DIR=$(dirname "$0")
     COURSIER_BIN="$BASE_BIN_DIR/%s"
-    /usr/libexec/java_home -v 1.8 -F -R --exec java \\
+    /usr/libexec/java_home -v 1.8+ -F -R --exec java \\
       %s"${java_opts[@]}" -jar "$COURSIER_BIN" launch %s \\
       -r bintray:scalameta/maven \\
       -r bintray:scalacenter/releases \\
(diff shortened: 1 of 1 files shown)


jvican:
This PR is good but as I said in the linked ticket we should probably figure out a way we can do this without regressing on performance, as the default GC in versions bigger than 8 is significantly (around 20%) slower in previous experiments.
  
So, that would entail adding the full parallel GC options to the jvmopts file after the installation of bloop in Homebrew. I recommend you first read the bloop server reference guide in the bloop website and then looking at the `ReleaseUtils` source in bloop's project build to modify the HomeBrew installation recipe to add the necessary jvm options.

fredfp:
Thanks for the explanation, I've quickly hacked a (naive) way of trying different versions at runtime. What are your thoughts about it? If you don't like it I'll try to have a look at the `ReleaseUtils` you mentioned

lavrov:
While slower compilation on later versions of java is a bummer bloop is used by other tools such as metals etc. I think it make sense to relax requirements and allow 1.8+ with notice that 1.8 gives the best performance.

jvican:
Sorry for my slow feedback, I'm still giving this more thought. I'll reach back whenever I've made up my mind what the best thing to do here is.

jvican:
Hey @fredfp, in `1.4.0` we're going to be changing the way installation works in macOS. Instead of using the script with coursier and forcing the version, we will be relying on the default Java version in your system. The CLI now has the capability to start the server in the background, so whenever we do start the server we will try to figure out the environment and use the right JVM options.
  
I think we can close this pull request since this change will no longer be required in 1.4.0. Thanks a lot for starting this discussion and caring about opening a pull request.

Issue #903, opened 2019-05-22
Execute tests in parallel


tkroman:
Nice to have: executing tests for separate projects in parallel. Should be done optionally at user's discretion (e.g. some integration tests probably can't even be run in parallel, but most unit tests can be).
As fas as I can tell, compilation phase handling is already delegated to compilation task which was solved by #11, so we should do something similar for tests.

cc @jvican

jvican:
Fixed by #935, thank you @tkroman :pray:

PR #935, opened 2019-06-11
Add parallel test execution support

tkroman:
Should address #903. Just something I figured on my own, hope this is the right approach.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
@@ -88,7 +88,8 @@ object Tasks {
       userTestOptions: List[String],
       testFilter: String => Boolean,
       testEventHandler: TestSuiteEventHandler,
-      failIfNoTestFrameworks: Boolean
+      failIfNoTestFrameworks: Boolean,
+      runInParallel: Boolean = false
   ): Task[State] = {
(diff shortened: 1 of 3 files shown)


frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
```diff
@@ -122,7 +123,14 @@ object Tasks {
     }
 
     // For now, test execution is only sequential.
-    Task.sequence(testTasks).map { exitCodes =>
+    val runAll: List[Task[Int]] => Task[List[Int]] =
+      if (runInParallel) {
+        Task.gather
```
tkroman: didn't go with `gatherUnordered` since not sure if users would find that behavior predictable.

jvican:
<3 I love the implementation of this is so short! I'm thinking about the implications of enabling this, I'll add a much better review tomorrow :smile:

tkroman:
Biggest doubt point for me is that test output is interleaved now which is not good for interactive tailing, but for ci purposes that shouldn’t matter as projects’ reports get written to separate files (iirc), and for cases where you have failed tests you will most probably rerun those separately. For most happy path cases this can be seen as a useful shortcut to save some seconds, maybe if we “sell” it as just that, there won’t be any extra expectations from users.

PR #938, opened 2019-06-13
Fix link reference in docs

ignasi35:
The syntax looked wrong in https://scalacenter.github.io/bloop/setup#generate-configuration-files.

Not sure what the reference should point to...

### DIFF
--- a/docs/build-tools/sbt/export.md
+++ b/docs/build-tools/sbt/export.md
@@ -17,7 +17,7 @@ written to a configuration directory. The default location of this directory in 
 `.bloop/` (you may want to add `.bloop/` to your `.gitignore` file).
 
 For example, a build with a single project `foo` generates two configuration files by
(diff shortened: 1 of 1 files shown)


ignasi35:
> Hey @ignasi35 this looks like valid markdown code. Was the link not working?

Yeah, you're right. Still, it's doesn't render as a link:

Screen Shot 2019-06-21 at 10 34 30

Go to https://scalacenter.github.io/bloop/setup#homebrew, and then select `Sbt` in section `4. Export your build`.

I guess the `sbt-configuration` id is not found.

jvican:
I forgot to update that last time I removed the section that link was pointing to, I took the liberty to update your PR to do so :smile: Thanks for the catch @ignasi35 🚀

ignasi35:
nice!

Issue #895, opened 2019-05-15
Unify build tool integrations with Metals


jvican:
## State of the art
  
Integrating with Metals requires several changes in any build to get the full IDE experience (find all references, go to definition, rename, etc). These include:
  
1. Adding semanticdb to every project definition and configuring it based on its Scala version. Dealing with semanticdb compiler plugin options such as target and source roots is also necessary.
2. Disabling `-Xfatal-warnings` so that warnings turned as errors don't cripple IDE features and enabling `-Yrangepos`.
3. Setting `bloopExportJarClassifiers` to `true` so that library dependencies have source artifacts that "Go to Definition" can access.
  
These changes are done per build tool. To support sbt, Metals has an sbt plugin that depends on `sbt-bloop` and configures the options above. To support Maven and Gradle, we are inlining these modifications in the plugin sources instead.
  
However, this has several disadvantages. Sbt is the best supported build tool and the "Import Build" experience is still:
  
1. Slow. A plugin is added to the global sbt directory which cleans previous caches and forces a new reload every time a build is imported (which affects sbt shells running in users' terminals).
1. Fragile. Adding a global sbt plugin creates problems around cache cleaning, corrupted settings and conflicts with other sbt plugins.
  
To those disadvantages, we need to add the complexity of modifying scalac options of build targets in build tools such as Maven and mill, where no easy way to modify the build settings is provided.
  
## Proposed solution
  
Me and @olafurpg have been working on an proposal to improve on the status quo and provide a better built-in experience, while eliminating the above drawbacks.
  
The idea is to centralize the modifications required by Metals in bloop itself instead of spreading them across all plugins. The solution will work as follows:
  
1. Metals adds the sbt-bloop plugin to `project/metals.sbt` and gitignores it.
2. Metals triggers `sbt bloopInstall`, the whole build is exported.
3. Metals connects to Bloop via BSP and communicates which semanticdb jars it should use for every supported Scala version.
4. Bloop will take those semanticdb jars and apply the build target modifications to its in-memory representation. It's important these changes are not written back to the .bloop configuration files, but that instead we create some infrastructure to apply these changes every time there's a change in a configuration file under `.bloop/`.
5. Bloop will save the choice for semanticdb artifacts in a workspace settings file under `.bloop`. Next time that the server is restarted, bloop will detect that setting is set and transform the build targets appropriately. This avoids invalidating compilation caches.
  
### Advantages
  
[...]

PR #942, opened 2019-06-25
Unify build tool integrations with Metals

tgodzik:
Add an additional mechanism to save information about the SemanticDB plugin version for Metals. That information is then used to add scala options to compiler instances for the SemanticDB plugin.

Fixes https://github.com/scalacenter/bloop/issues/895

@jvican Is this what you meant? I still need figure out some tests and a way to simulate failing fatal warnings, although the current solution works the same as the one in this PR.


tgodzik:
After talking with @jvican I think what we should do is:

1. Create a separate module something like integration-shared 
2. Add functionality to fetch SemanticDB plugin via Coursier and save the paths to settings file - based on Scala version and SemanticDB version
3. Depend on that shared module in all integrations and just invoke for each project's scala version - that will save all needed paths to the settings path.
4. The mapper inside bloop would read that settings file and use the appropriate SemanticDB plugin path

We would need to add an additional parameter with SemanticDB version to all integrations (we don't have that while running `bloopInstall`), but that shouldn't be a problem.

Just a question to @jvican : is there a reason that we don't use Scala 2.12 in all of the integrations? At least in Gradle and Maven I don't see any reason to do that. And Metals doesn't work with 2.10 modules :P

jvican:
I've given this a lot more thought and I think this PR is on the right track. Implementing specific semanticdb support in every build as I proposed in our call @tgodzik has the advantage that everything that requires downloading is done by the build tool but two big disadvantages:

1. By default, we enable semanticdb and we're exporting a build that is different from that in the build tool. That should be fine in theory but it has some compiler performance implications, especially for people using Bloop with IntelliJ.
2. It requires a lot more work. To be done safely, we need to shade ourselves the coursier dependency we pull in and we need to repeat this across all build tools.
  
By centralizing the download and setup in the server, we have two key advantages:
  
1. Faster to implement and reason about.
1. Build tools don't need to add support for semanticdb and apply the rest of changes in the build to play nicely with Metals.
2. By default, if Metals is not used, we don't experience the compiler performance overhead added by Semanticdb (important for CI workflows and IntelliJ users).
  
Therefore, we only need to tweak how this PR is implemented internally, I'll leave  few code comments after this long comment.

frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -6,12 +6,39 @@ import bloop.io.AbsolutePath
...
 import bloop.io.ByteHasher
 import monix.eval.Task
+import io.circe.Json
+import bloop.data.SettingsMapper
+import bloop.config.Config
 
 object BuildLoader {
 
+  /** File to store Metals specific settings*/
+  val settingsFileName = "settings.json"
```
jvican: `bloop.settings.json` :smile:

frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -6,12 +6,39 @@ import bloop.io.AbsolutePath
...
   private[bloop] final val JsonFilePattern: String = "glob:**.json"
 
+  /**
+   * Gets information from additional data in the initialisation request to determine the settings for SemanticDB plugin
+   *
+   * @param data       Additional data from initialise request to interpret.
+   * @param configDir  Main bloop configuration directory.
+   * @param logger     The logger that collects messages about saving the settings file.
+   */
+  def writeSettingsFile(data: Option[Json], configDir: AbsolutePath, logger: Logger) : Unit = {
```
jvican: I propose we move this to the file where we define settings (`SettingsMapper`) so that we don't clutter this API. Also, it might be a good idea to move the logic reading/writing this file to the caller of `BuildLoader` and add a `settings: Settings` field to `Build`. Therefore, all the build loader does is only loading the build instead of handling the settings :smile:

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -213,6 +213,8 @@ object Config {
       modules: List[Module]
   )
 
+  case class Settings(semanticDBVersion: String)
```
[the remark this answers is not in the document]
jvican: We might actually want to leave the `semanticdbVersion` field here but make it an `Option`. This could be a escape hatch for users that know what they are doing and want to force a concrete version, otherwise we pick `latest.release` if `semanticdbVersion` is missing.

tgodzik:
@jvican Thanks for the comments, I just about starting to work on it, since I spent some time on Metal's issues.

frontend/src/main/scala/bloop/data/SettingsMapper.scala
```diff
@@ -0,0 +1,106 @@
...
+      logger.debug(s"Loading workspace settings from ${BuildLoader.settingsFileName}")(
+        DebugFilter.All
+      )
+      val contents = new String(bytes, StandardCharsets.UTF_8)
+      parser.parse(contents) match {
+        case Left(failure) => throw failure
+        case Right(json) => Some(fromJson(json))
+      }
+    } else {
+      None
```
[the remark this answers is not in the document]
tgodzik: I will do it in another place, since we here if the file is not present it might mean we are actually running it not from Metals and we don't want to add SemanticDB plugin in that case

frontend/src/main/scala/bloop/data/SettingsMapper.scala
```diff
@@ -0,0 +1,106 @@
...
+import scala.util.Try
+
+object SettingsMapper {
+
+  /**
+    * @param settingsPath Path to the json settings file.
+    * @param logger The logger that collects messages about loading the settings.
+    * @return Settings if the file exists.
+    */
+  def fromFile(settingsPath: AbsolutePath, logger: Logger): Option[Settings] = {
```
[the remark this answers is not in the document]
tgodzik: Don't want to remove Settings always , because `None` will mean we don't want to add SemanticDB plugin. Will rename to apply.

frontend/src/main/scala/bloop/data/SettingsMapper.scala
```diff
@@ -0,0 +1,106 @@
...
+    */
+  def transformProject(
+      settings: Config.Settings,
+      project: Project,
+      logger: Logger
+  ): Project = {
+
+    def findSemanticDBPlugin(scalaVersion: String) =
+      Try {
+        DependencyResolution
```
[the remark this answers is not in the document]
tgodzik: @jvican Do you mean we should use ZincComponentManager directly? And I think this is in conflict to the `latest.release`, since we can't vache that reliably.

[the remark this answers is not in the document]
jvican: > Do you mean we should use ZincComponentManager directly?

We should probably use our own component manager only for semanticdb plugins. It'll be very similar to ZincComponentnManager.
  
> And I think this is in conflict to the latest.release, since we can't cache that reliably.

Why not?

frontend/src/main/scala/bloop/data/SettingsMapper.scala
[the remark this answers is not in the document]
jvican: If you rename to apply, let's return `Settings`, it's the convention. I propose instead of checking for `None`, you check for `settings == DefaultSettings`. It's effectively the same thing and IMO makes sense.

frontend/src/main/scala/bloop/data/SettingsMapper.scala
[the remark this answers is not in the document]
tgodzik: We don't know what is `latest.release` unless we resolve it . The other way is to never cache `latest.release`, so I guess it should be fine.

frontend/src/main/scala/bloop/data/SettingsMapper.scala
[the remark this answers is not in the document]
tgodzik: Not sure I understand, what would `settings == DefaultSettings` give us? `None` would mean that we don't have any settings, so we shouldn't apply them. If default settings was just `Settings(None)` then how would it be different from `Settings(None)` that means we should get `release.latest`?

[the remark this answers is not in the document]
jvican: Nevermind :smile: After having a look at the code again, I propose we keep it as is. `Settings.fromFile` is good. I was just mentioning that the convention I use for `apply` is that it always returns the same type of the companion class, returning `Option[Settings]` looks weird.

frontend/src/main/scala/bloop/data/SettingsMapper.scala
```diff
@@ -0,0 +1,106 @@
...
+      case Right(settings) => settings
+      case Left(failure) => throw failure
+    }
+  }
+
+    /**
+    * @param target Path to where the json settings file should be saved.
+    * @param settings The settings to save.
+    */
+  def save(target: AbsolutePath, settings: Settings): Unit = {
```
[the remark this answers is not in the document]
tgodzik: Added Either, but in most places if there is an error, we just ignore it - since there is not much we can do if writing fails. We could just log an error, does it make sense?

frontend/src/main/scala/bloop/engine/SemanticDBCache.scala
```diff
@@ -0,0 +1,59 @@
+package bloop.engine
```
tgodzik: @jvican  Still need to tidy up the PR and fix one test, but what do you think about `SemanticDBCache`? Is this something you had in mind?

frontend/src/it/scala/bloop/CommunityBuild.scala
```diff
@@ -88,8 +88,8 @@ abstract class CommunityBuild(val buildpressHomeDir: AbsolutePath) {
 
   def loadStateForBuild(configDirectory: AbsolutePath, logger: Logger): State = {
     assert(configDirectory.exists, "Does not exist: " + configDirectory)
-    val loadedProjects = BuildLoader.loadSynchronously(configDirectory, logger)
-    val build = Build(configDirectory, loadedProjects)
+    val (loadedProjects, buildSettings) = BuildLoader.loadSynchronously(configDirectory, logger)
```
marek1840: Just nitpicking, but can we maintain here the order of "buildSettings, loadedProjects" ? It will then correspond to the `Build.apply` method

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -180,28 +181,45 @@ final class BloopBspServices(
...
-            compileProvider = Some(BloopBspServices.DefaultCompileProvider),
-            testProvider = Some(BloopBspServices.DefaultTestProvider),
-            runProvider = Some(BloopBspServices.DefaultRunProvider),
-            inverseSourcesProvider = Some(true),
-            dependencySourcesProvider = Some(true),
-            resourcesProvider = Some(true),
-            buildTargetChangedProvider = Some(false)
-          ),
-          None
+    /*Metals*/
```
marek1840: Please, elaborate ;)

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -180,28 +181,45 @@ final class BloopBspServices(
...
+            val settings = BuildSettings.fromJson(json)
+            val savedSettings = BuildSettings.fromFile(configDir, callSiteState.logger)
+            if (savedSettings.isEmpty || savedSettings.exists(_ != settings)) {
+              BuildSettings.write(configDir, settings)
+            }
+          case _ =>
+        }
+      }
+    }
+    /*------*/
```
marek1840: Remove? 
```suggestion

```

frontend/src/main/scala/bloop/data/BuildSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+import io.circe.derivation._
+import java.nio.file.Path
+import scala.util.Failure
+import scala.util.Success
+
+case class BuildSettings(semanticDBVersion: Option[String])
+
+object BuildSettings {
+
+  /** File to store Metals specific settings*/
```
marek1840: Maybe we could call the var `metalsSettingsFileName` and avoid the comment?

frontend/src/main/scala/bloop/data/BuildSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+  def fromFile(configPath: AbsolutePath, logger: Logger): Option[BuildSettings] = {
+    val settingsPath = configPath.resolve(settingsFileName)
+    if (settingsPath.isFile) {
+      val bytes = Files.readAllBytes(settingsPath.underlying)
+      logger.debug(s"Loading workspace settings from $settingsFileName")(
+        DebugFilter.All
+      )
+      val contents = new String(bytes, StandardCharsets.UTF_8)
+      parser.parse(contents) match {
+        case Left(failure) => throw failure
```
marek1840: why not None? or even better, chaning the return type to a `Try`?

frontend/src/main/scala/bloop/data/BuildSettings.scala
tgodzik: This might be needed for something more - I thinkw e can keep this name

frontend/src/main/scala/bloop/engine/Build.scala
```diff
@@ -32,8 +34,11 @@ final case class Build private (
     val files = projects.iterator.map(p => p.origin.toAttributedPath).toSet
     val newFiles = BuildLoader.readConfigurationFilesInBase(origin, logger).toSet
 
+    val loadedSettings = WorkspaceSettings.fromFile(origin, logger)
+    val sameSettings = loadedSettings == settings
+
     // This is the fast path to short circuit quickly if they are the same
-    if (newFiles == files) Task.now(Build.ReturnPreviousState)
+    if (newFiles == files && sameSettings) Task.now(Build.ReturnPreviousState)
```
jvican: Can we instead just check that the semanticdb version is the same in both? This way, when we add new fields to settings that do not require transformations in the project, we will not be reloading needlessly.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -184,8 +194,23 @@ final class BloopBspServices(
...
     )
 
-    reloadState(configDir, client).map { state =>
-      callSiteState.logger.info("request received: build/initialize")
+    /* Metals specific settings that are used to store the
+     * SemanticDB version that will later be applied to all
+     * projects in the workspace. If the client is Metals but 
+     * the version is not specified we use `latest.release`
+     */
+    val metalsSettings =
```
jvican: Nits: move `else` clause to top and wrap this in curly braces

frontend/src/main/scala/bloop/data/WorkspaceSettings.scala
```diff
@@ -0,0 +1,71 @@
...
+import bloop.DependencyResolution
+import scala.util.Try
+import io.circe.ObjectEncoder
+import io.circe.Decoder
+import io.circe.derivation._
+import java.nio.file.Path
+import scala.util.Failure
+import scala.util.Success
+
+case class WorkspaceSettings(semanticDBVersion: String)
```
jvican: I'll add docs here and in several other places when we're doing with the review

frontend/src/main/scala/bloop/engine/Build.scala
```diff
@@ -70,7 +75,8 @@ object Build {
   case object ReturnPreviousState extends ReloadAction
   case class UpdateState(
       createdOrModified: List[ReadConfiguration],
-      deleted: List[AbsolutePath]
+      deleted: List[AbsolutePath],
+      settingsChanged: Option[WorkspaceSettings]
```
jvican: `changedSetings`?

shared/src/main/scala/bloop/io/AbsolutePath.scala
```diff
@@ -18,6 +19,10 @@ final class AbsolutePath private (val underlying: Path) extends AnyVal {
   def resolve(other: String): AbsolutePath = AbsolutePath(underlying.resolve(other))(this)
   def getParent: AbsolutePath = AbsolutePath(underlying.getParent)
 
+  def list: List[AbsolutePath] =
+    Files.list(underlying).iterator().asScala.toList.map(AbsolutePath.apply)
```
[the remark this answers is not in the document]
tgodzik: Actually that was needed without the ZincComponentManager, I will remove that.

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+        val classpath = compiledState.client.getUniqueClassesDirFor(
+          compiledState.build.getProjectFor(projectName).get
+        )
+        val semanticDBFile =
+          classpath.resolve("META-INF/semanticdb/src/main/scala/Foo.scala.semanticdb")
+        assert(semanticDBFile.exists)
+      }
+    }
+  }
+
```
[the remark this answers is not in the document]
tgodzik: I was thinking about adding that suite, but forgot :/ 

I don't think Metals needs to be sent the information, we can inspect that ourselves and show in the doctor. We already check if the plugin is added.

frontend/src/main/scala/bloop/engine/SemanticDBCache.scala
```diff
@@ -0,0 +1,75 @@
...
+    }.toOption.flatten
+  }
+
+  private def resolveCache(
+      organization: String,
+      module: String,
+      version: String,
+      logger: Logger,
+      additionalRepositories: Seq[Repository] = Nil
+  )(implicit ec: scala.concurrent.ExecutionContext): Option[AbsolutePath] = {
```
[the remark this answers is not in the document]
tgodzik: It actually is using ioScheduler only, but I will make it more explicit.

frontend/src/main/scala/bloop/engine/SemanticDBCache.scala
```diff
@@ -0,0 +1,75 @@
...
+    if (version == latestRelease) {
+      getFromResolution
+    } else {
+      val semanticDBId = s"$organization.$module.$version"
+      Try(zincComponentManager.file(semanticDBId)(IfMissing.Fail)) match {
+        case Failure(exception) =>
+          val resolved = getFromResolution
+          resolved match {
+            case None =>
+              logger.warn("Could not find semanticDB version:\n" + exception.getMessage())
```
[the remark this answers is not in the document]
tgodzik: We already report if the semanticDBplugin is available in each target and whether it's supported. We take that information from targets reported by Bloop, so we would show the correct warning - we might need to fix it up a bit and show up more, but I don't think we need to forward anything new from Bloop.

For spurious issues we should for sure try to resolve again and I will change the warning in Metals to say that we couldn't resolve the plugin although it should be available when it's not available even after retries.

frontend/src/main/scala/bloop/engine/SemanticDBCache.scala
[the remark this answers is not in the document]
jvican: Yeah, I'm suggesting to not pass the context as a parameter and instead use io scheduler inside `resolveFromCache`.

backend/src/main/scala/bloop/DependencyResolution.scala
```diff
@@ -62,6 +63,7 @@ object DependencyResolution {
     }
     val fetch = ResolutionProcess.fetch(repositories, Cache.default.fetch)
     val resolution = start.process.run(fetch).unsafeRun()
+    if (shouldReportErrors) reportErrors(resolution, logger)
```
tgodzik: Tested with Metals and now we will show a warning like:

warning

tgodzik:
Added some more tests and tested it with Metals - seems to work fine, but some further testing might be needed.

frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -90,6 +105,89 @@ object BuildLoader {
...
+        val optionsSet = project.scalacOptions
+        val containsSemanticDB = optionsSet.find(
+          setting => setting.contains("-Xplugin") && setting.contains("semanticdb-scalac")
+        )
+        val containsYrangepos = optionsSet.find(_.contains("-Yrangepos"))
+        val semanticDBAdded = if (containsSemanticDB.isDefined) {
+          logger.info(s"SemanticDB plugin already added: ${containsSemanticDB.get}")
+          optionsSet
+        } else {
+          val workspaceDir = project.origin.path.getParent.getParent
```
tgodzik: @jvican Not sure if that is the best way of getting the workspace root directory.

frontend/src/main/scala/bloop/data/LoadedProject.scala
```diff
@@ -0,0 +1,30 @@
+package bloop.data
+
+sealed trait LoadedProject {
```
jvican: @tgodzik This is a new key abstraction I have added to our build load infrastructure so that we can distinguish internally whether a project has been configured or it hasn't. Even if semanticdb is not supported in a project, we still consider it configured because we always add the range positions, so in `checkForChange` we detect configured projects that don't have semanticdb enabled and we retry the resolution when `newSettings` are passed.

frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -199,4 +200,70 @@ object Project {
...
+   * @param logger The logger responsible of tracking any transformation-related event.
+   * @return Either the same project as before or the transformed project.
+   */
+  def enableMetalsSettings(
+      project: Project,
+      semanticDBPlugin: Option[AbsolutePath],
+      workspaceDir: AbsolutePath,
+      logger: Logger
+  ): Project = {
+    def enableSemanticDB(options: List[String], pluginPath: AbsolutePath): List[String] = {
```
jvican: Instead of adding semanticdb and range pos in one pass, we do it in two different methods so that by default we add range pos and only if we have a plugin path we enable the semanticdb.

frontend/src/main/scala/bloop/data/WorkspaceSettings.scala
```diff
@@ -0,0 +1,116 @@
...
+)
+
+object WorkspaceSettings {
+
+  /** Represents the supported changes in the workspace. */
+  sealed trait DetectedChange
+  final case object SemanticDBVersionChange extends DetectedChange
+
+  /** File name to store Metals specific settings*/
+  private[bloop] val settingsFileName = RelativePath("bloop.settings.json")
```
jvican: Nit: i relative path instead of a string is the most idiomatic way of writing file names :smile:

frontend/src/main/scala/bloop/data/WorkspaceSettings.scala
```diff
@@ -0,0 +1,98 @@
...
+        case Right(json) => Option(fromJson(json))
+      }
+    }
+  }
+
+  def writeToFile(
+      configDir: AbsolutePath,
+      settings: WorkspaceSettings,
+      logger: Logger
+  ): AbsolutePath = {
```
jvican: Note how we're making this method throw just as `readFile` does. Inside `checkForChange`, we catch it, log the exception and warn the user that something really wrong happened. This could happen, for example, if the workspace settings file didn't have write permission for bloop. cI wish I had the time to go through all of the weird denied access exception we could get and fix them but so far this is the best we can do with the limited time we have.

frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -22,56 +27,136 @@ object BuildLoader {
...
    *
-   * @param configDir The base directory from which to load the projects.
-   * @param logger The logger that collects messages about project loading.
-   * @return The list of loaded projects.
+   * Some of the logic of this method has a duplicate version in
+   * `loadSynchronously` that doesn't use `Task` to resolve SemanticDB plugins
+   * in parallel and apply the Metals projects. Any change here **must** also
+   * be propagated there.
    */
-  def load(
```
jvican: The ordinary load that loads everything isn't implemented in terms of this but in terms of `checkForChange`, so that's why I removed this method.

frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -22,56 +27,136 @@ object BuildLoader {
...
-      .flatMap(fs => loadBuildFromConfigurationFiles(configDir, fs, logger))
+    val enableMetalsInProjectsTask = groupedProjectsPerScalaVersion.toList.map {
+      case (scalaVersion, projects) =>
+        def enableMetalsTask(plugin: Option[AbsolutePath]) = {
+          Task(
+            projects.map(p => Project.enableMetalsSettings(p, configDir, plugin, logger) -> Some(p))
+          )
+        }
+
+        // Recognize 2.12.8-abdcddd as supported if 2.12.8 exists in supported versions
```
jvican: Trick so that people using custom versions of Scala locally can have this working. This can happen sometimes to me :)

frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -80,16 +165,60 @@ object BuildLoader {
...
       Build.ReadConfiguration(Origin(ap, hash), bytes)
     }
 
     logger.debug(s"Loading ${configFiles.length} projects from '${configDir.syntax}'...")(
-      DebugFilter.Compilation
+      DebugFilter.All
     )
-    configFiles.map(f => Project.fromBytesAndOrigin(f.bytes, f.origin, logger))
+
+    configFiles.map { f =>
```
jvican: I don't like this deduplicated logic but I don't mind living with it. There's no way to reuse the logic in `loadBuildIncrementally` because everything there runs in parallel and it's implemented in terms of `Task`.

frontend/src/main/scala/bloop/engine/Feedback.scala
```diff
@@ -93,6 +93,13 @@ object Feedback {
   def unknownHostName(host: String): String =
     s"Host name '$host' could not be either parsed or resolved"
 
+  def skippedUnsupportedScalaMetals(scalaVersion: String): String =
+    s"Skipped configuration of SemanticDB in unsupported $scalaVersion projects"
+  def configuredMetalsProjects(projects: Traversable[Project]): String =
+    s"Configured SemanticDB in projects ${Project.pprint(projects)}"
+  def failedMetalsConfiguration(version: String, cause: String): String =
+    s"Stopped configuration of SemanticDB in Scala $version projects: $cause"
```
jvican: I usually add user-facing messages in `Feedback` objects so that I can easily know where to change a log message if I don't like it and so that it removes the distraction of reading what the output to the users is from the call-sites.

frontend/src/main/scala/bloop/engine/caches/SemanticDBCache.scala
```diff
@@ -0,0 +1,72 @@
...
+            case Some(pluginPath) => Right(pluginPath)
+            case None =>
+              Left(
+                s"Missing semanticdb plugin in resolved jars ${paths.map(_.syntax).mkString(",")}"
+              )
+          }
+      }
+    }
+
+    if (version == "latest.release") {
```
jvican: Mostly left this as you made it @tgodzik with the exception that if the version is `latest.release` we only resolve it once per instance of the bloop server.

frontend/src/test/scala/bloop/BuildLoaderSpec.scala
```diff
@@ -7,17 +7,129 @@ import bloop.engine.{Build, BuildLoader, State}
...
+          sys.error(s"Expected return updated state, got previous state")
+        case action: Build.UpdateState =>
+          assert(action.createdOrModified.size == 2)
+          assert(action.deleted.size == 0)
+          assert(action.invalidated.size == 0)
+          assert(action.settingsForReload == Some(sameSettings))
+      }
+  }
+
+  testLoad("reload if two file contents changed with same settings", Some(sameSettings)) {
```
jvican: I added a few more tests here.I **really** appreciate that you were the first one to add tons of tests here, I had almost forgotten about this test suite :sweat_smile:

frontend/src/test/scala/bloop/BuildLoaderSpec.scala
```diff
@@ -7,17 +7,129 @@ import bloop.engine.{Build, BuildLoader, State}
...
+
+  testLoad("reload if new settings are added and two file contents changed", Some(sameSettings)) {
+    (testBuild, logger) =>
+      changeHashOfRandomFiles(testBuild, 2)
+      val newSettings = WorkspaceSettings("4.1.11", List(BuildInfo.scalaVersion))
+      testBuild.state.build.checkForChange(Some(newSettings), logger).map {
+        case Build.ReturnPreviousState =>
+          sys.error(s"Expected return updated state, got previous state")
+        case action: Build.UpdateState =>
+          assert(action.deleted.size == 0)
```
jvican: I made some tests more specific

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
```diff
@@ -0,0 +1,399 @@
...
+      val semanticdbVersion = "4.2.0"
+      val extraParams = BloopExtraBuildParams(
+        clientClassesRootDir = None,
+        semanticdbVersion = Some(semanticdbVersion),
+        supportedScalaVersions = List(testedScalaVersion)
+      )
+
+      loadBspState(workspace, projects, logger, "Metals", bloopExtraParams = extraParams) { state =>
+        assertNoDiff(logger.warnings.mkString(System.lineSeparator), "")
+        assertNoDiffInSettingsFile(
```
jvican: In general, I really like testing against strings and showing diffs to know what's wrong opposed to doing concrete assertions. This way, it's obvious when reading the tests what the output of workspace settings should be. Just a FYI :smile:

integrations/gradle-bloop/src/test/scala/bloop/integrations/gradle/ConfigGenerationSuite.scala
```diff
@@ -22,11 +22,11 @@ import bloop.engine.BuildLoader
 
 import scala.collection.JavaConverters._
 
-class ConfigGenerationSuite481 extends ConfigGenerationSuite{
+class ConfigGenerationSuite481 extends ConfigGenerationSuite {
```
jvican: Looks like these sources were not formatted before and my commit formatted them instead

frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -199,4 +202,67 @@ object Project {
...
+      val hasYrangepos = options.exists(_.contains("-Yrangepos"))
+      if (hasYrangepos) options else options :+ "-Yrangepos"
+    }
+
+    val projectWithRangePositions =
+      project.copy(scalacOptions = enableRangePositions(project.scalacOptions))
+    semanticDBPlugin match {
+      case None => projectWithRangePositions
+      case Some(pluginPath) =>
+        // Recognize 2.12.8-abdcddd as supported if 2.12.8 exists in supported versions
```
tgodzik: We don't check it here, is the comment needed?

frontend/src/main/scala/bloop/engine/Build.scala
```diff
@@ -1,77 +1,237 @@
...
+  def findUpdateSettingsAction(
+      currentSettings: Option[WorkspaceSettings],
+      newSettings: Option[WorkspaceSettings]
+  ): Build.UpdateSettingsAction = {
+    (currentSettings, newSettings) match {
+      case (Some(currentSettings), Some(newSettings))
+          if currentSettings.semanticDBVersion != newSettings.semanticDBVersion =>
+        Build.ForceReload(newSettings, List(WorkspaceSettings.SemanticDBVersionChange))
+      case (Some(_), Some(newSettings)) => Build.AvoidReload(Some(newSettings))
+      case (None, Some(newSettings)) => Build.AvoidReload(Some(newSettings))
```
tgodzik: Shouldn't we force it here?

frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -80,16 +165,60 @@ object BuildLoader {
...
+            case Some(instance) =>
+              def enableMetals(plugin: Option[AbsolutePath]) = {
+                LoadedProject.ConfiguredProject(
+                  Project.enableMetalsSettings(project, configDir, plugin, logger),
+                  project,
+                  settings
+                )
+              }
+
+              val scalaVersion = instance.version
```
tgodzik: Since all paths end with metalsEnable, can't we put this into a function at least:
```
def enablePlugin[T](settingsEnable: Option[Path] => T): T 
```
a bit less of a duplciated code.

frontend/src/main/scala/bloop/engine/Build.scala
jvican: Good catch!

backend/src/main/scala/bloop/DependencyResolution.scala
```diff
@@ -15,72 +17,65 @@ object DependencyResolution {
...
-    }
-    if (fileErrors.isEmpty) {
-      localArtifacts.collect { case (_, Right(f)) => AbsolutePath(f.toPath) }.toArray
-    } else {
-      val moduleInfo = s"$organization:$module:$version"
-      val prettyFileErrors = fileErrors.map(_.describe).mkString(System.lineSeparator)
-      sys.error(
-        s"Resolution of module $moduleInfo failed with:${System.lineSeparator}${prettyFileErrors}"
-      )
+    try Right(fetch.run().map(f => AbsolutePath(f.toPath)).toArray)
```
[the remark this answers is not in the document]
jvican: `Try` catches everything, I now that coursier can only throw `CoursierError` so the try-catch only catches that

tgodzik:
LGTM

jvican:
Et voilà

PR #944, opened 2019-06-26
Implement the Debug Adapter Protocol

marek1840:


frontend/src/main/scala/bloop/dap/JavaDebugInterface.scala
```diff
@@ -0,0 +1,60 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+object JavaDebugInterface {
+
+  /**
+   * Checks if the Java Debug Interface is either already on classpath or if it can be automatically loaded.
+   */
+  def isAvailable: Try[_] = {
+    def loadJdi: Try[_] = Try(Class.forName("com.sun.jdi.Value"))
```
tgodzik: Can we make those `Try`s have an exact type?

frontend/src/main/scala/bloop/dap/JavaDebugInterface.scala
```diff
@@ -0,0 +1,60 @@
...
+    private val windowsJrePath = {
+      val version = javaHome.underlying.getFileName.toString.drop("jre".length)
+      javaHome.getParent.resolve("jdk" + version).resolve(tools)
+    }
+
+    val jar: Try[AbsolutePath] = {
+      val paths = List(jdkPath, jrePath, windowsJrePath)
+      paths.find(_.exists) match {
+        case None =>
+          val error = s"Could not find 'tools.jar'. Searched in: $paths"
```
tgodzik: print paths line by line?

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -0,0 +1,163 @@
...
+    val logger = new DebugSessionLogger(debugSession)
+    val task = underlying.runMain(cwd, mainClass, args, jvmOptions, logger, opts, extraClasspath)
+
+    task
+  }
+
+  private def findFreeSocketAddress: InetSocketAddress = {
+    val server = new ServerSocket(0)
+    server.close()
+    server.getLocalSocketAddress
```
tgodzik: So we open a socket and close it to find a free address? I don't think it's the best solution. Whathever runs should find and reserve that socket itself. You can have a race here, no?

frontend/src/main/scala/bloop/dap/DebugSession.scala
```diff
@@ -0,0 +1,182 @@
...
+      case _ => super.sendResponse(response)
+    }
+  }
+
+  override def sendEvent(event: Events.DebugEvent): Unit = {
+    if (event.`type` == "exited") {
+      debuggeeExited.success(())
+    }
+
+    super.sendEvent(event)
```
jvican: I moved this line after the `if` so that if this throws, for any reason, it doesn't affect the completion of `debuggeeExited`. Also, it looks like there might be a race condition here and we might never be completing `debuggeeExited` because the exited event is not generated if the socket was already closed. I haven't confirmed if this race condition indeed happens, it's just an intuition. What are your thoughts?

frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
```diff
@@ -0,0 +1,51 @@
...
+
+object DebuggeeRunner {
+  def forMainClass(
+      projects: Seq[Project],
+      mainClass: ScalaMainClass,
+      state: State
+  ): Either[String, DebuggeeRunner] = {
+    val project = projects.head
+    project.platform match {
+      case jvm: Platform.Jvm =>
```
jvican: Here, we should also be checking if the associated java home to the project is a JDK or not. Most of the times, there is no associated java home, but if there is bloop should be respecting it and using it to fork. There have been several people in Gitter that have complained this is not working correctly, I'm planning to look into it in another PR, but we should nonetheless check those java homes here.

frontend/src/main/scala/bloop/dap/JavaDebugInterface.scala
```diff
@@ -0,0 +1,60 @@
...
+
+object JavaDebugInterface {
+
+  /**
+   * Checks if the Java Debug Interface is either already on classpath or if it can be automatically loaded.
+   */
+  def isAvailable: Try[_] = {
+    def loadJdi: Try[_] = Try(Class.forName("com.sun.jdi.Value"))
+
+    def addToClasspath(jar: AbsolutePath): Unit = {
```
jvican: As discussed privately, I think all of this should be unnecessary because JDI is by default always accessible if the host JVM is a JDK. If that's not the case, I'd assume there's a problem with the installed JDK. I'm a little bit afraid of doing this `addURL` magic and I'm willing to bet this is not compatible with Java 9 and Java 11 where JDI is a module instead. The best is probably to remove all of this unnecesary logic and provide excellent error messages to users so that they don't feel lost if their Java machine is not a JDK.

frontend/src/main/scala/bloop/dap/DebugSession.scala
```diff
@@ -0,0 +1,182 @@
...
+import bloop.logging.Logger
+
+/**
+ *  This debug adapter maintains the lifecycle of the debuggee in separation from JDI.
+ *  The debuggee is started/closed together with the session.
+ *
+ *  This approach makes it necessary to handle the "launch" requests as the "attach" ones.
+ *  The JDI address of the debuggee is obtained through the [[DebugSessionLogger]]
+ */
+final class DebugSession(
```
jvican: I really like the `DebugSession` and `DebugServer` abstractions and their implementation, good work 👍

frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
```diff
@@ -0,0 +1,51 @@
...
+    project: Project,
+    mainClass: ScalaMainClass,
+    env: JavaEnv,
+    state0: State
+) extends DebuggeeRunner {
+  def run(logger: DebugSessionLogger): Task[Unit] = {
+    val stateForDebug = state0.copy(logger = logger)
+    val workingDir = state0.commonOptions.workingPath
+    val runState = Tasks.runJVM(
+      stateForDebug,
```
jvican: Let's rename `state0` to `state` and use `state` here, as `stateForDebug == state0`.

frontend/src/main/scala/bloop/dap/DebugSessionLogger.scala
```diff
@@ -0,0 +1,63 @@
...
+    listener: InetSocketAddress => Unit,
+    underlying: Logger
+) extends Logger {
+  @volatile private var initialized = false
+  override val name: String = s"${underlying.name}-debug"
+
+  override def isVerbose: Boolean = underlying.isVerbose
+  override def trace(t: Throwable): Unit = underlying.trace(t)
+  override def printDebug(msg: String): Unit = underlying.debug(msg)(DebugFilter.All)
+  override def warn(msg: String): Unit = underlying.warn(msg)
```
jvican: In which scenarios could we be calling `warn`, `trace` or `debug` on `DebugSessionLogger`? Or are we somehow ensured these will never be called? Currently we pass them to the underlying logger but we should make a good effort to give clients as much feedback as possible.

frontend/src/test/scala/bloop/bsp/BspBaseSuite.scala
```diff
@@ -303,11 +336,7 @@ abstract class BspBaseSuite extends BaseSuite with BspClientTest {
     }
   }
 
-  // We limit the scheduler on purpose so that we don't have any thread leak.
-  private val bspDefaultScheduler: Scheduler = Scheduler(
-    java.util.concurrent.Executors.newFixedThreadPool(4),
-    ExecutionModel.AlwaysAsyncExecution
-  )
+  private val bspDefaultScheduler: Scheduler = TestSchedulers.async("bsp-default", threads = 8)
```
jvican: Why are we increasing from 4 to 8? We should be using the previous value so that the old useres of `bspDefaultScheduler` are not affected by this. I propose to create a `DebugBspBaseSuite` class that extends `BspBaseSuite` and that it creates its own schedulers for the debug session, as well as any other utility required for testing the debugger.

frontend/src/test/scala/bloop/dap/DebugAdapterConnection.scala
```diff
@@ -0,0 +1,82 @@
...
+private[dap] final class DebugAdapterConnection(val socket: Socket)(
+    implicit proxy: DebugSessionProxy
+) {
+  def initialize(): Task[Capabilities] = {
+    val arguments = new InitializeArguments()
+    Initialize(arguments)
+  }
+
+  def configurationDone(): Task[Unit] = {
+    ConfigurationDone(())
```
jvican: Is this instantiation sending a request and then waiting for a response? If that's the case, can we avoid these side effects and have a clearer API?

frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
```diff
@@ -0,0 +1,142 @@
...
+          } yield output
+        }
+
+        assertNoDiff(output, "Hello, World!\n")
+      }
+    }
+  }
+
+  // when the session detaches from the JVM, the JDI once again writes to the standard output
+  test("restarted session does not contain JDI output") {
```
jvican: This is the test case that I've mentioned before in the other comment and that currently fails.

frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
```diff
@@ -0,0 +1,142 @@
+package bloop.dap
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+
+import bloop.bsp.BspBaseSuite
+import bloop.cli.BspProtocol
+import bloop.logging.RecordingLogger
+import bloop.util.{TestProject, TestUtil}
+
+object DebugProtocolSpec extends BspBaseSuite {
```
jvican: These test definitely look slick 👌 Great work!

frontend/src/test/scala/bloop/dap/DebugSessionProxy.scala
```diff
@@ -0,0 +1,107 @@
...
+import monix.execution.{Ack, Scheduler}
+import monix.reactive.observables.ObservableLike.Operator
+import monix.reactive.observers.Subscriber
+import monix.reactive.{Observable, Observer}
+
+import scala.collection.mutable
+import scala.concurrent.{Future, Promise}
+import scala.meta.jsonrpc.{BaseProtocolMessage, MessageWriter}
+
+private[dap] final class DebugSessionProxy(
```
jvican: This looks correct, but could you explain a little bit more how all of this testing infrastructure works at a high-level with in-depth Scaladocs?

project/build.sbt
```diff
@@ -18,6 +18,7 @@ val `bloop-build` = project
     addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.6.2"),
     addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.1.0-M13-2"),
     addSbtPlugin("org.scalameta" % "sbt-mdoc" % "1.2.10"),
+    addSbtPlugin("org.scala-debugger" % "sbt-jdi-tools" % "1.1.1"),
```
jvican: 🤔 Why is this required?

frontend/src/main/scala/bloop/dap/JavaDebugInterface.scala
marek1840: OK, "works on my machine" so I will drop this functionality

frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
marek1840: We must inject our logger, so those states are not equal

frontend/src/main/scala/bloop/dap/DebugServer.scala
```diff
@@ -0,0 +1,79 @@
...
+    val closedServer = AtomicBoolean(false)
+    val listeningPromise = Promise[Option[URI]]()
+    val ongoingSession = MultiAssignmentCancelable()
+
+    def listen(serverSocket: ServerSocket): Task[Unit] = {
+      val listenAndServeClient = Task {
+        listeningPromise.trySuccess(Some(handle.uri))
+        val socket = serverSocket.accept()
+
+        DebugSession.open(socket, runner.run, logger, ioScheduler).flatMap { session =>
```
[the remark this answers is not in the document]
marek1840: we can remove it altogether when we no longer load the JDI dynamically

frontend/src/test/scala/bloop/dap/DebugAdapterConnection.scala
marek1840: You mean instantiation of `()` ? I don't believe I follow :(

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -416,6 +429,65 @@ final class BloopBspServices(
...
+          compileProjects(mappings, state, Nil).flatMap {
+            case (state, Left(error)) =>
+              Task.now((state, Left(error)))
+            case (state, Right(result)) if result.statusCode != bsp.StatusCode.Ok =>
+              Task.now(
+                (state, Left(JsonRpcResponse.internalError("Compilation not successful")))
+              )
+            case (state, Right(_)) =>
+              val projects = mappings.map(_._2)
+              inferDebuggeeRunner(projects, state) match {
```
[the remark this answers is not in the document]
jvican: > What is the best approach to check if we are using JDK?
  
Usually a JDK implies that local java compiler APIs are present. I'd detect a JDK if `Option(javax.tools.ToolProvider.getSystemJavaCompiler)` produces a non empty value.
  
Then, we can detect if the debug APIs are present or not by querying for `com.sun.jdi.Value`.
  
> Also there is a case when we are running on a JDK but the tools are not present (the vendor is not providing them). In that case it is better to say the Java Debug Interface is not provided.
  
I agree, well-thought.
  
I propose we add the logic to detect debug APIs and JDK in the companion of `JavaEnv` and that we also tweak `bloop about` to print whether the detect java home is indeed a JDK and debugging is supporting (something like adding `(JDK, debug enabled)`, `(JDK, debug disabled!)` and `(JRE)` at the end of the java home line).

frontend/src/main/scala/bloop/dap/DebugServer.scala
```diff
@@ -0,0 +1,79 @@
...
+     * which can happen when a restart is request and the client immediately
+     * connects without waiting for the other session to finish.
+     */
+    val handle = ServerHandle.Tcp(backlog = 1)
+
+    val closedServer = AtomicBoolean(false)
+    val listeningPromise = Promise[Option[URI]]()
+    val ongoingSession = MultiAssignmentCancelable()
+
+    def listen(serverSocket: ServerSocket): Task[Unit] = {
```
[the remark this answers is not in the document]
jvican: > Could you include an error? Does it contain a JDI output?
  
Didn't attach an error because it's just a timeout. Can you not reproduce it?
  
> Also, could you point me to those hanging issues? In the end, I might miss some.
  
I'll have another look in the second review round, but the most glaring ones were in `DebugSession` and `DebugServer`. Let's make sure that in these two abstractions, we have a clear distinction between: closing resources when everything exits successfully, cancelling and starting the logic, and that none of these race with each other.

frontend/src/test/scala/bloop/dap/DebugAdapterConnection.scala
jvican: `ConfigurationDone` returns a `Task` and inside that task it does lots of IO. From reading the code, I would have never guessed that the instantiation of a class would trigger the creation and delivery of a request and then the blocking on a response.
  
`Request` should not be doing all of that work inside `apply`. My suggestion is that we decouple clearly the creation of a request with its inputs and the time when that request is sent. These are two different responsibilities that merit different API endpoints.

project/build.sbt
jvican: All of our running/testing in projects that require debugging/JDK APIs is forked, which means we system process to the installed JDK. If I read the motivation of the plugin correctly, the plugin is not useful when doing forked execution, so I'd be inclined to remove it. Did you get errors without it?

frontend/src/main/scala/bloop/dap/DebugSession.scala
```diff
@@ -30,112 +29,119 @@ final class DebugSession(
...
-   * 1. starts the debuggee
-   * 2. begins listening for debug clients
+   * Schedules the start of the debugging session.
+   *
+   * For a session to start, two executions must happen independently in a
+   * non-blocking way: the debuggee process is started in the background and
+   * the DAP server starts listening to client requests in an IO thread.
    */
-  override def run(): Unit = {
+  def startDebuggeeAndServer(): Unit = {
```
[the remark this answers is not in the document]
jvican: If you're worried about people seeing either `run` or `startDebuggeeAndServer` and being confused about which one they should use, we can override `run`, redirect it to `startDebuggeeAndServer` and call the `super.run` inside it.
  
There are some tricks to avoid the invocation of `run` from the outside but they are overkill. The general sentiment driving this change is that we don't want people to be using run but its more precise `startDebuggeeAndServer` because `run` is a huge misnomer about what's happening here :smile:.

frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
jvican: Good catch, I misread this! Can we do instead `state0.copy(logger = debugLogger)`? I thought `logger` was `state0.logger` and that it was in scope, as that's usually the case in all of our implementation, so that's what confused me.

frontend/src/main/scala/bloop/dap/DebugSessionLogger.scala
jvican: > Maybe we should forward all of the log levels to the underlying in addition to sending some of them to the DAP-client? Just a thought.
  
That's a great thought, we should definitely be doing that.

frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
jvican: >  the JVM on which we run the main class should not, as long as it implements the JDI.
  
It'd be good if you can verify it, but I think you do need to have a JDK in both the host VM and the forked VM for debugging to work.

marek1840:
As @jvican mentioned:
> we could be leveraging BSP’s showMessage if something fails spectacularly during the DAP session with the client. This way, Metals is never out-of-the-loop

BSP spec:
https://github.com/scalacenter/bsp/blob/master/docs/bsp.md#show-message

example:
https://github.com/scalacenter/bloop/blob/38a83ab74f6526b04e0d137bba34d87384736fb9/frontend/src/main/scala/bloop/logging/BspServerLogger.scala#L54-L63

frontend/src/main/scala/bloop/dap/DebugSession.scala
```diff
@@ -0,0 +1,182 @@
...
+
+  /**
+   * Cancels the background debuggee process, the DAP server and closes the socket.
+   */
+  def cancel(): Unit = {
+    if (isCancelled.compareAndSet(false, true)) {
+      runningDebuggee.get.cancel()
+      Task
+        .fromFuture(debuggeeExited.future)
+        .map(_ => DebugSession.disconnectRequest(DebugSession.InternalRequestId))
```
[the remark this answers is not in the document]
jvican: I'm down with having the DAP server send all of those exited and terminated events on its own, but then we need to figure out several important aspects that are not clear to me yet. Here's a few questions:
  
* How does cancelling the debuggee affect the DAP server running in the background? If the VM the DAP server is attached to exits because we cancel it, what exception in received by the DAP server and how is it handled?
* Why do we need to send this disconnect request manually? Is the exit of the VM not forcing the delivery of exited and terminated events?
* Should we not try to get the DAP server to disconnect from the VM first and then cancel the debuggee when the exited/terminated events have been sent out? At the point we should also be able to close the socket imediately.
* Why are we also cancelling the debuggee before the cancellation logic runs and inside  the handling of the internal disconnect in `dispatchRequest`?
* Can we get rid of the delay and ensure that we clean up resources at the very moment where we cancel everything so that we can instantiate another debug session (in case of a restart) from a clean slate?
* The handling of `dispatchRequest` in `ProtocolServer` is special casing some VM disconnected errors. If we were to do something special, should we be trying to handle benign VM-related errors and return successful responses to the client (assuming we might not be doing that already)?

jvican:
Another remaining issue is what we should do with the JUL (java logger) dependency of java-debug. This is quite problematic and it doesn't look like there's an obvious solution. Currently it prints output to stdout and stderr and that's not good because we want to be in control of what gets to be shown to the user, we don't want java-debug to suddenly spit out exceptions to users' console outputs. The whole notion of global loggers doesn't play well with bloop, which has client-specific logging infrastructure.
  
So, what should we do? Here are a few ideas that come to mind:
  
1. Completely disable JUL by resetting all loggers. This is the "bazooka" option, not ideal because we could be missing out on key log events that could be useful to track down debugging issues by the user.
2. Set up a file where JUL spits out all of the debugging logic. We would not miss log events but we would need to teach users to look at those files whenever there's something wrong. There are two core disadvantages: an unmaanged file will be filled in with debug events forever and the logs of different concurrent debug servers will be intermingled and will be hard to analyze by clients.
3. Fork java-debug, make top-level loggers non-static and allow a way to initialize them to loggers passed in class constructors. If no loggers are passed then we can default on the global ones. Then publish it, depend on this artifact and create create a JUL logger adapter like we have for slf4j. After having a look at the code, this isn't too bad of an idea, there are only 20 use sites of loggers inside the project, most of them are request handlers which could be solved if we add a `getLogger/setLogger` API in `IDebugAdapterContext`. The rest of call sites can be updated by adding the logger to secondary constructors. We could try to upstream this change. Disadvantage with this approach is that we'd need to maintain this fork if upstreaming doesn't work.
  
I'm listening to more ideas. I personally think forking java-debug is actually OK and it makes sense to reuse our current loggers for it, though the second option is also feasible... Thoughts @marek1840 ?

frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
marek1840: I have created a simple java class and then run it with JDI enabled on both JDK and JRE:
It prints the compiler it gets by using  `javax.tools.ToolProvider.getSystemJavaCompiler()`

```
$ java -version
openjdk version "1.8.0_222"
OpenJDK Runtime Environment (build 1.8.0_222-b05)
OpenJDK 64-Bit Server VM (build 25.222-b05, mixed mode)

$ java -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,quiet=n Foo 
Listening for transport dt_socket at address: 41419
Compiler: com.sun.tools.javac.api.JavacTool@1175e2db
```

```
$ /usr/lib/jvm/java-9-jre/jre/bin/java -version
java version "9.0.4"
Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)

$ /usr/lib/jvm/java-9-jre/jre/bin/java -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,quiet=n Foo 
Listening for transport dt_socket at address: 43201
Compiler: null
```
Since both of those are listening for `transport dt_socket...`, it looks like we don't need JDK on the debuggee VM, just the host.

project/build.sbt
marek1840: Sorry, should have been more clear:

It adds `tools.jar` to our development classpath and makes sure it is not published with bloop as it would be illegal.
We need this jar on classpath to work on `DebugExtensions` which heavily rely on the `com.sun.jdi` package. And we need `DebugExtensions` to even instantiate the Debug Adapter Server.

Without it I get this compilations error:
```
[error] bloop/frontend/src/main/scala/bloop/dap/DebugExtensions.scala:11:16: object jdi is not a member of package com.sun
[error] import com.sun.jdi.{ObjectReference, StackFrame, ThreadReference, Value}
```

frontend/src/test/scala/bloop/dap/DebugSessionProxy.scala
marek1840: I have given it a go, but I am not very proficient at creating good docs. Could you (or @tgodzik ) take a look and point me in the right direction (what to update etc.)?

project/build.sbt
jvican: Right, so I guess we get this error because sbt is blacklisting the load of `com.sun.jdi` when compiling these projects. If those errors com from inside the sbt shell, that makes sense.

frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
jvican: Good catch :+1: I'm happy we don't need to check the java homes we use to run/test.

frontend/src/main/scala/bloop/dap/DebugSession.scala
marek1840: I am once again completing this promise after the event is being sent. It must be this way, since we are closing the socket once it is completed. If we were completing before sending event, we would get a race (closing socket, sending event).

frontend/src/test/scala/bloop/bsp/BspBaseSuite.scala
marek1840: The tests for debug protocol were not releasing the threads (blocking operation in the java-debug server). Now it is reverted to 4.

marek1840: As it turns out, debug-protocol-spec requires at leasts 5 threads. I followed your suggestion and created the DebugBspBaseSuite

frontend/src/main/scala/bloop/dap/DebugServer.scala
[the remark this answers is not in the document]
marek1840: In the end I was not able to reproduce it on this particular test case. Let's hope the fixes I introduced  help.

frontend/src/main/scala/bloop/dap/DebugSession.scala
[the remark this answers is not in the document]
marek1840: It looks like this internal disconnect request is indeed redundant. Also, I think I have found a race condition in the IntelliJ UTestRunner (in Intellij it sometimes "hangs" with one test case marked "running" even though it clearly finished (last statement was executed)). I have run the same test suites in sbt and didn't encounter the issues in more than 30 runs (Intellij usually failed 1 out of 10 runs).

To answer your questions:
>  If the VM (...) exits because we cancel it (...) how is it handled?

If debuggee exits unexpectedly it emits VMDeath and VMDisconnected events. Those are handled by sending the Exited and Terminated events respectively.

> Why do we need to send this disconnect request manually?

It turns out we don't have to. There is some cleanup logic in DisconnectHandler but it is used only when the Server is launching the debuggee, not attaching to it like in our case. 

> Can we get rid of the delay and ensure that we clean up resources at the very moment where we cancel everything so that we can instantiate another debug session (in case of a restart) from a clean slate?

When we cancel the Session there is no restart. We stop the debuggee, client socket and the server.
Also, the timeout is necessary if the debuggee for some reason does not close. Without the timeout we would not close the socket in this case.

frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
```diff
@@ -0,0 +1,197 @@
...
+            _ <- client.disconnect()
+            output <- client.allOutput
+          } yield output
+        }
+
+        assertNoDiff(output, "Non-blocking Hello!")
+      }
+    }
+  }
+
```
jvican: Can we add a test when we restart the execution of a test suite while the program is still running tests suites? Maybe also think of other test cases we might want to be testing? Like what happens when things go wrong?

frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
```diff
@@ -0,0 +1,197 @@
...
+            previousSessionOutput <- previousSession.allOutput
+          } yield previousSessionOutput
+        }
+
+        assertNoDiff(output, "")
+      }
+    }
+  }
+
+  test("picks up source changes across sessions") {
```
jvican: While thinking about more scenarios for our test cases, I've realized that we're only picking up source changes across sessions. So, a restart action from vscode will not pick up any changes to the sources, it will instead just force a new start of the VM with the same class files. Let's have a call tomorrow about how we can work around this.

frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
marek1840: `DebugServerSpec.restarting closes current client and debuggee` should verify that both the connection and the debuggee (a task obtained from `TestTask.runTestSuites`) are cancelled.
It's true, I have not verified if this task cancels correctly - I will look into it soon.

marek1840: in short, I was thinking of `DebugServerSpec` as a test suite which abstracts over the debuggee and verifies the server itself - something closer to unit tests than integration tests. `DebugProtocolSpec` is used to verify that the actual protocol works correctly - those are the "real" integration tests.

frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
```diff
@@ -0,0 +1,95 @@
...
+    state: State
+) extends DebuggeeRunner {
+  def run(debugLogger: DebugSessionLogger): Task[Unit] = {
+    val filter = TestInternals.parseFilters(filters)
+    val handler = new LoggingEventHandler(state.logger)
+
+    val sequentialTests = projects.map { project =>
+      val debugState = state.copy(logger = debugLogger)
+      Tasks.test(
+        state,
```
marek1840: ah, the perils of extracting variables ;)
```suggestion
        debugState,
```

frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
```diff
@@ -0,0 +1,197 @@
...
+            _ <- client.exited
+            _ <- client.terminated
+            _ <- client.disconnect()
+            output <- client.allOutput
+          } yield output
+        }
+
+        val lines = output.lines.toSeq
+        assert(lines.size == 2)
+        assert(lines(0).contains("Running Tests"))
```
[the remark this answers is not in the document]
marek1840: 1. because of a bug ;)
2. Now, we use them to verify the output
3. those are exactly the same kind of filters as expected by the `buildTarget/test` endpoints
4. I am not sure, I could not reproduce. Can you verify it still happens with the newest version?

frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
marek1840: here is a small PoC for recompiling sources before starting the debuggee: https://github.com/marek1840/bloop/pull/4

jvican:
@marek1840 I think we're ready to merge this! Thanks for all of your hard work. 👍 
  
I have added a documentation page in our website talking about the DAP <> BSP implementation in Bloop. I have added a limitations section that explains what users might expect from this integration and current gotchas.
  
Given that there are some important things missing in our current debug implementation, I'll omit this feature in 1.3.3 release notes. Between 1.3.3 and 1.4.0, we can work on:
  
1. Adding breakpoint support.
2. Adding a BSP notification to return the response produced by the BSP action run during the DAP session. This is the first limitation I list in the documentation page; there is currently no way that bloop can send `TestResult` or `RunResult` to the BSP client. 

I'll take care of adding breakpoint support, I want to get more familiar with the debugging JDI APIs and make sure I understand them well enough to maintain them during the next years.

@marek1840 Could you look at the second item? We should have a chat to discuss what the best way to implement that is because there are several solutions. After that, I propose you work on the debugging UX side of things in Metals as well as the unified debug/test/run interface. That's where most of the impact is to be made.

Let me know what your thoughts are when you come back from vacation! :smile:

jvican:
One limitation I forgot to add to our docs is that restarting a session via DAP doesn't trigger compilation of the involved projects. This is an intentional design decision.
  
BSP clients can decide how to handle this particular scenario. For clients such as Metals that are proxying DAP communication from the BSP server to the end client (VS Code), this limitation can be avoided by intercepting the DAP restart request from VS Code, closing the current DAP session, triggering a compile and then starting a new DAP session. This is the ideal way to solve this problem because Metals will be aware there's a compilation going on and will correctly report diagnostics to the user as well as results and timings. If restarting a DAP session implied recompiling a project, its implementation in bloop and Metals would be more complicated. It's better to make BSP compile requests explicit in the protocol and not rely on action side effects of the DAP request.

olafurpg:
@jvican Good to know, thanks for sharing!

Issue #932, opened 2019-06-10
Fix `ClassNotFoundException` in IntelliJ's test runner


ArturGajowy:
Steps:

1. update to v1.3.0 or 1.3.2
2. sbt bloopInstall, reimport bsp project in intellij
3. Run any scalatest test
4. Observe a 'ClassNotFoundException' and a mention of possible wrong runpath.

Opening here (not in IntelliJ tracker), since it works with 1.2.5 :)

jvican:
Hey @ArturGajowy Thanks for reporting! This ticket is most likely caused by the IntelliJ BSP integration, we did some changes between 1.2.5 and 1.3.0 which could require updates in IntelliJ. Would you mind opening a ticket over there? cc @jastice

ArturGajowy:
Opened https://youtrack.jetbrains.com/issue/IDEA-216216, hope that's the right bug tracker? Couldn't find a scala subsystem. 

Having a link in issue_template.md to the right bug tracker for IntelliJ and other projects related to bloop would be helpful :)

jvican:
Yup I think that's the one, thank you!

jastice:
I looked into this and have to put the burden back on Bloop. Since 1.3, Bloop provides a different outputpath (`ScalacOptions.classDirectory`) for every bsp session. IntelliJ currently uses different sessions for loading the project and building, hence it will not find the classfile in the expected directory. 

Even if we used the same session, I would still expect problems: Any time the session needs to be restarted, for instance a client restart, this would require a refresh. For some projects and build systems, this can take minutes.

jvican:
After an offline dicussion with @jastice, I have clarified that behavior is expected and clarified our positions. The fix to this is going to be mostly on the Bloop side while it'll require a small arrangement in the intellij-scala code. I'm re-opening.

Igosuki:
Since using 1.3.2, I have to manually update the class paths of my modules after doing bloopInstall, it's pretty cumbersome

Igosuki:
If this isn't going to be fixed soon, you should definitely patch the website and advise all intellij users not to update to 1.3.2

fooblahblah:
Sorry to post to a closed issue, but I'm using Idea Ultimate 2019.2 and sbt-bloop to export. When I try and run tests I experience the issue with the classpath mentioned about. I was curious how #964 factors into this and if there's some option I need to pass to SBT?

jvican:
[...]

PR #964, opened 2019-07-15
Enable option for BSP stable classes directories

bloopoid:
The following commit adds support for a parameter that clients can pass
to bloop so that bloop creates stable client classes directories
(client-specific directories) for a given client. The status quo is to
create client-specific directories that are only valid during a given
session and then they are removed but IntelliJ's model unfortunately
assumes stable classes directories. If we create a client classes
directory per session, IntelliJ needs to update the whole build model
and reindex everything, which is apparentely a very costly action that
can take up more than 5 minutes in certain projects.

For the sake of having a good intellij compatibility story, we add this
option. This option implies that the client (e.g. IntelliJ) is 100%
responsible for managing the contents created inside a root directory
provided by the initialize build params option. Bloop will create
project-specific classes directories where it will write the compilation
products produced by every BSP compile request triggered by IntelliJ.
IntelliJ has, in turn, the responsibility of ensuring that a root
directory is only used by one client.

It's worth noting that IntelliJ can only access the contents of the
client classes directories when bloop is not running any action. If
bloop is running any action (compile, test, run), IntelliJ cannot access
the contents of the classes diretories because it has no guarantee that
they will not change when the bloop action is done. Only when bloop runs
no action, that guarantee is met by the build server.

If you're a BSP build client implementor and you've just found this
option, beware that you need to **ensure** that there's only one build
client with a given BSP root directory where client classes directories
are stored. If you cannot ensure this, please do not use this feature.

Fixes https://github.com/scalacenter/bloop/issues/932

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspDefinitions.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspDefinitions.scala
@@ -1,0 +1,16 @@
+package bloop.bsp
+
+import io.circe._
+import io.circe.derivation._
+import ch.epfl.scala.bsp.Uri
+
+object BloopBspDefinitions {
+  final case class BloopExtraBuildParams(
+      clientClassesRootDir: Option[Uri]
+  )
+
+  object BloopExtraBuildParams {
+    val encoder: RootEncoder[BloopExtraBuildParams] = deriveEncoder
+    val decoder: Decoder[BloopExtraBuildParams] = deriveDecoder
+  }
+}
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -7,6 +7,7 @@ import java.nio.file.{Files, FileSystems, Path}
 import java.util.concurrent.ConcurrentHashMap
 import java.util.stream.Collectors
 
+import bloop.bsp.BloopBspDefinitions.BloopExtraBuildParams
 import bloop.{CompileMode, Compiler, ScalaInstance}
 import bloop.cli.{Commands, ExitStatus, Validate}
 import bloop.data.{Platform, Project, ClientInfo}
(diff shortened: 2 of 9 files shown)


frontend/src/main/scala/bloop/bsp/BloopBspDefinitions.scala
```diff
@@ -0,0 +1,16 @@
...
+import io.circe.derivation._
+import ch.epfl.scala.bsp.Uri
+
+object BloopBspDefinitions {
+  final case class BloopExtraBuildParams(
+      clientClassesRootDir: Option[Uri]
+  )
+
+  object BloopExtraBuildParams {
+    val encoder: RootEncoder[BloopExtraBuildParams] = deriveEncoder
```
tgodzik: Do we plan to save that into a file or the plan is just to always send that to the bloop server?

jvican: This is not intended to be saved. `BloopExtraBuildParams` covers any kind of configuration that a client can pass in to tweak bloop's behaviour. That includes `semanticdbVersion` too --it's just a configuration option passed during the initialization handshake-- so that's why we should add it here.
  
Re your specific PR, here's how I'd make the semanticdb version fit in with the infrastructure in `BloopBspServices`:
  
1. Add `semanticdbVersion` to `BloopExtraBuildParams`.
2. Add an explicit parse step  to obtain the `semanticdbVersion` from a decoded `BloopExtraBuildParams`. You can do that inside `initialize` in `BloopBspServices`, in a similar fashion as we do it with `clientClassesRootDir`. If a semanticdb version is found, then we create a `WorkspaceSettings` instance with it.
  
There's a key difference between what is a workspace setting and what is a BSP configuration enabled by the client and that's why I'm being so adamant on modelling as different entities and instantiating the former based on the latter.

tgodzik: Cool, thanks for the explanation. I think we can merge this if everything is in order.

jvican:
Sounds good, given it’s somehow an important fix for IntelliJ and Tomasz’s PR, let’s merge it right away @marek1840 if you still have comments let me know and I’ll address them in an upcoming PR

PR #967, opened 2019-07-17
Export JVM platform options from Gradle project

### LINKED ISSUE #966: Export javac fork options from Gradle
Sometimes a project may want to use a different JDK than the default one.
In gradle this is accomplished by setting fork options:

```
javaCompile.options.fork = true 
javaCompile.options.forkOptions.javaHome = ...
javaCompile.options.forkOptions.memoryInitialSize = "2g"
javaCompile.options.forkOptions.memoryMaximumSize = "2g"
```

These should be exported from Gradle to .bloop/

pkolaczk:
Fixes #966

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -26,7 +26,7 @@ object Dependencies {
   val mavenApiVersion = "3.6.1"
   val mavenAnnotationsVersion = "3.5"
   val mavenScalaPluginVersion = "3.2.2"
-  val gradleVersion = "3.0"
+  val gradleVersion = "3.5"
   val groovyVersion = "2.5.0"
   val ipcsocketVersion = "1.0.1"
   val monixVersion = "2.3.3"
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -26,7 +26,7 @@ object Dependencies {
   val mavenApiVersion = "3.6.1"
   val mavenAnnotationsVersion = "3.5"
   val mavenScalaPluginVersion = "3.2.2"
-  val gradleVersion = "3.0"
+  val gradleVersion = "3.5"
```
jvican: Any reason why we're doing this? We want to be compatible with previous Gradle versions and test as old versions as possible, so that's why we use 3.0 here.

tgodzik: It's because of the fork options, apparently they were introduced in 3.5. I think it's reasonable to upgrade API to 3.5, we still support 3 this way and I doubt that anybody is using 3.0.0 anyway.

PR #969, opened 2019-07-18
Use Platform JVM when compiling

### LINKED ISSUE #968: Respect JVM platform.config settings
It looks like JDK home and other JVM settings are currently ignored.
This breaks multi-JDK projects.

I have this in my json config:
```        
        "platform" : {
            "name" : "jvm",
            "config" : {
                "home" : "/opt/jdk11",
                "options" : [
                    "-Xms2g",
                    "-Xmx2g"
                ]
            },
            "mainClass" : [
            ]
        }
```

but bloop still compiles this project with JDK8.

pkolaczk:
Fixes #968

This is not the final fix.
What works:
- Compiling a project with JDK8 / JDK11 subprojects

What doesn't work:
- Compilation of the JDK8 part (99% of my project) got 2x slower. 
- Compiler JVM options are still not set yet

@jvican can you take a look and give me a hint what causes the slowdown? 
Did I somehow accidentally break preserving compilers between the runs?

Thanks

### DIFF
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -35,6 +35,7 @@ import bloop.CompileMode.Sequential
 import monix.execution.ExecutionModel
 
 case class CompileInputs(
+    javaInstance: JavaInstance,
     scalaInstance: ScalaInstance,
     compilerCache: CompilerCache,
     sources: Array[AbsolutePath],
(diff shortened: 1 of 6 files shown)


pkolaczk:
Yes, I am aware of the fact you're working on this and you won't merge my PR this time :)
I just need a quick hack to make it work with my project until your fully-fledged PR is ready. I'd like to present Bloop to the rest of the company, but I don't want to present a long list of "sorry, bloop won't work for your subproject" items. 

I assumed wrongly that forked compilers stay in memory and are reused. Now I see the compiler object forks a fresh compiler on every run, so indeed, this needs redoing how forked compilers work. Until this is fixed, I'll just make a hack to use the default system compiler when java home of the project matches the system compiler java home.

jvican:
> Until this is fixed, I'll just make a hack to use the default system compiler when java home of the project matches the system compiler java home.
  
This is indeed not a hack, but the preferred solution here. It's what makes the most sense :smile:.
  
It's great you're presenting this in your company! It makes sense you want to work around the current shortcomings, so forking the project and publishing it locally makes sense while all of these issues get solved 👍

PR #975, opened 2019-07-21
Import Gradle test system properties to test framework config

pkolaczk:
This allows for some limited configurability of most common test frameworks.

### DIFF
--- a/config/src/main/scala/bloop/config/Config.scala
+++ b/config/src/main/scala/bloop/config/Config.scala
@@ -45,9 +45,12 @@ object Config {
 
   case class Test(frameworks: List[TestFramework], options: TestOptions)
   object Test {
-    def defaultConfiguration: Test = {
-      val junit = List(Config.TestArgument(List("-v", "-a"), Some(Config.TestFramework.JUnit)))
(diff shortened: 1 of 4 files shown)


pkolaczk:
Thank you so much!

pkolaczk:
One more thing that would be a nice to have is also exporting the `jvmArgs` and environment for the forked test JVM. Gradle has these properties, but currently I had no place to copy them to. I would have to change the JSON format - is it ok to add new fields?

jvican:
The story for supporting platform arguments for tests isn't clear at this moment and needs more thought before we decide that adding new fields to the JSON configuration file is okay.
  
The jvm args and environment for a test build target could be added to the platform field of the JSON configuration file in those cases where the test build targets are independent from the compile build targets. I think this is the case for gradle, so you could piggyback on serializing jvm args and home to the platform of the test build target and then modify bloop test internals to use the platform while running tests (if we don't already do, which I don't remember).

pkolaczk:
Yes, that makes sense. I forgot we have a separate platform settings for the test build target :)

PR #979, opened 2019-07-24
Fix missing dependencies to files referenced directly

### LINKED ISSUE #978: Direct jar dependencies not imported from Gradle
Create a project with the following dependency:

```
dependencies {
    implementation(files(org.gradle.internal.jvm.Jvm.current().toolsJar))
}
```

Expected:
The tools.jar is included in the bloop project classpath.

Actual:
The tools.jar is missing from the bloop project classpath.

pkolaczk:
In gradle it is possible to define a dependency to an external file
like this:

  dependencies {
    runtime files(Jvm.current().toolsJar)
  }

Unfortunately such dependencies do not come as ResolvedArtifacts
from the Gradle API, and therefore they were missed in the generated
bloop configuration. This commit fixes importing them.

Fixes #978


pkolaczk:
The force-push was for tests. Now it is ready!

pkolaczk:
Comment cleanup

PR #983, opened 2019-07-27
Support javac processorpath in Gradle

### LINKED ISSUE #962: Add an annotation processor classpath to bloop
Somehow I can't make Dagger2 work with bloop. Complains about missing generated classes like `DaggerSystemComponent` which should be generated for the following interface:

```java
@Singleton
@Component(modules = SystemModule.class)
public interface SystemComponent
...
```
Looks like annotation-processing is bugs all the way down :(

~~My guess is that annotation-processing for the component class somehow happens *after* an attempt to compile the code referencing the generated class.~~

Any ideas what to look at?

pkolaczk:
Fixes #962


pkolaczk:
That would also work, but would be more of a hack.

Processorpath is the same kind of thing like a classpath, so why would you like to configure classpath as a seprate config parameter and processorpath like a regular option? Also, if processorpath changes, the project should be recompiled, so processorpath must be a part of compile inputs / hashes.

pkolaczk:
Additionally, I realized annotation processors may actually be defined inside of the project, as a source form, e.g. in a dependent project. I didn't  implement that capability yet (because it is a rare use-case), but honestly that should work just like regular classpath, to be complete.

jvican:
> Processorpath is the same kind of thing like a classpath, so why would you like to configure classpath as a seprate config parameter and processorpath like a regular option?
  
I want to avoid changing the configuration file format. If I can encode the processor path in the classpath, I prefer that approach. In theory we can do that and then just parse the classpath to obtain the processor path and pass it to `bloop.data.Project` internally. It is not necessary to that though because for now we're not going to process the processor path in any way, we'll just pass it to `javac`, so just having it in the classpath is OK. 

> if processorpath changes, the project should be recompiled, so processorpath must be a part of compile inputs / hashes
  
Maybe Zinc will recompile it because the compiler options changed, I haven't checked, but I don't intend to fully support this use case. If users change something affecting the processor path, they can clean and then compile again. It should not be a normal thing to do anyway.
 
> Additionally, I realized annotation processors may actually be defined inside of the project, as a source form, e.g. in a dependent project. I didn't implement that capability yet (because it is a rare use-case), but honestly that should work just like regular classpath, to be complete.
  
Yeah, I agree, handling that and incremental changes would be ideal.
  
But you see, from my POV, this is kind of a niche feature only used by a few companies in the world and I don't intend to fully support it out-of-the-box because gracefully detecting changes in processor paths require time and dedication I don't have now. Even if someone was to put in the time to handle those corner cases gracefully, I don't even know I'd like to maintain it; it drags the evolution of the project and distracts me from more important things.
  
The companies having using processor paths can likely work around them on their own, more than I can afford to fully support this.
  
I'm happy to have basic support for it so that the original issue you had is gone. If people continue to stumble upon more uses of processor path and are willing to put in the time to create a ticket and work on fully supporting this with high-quality testing and verification of it working, then maybe we can consider doing that in the future. If they don't want to use Bloop because it doesn't fully support Java processor paths, I don't mind.
  
I personally appreciate how helpful you've been in catching this error and submitting a PR to fix it. Really, thank you for that. It's just that I prefer we do the most minimal change to fix this issue and move on. We'll see how support for this feature shapes up in the future.
  
Let me know if you understand my POV as a maintainer of the project!

pkolaczk:
Looks like changing `List[Path]` to `Option[List[Path]]` solves the backwards compatibility issue of the config. The old config can be read properly by a new version of bloop. Is it good enough? Or are there any other reasons you don't want to change the config file format?

pkolaczk:
I pushed an updated version of my patch. Now for modules that don't have custom `processorpath` configuration set in gradle, the config file will not get a separate `processorpath` entry in the json config.  Therefore this is now both backwards- and forwards-compatible and there is **no change for projects that don't use processorpath**. An old version of bloop will be able to read a new config as long as it doesn't use processorpath feature (if it uses, it wouldn't compile anyway). A new version is able to read an old config.

jvican:
@pkolaczk The state of the PR looks good but I'm still concerned about adding a field just for the processor path. I just don't see the need. The top-level fields in the configuration file are reserved for concepts that exist in all build tools. The notion of processor path on its own only exists in Gradle ("annotation processor"), it doesn't exist in neither sbt, Mill, Maven nor other build tools in the Scala community.
  
In all of these build tools, the processor path is represented with javac options. By adding processor path as a field, we force all of the other build tools to parse javac options and then move the arguments for processor paths in the top-level `processor path` field. This is redundant and I don't think it's the right direction.
  
To support processor paths, we should export them from Gradle as Javac options. Inside Bloop, we can parse processor path from the javac args just once and be done with support for `processor path`s for all of the build tools. This is the simplest solution that makes it work with all of the build tools.

jvican:
Let's open this PR when it's updated with the proposed solution :smile:

Issue #987, opened 2019-07-28
Incremental compilation of Java classes is non-deterministic and invalidates far too many classes


pkolaczk:
Compiling the same type of change on the same Java sources repeatedly is causing different invalidation sets each time.

On a real project (2000+ files) I observe the following erratic behavior:
Sometimes on the same sources, incremental compilation goes totally wild and ends up recompiling everything (2000+ files) and takes 3 minutes, and sometimes on the exactly same initial source set, but different prior compilations (different saved analysis state?) the exactly same change causes recompilation of only 40 classes and takes 3 seconds.

I tried to reproduce the problem at a small scale, and found something wrong with this tiny example (although I'm not 100% sure if the problems visible on this tiny example are the real cause of my weird compile times):

```
test("incremental compilation is deterministic") {
    TestUtil.withinWorkspace { workspace =>
      object Sources {
        val `A1.java` =
          """/A.java
            |public class A {
            |  public static void method() {}
            |  public static void dummy() {}
            |}
          """.stripMargin

        val `B1.java` =
          """/B.java
            |public class B {
            |  public void callA() {
            |    A.method();
            |  }
            |}
          """.stripMargin

        val `O.java` =
          """/O.java
            |
            |public class O {
            |  public void callA() {
            |    A.dummy();
            |  }
            |}
            |
            |""".stripMargin

        val `A2.java` =
          """/A.java
            |public class A {
            |  public static void methodRenamed() {}
            |  public static void dummy() {}
            |}
          """.stripMargin

        val `B2.java` =
          """/B.java
            |public class B {
            |  public void callA() {
            |    A.methodRenamed();
            |  }
            |}
          """.stripMargin

        val `A3.java` =
          """/A.java
            |public class A {
            |  public static void methodRenamed3() {}
            |  public static void dummy() {}
            |}
          """.stripMargin

        val `B3.java` =
          """/B.java
            |public class B {
            |  public void callA() {
            |    A.methodRenamed3();
            |  }
            |}
          """.stripMargin

        val `A4.java` =
          """/A.java
            |public class A {
            |  public static void methodRenamed4() {}
            |  public static void dummy() {}
            |}
          """.stripMargin

        val `B4.java` =
          """/B.java
            |public class B {
            |  public void callA() {
            |    A.methodRenamed4();
            |  }
            |}
            |
          """.stripMargin

        val `C.scala` =
          """/C.scala
            |object C {
            |  val b = new B()
            |  b.callA()
            |}
[...]

PR #988, opened 2019-07-29
Don't ever recompile already compiled classes

### LINKED ISSUE #986: Worst-case incremental Java compile is slower than full compile
I observed a very weird behavior.

```
$ bloop clean project
$ bloop compile project
Compiling project (2208 Java sources)
Compiled project (27052ms)

$ git apply <some patches>
$ bloop compile project
Compiling project (34 Java sources)
Compiled project (2572ms)
Compiling project (167 Java sources)
Compiled project (10006ms)
Compiling project (169 Java sources)
Compiled project (4016ms)
Compiling project (2208 Java sources)    // exceeded 50%, recompiling all sources
Compiled project (188091ms)                  // but why it takes soooo long now?
```

However, when I later unapply the patch, recompile project, then apply the patch again and recompile the project - this time it is crazy fast and recompiles just a few files in a few seconds!

Why is that? Is there any form of caching other than caching the direct previous compilation?

pkolaczk:
Fixes a highly critical problem that caused inconsistent
behavior of incremental compiler and could led to abysmal compile times 
as well as analysis corruption and not recompiling classes that 
should be compiled.

Invalidating a class that was already recompiled once and then letting
javac recompile it again caused javac to not actually recompile
the class and not emit the analysis for that class. And because the
class was already invalidated, its previous analysis was pruned,
hence lost completely. A class without analysis was then considered
removed and triggered many API changes, which invalidated classes
that shouldn't really be invalidated.

Impact of the bug:
- Recompiling far too many Java classes, sometimes
  orders of magnitude more
- Recompiling the same classes again and again
- Running too many rounds of incremental compilation,
  eventually triggering a slow "brute-force" transitive step
- **Saving incorrect analysis with many classes missing**
- **Not recompiling dependent classes on the next recompilation, due
  to broken, incomplete analysis loaded from the previous run**

Fixes #986, #987

@jvican I know that you asked me to stop submitting PRs, but this one is highly critical
and I believe the community will like it. Without this PR, Java compilation was just broken for any 
larger, non-trivial changeset.

### DIFF
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -256,7 +256,8 @@ object Compiler {
             f =>
               // A safety measure in case Zinc does not get a complete list of generated classes
               val filePath = f.getAbsolutePath
-              if (!filePath.startsWith(readOnlyClassesDirPath)) true
+              if (!filePath.startsWith(readOnlyClassesDirPath))
+                false
               else {
                 import compileInputs.generatedClassFilePathsInDependentProjects
                 val relativeFilePath = filePath.replace(readOnlyClassesDirPath, "")
--- a/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopIncremental.scala
+++ b/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopIncremental.scala
@@ -104,7 +104,7 @@ object BloopIncremental {
         } yield callback.get
       }
 
-      incremental.entrypoint(initialInvClasses, initialInvSources, setOfSources, binaryChanges, lookup, previous, doCompile, manager, 1)
+      incremental.entrypoint(initialInvClasses, initialInvSources, setOfSources, Set.empty, binaryChanges, lookup, previous, doCompile, manager, 1)
     }
 
     analysisTask.materialize.map {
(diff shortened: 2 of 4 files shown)


pkolaczk:
Ok, anyway, I'll keep the branch in case you want to look at it in a few weeks. Thank you.

jvican:
As I’ve said already in the linked tickets, this should not be fixed in bloop but sbt/zinc, if you want to continue investigating this you can make a PR against the incremental compiler itself. I have many thoughts and doubts about this particular fix attempt but as I said I won’t provide feedback to external contributions during the next weeks due to limited time.

pkolaczk:
I'm not convinced this is fixable on the Zinc side alone (zinc is actually doing what it should) and I'm also very curious on your "thoughts and doubts", but let's leave the conversation for later. In the meantime we'll be testing my fix and I'll let you know if we hit any further problems.

jvican:
This is definitely fixable in sbt/zinc.

pkolaczk:
If it can be fixed in sbt/zinc that's even better, because it would benefit more projects and more people :)  

> if you want to continue investigating this you can make a PR against the incremental compiler itself. 

I don't mind working on this further, but for this to happen, I need at least some tiny bit of help on what is actually broken in zinc (a description of its internal contracts / responsibilities would be a good start). And in order to test I need a way to modify / compile the fork of zinc used in bloop, which I haven't figured out yet. That stuff is not really very complex in terms of algorithms for me, but rather a bit undocumented and this makes it harder to come up with an elegant fix.

Issue #976, opened 2019-07-23
Unexpected compile dependency in sbt projects with configuration dependencies


zainab-ali:
The bloop sbt plugin does not respect [per configuration classpath dependencies](https://www.scala-sbt.org/release/docs/Multi-Project.html#Per-configuration+classpath+dependencies).

The following `build.sbt` file
```scala
lazy val foo = project in file("foo")

lazy val bar = (project in file("bar"))
.dependsOn(foo % "test->test")
```
In this example, bar should have a test dependency on foo, but not a compile dependency. 
The `.bloop/bar.json` file should not contain a dependency on foo.
However, the `bar.json` contains:

```json
    "version" : "1.1.2",
    "project" : 
        "name" : "bar",
        ....
        "dependencies" : [
            "foo"
        ],
```

`bloop compile bar`compiles `foo` as well as `bar`.

jvican:
Thanks for reporting @zainab-ali! This is a good catch 😊  `sbt-bloop` does support per configuration classpath dependencies but it looks like it's too conservative when pinning down normal compile dependencies between projects. `bar-test` correctly depends on `foo-test` but `bar` should, as you say, not depend on `foo`.

PR #989, opened 2019-07-29
Fixes unnecessary dependency in sbt extractor

jvican:
This commit removes a bit of logic that was inherited from the previous
extractor that was not configuration based and that was not spotted
before. This extra line was adding an unncessary compile dependency
between projects when the project instead had another dependency (e.g.
test, it) on that project.


jvican:
Fixes #976

Issue #880, opened 2019-05-06
Distinguish between pure Java/Scala BSP targets


olafurpg:
Followup from https://github.com/scalameta/metals/pull/694#issuecomment-489164128

To reproduce:

- import apache/kafka repository with Bloop Gradle plugin
- open Metals and run doctor, which warns that many targets in the build are misconfigured
![](https://user-images.githubusercontent.com/1408093/57095217-9c433580-6d12-11e9-884f-5b28ab15db7c.png)
- I expected there to be no warning for the Java-only targets

The reason Metals warns about the Java-only targets is because `BuildTarget.languageIds=["scala","java"]` and `BuildTarget.data` includes Scala-specific information

```json
{
        "id": {
          "uri": "file:/Users/olafurpg/dev/kafka/clients/?id\u003dclients"
        },
        "displayName": "clients",
        "baseDirectory": "file:///Users/olafurpg/dev/kafka/clients/",
        "tags": [
          "library"
        ],
        "languageIds": [
          "scala",
          "java"
        ],
        "dependencies": [],
        "capabilities": {
          "canCompile": true,
          "canTest": true,
          "canRun": true
        },
        "data": {
          "scalaOrganization": "org.scala-lang",
          "scalaVersion": "2.12.8",
          "scalaBinaryVersion": "2.12",
          "platform": 1,
          "jars": [
            "file:///Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.12.8/scala-library-2.12.8.jar",
            "file:///Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-reflect/2.12.7/scala-reflect-2.12.7.jar",
            "file:///Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.12.7/scala-compiler-2.12.7.jar",
            "file:///Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-xml_2.12/1.1.0/scala-xml_2.12-1.1.0.jar",
            "file:///Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/jline/jline/2.14.4/jline-2.14.4.jar"
          ]
        }
```

[...]

PR #991, opened 2019-07-30
Identify Java-only projects inside Bloop

tgodzik:
Fixes https://github.com/scalacenter/bloop/issues/880

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -616,12 +616,14 @@ final class BloopBspServices(
                 canTest = true,
                 canRun = true
               )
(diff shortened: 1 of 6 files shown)


frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -598,12 +598,14 @@ final class BloopBspServices(
                 canTest = true,
                 canRun = true
               )
+              val isJavaOnly = p.scalaInstance == ScalaInstance.scalaInstanceFromBloop(bspLogger)(ioScheduler)
```
jvican: What do you think about renaming `scalaInstanceFromBloop` to `scalaInstanceForJavaProjects` so that it's clearer for future readers that we assign this instance to all java projects?

tgodzik:
Looks like the issues in build are not connected.

PR #993, opened 2019-07-31
Set proper compile order in projects imported from Gradle

### LINKED ISSUE #992: Incorrect compile order after exporting project from Gradle
Gradle compile order semantics are different than Bloop's.
Consider the following config:
```
sourceSets {
    main {
        java {  srcDirs = ['src/java'] }
        scala {  srcDirs = ['src/scala'] }
    }
}
```

This means: compile `src/java` with the Java compiler first, then compile `src/scala` with the Scala compiler. For such config, we should set compile order to JavaThenScala.

However, it is possible to specify joint compilation in Gradle like this:

```
sourceSets {
    main {
        java {  srcDirs = [] }
        scala {  srcDirs = ['src/java', 'src/scala'] }
    }
}
```

This means, mixed compilation should be used.
You don't always want mixed compilation, because Java annotation processors don't work reliably if they generate classes that are used in the same project when mixed compilation is used (probably something that should be addressed by sbt/zinc anyway, but this is a different issue). 

Unfortunately I can see that Bloop's project model is not flexible enough to reflect Gradle model fully, because Bloop throws all sources into the same bin, and the only choice is compile order. 

For example this is impossible to model in Bloop now:

```
sourceSets {
    main {
        java {  srcDirs = ['src/java-that-must-be-compiled-first] }
        scala {  srcDirs = ['src/scala', 'src/java-that-should-be-mixed-with-scala] }
    }
}
```

I believe in this case we maybe could just generate 2 projects instead of one, similar way how we're doing with tests. But this looks like a complex change to BloopConverter.

For now I just do some simple hack to set compile order based on existence of non-empty java source set.

pkolaczk:
If java source set is non-empty, a java-then-scala order is used.

This config results in java-then-scala order:

    sourceSets {
      main {
        java {  srcDirs = ['src/java'] }
        scala {  srcDirs = ['src/scala'] }
      }
    }

This config results in mixed order:

    sourceSets {
      main {
        java {  srcDirs = [] }
        scala {  srcDirs = ['src/java', 'src/scala'] }
      }
    }

Fixes #992

### DIFF
--- a/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
+++ b/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
@@ -6,7 +6,7 @@ import java.nio.file.Paths
 import java.nio.file.Files
 
 import bloop.config.Config
-import bloop.config.Config.{JvmConfig, Platform, TestArgument, TestOptions}
+import bloop.config.Config.{CompileOrder, CompileSetup, JavaThenScala, JvmConfig, Mixed, Platform, TestArgument, TestOptions}
 import bloop.integrations.gradle.BloopParameters
 import bloop.integrations.gradle.model.BloopConverter.SourceSetDep
 import bloop.integrations.gradle.syntax._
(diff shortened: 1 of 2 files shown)


integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
```diff
@@ -592,11 +592,15 @@ final class BloopConverter(parameters: BloopParameters) {
           val opts = scalaCompileTask.getScalaCompileOptions
           val options = optionList(opts)
           val compilerName = parameters.compilerName
+          val compileOrder =
+            if (!sourceSet.getJava.getSourceDirectories.isEmpty) JavaThenScala
```
jvican: Couldn't find any official documentation about this, but looks like this SO defines the behaviour: https://stackoverflow.com/questions/23261075/compiling-scala-before-alongside-java-with-gradle

It's weird that the default in Gradle when java sources are defined is `JavaThenScala`. In build tools such as sbt the default is always mixed. Anyhow, I'm trusting this is fine 👍

PR #998, opened 2019-08-01
Fix semantics of `isSameFile` in class file manager

### LINKED ISSUE #956: Unexpected error using a Java annotation processor
```
[E] [E-1] Errors during annotation processing
[E] [E18] XXXX.java:17
[E]       Exception during annotation processing: java.lang.IllegalArgumentException: Not supported: sbt.internal.inc.javac.WriteReportingJavaFileObject@3e42b4ef

// last error repeats many times, followed by errors about missing symbols

[E] [E12] YYYYYY.java:41
[E]       cannot find symbol
[E]         symbol:   class .....
[E]         location: class ....
```

pkolaczk:
Passing a WriteReportingFileObject to the original, wrapped
JavaFileManager may cause InvalidArgumentException. To avoid this,
we unwrap the original FileObject.

Fixes #956

### DIFF
--- a/backend/src/main/scala/bloop/CompilerCache.scala
+++ b/backend/src/main/scala/bloop/CompilerCache.scala
@@ -4,35 +4,33 @@ import java.io.File
 import java.lang.Iterable
 import java.io.PrintWriter
 import java.util.concurrent.ConcurrentHashMap
+
 import javax.tools.JavaFileManager.Location
 import javax.tools.JavaFileObject.Kind
 import javax.tools.{
   FileObject,
   ForwardingJavaFileManager,
-  StandardJavaFileManager,
-  ForwardingJavaFileObject,
   JavaFileManager,
   JavaFileObject,
   JavaCompiler => JavaxCompiler
 }
-
 import bloop.io.{AbsolutePath, Paths}
 import bloop.logging.Logger
-
 import sbt.librarymanagement.Resolver
-
 import xsbti.ComponentProvider
 import xsbti.compile.Compilers
 import xsbti.compile.{JavaCompiler => XJavaCompiler, JavaTool => XJavaTool}
 import xsbti.compile.ClassFileManager
 import xsbti.{Logger => XLogger, Reporter => XReporter}
-
 import sbt.internal.inc.bloop.ZincInternals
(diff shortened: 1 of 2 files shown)


backend/src/main/scala/bloop/CompilerCache.scala
```diff
@@ -4,35 +4,22 @@ import java.io.File
...
-  FileObject,
-  ForwardingJavaFileManager,
-  StandardJavaFileManager,
-  ForwardingJavaFileObject,
-  JavaFileManager,
-  JavaFileObject,
-  JavaCompiler => JavaxCompiler
-}
-
+import javax.tools.{FileObject, ForwardingJavaFileManager, ForwardingJavaFileObject, JavaFileManager, JavaFileObject, StandardJavaFileManager, JavaCompiler => JavaxCompiler}
```
jvican: Can you leave the imports as they were and disable the formatter in intellij when hacking on bloop? You should be using the scalafmt formatter instead.

pkolaczk: The formatter touches only the modified lines. But, yeah, I'll try to set up scalafmt.
Thanks for heads up.

pkolaczk: How do I run `sbt scalafmt` on a single file? It reformats many other files if invoked on the whole project. :(

jvican: Try setting up scalafmt with IntelliJ. It has an option to reformat with save. Save only the file and it will be reformatted.

pkolaczk:
Ok, formatted with `sbt scalafmt`

backend/src/main/scala/bloop/CompilerCache.scala
```diff
@@ -295,6 +293,15 @@ final class CompilerCache(
           }
         }.asJava
       }
+
+      // Needed because JavaFileManager doesn't expect WriteReportingJavaFileObjects in isSameFile, fixes #956
+      override def isSameFile(a: FileObject, b: FileObject): Boolean = {
```
jvican: Can you add a test case?

pkolaczk: Ok, done

pkolaczk:
Rebased

PR #1001, opened 2019-08-05
Implement `-Xfatal-warnings` manually

jvican:
Having fatal warnings implemented inside the compiler has several
drawbacks:

1. Bloop receives a failed compilation when there is a warning. A failed
compilation doesn't allow us to cache the work done during an
incremental compiler cycle and the compilation products generated are
thrown away.  This means next time we compile we need to redo the work
of compiling all changed source files when we would only want to
recompile the files that produced a fatal warning.

2. When warnings are produced, Metals doesn't have access to the class
files generated by the failed compilation. This affects negatively the
editing experience in the editor.

To work around these limitations, we remove `-Xfatal-warnings` from the
scalac options passed to the compiler and we implement fatal warnings
ourselves in our reporter implementation.

Our implementation promotes warnings to errors just before we're going
to report them but still represents warnings as warnings in the analysis
file. At the same time, it also treats successful compilations with
fatal warnings as errors at the very end of the compilation pipeline,
after all the compilation passes have been done in a build graph, and
guarantees that it works for clients that have their compilations
deduplicated.

All in all, implementing fatal warnings reproduces the previous behavior
in the compiler but without the inconveniences of handling compilation
results as failures when only fatal warnings are solely responsible for
the failed result.

Our implementation makes sure that the next compile request will
recompile only those files with fatal warnings because we empty the hash
of those sources that contain at least a fatal warning before storing it
in our in-memory store and in disk. As the hashes will be empty, the
next recompilation will always attempt to compile those sources in the
first pass, hence giving the expected behavior.


backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -371,13 +374,24 @@ object Compiler {
     def getCompilationOptions(inputs: CompileInputs): CompileOptions = {
       val sources = inputs.sources // Sources are all files
       val classpath = inputs.classpath.map(_.toFile)
+      val optionsWithoutFatalWarnings = inputs.scalacOptions.flatMap { option =>
```
marek1840: why not just a filter?
```scala
inputs.scalacOptions.filter {
  case "-Xfatal-warnings" =>
    isFatalWarningsEnabled = true
    false
  case _ =>
    true
}
```

backend/src/main/scala/bloop/reporter/Reporter.scala
```diff
@@ -44,13 +44,39 @@ abstract class Reporter(
...
+    hasFatalWarningsEnabled = true
+  }
+
+  private val sourceFilesWithFatalWarnings = TrieMap.empty[File, Boolean]
+  override def getSourceFilesWithFatalWarnings: Set[File] = {
+    sourceFilesWithFatalWarnings.keysIterator.toSet
+  }
+
+  protected def liftFatalWarning(problem: Problem): Problem = {
+    if (!hasFatalWarningsEnabled || problem.severity != Severity.Warn) problem
```
marek1840: I find not-negated conditions easier to reason about:
`!hasFatalWarningsEnabled || problem.severity != Severity.Warn`
vs
`hasFatalWarningsEnabled && problem.severity == Severity.Warn`

What do you think?

frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
```diff
@@ -280,6 +280,8 @@ object CompileTask {
           case FinalNormalCompileResult(p, results) =>
             results.fromCompiler match {
               case Compiler.Result.NotOk(_) => List(p)
+              // Consider success with reported fatal warnings as error to simulate -Xfatal-warnings
+              case Compiler.Result.Success(_, _, _, _, _, _, true) => List(p)
```
marek1840: just wondering if something like that wouldn't be more readable? it makes it obvious what happens here (without reading the comment)

```suggestion
              case s@Compiler.Result.Success if s.reportedFatalWarnings => List(p)
```

frontend/src/main/scala/bloop/engine/tasks/compilation/CompileResult.scala
```diff
@@ -138,8 +138,13 @@ object FinalCompileResult {
             case Compiler.Result.Empty => s"${projectName} (empty)"
             case Compiler.Result.Cancelled(problems, ms, _) =>
               s"${projectName} (cancelled, failed with ${Problem.count(problems)}, ${ms}ms)"
-            case Compiler.Result.Success(_, _, _, ms, _, isNoOp) =>
-              val noOp = if (isNoOp) " no-op" else ""
+            case Compiler.Result.Success(_, _, _, ms, _, isNoOp, reportedFatalWarnings) =>
+              val noOp = {
```
marek1840: `noOp` no longer seems a valid name

```suggestion
              val mode = {
```

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -463,13 +477,17 @@ object Compiler {
         case Success(_) if cancelPromise.isCompleted => handleCancellation
         case Success(result) =>
           // Report end of compilation only after we have reported all warnings from previous runs
-          reporter.reportEndCompilation(previousSuccessfulProblems, bsp.StatusCode.Ok)
+          val sourcesWithFatal = reporter.getSourceFilesWithFatalWarnings
+          val reportedFatalWarnings = isFatalWarningsEnabled && reporter.getSourceFilesWithFatalWarnings.nonEmpty
```
marek1840:
```suggestion
          val reportedFatalWarnings = isFatalWarningsEnabled && sourcesWithFatal.nonEmpty
```

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -502,7 +520,7 @@ object Compiler {
           }
 
           val definedMacroSymbols = mode.oracle.collectDefinedMacroSymbols
-          val isNoOp = previousAnalysis.exists(_ == result.analysis())
+          val isNoOp = previousAnalysis.exists(_ == analysis)
```
marek1840:
```suggestion
          val isNoOp = previousAnalysis.contains(analysis)
```

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -602,7 +621,7 @@ object Compiler {
           cause match {
             case f: StopPipelining => Result.Blocked(f.failedProjectNames)
             case f: xsbti.CompileFailed =>
-              // We cannot assert reporter.problems == f.problems, so we aggregate them together
+              // We cannot guarantee reporter.problems == f.problems, so we aggregate them together
```
marek1840: sidenote: could use an explanation why we cannot

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -702,13 +722,20 @@ object Compiler {
     val newStamps = {
       import sbt.internal.inc.Stamps
       val oldStamps = analysis.stamps
-      val oldProducts = oldStamps.products.map {
+      // Use empty stamps for files that have fatal warnings so that next compile recompiles them
+      val rebasedSources = oldStamps.sources.map {
+        case t @ (file, _) =>
+          // Assumes file in reported diagnostic matches path in here
+          if (!sourceFilesWithFatalWarnings.contains(file)) t
```
marek1840: maybe reversing the condition to make it easier to read?

backend/src/main/scala/bloop/reporter/Reporter.scala
tgodzik: Wanted to suggest the same thing. It's the prefect De Morgan's law example :P

jvican: I can hoist up this value to another variable and then negate it. I always like to handle the happy path in the first if branch and then handle the complicated case in the else branch, that's why I'm using here a negated condition.

backend/src/main/scala/bloop/Compiler.scala
jvican: `filter` is implemented in terms of `flatMap` so it's one less level of indirection

frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
jvican: Sounds good! Have no strong opinions here, will incorporate your suggestion myself.

frontend/src/main/scala/bloop/engine/tasks/compilation/CompileResult.scala
jvican: Will change manually as it requires updating its reference.

backend/src/main/scala/bloop/Compiler.scala
jvican: Sounds good, have no strong opinions here

backend/src/main/scala/bloop/Compiler.scala
jvican: IIRC there were some cases when the compiler could return different problems in these two branches, hence the fix. Didn't elaborate when I introduced the fix because the root of the bug was not clear to me. I prefer to leave it as is for the moment

backend/src/main/scala/bloop/Compiler.scala
jvican: Will do the same as in the other suggestion :wink:

jvican:
@marek1840 Feedback addressed!

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
```diff
@@ -382,8 +382,11 @@ class BspMetalsClientSpec(
     val scalacOptions = state.scalaOptions(project)._2.items.flatMap(_.options).map { opt =>
       if (!opt.startsWith("-Xplugin:")) opt
       else {
-        opt.split(":") match {
-          case Array(key, value) => s"$key:${value.split(java.io.File.separator).last}"
+        val idx = opt.indexOf(":")
+        opt.splitAt(idx) match {
+          case (key, value) =>
+            val pluginPath = Paths.get(value.tail).getFileName().toString
```
marek1840: `toString` will be covered by the string interpolation
```suggestion
            val pluginPath = Paths.get(value.tail).getFileName()
```

jvican:
Thanks for the review guys @marek1840 @tgodzik !

Issue #679, opened 2018-10-16
Preserve origin id in BSP buildTarget/compile


olafurpg:
I triggered BSP `buildTarget/compile` with an `originId=example` and the response message and `build/logMessage` notifications has `originId=null`.

Request JSON
```json
{
  "jsonrpc": "2.0",
  "id": "3",
  "method": "buildTarget/compile",
  "params": {
    "targets": [
      {
        "uri": "file:///Users/ollie/dev/bsp/bsp4j/src/test/bloop/b/?id\u003db-test"
      },
...
    ],
    "originId": "example",
    "arguments": []
  }
}
```
Response notifications
```diff
- obtained
+ expected
{
  "jsonrpc": "2.0",
  "method": "build/logMessage",
  "params": {
    "type": 3,
    "message": "Compiling 1 Scala source to /Users/ollie/dev/bsp/bsp4j/src/test/bloop/.bloop/a/scala-2.12/classes ...",
+  "originId": "example"
  }
}
{
  "jsonrpc": "2.0",
  "method": "build/publishDiagnostics",
  "params": {
    "uri": "file:/Users/ollie/dev/bsp/bsp4j/src/test/bloop/a/src/main/scala/example/Example.scala",
+  "originId": "example",
    "diagnostics": [
      {
        "range": {
          "start": {
            "line": 3,
            "character": 24
          },
          "end": {
            "line": 3,
            "character": 24
          }
        },
        "severity": 2,
        "message": "Unused import"
      }
    ]
  }
}
```

PR #1004, opened 2019-08-08
Propagate origin id from BSP requests to loggers

jvican:
Makes sure that bloop bsp services handle the origin id field set in
certain BSP requests to let clients trace the origin of logMessage,
showMessage or diagnostics.

Fixes https://github.com/scalacenter/bloop/issues/679

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -77,9 +77,17 @@ final class BloopBspServices(
 
   // Disable ansi codes for now so that the BSP clients don't get unescaped color codes
   private val taskIdCounter: AtomicInt = AtomicInt(0)
-  private val bspLogger = BspServerLogger(callSiteState, client, taskIdCounter, false)
+  private val baseBspLogger = BspServerLogger(callSiteState, client, taskIdCounter, false)
+
+  def bspLoggerForRequest(originId: Option[String]): BspServerLogger = {
+    originId match {
+      case None => baseBspLogger
+      case Some(originId) => baseBspLogger.withOriginId(originId)
+    }
+  }
+
   final val services = JsonRpcServices
-    .empty(bspLogger)
+    .empty(baseBspLogger)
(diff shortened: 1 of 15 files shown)


frontend/src/main/scala/bloop/logging/BspServerLogger.scala
```diff
@@ -27,16 +27,20 @@ final class BspServerLogger private (
...
 
   override def isVerbose: Boolean = underlying.isVerbose
   override def asDiscrete: Logger =
-    new BspServerLogger(name, underlying.asDiscrete, client, taskIdCounter, ansiSupported)
+    new BspServerLogger(name, underlying.asDiscrete, client, taskIdCounter, ansiSupported, originId)
   override def asVerbose: Logger =
-    new BspServerLogger(name, underlying.asVerbose, client, taskIdCounter, ansiSupported)
+    new BspServerLogger(name, underlying.asVerbose, client, taskIdCounter, ansiSupported, originId)
+
+  def withOriginId(originId: String): BspServerLogger =
```
marek1840: we could make it like this and drop the `bspLoggerForRequest`
```scala
def withOriginId(originId: Option[String]): BspServerLogger =
  originId match {
    case None => this
    case _ => new BspServerLogger(name, underlying, client, taskIdCounter, ansiSupported, originId)
  }

File: frontend/src/main/scala/bloop/bsp/BloopBspServices.scala [hunk 2]
```diff
@@ -266,7 +276,9 @@ final class BloopBspServices(
     callSiteState.logger.info("BSP initialization handshake complete.")
   }
 
-  def ifInitialized[T](compute: BspComputation[T]): BspEndpointResponse[T] = {
+  def ifInitialized[T](
+      bspLogger: BspServerLogger
```
marek1840: maybe it could accept an `originId: Option[String]` ? that way each invocation would only have to consider if the originId is in scope.

jvican:
@marek1840 Those are good suggestions, good review! I've just pushed a commit addressing that as well as adding and testing `originId` in the BSP reports, which I had missed in the first implementation of this.

PR #1009, opened 2019-08-11
Assume column = 0 for diagnostics reported from Javac

### LINKED ISSUE #1007: errors in Java files are ignored in BSP publishDiagnostics
1. create a Java file with some compile error
2. run `buildTarget/compile` over BSP in bloop
3. Bloop will send an empty `build/publishDiagnostics`

```json
{
  "jsonrpc": "2.0",
  "method": "build/publishDiagnostics",
  "params": {
    "textDocument": {
      "uri": "file:///Users/jast/playspace/bloop-bsp-example/example/src/Joy.java"
    },
    "buildTarget": {
      "uri": "file:/Users/jast/playspace/bloop-bsp-example/example?id\u003dbloop-bsp-example"
    },
    "diagnostics": [],
    "reset": true
  }
}
```

See https://youtrack.jetbrains.com/issue/SCL-15984

pkolaczk:
Javac doesn't return column information in error messages.
Because of that, Javac diagnostic messages weren't properly published.

Additionally fixes publishing of diagnostics with no position at all,
but with file information present.

Fixes #1007

### DIFF
--- a/backend/src/main/scala/sbt/internal/inc/bloop/ZincInternals.scala
+++ b/backend/src/main/scala/sbt/internal/inc/bloop/ZincInternals.scala
@@ -5,11 +5,11 @@ import java.nio.file.Files
 
 import bloop.ScalaInstance
 import bloop.io.AbsolutePath
-import xsbti.{ComponentProvider, Position}
 import sbt.internal.inc.ZincComponentCompiler
-import sbt.internal.inc.javac.{AnalyzingJavaCompiler, DiagnosticsReporter}
+import sbt.internal.inc.javac.AnalyzingJavaCompiler
 import sbt.librarymanagement.{Configurations, ModuleID}
 import xsbti.compile.{ClasspathOptions, JavaCompiler}
+import xsbti.{ComponentProvider, Position}
 
 object ZincInternals {
   def latestVersion: String = ZincComponentCompiler.incrementalVersion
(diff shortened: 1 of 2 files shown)


frontend/src/main/scala/bloop/logging/BspServerLogger.scala
```diff
@@ -114,6 +117,23 @@ final class BspServerLogger private (
             event.clear
           )
         )
+      case (_, Some(file)) =>
```
jvican: How do we know this is coming from `javac` only? It would be great to check that manually and not assume that the position coming from javac will have a very concrete shape.

backend/src/main/scala/sbt/internal/inc/bloop/ZincInternals.scala
```diff
@@ -65,16 +65,12 @@ object ZincInternals {
...
-        startLine =>
-          position.startColumn().toOption.map(startColumn => (startLine.toInt, startColumn.toInt))
-      )
-
-      rangePosition.orElse {
-        position.line.toOption
-          .flatMap(line => position.pointer.toOption.map(column => (line.toInt, column.toInt)))
-      }
+      // Javac doesn't provide column information, so we just assume beginning of the line
+      val rangePosition = for { startLine <- position.startLine().toOption.map(_.toInt) }
```
jvican: Can we please use the previous style with map/flatmaps instead to minimise the change diff as much as possible?

PR #1016, opened 2019-08-18
Improve BSP services performance

pkolaczk:
1. BSP services that generate output from many projects now map pojects in parallel - reduces service call latency.
2. Writing to an `OutputStream` by single bytes was not the most efficient way of sending out data - replaced by channel. Could go probably further and avoid copying buffers at all if we didn't use `OutputStream`s, but a bigger refactoring would be needed and I don't have the time now ;)

Impact of this change:

Before this change importing our multi-module project took ~13-15 seconds with Intellij IDEA, and Bloop was using 100% of one CPU core for ~ 10 s.

After this change it takes ~ 2.5 s (most of that probably on the IDEA side), and the CPU spike on the Bloop side is hardly noticeable.

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopLanguageClient.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopLanguageClient.scala
@@ -6,12 +6,12 @@ import scala.meta.jsonrpc.CancelParams
 import scala.meta.jsonrpc.Notification
 import scala.meta.jsonrpc.JsonRpcClient
 import scala.meta.jsonrpc.MessageWriter
-
 import io.circe.Decoder
 import io.circe.Encoder
 import io.circe.syntax._
 import java.io.OutputStream
 import java.nio.ByteBuffer
+import java.nio.channels.Channels
 
 import monix.eval.Task
 import monix.eval.Callback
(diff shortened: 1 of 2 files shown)


jvican:
Let’s hold off from merging this until I’m back and can review this pull request.

frontend/src/main/scala/bloop/bsp/BloopLanguageClient.scala
```diff
@@ -157,6 +157,7 @@ object BloopLanguageClient {
       override def onError(ex: Throwable): Unit = ()
       override def onComplete(): Unit = {
         out.synchronized {
+          channel.close()
```
jvican: Should we not close the channel if `onError` is called too?

jvican:
When will you be able to look at this? @pkolaczk If you are busy and cannot look at this any time soon I will open a pull request incorporating these changes myself.

pkolaczk:
I'm on vacation till 16th

pkolaczk:
BTW, most of the performance gain comes from the second commit, the one fixing writing.

jvican:
> I'm on vacation till 16th
  
Hope you enjoy.
  
> BTW, most of the performance gain comes from the second commit, the one fixing writing.
  
Great, I'm really excited about this improvement. I'll quickly get it ready for the release I'm planning this week and try to collect some definitive numbers.

frontend/src/main/scala/bloop/bsp/BloopLanguageClient.scala
jvican: Nevermind, I think it's not necessary!

PR #1023, opened 2019-08-29
Close sockets after the BSP connection has been terminated

odisseus:
When the BSP connection is not used anymore, the socket remains open indefinitely. If the application that spawned the connection doesn't terminate immediately after `LauncherMain.runLauncher` returns, this causes a resource leak.


odisseus:
I agree; shutdownInput and shutdownOutput is redundant.

PR #1025, opened 2019-09-06
Support running custom scala main class in BSP

marek1840:
Previously the `data` and `dataKind` fields in `RunParams` were ignored and the default class in the build target was run. 

Now, if those fields are specified, we use the information (class name, arguments and jvmOptions) to run the correct class. If `dataKind` is not specified, the old approach is used for backward compatibility

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -614,8 +614,7 @@ final class BloopBspServices(
   def run(params: bsp.RunParams): BspEndpointResponse[bsp.RunResult] = {
     def parseMainClass(project: Project, state: State): Either[Exception, bsp.ScalaMainClass] = {
       params.dataKind match {
-        // TODO replace with RunParamsDataKind.ScalaMainClass once bsp updates
-        case Some("scala-main-class") =>
+        case Some(bsp.RunParamsDataKind.ScalaMainClass) =>
           params.data match {
             case None =>
               Left(new IllegalStateException(s"Missing data for $params"))
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -612,7 +612,7 @@ final class BloopBspServices(
   }
 
   def run(params: bsp.RunParams): BspEndpointResponse[bsp.RunResult] = {
-    def getMainClass(project: Project, state: State): Either[Exception, bsp.ScalaMainClass] = {
+    def parseMainClass(project: Project, state: State): Either[Exception, bsp.ScalaMainClass] = {
       params.dataKind match {
         // TODO replace with RunParamsDataKind.ScalaMainClass once bsp updates
         case Some("scala-main-class") =>
@@ -643,7 +643,7 @@ final class BloopBspServices(
       import bloop.engine.tasks.LinkTask.{linkMainWithJs, linkMainWithNative}
       val cwd = state.commonOptions.workingPath
 
-      getMainClass(project, state) match {
+      parseMainClass(project, state) match {
         case Left(error) =>
           Task.now(sys.error(s"Failed to run main class in $project due to: ${error.getMessage}"))
         case Right(mainClass) =>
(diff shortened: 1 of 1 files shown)


frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -525,37 +526,65 @@ final class BloopBspServices(
...
       import bloop.engine.tasks.LinkTask.{linkMainWithJs, linkMainWithNative}
       val cwd = state.commonOptions.workingPath
-      val cmd = Commands.Run(List(project.name))
-      Interpreter.getMainClass(state, project, cmd.main) match {
-        case Left(state) =>
-          Task.now(sys.error(s"Failed to run main class in ${project.name}"))
+
+      getMainClass(project, state) match {
+        case Left(error) =>
+          Task.now(sys.error(s"Failed to run main class in $project due to: ${error.getMessage}"))
```
tgodzik: Will it show up for the client as in `logger.showWarning`?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -525,37 +526,65 @@ final class BloopBspServices(
...
         case Right(mainClass) =>
           project.platform match {
             case Platform.Jvm(javaEnv, _, _) =>
-              Tasks.runJVM(state, project, javaEnv, cwd, mainClass, cmd.args.toArray, false)
+              val env = JavaEnv(javaEnv.javaHome, javaEnv.javaOptions ++ mainClass.jvmOptions)
+              val args = mainClass.arguments.toArray
+              Tasks.runJVM(state, project, env, cwd, mainClass.`class`, args, false)
```
tgodzik: Maybe named parameter for boolean value?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -525,37 +526,65 @@ final class BloopBspServices(
...
             case platform @ Platform.Native(config, _, _) =>
+              val cmd = Commands.Run(List(project.name))
               val target = ScalaNativeToolchain.linkTargetFrom(project, config)
-              linkMainWithNative(cmd, project, state, mainClass, target, platform).flatMap {
-                state =>
+              linkMainWithNative(cmd, project, state, mainClass.`class`, target, platform)
```
tgodzik: Why change the change:  "mainClass.`class`" ?

marek1840: previously `mainClass` was a string, now it is a `ScalaMainClass`

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: It should be shown to the client in the logs, but I'm not sure this will be shown via bap's showMessage. Should this be Metals responsibility?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -525,37 +526,65 @@ final class BloopBspServices(
   }
 
   def run(params: bsp.RunParams): BspEndpointResponse[bsp.RunResult] = {
+    def getMainClass(project: Project, state: State): Either[Exception, bsp.ScalaMainClass] = {
+      params.dataKind match {
+        // TODO replace with RunParamsDataKind.ScalaMainClass once bsp updates
```
jvican: Because of this, I propose we merge this right after the DAP implementation which upgrades the version. Otherwise we will most likely forget to fix this TODO :smile:

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -525,37 +526,65 @@ final class BloopBspServices(
   }
 
   def run(params: bsp.RunParams): BspEndpointResponse[bsp.RunResult] = {
+    def getMainClass(project: Project, state: State): Either[Exception, bsp.ScalaMainClass] = {
```
jvican: Nit: maybe a more appropriate name would be `parseMainClass` or `readMainClass`? It makes it clearer the main class is coming from the request parameters.

PR #1026, opened 2019-09-06
Support running custom scala test suite in BSP

marek1840:
Previously the data and dataKind fields in TestParams were ignored and all of the test suites in given build targets were launched.

Now, if those fields are specified, we use the information (test suite filters and jvmOptions) to run only the specified test suites. If dataKind is not specified, the old approach is used for backward compatibility.

depends on:
https://github.com/scalacenter/bsp/pull/97

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -36,6 +36,7 @@ import scala.collection.concurrent.TrieMap
 import scala.collection.JavaConverters._
 import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
+import scala.util.Try
 import scala.util.Success
 import scala.util.Failure
 import monix.execution.Cancelable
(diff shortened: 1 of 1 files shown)


frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -36,8 +36,7 @@ import scala.collection.concurrent.TrieMap
 import scala.collection.JavaConverters._
 import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
-import scala.util.Success
-import scala.util.Failure
+import scala.util.{Try, Success, Failure}
```
tgodzik: Let's not change imports - I think single import on line is the standard in the repo.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -454,44 +453,64 @@ final class BloopBspServices(
...
   def test(params: bsp.TestParams): BspEndpointResponse[bsp.TestResult] = {
-    def test(
-        id: BuildTargetIdentifier,
-        project: Project,
-        state: State
-    ): Task[State] = {
-      val testFilter = TestInternals.parseFilters(Nil) // Don't support test only for now
-      val handler = new BspLoggingEventHandler(id, state.logger, client)
-      Tasks.test(state, List(project), Nil, testFilter, handler, false)
+    def parseParameters: Try[bsp.ScalaTestParams] = {
```
tgodzik: `parseParameters` would indicate a side effect function. Maybe `parsedParameters`?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -454,44 +453,64 @@ final class BloopBspServices(
...
-                case Right(result) =>
-                  val sequentialTestExecution = mappings.foldLeft(Task.now(newState)) {
-                    case (taskState, (tid, p)) => taskState.flatMap(state => test(tid, p, state))
-                  }
-
-                  sequentialTestExecution.materialize.map(_.toEither).map {
-                    case Right(newState) =>
-                      (newState, Right(bsp.TestResult(originId, bsp.StatusCode.Ok, None, None)))
-                    case Left(e) =>
-                      //(newState, Right(bsp.TestResult(None, bsp.StatusCode.Error, None)))
```
tgodzik: Remove?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -454,44 +453,64 @@ final class BloopBspServices(
...
+                      val filters = testParams.testClasses
+                        .map(x => (x.target, TestInternals.parseFilters(x.classes)))
+                        .toMap
+
+                      val userOptions = testParams.jvmOptions.map(opt => s"-J$opt")
+
+                      val sequentialTestExecution = mappings.foldLeft(Task.now(newState)) {
+                        case (taskState, (tid, project)) =>
+                          taskState.flatMap(state => {
+                            val testFilter = filters.getOrElse(tid, (_: String) => true)
```
tgodzik: `: String` is needed?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: It usually is the standard because almost all imports come from Metals, aggregating all of them in one import doesn't bother me, though I agree with the sentiment :smile: For the record, it is customary to avoid wildcard imports in the codebase, leaning always towards explicit imports.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: I'd be fine with either `parseTestParams` or `readTestParams` 👍 I think using `parsedParameters` is misleading because it's a `def` while I would expect it to be a value.

jvican:
Let's reopen this when we need this given that this is not critical and some BSP changes are still required upstream

PR #1029, opened 2019-09-07
Cross support for Scala Native 0.3 and 0.4.0-M2

lolgab:
This PR splits Scala Native bridge in two bridge, one for the 0.3.9 (stable) version and one for the 0.4.0-M2 (latest) version.


lolgab:
Seems like I can't reproduce the errors locally. Do you have any ideas?

.drone.yml
```diff
@@ -1,7 +1,6 @@
 # We publishLocal the maven plugin instead of compile to check the maven plugin integration
 # Only test sbt-bloop 1.x because of bug in `^`
 
-# TODO: re-add `nativeBridge/test`
```
jvican: The CI for JDK 11 is failing because we added back these tests, which is what we should do :smile: Would you mind trying to dig on the reasons why JDK 11 is doesn't work well with our Scala Native tests? Everything seems to work on JDK 8 (ignore Windows).

build.sbt
```diff
@@ -322,7 +322,7 @@ val docs = project
     name := "bloop-docs",
     moduleName := "bloop-docs",
     skip in publish := true,
-    scalaVersion := "2.12.6",
+    scalaVersion := "2.12.9",
```
jvican: Maybe replace by the almost always up-to-date `Dependencies.Scala212Version`? We can update the codebase to 2.12.9 in another PR.

bridges/scala-native-0.4/src/test/scala/bloop/scalanative/ScalaNativeToolchainSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+
+import bloop.data.{Platform, Project}
+import bloop.engine.tasks.toolchains.{ScalaJsToolchain, ScalaNativeToolchain}
+import bloop.util.TestUtil
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.experimental.categories.Category
+
+@Category(Array(classOf[bloop.FastTests]))
+class ScalaNativeToolchainSpec {
```
jvican: Would it be possible to change the project build definition so that we reuse the tests from the native bridge 0.3? Given that the tests are the same, we would avoid copying the toolchain spec.

jvican:
@lolgab Yes, let me look into the windows CI issue, it sometimes goes crazy and needs some caring.

jvican:
We’re good to go!

.drone.yml
ekrich: There are some issues around JDK 11, e.g. https://github.com/scala-native/scala-native/issues/1681

lolgab: Since JDK 8 is the official supported JDK by Scala Native I'm skipping the tests for that target.
When the bug is solved I can re-enable that.
Thank you for pointing out @ekrich

PR #1032, opened 2019-09-12
Add bloopgun CLI library and binary

jvican:
Bloopgun is a Scala library to talk to bloop via Nailgun. It can be compiled to a binary via GraalVM and be used as a (shaded) library. The first use case is essential for users that need a fast CLI experience, the second use case is essential for build clients that want to connect to Bloop quickly and robustly.
  
This pull request adds a lot of infrastructure necessary to make bloopgun possible as well as `bloop4j`, a bsp4j wrapper to talk to bloop via BSP which will be necessary for the sbt offloading.
  
On the bloopgun side, there is a new innovation compared to the previous Python bloop nailgun client: `bloopgun` will start the server if there is no server running. This is a game changer for users of the library because it means neither users nor library authors will need to manager a server running in the background.
  

Given that some of this was done previously by the launcher, that logic has been removed and the launcher instead relies on bloopgun. The launcher has become less of a launcher --now it only takes care of establishing a bap connection--, so I'm considering changing its name to something else.


jvican:
I'm not asking for review yet given that I'm experimenting/finishing some stuff on the bloop4j front. I'm probably going to leave this out of the minor release this week.

jvican:
Everything is more or less in place except for the GraalVM-native infrastructure in the CI. I'm going to hold off doing that for now given that it's quite difficult. I'll focus on modifying sbt-bloop to keep the resolution modules in the configuration file and this is good to go.

alexarchambault:
@jvican It seems sbt-scalafix and sbt-scalafmt are around in some projects here. They depend on coursier-small, itself depending on old versions of coursier, which don't play well concurrently. sbt-scalafmt and sbt-scalafix should switch to [coursier-interface](https://github.com/coursier/interface).

olafurpg:
I opened https://github.com/scalacenter/sbt-scalafix/pull/43 to migrate sbt-scalafix from coursier-small. I'll do the same for sbt-scalafmt.

jvican:
@alexarchambault Thanks for the pointers. We don't use sbt-scalafix but we do use sbt-scalafmt. Unfortunately, disabling sbt-scalafmt doesn't remove the error. Disabling the `CoursierPlugin` on the offending project did work in a previous attempt, though, for a part of the build. Any ideas what could be wrong here? I tried to see where another coursier version gets pulled in but I couldn't find it.
  
I should say that `/root/.coursier` is a symlink and it has been working just fine previously. After the upgrade to 2.0.0-RC3-3, things have started to go wrong, so I really wonder what's happening here :thinking:

olafurpg:
PR migrating scalafmt to coursier/interface https://github.com/scalameta/scalafmt/pull/1510

jvican:
Alright, I've run out of hacks. I downgraded coursier and the issue persists, this looks like a bug inside sbt-shading. I'm going to reimplement the functionality I care about in my build and remove the dependency of sbt-shading on coursier by just populating `toShadeJars` with files from `dependencyClasspath` and then filtering which jars should be shaded.

alexarchambault:
@jvican Did you check [sbt-declarative-packaging](https://github.com/JetBrains/sbt-idea-plugin#packaging) (via [this](https://twitter.com/IntelliJScala/status/1176520137194844162))?

alexarchambault:
but yeah, doing that manually might actually be more straightforward

olafurpg:
sbt-scalafmt upgrade to a no-coursier-small dependency https://github.com/scalameta/sbt-scalafmt/pull/50

jvican:
From what I've gathered in the whole day, the error here seems to be caused by a coursier version higher than 1.1.0-M3-2. My intuition is that coursier doesn't like the fact that the coursier cache is a symlink in those versions, while in the current version seems to work OK.
  
I ended up anyway implementing sbt-shading inside bloop, it makes it easier to modify what jars should be shaded and which should be not. This is just a temporary workaround, in the future I'd like to have an independent sbt-plugin built on top of coursier's sbt-shading. For now, we're taking this shortcut.
  
CI seems happy after the changes I did. This is good to go, so I'm merging as soon as the CI is green.

olafurpg:
🥳👏👏👏

PR #1033, opened 2019-09-12
Play around with Github actions

jvican:
Let's see if we can migrate away from Drone and Jenkins once and for all.


jvican:
Closing for now, I'll reopen whenever GitHub implements caching.

PR #1036, opened 2019-09-24
Assort some fixes to the compilation engine

jvican:


frontend/src/test/scala/bloop/bsp/BspBaseSuite.scala
```diff
@@ -340,7 +340,7 @@ abstract class BspBaseSuite extends BaseSuite with BspClientTest {
   def createBspCommand(configDir: AbsolutePath): Commands.ValidatedBsp = {
     protocol match {
       case BspProtocol.Tcp =>
-        val portNumber = 5202 + scala.util.Random.nextInt(2000)
+        val portNumber = 7001 + scala.util.Random.nextInt(20000)
```
olafurpg: Would it be possible to find an open port using `portNumber=0`?

jvican: We're testing the bsp server as a blackbox, if we pick 0 here we don't know what port will the server pick from the client side and therefore cannot materialise the connection.

jvican:
I'll address the flaky tests on the Windows side tomorrow in another PR. Let's merge this for now.

PR #1037, opened 2019-09-25
Increase the debuggee startup timeout

marek1840:
In some cases this timeout might be too low (e.g. when the task was not yet started). This seems like a probable cause for failing tests on [metals CI](https://github.com/scalameta/metals/pull/942). Increasing timeout here does not seem to be a major issue, since previous value already was quite arbitrary.

### DIFF
--- a/frontend/src/main/scala/bloop/dap/DebugSession.scala
+++ b/frontend/src/main/scala/bloop/dap/DebugSession.scala
@@ -109,7 +109,7 @@ final class DebugSession(
           .map(DebugSession.toAttachRequest(requestId, _))
           .foreachL(super.dispatchRequest)
           .timeoutTo(
-            FiniteDuration(5, TimeUnit.SECONDS),
+            FiniteDuration(15, TimeUnit.SECONDS),
             Task {
               val response = DebugSession.failed(request, "Could not start debuggee")
               this.sendResponse(response)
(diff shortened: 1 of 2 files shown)


frontend/src/main/scala/bloop/dap/DebugSession.scala
```diff
@@ -109,7 +109,7 @@ final class DebugSession(
           .map(DebugSession.toAttachRequest(requestId, _))
           .foreachL(super.dispatchRequest)
           .timeoutTo(
-            FiniteDuration(5, TimeUnit.SECONDS),
+            FiniteDuration(15, TimeUnit.SECONDS),
```
tgodzik: Is it possible to send this as a parameter?

marek1840: I would love to make it a parameter. @jvican is there a way of achieving this without a lot of hassle? There is more than one timeout in the debug adapter, so being able to set it all would be nice (also would speed up some tests probably).

jvican: >  is there a way of achieving this without a lot of hassle?
  
I would prefer not to allow any customisation here. It's very unlikely a user would like to modify this parameter. If you have a better reason why someone would want to change it then I'm all ears.
  
>  (also would speed up some tests probably).
  
How would customisation help speeding up tests? IIRC this is a timeout that only runs when attaching to the JVM failed. I don't see why a test case would be blocked on this timeout.

PR #1039, opened 2019-09-27
[debug] Switch to sbt-coursier 1.1.0-M14-3

alexarchambault:
Debugging the issues we were seeing in https://github.com/scalacenter/bloop/pull/1032.

sbt-coursier `1.1.0-M14-3` depends on coursier `1.1.0-M14-5`, which touches code handling
the creation of the coursier cache directory.

The CI of bloop was running into these issues rather consistently, right?

PR #1041, opened 2019-09-28
[debug] Switch to sbt-coursier 1.1.0-M14-2

alexarchambault:
Just a check, as a follow up to https://github.com/scalacenter/bloop/pull/1039. This is the coursier and sbt-coursier versions right before `1.1.0-M14-5` (coursier) / `1.1.0-M14-3` (sbt-coursier). So if the changes in those versions are really the problem, I would expect the PR here to pass.


alexarchambault:
Ok, seems the scripted job ran fine.

PR #1046, opened 2019-10-01
Migrate Windows CI to Github Actions

jvican:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,18 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest]
+    steps:
+      - uses: actions/checkout@v1
+      - uses: olafurpg/setup-scala@v4
+        with:
+          java-version: graalvm@
+      - run: git submodule update --init --recursive
```
gabro: `actions/checkout` has a `submodules` parameter that recursively clones also submodules, could that help?

olafurpg: or run the step separately

```yml
      - run: git submodule update --init --recursive
```

https://github.com/scalacenter/bloop/blob/0536c40217da480c4ae39f3b1cf47be77f144f14/.github/workflows/bloopgun.yml#L16

PR #1049, opened 2019-10-02
Redesign deduplication spec to avoid timeouts

jvican:


jvican:
Fixes https://github.com/scalacenter/bloop/issues/1051

PR #1063, opened 2019-10-11
Add all necessary SemanticDB options

tgodzik:
This is important if somebody already added a SemanticDB plugin, but not all the options required by metals.

Also added a check to not add `-Yrangepos` to Dotty.

### DIFF
--- a/frontend/src/main/scala/bloop/data/Project.scala
+++ b/frontend/src/main/scala/bloop/data/Project.scala
@@ -237,13 +237,10 @@ object Project {
       // engine so that semanticdb files are replicated in those directories
       val hasSemanticDB = hasSemanticDBEnabledInCompilerOptions(options)
       val pluginOption = if (hasSemanticDB) Nil else List(s"-Xplugin:$pluginPath")
-      val sourceRootOption =
-        if (hasSemanticdbSourceRoot(options)) {
-          Nil
-        } else {
-          List(s"-P:semanticdb:sourceroot:$workspaceDir")
-        }
-      (options ++ baseSemanticdbOptions ++ pluginOption ++ sourceRootOption).distinct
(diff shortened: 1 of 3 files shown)


frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
```diff
@@ -142,8 +143,8 @@ class BspMetalsClientSpec(
              |""".stripMargin
         )
 
-        val scalacOptions = state.scalaOptions(`A`)._2.items.head.options
-        val expected = defaultScalacOptions :+ "-Yrangepos"
+        val scalacOptions = state.scalaOptions(`A`)._2.items.head.options.toSet
```
marek1840: it would be usesful to also include the test for dotty

frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -227,27 +227,29 @@ object Project {
...
-          "-P:semanticdb:synthetics:on",
-          "-Xplugin-require:semanticdb",
-          s"-Xplugin:$pluginPath"
-        )
-
-        (options ++ semanticdbScalacOptions.toList).distinct
-      }
+      val pluginOption = if (hasSemanticDB) Nil else List(s"-Xplugin:$pluginPath")
+      val sourceRootOption =
+        if (hasSemanticdbSourceRoot(options)) {
```
olafurpg: I would filter out existing sourceroot settings if there are any. It's critical that the sourceroot matches the same directory as `.bloop`, otherwise code navigation wouldn't work.

tgodzik: Sure can do. @jvican what do you think?

tgodzik: Ok, change to replacing sourceroot.

olafurpg:
🎉

Issue #1059, opened 2019-10-09
Include Dotty fixes in Zinc


olafurpg:
Would it be possible to get fixes from https://github.com/sbt/zinc/pull/700 into the Bloop fork of Zinc?

PR #1065, opened 2019-10-14
Adjust only 1-based diagnostic positions

marek1840:
This fixes an issue with dotty diagnostics being treated as having 1-indexed lines, while dotty uses 0-based indexing.

Before:
before

After:
after


jvican:
I think this PR is not necessary, all we need is https://github.com/scalacenter/bloop/issues/1059

tgodzik:
> I think this PR is not necessary, all we need is #1059

Let's confirm this with new zinc released maybe? I can check that out as soon as we have that version merged.

jvican:
I need to do that myself as I need to publish it. Let's close this for the moment. If upgrading zinc doesn't work then we'll reopen.

PR #1066, opened 2019-10-14
Switch to sbt-coursier 2.0.0-RC4

alexarchambault:
This version includes https://github.com/coursier/coursier/pull/1389, that I would expect to fix the issues seen in https://github.com/scalacenter/bloop/pull/1032.


alexarchambault:
Ok, that won't fix. We're just running into https://bugs.openjdk.java.net/browse/JDK-8130464…

PR #1070, opened 2019-10-25
Fix spelling mistakes in documentation and code

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

### DIFF
--- a/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
+++ b/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
@@ -68,8 +68,8 @@ final class BloopAnalysisCallback(
   private[this] val classPublicNameHashes = new mutable.HashMap[String, Array[NameHash]]
   private[this] val objectPublicNameHashes = new mutable.HashMap[String, Array[NameHash]]
   private[this] val usedNames = new mutable.HashMap[String, mutable.HashSet[UsedName]]
-  private[this] val unreporteds = new mutable.HashMap[File, mutable.ListBuffer[Problem]]
-  private[this] val reporteds = new mutable.HashMap[File, mutable.ListBuffer[Problem]]
+  private[this] val unreportedProblems = new mutable.HashMap[File, mutable.ListBuffer[Problem]]
+  private[this] val reportedProblems = new mutable.HashMap[File, mutable.ListBuffer[Problem]]
   private[this] val mainClasses = new mutable.HashMap[File, mutable.ListBuffer[String]]
   private[this] val binaryDeps = new mutable.HashMap[File, mutable.HashSet[File]]
 
@@ -112,7 +112,7 @@ final class BloopAnalysisCallback(
       reported: Boolean
   ): Unit = {
     for (source <- InterfaceUtil.jo2o(pos.sourceFile)) {
-      val map = if (reported) reporteds else unreporteds
+      val map = if (reported) reportedProblems else unreportedProblems
       map
         .getOrElseUpdate(source, new mutable.ListBuffer())
         .+=(InterfaceUtil.problem(category, pos, msg, severity))
@@ -303,8 +303,8 @@ final class BloopAnalysisCallback(
           classNames.getOrElse(src, new mutable.HashSet[(String, String)]()).map(_._1)
         val analyzedApis = classesInSrc.map(analyzeClass)
         val info = SourceInfos.makeInfo(
-          getOrNil(reporteds, src),
-          getOrNil(unreporteds, src),
+          getOrNil(reportedProblems, src),
+          getOrNil(unreportedProblems, src),
           getOrNil(mainClasses, src)
         )
         val binaries = binaryDeps.getOrElse(src, Nil: Iterable[File])
--- a/docs/tools/scoop/usage.md
+++ b/docs/tools/scoop/usage.md
@@ -34,11 +34,11 @@ To learn more about `bloop server` and managing the server lifecycle automatical
 
 ### Command-line completions
 
-Bloop does not support Powershell completions at the moment. However, it does support Bash/Zsh
+Bloop does not support PowerShell completions at the moment. However, it does support Bash/Zsh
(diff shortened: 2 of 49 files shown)


frontend/src/test/scala/bloop/CliSpec.scala
```diff
@@ -154,7 +154,7 @@ object CliSpec extends BaseSuite {
   }
 
   test("fail at non-sensical host address") {
-    checkInvalidAddress("localhos")
+    checkInvalidAddress("localhost-")
```
jsoref: This is probably one of the more "interesting" changes. (Very few of the changes here are remotely interesting.)

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
```diff
@@ -68,8 +68,8 @@ final class BloopAnalysisCallback(
   private[this] val classPublicNameHashes = new mutable.HashMap[String, Array[NameHash]]
   private[this] val objectPublicNameHashes = new mutable.HashMap[String, Array[NameHash]]
   private[this] val usedNames = new mutable.HashMap[String, mutable.HashSet[UsedName]]
-  private[this] val unreporteds = new mutable.HashMap[File, mutable.ListBuffer[Problem]]
-  private[this] val reporteds = new mutable.HashMap[File, mutable.ListBuffer[Problem]]
+  private[this] val unreportedProblems = new mutable.HashMap[File, mutable.ListBuffer[Problem]]
```
jsoref: I'm not terribly certain about this change set. I could be wrong about what the objects are.

docs/tools/scoop/usage.md
```diff
@@ -34,11 +34,11 @@ To learn more about `bloop server` and managing the server lifecycle automatical
 
 ### Command-line completions
 
-Bloop does not support Powershell completions at the moment. However, it does support Bash/Zsh
+Bloop does not support PowerShell completions at the moment. However, it does support Bash/Zsh
```
jsoref: This is a brand.

frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
```diff
@@ -459,14 +459,14 @@ object CompileTask {
 
   /**
    * Prepares a clean up task that will delete the previous used read-only
-   * classes directory used as the read-only directory of an effectful compile.
+   * classes directory used as the read-only directory of an effective compile.
```
jsoref: I'm not at all certain about this.

launcher/src/main/scala/bloop/launcher/bsp/BspBridge.scala
```diff
@@ -183,7 +183,7 @@ final class BspBridge(
   def wireBspConnectionStreams(socket: Socket): Unit = {
     @volatile var isConnectionOpen: Boolean = true
 
-    def closeInconditionally(c: Closeable): Unit = {
+    def closeUnconditionally(c: Closeable): Unit = {
```
jsoref: I hope this isn't a problem.

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
jvican: Yeah, this is fine :+1:

frontend/src/test/scala/bloop/CliSpec.scala
jvican: LGTM, checking for `localhost-` should be the same error as checking for `localhos`.

frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
jvican: This one should actually be `effectful` instead of `effective`. Effectful is an adjective used to mean some computation that side-effects. Maybe we could replace it by side-effecting instead.

launcher/src/main/scala/bloop/launcher/bsp/BspBridge.scala
jvican: This is fine, thanks!

frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
jsoref: I'd like to replace it with something, but I think I need more information (I'm not using bloop right now, we're merely investigating a number of technologies and one of the things I do is investigate their ecosystems).

It sounds like you have a source directory, and an objdir (to avoid dirtying the source dir), and then you take that objdir and toss it into the classpath while compiling a new objdir.

And here you're merely talking about that intermediate directory.

There are other issues w/ this paragraph:
* I suspect that `previous` should be `previously`.

* The right hand side of the either-or here is flawed:
> If a compilation is a no-op or its used counter is bigger than zero, the
> deletion is skipped because in both scenarios the classes directory is
> still useful for either a next compile or is being used by an alternative
> running compilation process.

It should be possible to read it as:

> If a compilation is a no-op or its used counter is bigger than zero, the
> deletion is skipped because in both scenarios the classes directory is
> still useful for ... is being used by an alternative
> running compilation process.

But that's clearly nonsensical. (Also, the left hand side of that either-or isn't much better.)

For now, I'm going to take the pragmatic approach and drop the commit entirely. That will leave this review comment thread orphaned, but someone can still fix it. They just need to spend real time on it, and that's inappropriate for a spelling PR.

jsoref: For the record, this commit is now dropped. I've moved the problem to #1075

jvican: Sorry, just saw this message. What you've done in that PR looks good to me, thanks! @jsoref Let me know if you have any questions regarding Bloop as a technology. Feel free to drop by our Discord channel https://discord.gg/vys6d43 and ask questions

PR #1071, opened 2019-10-28
Set Bloopgun to use deamons

tgodzik:
As a follow up from https://gist.github.com/jvican/a4a081718ac330de3aa734fdfd665085

It seems those threads were keeping the connection alive and not allowing Metals to close.

@jvican Not exactly sure why we didn't run those as deamons, since anything starting Bloop launcher will not be able to kill those threads.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: Continuous Integration
 on:
   push:
+  pull_request:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


jvican:
It looks like the launcher tests are failing reliably. Can you  look into what's going on @tgodzik ? Maybe now that these threads are daemons they outlive the tests and we should find a way to reliably kill them?

tgodzik:
> It looks like the launcher tests are failing reliably. Can you look into what's going on @tgodzik ? Maybe now that these threads are daemons they outlive the tests and we should find a way to reliably kill them?

I will try to figure that out after I come back.

tgodzik:
The issue seems to be caused by setting daeomn in `establishBspConnectionViaBinary`, which wasn't neccessary.  The issues come from `"bsp-client-to-server"` shell thread, so I will leave daemon there.

tgodzik:
@jvican seems to be fixed - we might want to merge this and rebase @marek1840 PRs - the tests are not working there either.

PR #1073, opened 2019-10-30
Detect JDI initial notification on Windows

### LINKED ISSUE #1072: Make forker aware of JDI hardcoded '\n' message
Currently, the output from the JVM is being split into lines using the `System.lineSeparator` as a separator. 

Unfortunately, this stopped being a correct approach with the [introduction of a Debug Adapter Protocol](https://github.com/scalacenter/bloop/pull/944) where we depend on a first line from the JVM - a JDI notification telling us when we can attach to the debuggee. 

Regrettably, the JVM itself uses [hardcoded unix-style newline](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/back/error_messages.c#l119) when [sending the desired notification](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/back/transport.c#l508). It makes the debugging [not working on Windows](https://github.com/scalameta/metals/issues/1021) because of a deadlock:
1. we are waiting for `\r\n` to break the first line and attach to the debuggee
2. the debuggee cannot produce more output, since it is suspended until we attach to it

marek1840:
This makes the `Forker` split the process output on both unix- and windows- style newlines, regardless of the underlying OS.

Motivation here is that we cannot make the assumption on the basis of the underlying OS, because there is no guarantee that the application is intended to run on the same OS it is being developed on.

It changes the current behavior only in cases where the application produces an output with mixed newline style.

### DIFF
--- a/frontend/src/main/scala/bloop/dap/DebugSession.scala
+++ b/frontend/src/main/scala/bloop/dap/DebugSession.scala
@@ -5,6 +5,7 @@ import java.util.concurrent.CancellationException
 import java.util.concurrent.TimeUnit
 import java.util.logging.{Handler, Level, LogRecord, Logger => JLogger}
 import bloop.logging.{DebugFilter, Logger}
+import bloop.cli.{ExitStatus => DebuggeeExitStatus}
 import com.microsoft.java.debug.core.adapter.{ProtocolServer => DapServer}
 import com.microsoft.java.debug.core.protocol.Messages.{Request, Response}
 import com.microsoft.java.debug.core.protocol.Requests._
@@ -84,12 +85,19 @@ final class DebugSession(
           }
         }
 
+        def cancelIfError(exitStatus: DebuggeeExitStatus): Unit = {
+          if (!exitStatus.isOk) {
+            cancelPromises(new Exception(exitStatus.name))
+          }
+        }
+
         Task(super.run()).runAsync(ioScheduler)
 
         val logger = new DebugSessionLogger(this, addr => debugAddress.success(addr), initialLogger)
 
         // all output events are sent before debuggee task gets finished
         val debuggee = runner(logger)
+          .transform(cancelIfError, cancelPromises)
           .doOnFinish(terminateGracefully)
           .doOnCancel(terminateGracefully(None))
           .runAsync(ioScheduler)
@@ -113,7 +121,7 @@ final class DebugSession(
             case Success(address) =>
               super.dispatchRequest(DebugSession.toAttachRequest(requestId, address))
             case Failure(exception) =>
-              val cause = s"Could not start debuggee due to ${exception.getCause}"
+              val cause = s"Could not start debuggee due to: ${exception.getMessage}"
               this.sendResponse(DebugSession.failed(request, cause))
               attachedPromise.tryFailure(new IllegalStateException(cause))
               ()
@@ -147,6 +155,7 @@ final class DebugSession(
 
           state.transform {
             case DebugSession.Started(debuggee) =>
+              cancelPromises(new CancellationException("Client disconnected"))
               cancelDebuggee(debuggee)
               super.dispatchRequest(request)
               DebugSession.Cancelled
@@ -230,6 +239,7 @@ final class DebugSession(
         DebugSession.Cancelled
 
       case DebugSession.Started(debuggee) =>
+        cancelPromises(new CancellationException("Debug session cancelled"))
         cancelDebuggee(debuggee)
         scheduleForcedEndOfConnection()
         DebugSession.Cancelled
@@ -239,11 +249,14 @@ final class DebugSession(
   }
 
   private def cancelDebuggee(debuggee: Cancelable): Unit = {
-    val cause = new CancellationException("Debug session cancelled")
-    debugAddress.tryFailure(cause)
-    attachedPromise.tryFailure(cause)
     loggerAdapter.onDebuggeeFinished()
     debuggee.cancel()
+  }
+
+  private def cancelPromises(cause: Throwable): Unit = {
+    debugAddress.tryFailure(cause)
+    attachedPromise.tryFailure(cause)
+    ()
   }
 }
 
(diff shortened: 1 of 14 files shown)


marek1840:
> modify our waiting logic to detect the JDI notification (with both `System.lineSeparator` and `\n`)

Which waiting logic do you have in mind?

jvican:
> Which waiting logic do you have in mind?
  
I had in mind having a check before adding to `remaining` that would check if the buffer has the contents of `DebugSessionLogger.ListeningMessagePrefix`.
  
I like your reimplementation of this method, but I'm concerned about its implications in Windows. Except for the hardcoded message we're looking for, Windows users would expect lines with only LF characters to not be considered as  a new line character and this PR changes that expectation.
  
Maybe we can continue using this implementation, but modify it so that if the line is `\n` only and it's not the message we're looking for, we don't consider it a new line?

jvican:
@marek1840 In case I didn't explain it well, I meant reusing your new implementation.

frontend/src/main/scala/bloop/exec/Forker.scala
```diff
@@ -188,43 +190,46 @@ object Forker {
...
+      carry.clear()
+
+      @tailrec
+      def traverseLines(start: Int): Unit = {
+        if (start < msg.length) {
+          val lineEnd = msg.indexOf(System.lineSeparator(), start)
+          if (lineEnd < 0) {
+            // JVM send the JDI notification with "\n" as newline delimiter even on windows
+            if (CrossPlatform.isWindows && msg.startsWith(DebugSessionLogger.JDINotificationPrefix)) {
+              op(msg.stripLineEnd)
```
jvican: The message could be partially filled in the buffer, so I'm not sure this line would work as expected. I think we could make it work by doing `val jdiIdx = msg.indexOf(DebugSessionLogger.JDINotificationPrefix)` and then checking that LF exists with `msg.indexOf("\n", jdiIdx)`. Then, I think `op(msg.stripLineEnd)` should be correct

frontend/src/main/scala/bloop/exec/Forker.scala
```diff
@@ -188,43 +190,52 @@ object Forker {
...
           }
-        } else Array.empty[String]
+        }
+      }
+
+      traverseLines(0)
     }
   }
+
+  private def containsFullJdiNotification(msg: StringBuilder): Boolean = {
```
jvican: Very good :+1:

jvican:
This change seems to be reliably triggering a test error in Windows: https://github.com/scalacenter/bloop/pull/1073/checks?check_run_id=287902704#step:7:763 I've run the CI two times, let's see what happens with the third, but two times already hints that there's something wrong here.

frontend/src/main/scala/bloop/dap/DebugSession.scala
```diff
@@ -113,7 +121,7 @@ final class DebugSession(
             case Success(address) =>
               super.dispatchRequest(DebugSession.toAttachRequest(requestId, address))
             case Failure(exception) =>
-              val cause = s"Could not start debuggee due to ${exception.getCause}"
```
jvican: Was the cause not relevant?

frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
```diff
@@ -158,7 +154,6 @@ object DebugProtocolSpec extends DebugBspBaseSuite {
             _ <- client.initialize()
             _ <- client.launch()
             _ <- client.configurationDone()
-            _ <- client.exited
```
jvican: Why this removal?

frontend/src/test/scala/bloop/dap/DebugServerSpec.scala
```diff
@@ -1,28 +1,25 @@
 package bloop.dap
+
 import java.net.{ConnectException, SocketException, SocketTimeoutException}
 import java.util.NoSuchElementException
 import java.util.concurrent.TimeUnit.{MILLISECONDS, SECONDS}
-
-import bloop.bsp.BspBaseSuite
-import bloop.cli.BspProtocol
+import bloop.cli._
```
jvican: Let's use explicit instead of wildcard imports

frontend/src/main/scala/bloop/dap/DebugSession.scala
marek1840: `cause` is a throwable which underlies the `exception`. When printed, it would append the entire stack trace instead of just the message.

It should have been `getMessage` all along, since that is what we want to show to the user.

frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
marek1840: as per the commit message:
>It seems that graalvm is not sending the 'exited' event when running the test suite

Why it is not sending it? Unfortunately I don't know and investigating it further would quite too time consuming.

frontend/src/main/scala/bloop/dap/DebugSession.scala
jvican: But `getMessage` doesn't throw the `getMessage` of the cause exception, so we could be showing only the superficial `getMessage` here, not the one that actually caused the issue?

frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
jvican: This is a pity, I agree investigating this would be too time consuming. Let's hope this change doesn't introduce regressions on the way we handle `client.exited` in the future.

frontend/src/main/scala/bloop/dap/DebugSession.scala
marek1840: That is a question about building exceptions on top of other exceptions, so let me answer with an example.

Lets say, we want to parse a config file and we encounter a malformed integer "1foo2". The first exception here would be NumberFormatException with message like `invalid number "1foo2"`.
That is the underlying cause. But out parser knows that the property being deserialized is called "prop1", so we wrap the exception in another one with message: `Could not parse prop1 due to: ${underlying.getMessage}`. 

Of course, someone could have wrapped in an exception with message "parsing failed", but we cannot check that in the code and we have to trust the library authors and other developers to provide meaningful messages.

In summary, the top-level exception should contain a message with the broadest context, while the cause at the bottom should contain a very specific description of the problem but without any context.

jvican: You can't assume the wrapped exceptions are always good, they aren't, for example I've run lately into many exceptions whose cause were NPEs but that were not properly handled by the libraries (in this case, lsp4j). I agree with you that in principle this should be fine.
  
Let's trust we won't run into these issues here for now.

PR #1075, opened 2019-10-31
Doc cleanup clean up previous result

jsoref:
@jvican: Here's my attempt to address the thread in https://github.com/scalacenter/bloop/pull/1070#discussion_r340926145

Note: this will conflict with that PR. It's more as a placeholder for a discussion. I don't expect it to be merged as-is.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
@@ -462,7 +462,7 @@ object CompileTask {
   }
 
   /**
-   * Prepares a clean up task that will delete the previous used read-only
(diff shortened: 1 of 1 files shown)


frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
```diff
@@ -458,16 +458,21 @@ object CompileTask {
   }
 
   /**
-   * Prepares a clean up task that will delete the previous used read-only
-   * classes directory used as the read-only directory of an effectful compile.
+   * Prepares a clean-up task that will delete the previous classes directory
+   * used as the read-only directory for a compile which has generates artifacts
```
jvican: s/generates/generated/g

Issue #315, opened 2018-03-08
Offload compilation from sbt to bloop


jvican:
The idea is to have our sbt integration delegate the `compile`, `test` and `run` tasks to bloop via BSP. The benefit of this is that sbt can always reuse hot compilers, no matter if users `reload` their builds or open multiple sbt shells in different projects.
  
While `bloop` can still be used via CLI (and it's the recommended workflow, since sbt is just too memory consuming and blows up from time to time), this issue tries to integrate as much as possible with people's workflow so that they don't need to get used to working with the bloop CLI if they feel lazy or don't like it.

julienrf:
Is there any update on this? Do you plan to address this issue, and do you have an ETA?

jvican:
I'm narrowing down the scope of this ticket to only compilation, integrating bloop's test and bloop's run with sbt is a non-goal of this ticket.

rtar:
Would not this feature be awesome for using on continuous integration servers such as Jenkins? I.e. you have various sbt tasks one after another, interleaved with deployments etc. and then you can have your compiler there hot and waiting for sbt to use it.

PR #1078, opened 2019-11-04
Offload compilation from sbt to Bloop

jvican:
Fixes https://github.com/scalacenter/bloop/issues/315


jvican:
CI is passing. I'm going to merge this PR because all the documentation and the gifs will be in a follow-up PR. I want to start rebasing every change in the PR pipeline onto this one, too, which seems to fix some spurious errors we were experiencing in master. This PR doesn't include scripted tests for offloading compilation and there are still some missing docs which will be done in a follow-up PR.

kubukoz:
Super excited about this!

kluen:
@jvican Hi, I'm also very excited about this. How exactly can offloading be enabled? I wasn't able to find anything in the Bloop documentation, only [this concept](https://scalacenter.github.io/bloop/docs/offloading-compilation-sbt#offloading-compilation-from-sbt).
If this is not ready yet, are there any follow-up issues I could subscribe to?

tgodzik:
@kluen unfortunately it never got out of being experimental and there are not current plans to work on it further. Jorge is no longer working in the Scala Center, so it's doubtful he will have time to continue working on this.

I would love to take a look, but unfortunately there are much more pressing issues currently I need to deal with. If there is a broader community interest it would be possible to raise it as a proposal to the Scala Center advisory board through the community representatives.

igor-ramazanov:
The idea sounds really good, because it'll allow to waste much less resources.

I often use bloop for the local development and import projects into the IntelliJ IDEA as a BSP projects, so I use the same bloop process across multiple projects keeping it hot.

However, sometimes I need to run various sbt related tasks, like building Docker containers and I **have to** run another sbt process for that, basically, duplicating the work which could be avoided.

igor-ramazanov:
@badeball Have you taken a look at `sbtn`?

igor-ramazanov:
@badeball I'd say that the main purpose of `bloop` - is to be leaked.

It's a long-running background process which heavily utilizes incremental compilation and shares the same process across multiple projects - that's what makes it fast.

For CI:
* If it's ok to you to have a long-term running background `bloop` process in the CI, surviving multiple CI builds, then it'll speed up your builds. However, starting from the clean state could be a better idea.
* You can also keep the `bloop` process alive only during the CI session, but then it doesn't differ with having an `sbt` process. E.g. `sbt compile my-project/test` (actually, `sbt my-project/test` most probably depends on the `sbt compile`, so `sbt compile my-project/test` will have almost the same behaviour `sbt my-project/test`). You can spin up a background `sbt` process using `sbtn`.
* With CIs I think it's reasonable to take a look at caching compilation output. There's some progress made by @eed3si9n recently with remote caching.

https://eed3si9n.com/remote-caching-sbt-builds-with-bintray
https://databricks.com/blog/2019/02/27/speedy-scala-builds-with-bazel-at-databricks.html
https://github.com/sbt/sbt/releases/tag/v1.4.0-M1

badeball:
Hey, @igor-ramazanov, thanks for your tips! We've actually just implemented and utilized remote caching and sbtn now and it seems to work great. We're mounting machine-directories into each CI-job, to persist the "remote cache".

My idea / hope with bloop was to kind of have a dedicated & capable machine running that, in which compiling could be offloaded to, while still running tests in container-isolation. But my impression is that that is not possible (yet?).

tgodzik:
Unfortunately, that is not possible yet, but we had some ideas about it and we mostly lack the time currently.

PR #1086, opened 2019-11-10
Fix class file invalidation issue when moving sources

jvican:
This commit fixes an invalidation issue that occured when moving sources
around projects and then compiling these projects independently.  The
previous invalidation logic was only handling the case where users move
sources around when all projects involved are compiled together.
However, when they are compiled in different steps, the invalidation
logic was no longer acting correctly.

The fix included here uses the generated products in all dependent class
files to avoid invalidating class files mapped to symbols that have
already been found in at least one of the project dependencies. This fix
doesn't rely on any in-memory information and therefore correctly
handles the stepped compilation scenario.

There is more information about the fix in the source code via comments.


jvican:
Error in CI is spurious, this is good to go!

Issue #1057, opened 2019-10-09
Support debugging breakpoints in Bloop


jvican:
This ticket collects my thoughts on how we can implement breakpoint support in Bloop, what are the different implementation avenues, their pros/cons and specifies what is the relationship between Bloop as a DAP server and BSP clients.
  
### Constraints
  
1. Breakpoint support should not depend on information from the BSP/DAP frontends unless it's strictly necessary.
2. Bloop should handle breakpoints in user code as well as third-party dependencies for both Java and Scala source code.
3. This design document doesn't explore supporting fancy breakpoints that allow to evaluate expressions upon hits.
4. Exception breakpoints are outside the scope of this document (and they are already supported in the current implementation if the BSP/DAP frontend allows the user to set them).

### Implementation walkthrough
  
To implement *basic* breakpoint support in [`java-debug`](https://github.com/microsoft/java-debug), we must implement the `getFullyQualifiedName` method of the `ISourceLookupProvider`. This method maps a `(source uri, source position)` to a set of fully qualified names. These fully qualified names are then used by the java-debug implementation to look at all the loaded classes in the JDI `VirtualMachine` and set a breakpoint in them by resolving the source positions with the `LineNumberTable` and `SourceFile` information (documented [here](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.10)).
  
There are two important kind of URIs we need to support in `getFullyQualifiedName`:
  
1. Source file URIs located in the user workspace.
1. Source file URIs coming from a library dependency.
  
BSP clients such as Metals and IntelliJ might differ on how they present to us sources coming from binary dependencies. This might be an important detail to drive the implementation. For example, Metals will give us an actual file URI living under `$WORKSPACE/.metals/readonly` while IntelliJ might give us another kind of URI format.
  
### Finding FQNs for workspace sources
  
Implementing FQNs for workspace sources can be done through some of the information we host in the Zinc analysis file. The Zinc analysis file contains a mapping between source files and a list of the class files names generated from that source.
  
We have this information because Zinc has compiled the projects before debugging. This information is reliable and thus we can use it to power `getFullyQualifiedName`.

### Finding FQNs for third-party dependency sources
  
Mapping sources from third-party dependency sources to FQNs is more complicated. Given that we don't compile these projects, we don't have the Zinc mapping between sources and class file names. Heuristics such as relying on the source file path can only be used for Java source files, not Scala source files.
  
[...]

PR #1087, opened 2019-11-10
Implement breakpoint support for user Java/Scala sources

jvican:
This commit implements breakpoint support for user-provided Java and
Scala sources. It does not support breakpoint support for third-party
sources because those require the intervention of the tool managing them
(i.e. IDEs such as IntelliJ and Metals), see
https://github.com/scalacenter/bloop/issues/1057 for more information.

The way this support works is by reusing the mapping between sources and
class files found in the analysis files. Once we have the analysis files
mapped to a source, we parse them with ASM and extract all the lines in
the `LineNumberTable`, which effectively contain all lines containing
instructions that the user can add breakpoints on. We build a map of
these instruction lines to the fully qualified class name that owns it
and then return back to java-debug, which will do its job.

IDEs that are confident in the reliability of their data can always pass
us the fully qualified class names for some positions themselves via the
URI format "dap-fqcn:$FULLY_QUALIFIED_CLASS_NAME". This is the URI
encoding that should be used to tell Bloop what names should be used for
third-party sources. It is optional and not recommended to pass these
names for user-defined sources and instead let Bloop do its job as its
implementation is as reliable as it can get.

In addition to this, the commit changes some fundamental parts of the
test code to ensure that we can listen for several DAP event
notifications. The previous code only allowed to do `first` and block on
all types of an event. The introduced test changes in this PR allow a
richer, Monix-friendly way of querying the streams of events coming from
the DAP server and block on them only if necessary.


jvican:
Paging @marek1840 who has been working on integrating with these changes from the Metals side. Keep in mind that the format Bloop has finally chosen to allow clients to specify fully qualified names is `dap-fqcn:$FULLY_QUALIFIED_CLASS_NAME` because it's easier to parse than the proposed `jvm:fqcn:$FULLY_QUALIFIED_CLASS_NAME` in https://github.com/scalameta/metals/issues/1010.

Paging @olafurpg who might want to be in the loop about this.

olafurpg:
🥳

PR #1097, opened 2019-11-17
Resolve correct linker versions in non-JVM builds

### LINKED ISSUE #1062: Use linker version set in Bloop configuration
When a library was compiled with a newer Scala.js version than the [one](https://github.com/scalacenter/bloop/blob/master/project/Dependencies.scala#L43) used by Bloop, the module cannot be linked.

For example, https://github.com/sparsetech/toml-scala/pull/16 fails as it updates ScalaCheck from 1.14.0 to 1.14.2 which was compiled for Scala.js 0.6.29, however the linker version used by Bloop is 0.6.28. From the [CI log](http://ci.sparse.tech/sparsetech/toml-scala/67/1/2):

> [E] JavaScript linking failed with 'Failed to deserialize info of file compiled with Scala.js 0.6.29 (supported: 0.6.0, 0.6.13, 0.6.3, 0.6.17, 0.6.4, 0.6.6, 0.6.14, 0.6.8, 0.6.5, 0.6.15): /drone/src/build/bloop/toml-js-test/toml/CodecSpec$$anonfun$75$Root$101.sjsir' 

In any case, the linker version set in the Bloop configuration should be honoured as this would allow users to downgrade in case of regressions.

tindzk:
The linker versions are still hard-coded in `Dependencies.scala`, but have
been annotated with sbt's `Provided` scope. When the bridges are resolved
now, the platform-specific libraries have to be included explicitly which
allows us to specify the same version as in the user's build configuration.

Closes #1062.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/toolchains/ScalaJsToolchain.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/toolchains/ScalaJsToolchain.scala
@@ -4,6 +4,7 @@ import java.lang.reflect.InvocationTargetException
 import java.net.URLClassLoader
 import java.nio.file.Path
 
+import bloop.DependencyResolution
 import bloop.config.Config
 import bloop.config.Config.JsConfig
 import bloop.data.Project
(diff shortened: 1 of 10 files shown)


frontend/src/main/scala/bloop/engine/tasks/toolchains/ScalaJsToolchain.scala
```diff
@@ -121,13 +122,27 @@ object ScalaJsToolchain extends ToolchainCompanion[ScalaJsToolchain] {
   }
 
   override def artifactNameFrom(version: String): String = {
-    if (version.startsWith("0.6")) BuildInfo.jsBridge06
+    if (version.length == 3) sys.error("The full Scala.js version must be provided")
```
tindzk: I added this check because up until now it was sufficient to only specify 0.6 or 1.0 in the build configuration. Seed uses this convention, so old builds will be incompatible. The sbt plug-in is not affected, so I think we can avoid adding code to fall back to a specific Scala.js version.

tindzk:
Do we need a test case for this? I could copy `cross-test-build-scalajs-0.6` and change the Scala.js version to 0.6.29 which has a different IR.

frontend/src/test/scala/bloop/io/ClasspathHasherSpec.scala
```diff
@@ -22,11 +22,15 @@ object ClasspathHasherSpec extends bloop.testing.BaseSuite {
     val cancelPromise = Promise[Unit]()
     val cancelPromise2 = Promise[Unit]()
     val tracer = BraveTracer("cancels-correctly-test")
-    val jars = {
-      DependencyResolution.resolve("org.apache.spark", "spark-core_2.11", "2.4.4", logger) ++
-        DependencyResolution.resolve("org.apache.hadoop", "hadoop-main", "3.2.1", logger) ++
-        DependencyResolution.resolve("io.monix", "monix_2.12", "3.0.0", logger)
-    }
+    val jars =
```
jvican: The goal of the previous code was to force independent resolutions for every artifact so that the resolution of shared libraries is not shared. Is there a motivation to change that?

tindzk: I can change it back in this file. The `resolve()` function was modified to take a list of artefacts such that we can resolve platform libraries in one go.

jvican: The changes in the resolve function are good, but this use site of that should not resolve everything together, so let’s pass a list of one element per every invocation.

tindzk:
I ran `nativeBridge03` and `nativeBridge04` locally and both suites passed.

PR #1098, opened 2019-11-18
Update cwd to be project.workSpaceDirectory.getOrElse(project.baseDirectory) - modifies fix for #724

er1c:
Haven't gotten my local compile/test config setup, so hoping the CI will work :)

### DIFF
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -541,7 +541,7 @@ object Interpreter {
 
   private def run(cmd: Commands.Run, state: State): Task[State] = {
     def doRun(project: Project)(state: State): Task[State] = {
-      val cwd = project.baseDirectory
+      val cwd = project.workspaceDirectory.getOrElse(project.baseDirectory)
       compileAnd(cmd, state, List(project), false, "`run`") { state =>
         getMainClass(state, project, cmd.main) match {
           case Left(failedState) => Task.now(failedState)


frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -541,7 +541,7 @@ object Interpreter {
 
   private def run(cmd: Commands.Run, state: State): Task[State] = {
     def doRun(project: Project)(state: State): Task[State] = {
-      val cwd = project.baseDirectory
+      val cwd = project.workspaceDirectory.getOrElse(project.baseDirectory)
```
olafurpg: I believe the current Bloop behavior is working as expected, it mirrors the sbt behavior with `fork := true`. What happens if you configure the working directory with `javaOptions` in build.sbt?

```scala
    javaOptions += s"-Duser.dir=${baseDirectory.in(ThisBuild).value}",
```

With this change, I was able to get the working directory in `bloop run` to the same directory as where build.sbt is instead of the base directory of the project.

er1c: IIRC, the sbt `fork := true` still honors the `baseDirectory` configuration setting (which also seems to be supported via https://www.scala-sbt.org/1.x/docs/Forking.html) . I have some additional details here: https://github.com/scalacenter/bloop/issues/724

But the issue is in a project with multiple `.dependsOn`, the `baseDirectory` sbt `baseDirectory` setting should still allow you to change the `cwd` via the `baseDirectory` setting.  

```
  baseDirectory in (Compile, run) := file("."),
  baseDirectory in (Test) := file("."), // fork sets the working to the subproject, so a test can't read the config_crypto_keys
```

(I just noticed I even have a comment about the fork behavior above)

I can try the javaOptions configuration, but this behavior still diverges from `sbt`

olafurpg: If configuring `javaOptions` is a suitable workaround then I think it would be fine to recommend that as the official solution for customizing the working directory. The changes in this PR also diverge from sbt behavior. I don't think it would be a desirable to port the whole task/config/project axis scoping rules from sbt to bloop.

er1c: Should I close this PR and create a separate issue? perhaps a new "cwd" property needs to get added to the `.bloop/foo.json`?

er1c: This appears to work:

```
  baseDirectory in (Compile, run) := file("."),
  baseDirectory in (Test) := file("."), // fork sets the working to the subproject, so a test can't read the config_crypto_keys
  javaOptions += s"-Duser.dir=${baseDirectory.in(ThisBuild).value}", // metals support for baseDirectory 
```

PR #1100, opened 2019-11-18
Finish up hydra compiler support

dotta:
@jvican Here are the changes needed in Bloop for Hydra integration to be complete and fully working 🎉 

There is one `FIXME` note as I think it would be nice to allow users to configure the number of Hydra workers used, but I don't think it's a blocker for this PR. We can address it later on.

By the way, the Zinc commit tracked by Bloop should be upgraded to include https://github.com/scalacenter/zinc/pull/12. And, one more thing, I can release the bloop-hydra-bridge 0.0.2 as soon as https://github.com/scalacenter/zinc/pull/12 is merged.

\cc @dragos


frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -285,14 +285,12 @@ object Project {
     // Project is unique and derived from project name and ivy project configuration
     val hydraTag = project.name
 
-    val hydraRootBaseDir = workspaceDir.resolve("hydra").createDirectories
+    val hydraRootBaseDir = workspaceDir.resolve(".hydra").createDirectories
     val hydraBaseDir = hydraRootBaseDir.resolve("bloop").createDirectories
```
dotta: @jvican Ideally, we should also set the location of the `hydra.log`, as currently the hydra.log is stored inside the bloop installation folder. The difficulty is that to configure it we use a system property `-Dhydra.logFile`, and since Bloop spawns a single JVM I'm not sure how we can properly set the location as it can be different depending on the project being compiled. In essence, for a given `Project`, here is the value to set for the system property:

```
System.setProperty("hydra.logFile", hydraBaseDir + File.separator + "hydra.log")
```

Do you have any suggestion on how to do it? This is definitely not a blocker, but it'd be nice we sort it out eventually :)

backend/src/main/scala/sbt/internal/inc/test.worksheet.sc
```diff
@@ -0,0 +1,5 @@
+val a: Vector[String] = Vector()
+a match {
```
jvican: @dotta When you add the tests, could you remove this file too please? :smile:

PS. I rebased the whole PR so that when merged it goes directly to master.

dotta: Sounds good! I'm OOO atm and will be back on Thu. Hopefully, I can have it all done by the end of the week!

backend/src/main/scala/sbt/internal/inc/BloopComponentCompiler.scala
```diff
@@ -38,6 +39,39 @@ object BloopComponentCompiler {
   final val javaClassVersion = System.getProperty("java.class.version")
 
   final lazy val latestVersion: String = BloopComponentManager.version
+
+  object Hydra {
+    val bridgeNamePrefix =
+      sys.props.get("bloop.hydra.bridgeNamePrefix").getOrElse("bloop-hydra-bridge")
```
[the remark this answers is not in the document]
dotta: Great. I'll try this out and will make sure to include this in our doc. Thanks!

frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -303,8 +301,10 @@ object Project {
     val metricsDirOption = ("-YhydraMetricsDirectory", hydraMetricsDir.syntax)
     val hydraTagOption = ("-YhydraTag", hydraTag)
     val sourcepathOption = ("-sourcepath", hydraSourcepath)
-    val sourcePartitionerOption = ("-YsourcePartitioner", hydraSourcePartitioner)
+    val sourcePartitionerOption = (s"-YsourcePartitioner:$hydraSourcePartitioner", "")
 
+    // FIXME: Make the number of workers configurable (2 should stay the default)
+    val hydraCpus = ("-cpus", "2")
```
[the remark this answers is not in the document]
dotta: That works for me!

frontend/src/main/scala/bloop/data/Project.scala
dotta: Adding a setting is simple, the difficulty is propagating the value so that it's used by logback. I think this is a non trivial effort for little gain, so I'd say we can just keep things as they are and we will document where the hydra log is stored when using Bloop. Let's keep it simple :)

jvican: Agreed! Let's make sure that the logback setup you have in Hydra is friendly to concurrent Hydra compilations in another build. For example, if I was compiling two projects in two different builds at the same time, would logback correctly redirect the logs of each compilation to independent files?

dotta: No, right now it would all end up in the same log file. But I think we can live with this limitation. I believe the alternative would be to take advantage of classloaders' isolation and have a different logback.xml in each compilation process, configured so that the hydra.log is stored in the `<project-root>/.hydra/bloop` dir. It's a bit of work and I'd prefer not to make this a blocker. The hydra.log is only relevant if there are issues, and if that occurs we can always guide users to remove any artifact due to concurrent compilation (if that is needed, which I really think it won't be needed in most if not all cases). Sounds good?

jvican: Another way to work around this is not to use globally configured logging frameworks such as logback. I don't think it's a blocker, but it's definitely nicer to have no global state in the compiler class loader. Specifically, I'm concerned about that global state leaking out to the bloop classloader. It doesn't look like that should happen... so I'm fine proceeding as you point out. But this interaction is something to have in mind in case people start opening tickets about lost messages or logging-related issues.

dotta:
@jvican Just in case, I haven't forgot about this :) I have one failing test that I need to debug, but I'm almost there. Hopefully, on Monday!

jvican:
Looking forward to getting this PR done tomorrow then @dotta I'd like to cut a RC1 release and would love to include Hydra support as a highlight.
  
Could you rebase this PR and remove the Zinc version change? I've just released a version that contains the changes we merged in scalacenter/zinc + another change to get support for Dotty rendered problems. That version is now used in this PR https://github.com/scalacenter/bloop/pull/1124 and intend to merge it today if everything is green.

dotta:
> Looking forward to getting this PR done tomorrow then @dotta I'd like to cut a RC1 release and would love to include Hydra support as a highlight.

Excellent! And I have good news: I found and fixed the bug that was making that one test I mentioned failing. I'll push everything by this evening (I just need to do some cleanup first).

> Could you rebase this PR and remove the Zinc version change? I've just released a version that contains the changes we merged in scalacenter/zinc + another change to get support for Dotty rendered problems. That version is now used in this PR #1124 and intend to merge it today if everything is green.

Will do!

dotta:
@jvican I'm done. I've rebased on top of master and force pushed. But to make it easier for you to review what changed, I've pushed additional commits without doing any squashing. I also have a scripted test ready, but it requires hydra v2.1.14 to be released first (we are aiming for tomorrow or Wednesday at latest), so I'll open a PR the moment it will work.

dotta:
Btw, here is the scripted test: https://github.com/dotta/bloop/commit/93842b0590a864aeb0bb23ee2a16e0ea0e2e9637

jvican:
Thanks, will take a look, it's not customary to squash in bloop, so no worries about that :+1:

dotta:
> Thanks, will take a look, it's not customary to squash in bloop, so no worries about that 👍

Sounds good. And of course I didn't re-compile after rebasing, so I'm having a look now to resolve the compilation errors...

dotta:
@jvican I've pushed another commit that should address the compilation failure introduced by the rebase. However, I noticed that running `test:compile` on master fails because of several errors in the module `gradle-bloop` (as far as I can tell this module is not compiled by the CI). Just letting you know in case you weren't aware.

frontend/src/main/scala/bloop/data/Project.scala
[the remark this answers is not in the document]
dotta: Alright, I've included a system property as it was the simpler and quickest thing I could do. However, I'd still like to include a CLI option as the one you suggested, but this can come in a separate PR. I checked how `--pipeline` is processed in `Bloop.scala` and `BloopBspServices.scala`, and I guess I need to do something similar to support a configurable number of Hydra cpus. With the additional complexity in `Bloop.scala` that I should handle the combination of `--pipeline` with `--hydra-cpus` as the current implementation of the CLI options parsing is fairly ad-hoc. Does it sound right?

dotta:
@jvican This PR is ready for review. There is one failure on the CI but I assume it's a spurious one.

frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -30,7 +30,7 @@ object BuildLoader {
   def readConfigurationFilesInBase(base: AbsolutePath, logger: Logger): List[AttributedPath] = {
     val workspaceFileName = WorkspaceSettings.settingsFileName.toFile.getName()
     bloop.io.Paths
-      .attributedPathFilesUnder(base, JsonFilePattern, logger, 1)
+      .attributedPathFilesUnder(base, JsonFilePattern, logger, 1, ignoreVisitErrors = true)
```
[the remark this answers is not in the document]
dotta: That change was made by you :) See commit https://github.com/scalacenter/bloop/pull/1100/commits/9943709b349bff42b92d2474e62d940c80f7a449

jvican:
Hydra compilation is failing @dotta

frontend/src/main/scala/bloop/engine/BuildLoader.scala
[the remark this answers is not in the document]
jvican: LOL, definitely did not remember neither the change nor the rationale. I'll try to think more about it later when I'm less sick.

jvican:
Merging, I think secrets are not being exposed and that's what makes the CI fail. Let's hope that upon merging the secret is available and the CI jobs succeed.

jvican:
@dotta Thanks for all the work you've done here, I appreciate it. I'm excited to have first-citizen Hydra support in Bloop.

dotta:
@jvican It's been a true delight to work on this together. A sincere thanks for all the help and guidance.

Very excited to have Bloop and Hydra working hand in hand. We are ready to melt all those CPUs! ;-)

PR #1102, opened 2019-11-18
Avoid classpath separator in classes output directory.

### LINKED ISSUE #1101: Classes directory should not contain colon
Steps to reproduce: follow instructions in this repo here https://github.com/olafurpg/intellij-bsp-pants

The problem manifests in error messages like

```
Error: Could not find or load main class src.Main
```

when testing and running a Pants build that was exported to Bloop and then imported into IntelliJ via BSP. The problem seems caused by the combination of

* Pants targets can have colons `:` in their names such as `src:src` and the Pants target name is used verbatim as the Bloop project name
* IntelliJ BSP uses `"clientClassesRootDir": "file:/Users/lgeirsson/dev/pants/out/"` as an initialization parameter

I believe this issue would be fixed if Bloop processes away the path separator `:` when creating new classes directories for `clientClassesRootDir`

olafurpg:
Fixes #1101. Previously, when project names included classpath
separators like `:` then it was not possible to run/test via IntelliJ.
Now, we replace path separators in the classes output directory with
underscore `_`.

### DIFF
--- a/frontend/src/main/scala/bloop/data/ClientInfo.scala
+++ b/frontend/src/main/scala/bloop/data/ClientInfo.scala
@@ -3,6 +3,7 @@ package bloop.data
 import bloop.io.AbsolutePath
 import bloop.util.UUIDUtil
 
+import java.io.File
 import java.nio.file.Files
 import java.util.concurrent.ConcurrentHashMap
 import java.io.PrintStream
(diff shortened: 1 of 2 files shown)


olafurpg:
Manually tested that this fixes the IntelliJ issue

Screenshot 2019-11-18 at 23 17 39

PR #1110, opened 2019-12-01
Add missing dependency on scala-library in shaded bloopgun

### LINKED ISSUE #1109: Shaded Bloopgun is missing a dependency on scala-xml
In a computer without Bloop installed:
```
coursier launch ch.epfl.scala:bloopgun_2.12:1.3.4+226-153862a1 -- --help
No server running at 127.0.0.1:8212, let's fire one...
Resolving ch.epfl.scala:bloop-frontend_2.12:1.3.4+226-153862a1...
java.lang.NoClassDefFoundError: scala/xml/Node
	at bloop.shaded.coursier.core.Parse$.<init>(Parse.scala:62)
```

A workaround is to add scala-xml to the dependencies
```
coursier launch ch.epfl.scala:bloopgun_2.12:1.3.4+226-153862a1 org.scala-lang.modules:scala-xml_2.12:1.2.0 -- --help
```

Another workaround is to use the unshaded bloopgun

```
coursier launch ch.epfl.scala:bloopgun-core_2.12:1.3.4+226-153862a1 -- --help
```

olafurpg:
Fixes #1109. The shaded bloopgun already depends on scala-library so it
should be safe to also include scala-xml on the classpath.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -363,7 +363,8 @@ lazy val bloopgunShaded = project
     fork in run := true,
     fork in Test := true,
     bloopGenerate in Compile := None,
-    bloopGenerate in Test := None
+    bloopGenerate in Test := None,
+    libraryDependencies += Dependencies.scalaXml
   )
 
 lazy val launcher: Project = project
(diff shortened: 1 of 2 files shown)


olafurpg:
I believe the issue happens because the Coursier version used in Bloopgun has an undeclared dependency on scala-xml. We hit on the same issue in Metals.

olafurpg:
Ignoring flaky test failure on macOS GraalVM

PR #1111, opened 2019-12-01
Don't use blp-server from local installation in bloopgun and launcher.

olafurpg:
Previously, the Bloop launcher and bloopgun would use a `blp-server`
binary from the local computer to launch Bloop.  This behavior was
especially surprising when using the Bloop launcher since the explicitly
provided version would be ignored.  In the example log below, the Bloop
version v1.3.4 was picked from my local Homebrew installation over the
explicitly provided version number 1.3.4+226-153862a1.
```
$ cs launch ch.epfl.scala:bloop-launcher_2.12:1.3.4+226-153862a1 -- 1.3.4+226-153862a1 --skip-bsp-connection
Starting the bsp launcher for bloop...
Attempting a connection to the server...
No server running at 127.0.0.1:8212, let's fire one...
A bloop installation has been detected either in the PATH or $HOME/.bloop
Starting bloop server at 127.0.0.1:8212...
-> Command: List(sh, /usr/local/Cellar/bloop/1.3.4/bin/blp-server, 8212, -J-XX:MaxInlineLevel=20, -J-XX:+UseParallelGC)
Attempting a connection to the server...
bloop v1.3.4
```

Now, with this change

* the launcher always launches Bloop with the provided version number
* bloopgun always launches Bloop with the same version as bloopgun itself

### DIFF
--- a/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
+++ b/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
@@ -363,8 +363,7 @@ class BloopgunCli(
       logger: SnailgunLogger
   ): Option[mode.Out] = {
     ServerStatus
-      .findServerToRun(bloopVersion, config.serverLocation, shell, logger)
-      .orElse(ServerStatus.resolveServer(bloopVersion, logger))
+      .resolveServer(bloopVersion, logger)
       .map(mode.fire(_, config, logger))
   }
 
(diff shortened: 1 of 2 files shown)


bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
```diff
@@ -362,9 +362,7 @@ class BloopgunCli(
       bloopVersion: String,
       logger: SnailgunLogger
   ): Option[mode.Out] = {
-    ServerStatus
-      .findServerToRun(bloopVersion, config.serverLocation, shell, logger)
```
olafurpg: I can remove `findServerToRun` along with accompanying tests if we agree that this change is desirable, I wanted first to get some feedback

Issue #1094, opened 2019-11-13
Test fails to run because Java argument list is too long 


olafurpg:
When running `bloop test` for a project with a large classpath, I get the following output
```
===============================================
Total duration: 0ms
No test suites were run.
===============================================
```

When I re-run the same command with `--verbose` and pull out the executed command `java -classpath ...` and run it directly from my terminal I get the following output

```
./test.sh: line 1: /Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/jre/bin/java: Argument list too long
```

One possible workaround would be to use `-classpath manifest.jar` where the manifest jar links to all the other jars on the classpath.

Issue #1084, opened 2019-11-08
Unexpected `CreateProcessW` error in Windows


schleumer:
There's an issue i've found running my project on Windows, when running(`bloop run`) large projects bloop generates really big command which causes CreateProcessW to reject the execution thus making `bloop run fails`, command limitation is 32768 characters, bloop generated one with 44540 characters for my project. It's actually just a report of a compatibility issue on Windows, running it on Linux just runs fine.

jvican:
Thanks for reporting! @schleumer Is this something you can look into and provide a fix? I can give you pointers, I think I know where this is happening.

schleumer:
I forgot do add the stack trace:

```
java.lang.RuntimeException: CreateProcessW() failed, error: 206
        at com.zaxxer.nuprocess.windows.WindowsProcess.start(WindowsProcess.java:248)
        at com.zaxxer.nuprocess.windows.WinProcessFactory.createProcess(WinProcessFactory.java:42)
        at com.zaxxer.nuprocess.NuProcessBuilder.start(NuProcessBuilder.java:266)
        at bloop.exec.Forker$.$anonfun$run$6(Forker.scala:183)
        at monix.eval.internal.TaskRunLoop$.monix$eval$internal$TaskRunLoop$$loop$1(TaskRunLoop.scala:187)
        at monix.eval.internal.TaskRunLoop$RestartCallback$1.onSuccess(TaskRunLoop.scala:119)
        at monix.eval.Task$.$anonfun$forkedUnit$2(Task.scala:1463)
        at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1425)
        at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
        at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
        at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
        at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
        at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
```

Also:

OpenJDK 12
Scala(global) 2.13.1
Scala(local) 2.12.9
SBT(global) 1.3.3
SBT(local) 1.3.0

If you can point me, i can work it out as soon as possible.

[...]

Issue #1093, opened 2019-11-12
Improve runtime performance of no-op compilations


olafurpg:
Currently, no-op compilation for targets with a large classpath can sometimes be slow. The profile below hints that slow no-op compilation is bottlenecked by `java.io.File.isDirectory()`, which invokes the native system call `stat64` 

<img width="798" alt="Screenshot 2019-11-12 at 10 37 37" src="https://user-images.githubusercontent.com/1408093/68713233-3aca1200-0552-11ea-9a37-8e95139d17c7.png">

Full profile: [nop.svg.zip](https://github.com/scalacenter/bloop/files/3838353/nop.svg.zip)

It would be nice to use alternative native APIs to testing which files are directories. For example, https://github.com/swoval/swoval.

jvican:
Thanks for reporting! Yes, I've wanted to use swoval for a few weeks now. We should avoid expensive stat64 operations as much as possible to speed up no-op compilations, class path hashing and other IO-intensive tasks the compilation engine.
  
I don't think I'll have the time to look into this any time soon but I'm very open to accepting PRs contributing this feature. I'd recommend going through all the call-sites of `isDirectory` in the codebase and replacing them with swoval APIs.
  
Paging @eatkins in case he wants to be in the loop.

eatkins:
I’d be happy to provide guidance and PR feedback but I’m mostly taking a break from open source right now.

PR #1118, opened 2019-12-10
Use MANIFEST jar to run commands that exceed OS limits

wiwa:
Previously, `JvmForker` passed the full classpath into the `java` invocation via `-cp`, without considering the implications for very long classpaths which break the command line length limits.

Now, it will create a temporary manifest jar to pass on classpath information in a command line-friendly way to the `java` invocation. This will allow us to run commands which exceed OS limits due to the length of the classpath. This will *not* address commands whose argument list to the main class causes it to fail despite shortening the classpath string.

Fixes https://github.com/scalacenter/bloop/issues/1094
Fixes https://github.com/scalacenter/bloop/issues/1084

### DIFF
--- a/frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
+++ b/frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
@@ -1,17 +1,18 @@
 package bloop.exec
 
-import java.io.File.pathSeparator
-import java.net.{InetSocketAddress, ServerSocket, URLClassLoader}
-
 import bloop.cli.CommonOptions
-import bloop.io.AbsolutePath
 import bloop.data.JdkConfig
 import bloop.engine.tasks.RunMode
+import bloop.io.{AbsolutePath, Paths, RelativePath}
 import bloop.logging.{DebugFilter, Logger}
-import monix.eval.Task
-
-import scala.util.{Failure, Success, Try}
 import bloop.util.CrossPlatform
+import java.io.File
+import java.io.File.pathSeparator
+import java.net.URLClassLoader
+import java.nio.file.Files
+import java.util.jar.{Attributes, JarEntry, JarOutputStream, Manifest}
+import monix.eval.Task
+import scala.util.{Failure, Success, Try}
 
 /**
  * Collects configuration to start a new program in a new process
@@ -71,6 +72,9 @@ trait JvmProcessForker {
 }
 
 object JvmProcessForker {
+  // Windows max cmd line length is 32767, which seems to be the least of the common shells.
+  val classpathCharLimit: Int = 30000
+
   def apply(config: JdkConfig, classpath: Array[AbsolutePath]): JvmProcessForker =
     new JvmForker(config, classpath)
 
@@ -113,22 +117,64 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
       extraClasspath: Array[AbsolutePath]
   ): Task[Int] = {
     val jvmOptions = jargs.map(_.stripPrefix("-J")) ++ config.javaOptions
-    val fullClasspath = (classpath ++ extraClasspath).map(_.syntax).mkString(pathSeparator)
-    Task.fromTry(javaExecutable).flatMap { java =>
-      val classpathOption = "-cp" :: fullClasspath :: Nil
-      val appOptions = mainClass :: args.toList
-      val cmd = java.syntax :: jvmOptions.toList ::: classpathOption ::: appOptions
-      val logTask = if (logger.isVerbose) {
+    val fullClasspath = classpath ++ extraClasspath
+
+    for {
+      fullClasspathStr <- Task.fromTry(ensureClasspathLength(fullClasspath))
+      java <- Task.fromTry(javaExecutable)
+
+      classpathOption = "-cp" :: fullClasspathStr :: Nil
+      appOptions = mainClass :: args.toList
+      cmd = java.syntax :: jvmOptions.toList ::: classpathOption ::: appOptions
+
+      logTask <- if (logger.isVerbose) {
         val debugOptions =
           s"""
              |Fork options:
              |   command      = '${cmd.mkString(" ")}'
              |   cwd          = '$cwd'""".stripMargin
         Task(logger.debug(debugOptions)(DebugFilter.All))
-      } else Task.unit
-      logTask.flatMap(_ => Forker.run(cwd, cmd, logger, opts))
+      } else {
+        Task.unit
+      }
+      res <- Forker.run(cwd, cmd, logger, opts)
+    } yield {
+      res
     }
   }
+
+  private def ensureClasspathLength(classpath: Array[AbsolutePath]): Try[String] = {
+    val fullClasspathStr = classpath.map(_.syntax).mkString(pathSeparator)
+
+    if (fullClasspathStr.length > JvmProcessForker.classpathCharLimit)
+      createTempManifestJar(classpath).map(_.syntax)
+    else
+      Try(fullClasspathStr)
+  }
+
+  private def createTempManifestJar(classpath: Array[AbsolutePath]): Try[AbsolutePath] = Try {
+    val prefix = "jvm-forker-manifest"
+    val manifestJar = Files.createTempFile(prefix, ".jar").toAbsolutePath
+
+    val classpathStr = classpath.map(addTrailingSlashToDirectories).mkString(" ")
+
+    val manifest = new Manifest()
+    manifest.getMainAttributes.put(Attributes.Name.MANIFEST_VERSION, "1.0")
+    manifest.getMainAttributes.put(Attributes.Name.CLASS_PATH, classpathStr)
+
+    val jos = new JarOutputStream(Files.newOutputStream(manifestJar), manifest)
+    jos.close()
+
+    AbsolutePath(manifestJar)
+  }
+
+  // Turns out manifest files can use absolute path directories in the classpath
+  // But they need a trailing slash
+  private def addTrailingSlashToDirectories(path: AbsolutePath): String =
+    if (path.isDirectory)
+      s"${path.syntax}${File.separatorChar}"
+    else
+      path.syntax
 
   private def javaExecutable: Try[AbsolutePath] = {
     val javaPath = config.javaHome.resolve("bin").resolve("java")
--- a/frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
+++ b/frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
@@ -189,16 +189,18 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
 
   private def addTrailingSlashToDirectories(path: AbsolutePath): String = {
     val syntax = path.syntax
-    val separatorAdded = if (syntax.endsWith(".jar")) {
-      syntax
-    } else {
-      syntax + File.separator
+    val separatorAdded = {
+      if (syntax.endsWith(".jar")) {
+        syntax
+      } else {
+        syntax + File.separator
+      }
     }
+
     if (Properties.isWin) {
-      // On Windows, the drive letter is prepended to absolute paths
-      // Even though the official Java docs say to do this,
-      // it seems this fails for MANIFEST files, so we remove the drive letter
-      separatorAdded.substring(separatorAdded.indexOf(":") + 1)
+      // Remove drive letter as MANIFEST files don't support them on Windows
+      if (separatorAdded.indexOf(":") != 1) separatorAdded
+      else separatorAdded.substring(separatorAdded.indexOf(":") + 1)
     } else {
       separatorAdded
     }
--- a/frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
+++ b/frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
@@ -6,7 +6,7 @@ import bloop.engine.tasks.RunMode
 import bloop.io.{AbsolutePath, Paths}
 import bloop.logging.{DebugFilter, Logger}
 import bloop.util.CrossPlatform
-import java.io.File.pathSeparator
+import java.io.File
 import java.net.URLClassLoader
 import java.nio.file.Files
 import java.util.jar.{Attributes, JarOutputStream, Manifest}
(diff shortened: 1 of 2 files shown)


olafurpg:
Thank you for this contribution! I'm very excited to see this bug fixed.

I didn't know that manifest jars were not able to reference directory classpath entries. That's a shame...

I am concerned that copying classfiles into jars will introduce non-trivial overhead when forking. I took a stab at using another workaround:

* launch a custom main class `ForkLauncher`, which is a zero-dependency Java main class written in the Bloop `frontend` project
* pass in a single argument which is  a temporary file containing the full classpath, main class to be launched and arguments (if any)

I prototyped this approach in https://github.com/scalacenter/bloop/compare/master...olafurpg:f/long-classpath?expand=1 and the tests are at least passing.

olafurpg:
I took the liberty to push the commit into this PR just to trigger the CI.

jvican:
Hi @wiwa, thanks a lot for this contribution! I appreciate you helping out :smile:
  
I've just double-checked and it looks like you can reference directories in the `Class-Path` field of the MANIFEST.mf. The official [Java docs don't help clarify this point](https://docs.oracle.com/javase/tutorial/deployment/jar/downman.html), but I've just tested it locally and it works.
  
Instructions to reproduce:
  
1. Create the following java class in a directory `classes`:
```java
public class A {
  public static void main(String[] args) {
    System.out.println("Hello, World");
  }
}
```
2. Compile it with `javac A.java`
3. Then, step out of the `classes` directory and create the following `MANIFEST.mf`:
```
Manifest-Version: 1.0
Archiver-Version: Plexus Archiver
Built-By: john.doe
Class-Path: classes/
Created-By: Apache Maven 3.3.9
Main-Class: A
```
4. Package it in a jar with `jar cfm myjar.jar MANIFEST.mf`
5. Then, run it with `java myjar.jar`. The application will print `Hello, World!`
  
I believe we should modify this PR to add directories to the classpath and avoid all of the expensive file copying.

olafurpg:
Nice catch @jvican ! I was also able to validate locally that it's possible to point to directory entries in manifest jars. It's probably best to discard the commit adding `ForkLauncher` and use manifest jars instead.

wiwa:
Thanks for the detailed comments! Turns out I needed to add a trailing slash 🤕

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -113,23 +117,65 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
...
+      res <- Forker.run(cwd, cmd, logger, opts)
+    } yield {
+      res
     }
   }
 
+  private def ensureClasspathLength(classpath: Array[AbsolutePath]): Try[String] = {
+    val fullClasspathStr = classpath.map(_.syntax).mkString(pathSeparator)
+
+    if (fullClasspathStr.length > JvmProcessForker.classpathCharLimit)
```
olafurpg: Should we include the length of JVM options and program arguments instead of only `fullClasspathStr.length`? This check may still not insure us against "argument list too large" if the program argument list is very large.

olafurpg: we might want to write another test with a very large argument list

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -113,23 +117,65 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
...
+    val jos = new JarOutputStream(Files.newOutputStream(manifestJar), manifest)
+    jos.close()
+
+    AbsolutePath(manifestJar)
+  }
+
+  // Turns out manifest files can use absolute path directories in the classpath
+  // But they need a trailing slash
+  private def addTrailingSlashToDirectories(path: AbsolutePath): String =
+    if (path.isDirectory)
```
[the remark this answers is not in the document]
olafurpg: See https://github.com/scalacenter/bloop/issues/1093 for an example flamegraph where `file.isDirectory` is the bottleneck

jvican:
The latest change I pushed to master seems to make the CI work. Rebasing the PR should help.

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -71,6 +72,9 @@ trait JvmProcessForker {
 }
 
 object JvmProcessForker {
+  // Windows max cmd line length is 32767, which seems to be the least of the common shells.
+  val classpathCharLimit: Int = 30000
```
jvican: The char limit is associated with the length of the command that we use to sysprocess, not the classpath char itself. What about we fix that and we make it private?

```suggestion
  private val processCmdCharLimit: Int = 30000
```
  
Also, what about moving this definition as close to the use site as possible?

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -1,17 +1,18 @@
...
-
 import bloop.cli.CommonOptions
-import bloop.io.AbsolutePath
 import bloop.data.JdkConfig
 import bloop.engine.tasks.RunMode
+import bloop.io.{AbsolutePath, Paths, RelativePath}
 import bloop.logging.{DebugFilter, Logger}
+import bloop.util.CrossPlatform
+import java.io.File
+import java.io.File.pathSeparator
```
jvican: Looks like it's unused now

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
jvican: Let's wrap up bodies of if-else expressions with curly braces for consistency with the whole codebase, both here and in the other ifs in the PR :smile:

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -113,23 +117,65 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
...
+    val manifestJar = Files.createTempFile(prefix, ".jar").toAbsolutePath
+
+    val classpathStr = classpath.map(addTrailingSlashToDirectories).mkString(" ")
+
+    val manifest = new Manifest()
+    manifest.getMainAttributes.put(Attributes.Name.MANIFEST_VERSION, "1.0")
+    manifest.getMainAttributes.put(Attributes.Name.CLASS_PATH, classpathStr)
+
+    val jos = new JarOutputStream(Files.newOutputStream(manifestJar), manifest)
+    jos.close()
```
jvican: Should this not be in a try-finally?

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -113,23 +117,65 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
...
+
+    AbsolutePath(manifestJar)
+  }
+
+  // Turns out manifest files can use absolute path directories in the classpath
+  // But they need a trailing slash
+  private def addTrailingSlashToDirectories(path: AbsolutePath): String =
```
jvican: Let's wrap up method body with curly braces too

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -113,23 +117,65 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
...
+
+    if (fullClasspathStr.length > JvmProcessForker.classpathCharLimit)
+      createTempManifestJar(classpath).map(_.syntax)
+    else
+      Try(fullClasspathStr)
+  }
+
+  private def createTempManifestJar(classpath: Array[AbsolutePath]): Try[AbsolutePath] = Try {
+    val prefix = "jvm-forker-manifest"
+    val manifestJar = Files.createTempFile(prefix, ".jar").toAbsolutePath
```
jvican: This creates a temporary file per invocation, but are we removing these files after running?

jvican: Should we not be comparing the chars of the full cmd invocation instead of the classpath? We could have a classpath that doesn't violate the chars limit but that combined with the other jvm options does. In that case, we would not create a MANIFEST but we should have, as the underlying OS will fail.

frontend/src/test/scala/bloop/ForkerSpec.scala
```diff
@@ -116,6 +127,20 @@ class ForkerSpec {
     }
   }
 
+  @Test
+  def canHandleLongClasspaths(): Unit = TestUtil.withinWorkspace { tmp =>
+    val longCp = Array.fill(JvmProcessForker.classpathCharLimit / 10) {
+      AbsolutePath(Files.createTempFile("forkerspec-temp", ".jar"))
+    }
+
+    run(tmp, Array("foo", "bar", "baz"), longCp) {
```
jvican: Could we double-check in this test that we're indeed using the MANIFEST to run the code (maybe a debug log that we detect via `logger`)? Otherwise we're not sure if this test is actually checking the process char limit logic.

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
wiwa: Is there a way to stuff arguments into a MANIFEST jar, because the answer to that seems to be "no".

wiwa: Might have to use your solution with a wrapper class!

wiwa:
Not sure if my strategy can deal with large argument lists. Resolving previous comments and pivoting towards the solution @olafurpg wrote.

jvican:
Wait, why? I don’t understand. We should merge the manifest jar solution, I believe the issues raised have straightforward solutions.

wiwa:
Couldn't find a good solution for when the arguments exceed the OS cmd limit -- is there one?

wiwa:
Following up from offline conversation with @olafurpg: we'll continue with scoping this PR only around shortening the classpath via MANIFEST.

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
wiwa: Descoping to just classpath shortening

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
wiwa: done

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
wiwa: done

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
wiwa: done

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
wiwa: done

wiwa: will look at full cmd, but descoping to just classpath shortening

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
wiwa: done

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -112,22 +112,111 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
...
-      val logTask = if (logger.isVerbose) {
-        val debugOptions =
-          s"""
-             |Fork options:
-             |   command      = '${cmd.mkString(" ")}'
-             |   cwd          = '$cwd'""".stripMargin
-        Task(logger.debug(debugOptions)(DebugFilter.All))
-      } else Task.unit
-      logTask.flatMap(_ => Forker.run(cwd, cmd, logger, opts))
+      val cmdStr = cmd.mkString(" ")
```
olafurpg: Should we maybe count the total length without allocating the full string?

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -112,22 +112,111 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
...
+  }
+
+  private def withTempManifestJar[A](
+      classpath: Array[AbsolutePath],
+      logger: Logger
+  )(op: AbsolutePath => Task[A]): Task[A] = {
+
+    val manifestJar = Files.createTempFile("jvm-forker-manifest", ".jar").toAbsolutePath
+    val manifestJarAbs = AbsolutePath(manifestJar)
+    def cleanup = {
```
olafurpg: Not: I think this is safe to turn into a val. Task is lazy

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -112,22 +112,111 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
...
+  // Turns out manifest files can use absolute path directories in the classpath
+  // But they need a trailing slash
+  private def addTrailingSlashToDirectories(path: AbsolutePath): String = {
+    val syntax = path.syntax
+    val separatorAdded = if (syntax.endsWith(".jar")) {
+      syntax
+    } else {
+      syntax + File.separator
     }
+    // On Windows, the drive letter is prepended to absolute paths
```
olafurpg: Thank you for tracking down the windows failures and fixing the issue! It’s great to see the CI green 😁 

Should we maybe guard this logic for Windows only? Im not sure if Unix file systems support file names with colons.

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -112,21 +112,112 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
...
+
+    def logDebugIfVerbose(debugMsg: => String): Task[Unit] =
+      if (logger.isVerbose) {
+        Task(logger.debug(debugMsg)(DebugFilter.All))
+      } else {
+        Task.unit
+      }
+
+    def runCmd(cmd: List[String]): Task[Int] = {
+      def debugOptions =
```
jvican: @olafurpg Could you rebase on top of latest master and remove the debug logs? I removed these logs in another PR since they were redundant. The `Forker.run` expression and `logDebugIfVerbose` could now be inlined

jvican:
I've already done the merge myself, will merge as soon as it's green

PR #1129, opened 2019-12-18
Replace directory hash from loaded analysis

jvican:
The directory hash is random to avoid classpath hash conflicts and it's
a value initialized when the bloop server runs for the first time. This
means the directory hashes from different bloop instances in time is
different.

To work around this issue, we make sure we replace the directory hashes
from the previous bloop server instance with the new one. This avoid
classpath hash invalidations that might trigger recompiles upon
restarting the bloop server.

Thanks @olafurpg for reporting this misinteraction


jvican:
This should no longer be necessary, this issue was actually fixed in the latest master but didn't realize that was the case.

PR #1135, opened 2019-12-29
Remove 'code' property of bsp.Diagnostic

### LINKED ISSUE #1134: Remove source code from 'code' property in BSP diagnostic
See https://github.com/scalameta/metals/issues/1220 for reference.

The `code` property of `Diagnostic` seems to be intended for "error code", rather than "source code"

gabro:
Fixes #1134

### DIFF
--- a/frontend/src/main/scala/bloop/logging/BspServerLogger.scala
+++ b/frontend/src/main/scala/bloop/logging/BspServerLogger.scala
@@ -112,8 +112,7 @@ final class BspServerLogger private (
         val source = Some("bloop")
         val uri = bsp.Uri(file.toPath.toUri)
         val severity = bspSeverity(problemSeverity)
-        val code = Some(problemPos.lineContent())
-        val diagnostic = bsp.Diagnostic(pos, Some(severity), code, source, message, None)
+        val diagnostic = bsp.Diagnostic(pos, Some(severity), None, source, message, None)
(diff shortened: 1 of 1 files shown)


gabro:
that would also require to change the BSP spec then, since it says that Diagnostic is defined as in LSP.
I’m not opposed to it, just pointing it out

jvican:
@gabro Thank you for submitting this fix, I thought the code here was the actual code snippet sometimes added in diagnostics, I didn't think it was a code number.
  
I'm not going to merge it just yet as the implementation of offloading compilation in sbt depends on this so I need to fix that first before we can remove the use of `code` in Bloop. Will keep you in the loop when that's ready.

olafurpg:
We have already updated Metals to discard the `code` field. A long-term solution would be to revert that change in Metals and we either extend BSP diagnostics with a `lineContent` field or Bloop uses some other custom field for its own BSP extension.

gabro:
Yes, we discussed this offline briefly and @jvican mentioned he’s thinking of extending BSP for that purpose.

jvican:
Indeed, I'll modify BSP for this use case as it looks like the code property isn't really meant for this.

tgodzik:
Should we actually merge this? I don't think offloading compilation got anywhere yet and we will need to overhaul quite a bit when updating to newest zinc anyway.

PR #1145, opened 2020-01-14
Silence error from _arguments:comparguments:325

tovbinm:
When loading zsh completion script on Mac OSX the following error occurs:
```bash
_arguments:comparguments:325: can only be called from completion function
```
This change simply omits it. Are there any other proposals?

### DIFF
--- a/etc/zsh/_bloop
+++ b/etc/zsh/_bloop
@@ -81,7 +81,7 @@ _project_or_flags() {
     fi
 }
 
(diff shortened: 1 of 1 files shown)


tovbinm:
well, I have no clue ;) perhaps there is a better way?

jvican:
Sounds good, thank you @tovbinm  Let's go ahead and merge this

PR #1151, opened 2020-01-22
Fix Windows drive letter handling in jar manifest

Arthurm1:
Bloop has manifest classpath entries in Windows going from...
```c:/users/...```
to...
```/users/...```
which doesn't work when using multiple drive letters e.g.
```c:/users/... d:/dev/...```

Changed to produce this...
```/c:/users/... /d:/dev/...```

[reference](https://stackoverflow.com/questions/18659774/how-do-i-handle-files-with-spaces-in-the-classpath-in-manifest-mf)


Arthurm1:
Yes - deployed it locally, ran an app via Bloop (which didn't work before) and checked the manifest contents held drive letters.

Only tested on Windows but it looked like this code area was Windows specific.

I confess I didn't add a unit test.

Issue #1158, opened 2020-01-28
References to undefined settings `compileIncremental`


tgodzik:
Managed to reproduce an issue that some of the users are having with sbt 0.13.18
```
lazy val commonSettings = Seq(
  scalaVersion := "2.11.12",
  scalacOptions ++= Seq(
    "-deprecation",
    "-unchecked",
    "-encoding", "UTF-8",
    "-Yclosure-elim",
    "-Yinline",
    "-Xverify",
    "-feature"
  )
)

lazy val jmh = project
  .settings(commonSettings: _*)
```

It seems `.settings` is removing `compileIncremental`

tgodzik:
The log we are getting:
```
[info] Done updating.
INFO  References to undefined settings: 
INFO  
INFO    jmh/it:compileIncremental from jmh/it:bloopCompile ((bloop.integrations.sbt.Offloader) Offloader.scala:19)
INFO  
INFO    sbt-old/it:compileIncremental from sbt-old/it:bloopCompile ((bloop.integrations.sbt.Offloader) Offloader.scala:19)
INFO   
INFO  	at sbt.Init$class.Uninitialized(Settings.scala:270)
INFO  	at sbt.Def$.Uninitialized(Def.scala:10)
INFO  	at sbt.Init$class.delegate(Settings.scala:191)
INFO  	at sbt.Def$.delegate(Def.scala:10)
INFO  	at sbt.Init$class.compiled(Settings.scala:138)
INFO  	at sbt.Def$.compiled(Def.scala:10)
INFO  	at sbt.Init$class.make(Settings.scala:144)
INFO  	at sbt.Def$.make(Def.scala:10)
INFO  	at sbt.Load$$anonfun$8.apply(Load.scala:161)
INFO  	at sbt.Load$$anonfun$8.apply(Load.scala:156)
INFO  	at sbt.Load$.timed(Load.scala:1039)
INFO  	at sbt.Load$.apply(Load.scala:156)
INFO  	at sbt.Load$.defaultLoad(Load.scala:39)
INFO  	at sbt.BuiltinCommands$.liftedTree1$1(Main.scala:504)
INFO  	at sbt.BuiltinCommands$.doLoadProject(Main.scala:504)
INFO  	at sbt.BuiltinCommands$$anonfun$loadProjectImpl$2.apply(Main.scala:496)
INFO  	at sbt.BuiltinCommands$$anonfun$loadProjectImpl$2.apply(Main.scala:496)
[...]

PR #1159, opened 2020-01-29
Fix an issue with non existing compileIncremental key

tgodzik:
Fixes https://github.com/scalacenter/bloop/issues/1158

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,6 @@ website/build/bloop-gh-pages/
 node_modules/
 package-lock.json
 .metals/
-project/metals.sbt
 *.lock
 benchmark-bridge/corpus/
 
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ website/build/bloop-gh-pages/
 node_modules/
 package-lock.json
 .metals/
+project/metals.sbt
 *.lock
 benchmark-bridge/corpus/
 
--- a/integrations/sbt-bloop/src/main/scala-sbt-0.13/bloop/integrations/sbt/Offloader.scala
+++ b/integrations/sbt-bloop/src/main/scala-sbt-0.13/bloop/integrations/sbt/Offloader.scala
@@ -15,7 +15,5 @@ object Offloader {
 
   val bloopCompileProjectSettings: Seq[Def.Setting[_]] = List()
   val bloopCompileGlobalSettings: Seq[Def.Setting[_]] = List()
-  val bloopCompileConfigSettings: Seq[Def.Setting[_]] = List(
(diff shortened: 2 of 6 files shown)


integrations/sbt-bloop/src/main/scala-sbt-0.13/bloop/integrations/sbt/Offloader.scala
```diff
@@ -15,7 +15,5 @@ object Offloader {
 
   val bloopCompileProjectSettings: Seq[Def.Setting[_]] = List()
   val bloopCompileGlobalSettings: Seq[Def.Setting[_]] = List()
-  val bloopCompileConfigSettings: Seq[Def.Setting[_]] = List(
-    BloopKeys.bloopCompile := Keys.compileIncremental.value
```
olafurpg: Can you elaborate on why this change is safe?

tgodzik: Not 100% sure, I tested it out on a `0.13` project and it worked. `Keys.compileIncremental` can  sometimes be not  there and it will break some projects when they include the Bloop plugin.

I will work on a test case for that though.

.gitignore
```diff
@@ -33,6 +33,7 @@ website/build/bloop-gh-pages/
 node_modules/
 package-lock.json
 .metals/
+project/metals.sbt
```
jvican: Hmmm, Bloop loads the sbt-bloop plugin itself. You should not have this directory in your setup, otherwise bad things can happen (TM). The `-Dbloop.generate.sbt=false` option in `.jvmopts` is there so that Metals doesn't write this file and instead blindly assumes the plugin is already added in the project. Is that not being picked up somehow in your computer?

integrations/sbt-bloop/src/main/scala-sbt-0.13/bloop/integrations/sbt/Offloader.scala
jvican: I expect making this change is fine because we're not using `bloopCompile` directly in `SbtBloop` yet. I removed that to disable compile offloading. It's really surprising to me that `compileIncremental` is not scoped in sbt 0.13, most probably it's a bug.

jvican:
Can you remove the .gitignore addition, push again and we merge? Feel free to merge when you do that, CI is failing right now but it's spurious.

.gitignore
tgodzik: Och, that must have been an old file, sorry about that.

lionello: For the record: removing `project/metals.sbt` actually resolved the issue for me!

PR #1165, opened 2020-02-04
Add options for Dotty to create semanticDB when Metals is present

tgodzik:
This changes are needed so that user experience for Dotty will be exactly the same as it in Scala 2.

Previously, user would have to add the settings manually.

We only need 2 settings in this case: `-Ysemanticdb` and `-sourceroot`

@jvican Let me know if this makes sense.

### DIFF
--- a/frontend/src/main/scala/bloop/data/Project.scala
+++ b/frontend/src/main/scala/bloop/data/Project.scala
@@ -257,7 +257,7 @@ object Project {
     val workspaceDir = project.workspaceDirectory.getOrElse(configDir.getParent)
     val isDotty = project.scalaInstance.exists(_.isDotty)
 
-    def enableSemanticDB(options: List[String], pluginPath: AbsolutePath): List[String] = {
+    def enableSemanticdb(options: List[String], pluginPath: AbsolutePath): List[String] = {
       val baseSemanticdbOptions = List(
         "-P:semanticdb:failures:warning",
         "-P:semanticdb:synthetics:on",
@@ -273,7 +273,7 @@ object Project {
       (baseOptions ++ baseSemanticdbOptions ++ pluginOption).distinct
     }
 
-    def enableDottySemanticDB(options: List[String]) = {
+    def enableDottySemanticdb(options: List[String]) = {
       val ysemanticdb = if (!options.contains("-Ysemanticdb")) List("-Ysemanticdb") else Nil
       val sourceRoot =
         if (!options.contains("-sourceroot")) List("-sourceroot", workspaceDir.toString()) else Nil
@@ -291,12 +291,15 @@ object Project {
     val options = projectWithRangePositions.scalacOptions
 
     semanticDBPlugin match {
-      case None if isDotty =>
-        val optionsWithSemanticDB = enableDottySemanticDB(options)
-        projectWithRangePositions.copy(scalacOptions = optionsWithSemanticDB)
-      case None => projectWithRangePositions
+      case None =>
+        if (isDotty) {
+          val optionsWithSemanticDB = enableDottySemanticdb(options)
+          projectWithRangePositions.copy(scalacOptions = optionsWithSemanticDB)
+        } else {
+          projectWithRangePositions
+        }
       case Some(pluginPath) =>
-        val optionsWithSemanticDB = enableSemanticDB(options, pluginPath)
+        val optionsWithSemanticDB = enableSemanticdb(options, pluginPath)
         projectWithRangePositions.copy(scalacOptions = optionsWithSemanticDB)
     }
   }
--- a/frontend/src/main/scala/bloop/data/Project.scala
+++ b/frontend/src/main/scala/bloop/data/Project.scala
@@ -254,8 +254,10 @@ object Project {
       semanticDBPlugin: Option[AbsolutePath],
       logger: Logger
   ): Project = {
+    val workspaceDir = project.workspaceDirectory.getOrElse(configDir.getParent)
+    val isDotty = project.scalaInstance.exists(_.isDotty)
+
     def enableSemanticDB(options: List[String], pluginPath: AbsolutePath): List[String] = {
-      val workspaceDir = project.workspaceDirectory.getOrElse(configDir.getParent)
       val baseSemanticdbOptions = List(
         "-P:semanticdb:failures:warning",
         "-P:semanticdb:synthetics:on",
@@ -271,18 +273,29 @@ object Project {
       (baseOptions ++ baseSemanticdbOptions ++ pluginOption).distinct
     }
 
+    def enableDottySemanticDB(options: List[String]) = {
+      val ysemanticdb = if (!options.contains("-Ysemanticdb")) List("-Ysemanticdb") else Nil
+      val sourceRoot =
+        if (!options.contains("-sourceroot")) List("-sourceroot", workspaceDir.toString()) else Nil
+      ysemanticdb ++ sourceRoot
+    }
+
     def enableRangePositions(options: List[String]): List[String] = {
       val hasYrangepos = options.exists(_.contains("-Yrangepos"))
-      val isDotty = project.scalaInstance.exists(_.isDotty)
       if (hasYrangepos || isDotty) options else options :+ "-Yrangepos"
     }
 
     val projectWithRangePositions =
       project.copy(scalacOptions = enableRangePositions(project.scalacOptions))
+
+    val options = projectWithRangePositions.scalacOptions
+
     semanticDBPlugin match {
+      case None if isDotty =>
+        val optionsWithSemanticDB = enableDottySemanticDB(options)
+        projectWithRangePositions.copy(scalacOptions = optionsWithSemanticDB)
       case None => projectWithRangePositions
       case Some(pluginPath) =>
-        val options = projectWithRangePositions.scalacOptions
         val optionsWithSemanticDB = enableSemanticDB(options, pluginPath)
         projectWithRangePositions.copy(scalacOptions = optionsWithSemanticDB)
     }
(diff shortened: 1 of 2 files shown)


frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -271,18 +273,29 @@ object Project {
...
       if (hasYrangepos || isDotty) options else options :+ "-Yrangepos"
     }
 
     val projectWithRangePositions =
       project.copy(scalacOptions = enableRangePositions(project.scalacOptions))
+
+    val options = projectWithRangePositions.scalacOptions
+
     semanticDBPlugin match {
+      case None if isDotty =>
```
olafurpg: nit: instead of guarding in the pattern match, I find it more readable to put the guard inside the body
```suggestion
      case None =>
        if (isDotty) ...
        else ...
```

frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -271,18 +273,29 @@ object Project {
       (baseOptions ++ baseSemanticdbOptions ++ pluginOption).distinct
     }
 
+    def enableDottySemanticDB(options: List[String]) = {
```
olafurpg: nit: the convention we use is lowercase db for better or worse
```suggestion
    def enableDottySemanticdb(options: List[String]) = {
```

tgodzik:
There seems to be an issue running the new tests on JDK 11 :/

```
java.lang.ClassCastException: class jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to class java.net.URLClassLoader (jdk.internal.loader.ClassLoaders$AppClassLoader and java.net.URLClassLoader are in module java.base of loader 'bootstrap')
	at xsbt.CompilerClassLoader.computeFixedLoader(CompilerClassLoader.java:113)
	at xsbt.CompilerClassLoader.lambda$fixBridgeLoader$0(CompilerClassLoader.java:98)
	at java.base/java.util.Map.computeIfAbsent(Map.java:1003)
	at xsbt.CompilerClassLoader.fixBridgeLoader(CompilerClassLoader.java:98)
	at xsbt.CompilerInterface.newCompiler(CompilerInterface.java:31)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at sbt.internal.inc.AnalyzingCompiler.call(AnalyzingCompiler.scala:252)
	at sbt.internal.inc.AnalyzingCompiler.newCachedCompiler(AnalyzingCompiler.scala:144)
	at sbt.internal.inc.AnalyzingCompiler.newCachedCompiler(AnalyzingCompiler.scala:131)
	at sbt.internal.inc.FreshCompilerCache.apply(CompilerCache.scala:100)
	at sbt.internal.inc.AnalyzingCompiler.compile(AnalyzingCompiler.scala:91)
	at sbt.internal.inc.bloop.internal.BloopHighLevelCompiler.compileSources$1(BloopHighLevelCompiler.scala:142)
	at sbt.internal.inc.bloop.internal.BloopHighLevelCompiler.$anonfun$compile$7(BloopHighLevelCompiler.scala:160)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at sbt.internal.inc.bloop.internal.BloopHighLevelCompiler.$anonfun$compile$1(BloopHighLevelCompiler.scala:72)
	at bloop.tracing.BraveTracer.trace(BraveTracer.scala:31)
	at sbt.internal.inc.bloop.internal.BloopHighLevelCompiler.timed$1(BloopHighLevelCompiler.scala:71)
	at sbt.internal.inc.bloop.internal.BloopHighLevelCompiler.$anonfun$compile$6(BloopHighLevelCompiler.scala:160)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
```

Probably something we need to fix in zinc?

romanowski:
I think this line is to blame: https://github.com/lampepfl/dotty/blob/master/sbt-bridge/src/xsbt/CompilerClassLoader.java#L102

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
```diff
@@ -331,6 +331,53 @@ class BspMetalsClientSpec(
...
+      val configDir = TestProject.populateWorkspace(workspace, projects)
+      val logger = new RecordingLogger(ansiCodesSupported = false)
+      WorkspaceSettings.writeToFile(
+        configDir,
+        WorkspaceSettings("4.3.0", List()),
+        logger
+      )
+      loadBspState(workspace, projects, logger) { state =>
+        val compiledState = state.compile(`A`).toTestState
+        assert(compiledState.status == ExitStatus.Ok)
```
jvican: Let's try to add some of the other checks we have in `CompileSpec` for simple hello world compilations in Scala 2. You can find these checks at the beginning of the `CompileSpec` definition.

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
```diff
@@ -331,6 +331,53 @@ class BspMetalsClientSpec(
...
+        scalaOrg = Some("ch.epfl.lamp"),
+        scalaCompiler = Some("dotty-compiler_0.21")
+      )
+      val projects = List(`A`)
+      val configDir = TestProject.populateWorkspace(workspace, projects)
+      val logger = new RecordingLogger(ansiCodesSupported = false)
+      loadBspState(workspace, projects, logger) { state =>
+        val compiledState = state.compile(`A`).toTestState
+        assert(compiledState.status == ExitStatus.Ok)
+        assertSemanticdbFileFor("Foo.scala", compiledState, isPresent = false)
```
jvican: Same here as before :smile:

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
```diff
@@ -357,13 +404,16 @@ class BspMetalsClientSpec(
...
+      isPresent: Boolean = true
+  ): Unit = {
     val projectA = state.build.getProjectFor("A").get
     val classesDir = state.client.getUniqueClassesDirFor(projectA, forceGeneration = true)
     val sourcePath = if (sourceFileName.startsWith("/")) sourceFileName else s"/$sourceFileName"
-    assertIsFile(
-      classesDir.resolve(s"META-INF/semanticdb/A/src/$sourcePath.semanticdb")
-    )
+    val file = classesDir.resolve(s"META-INF/semanticdb/A/src/$sourcePath.semanticdb")
+    if (isPresent) assertIsFile(file) else assertNotFile(file)
```
jvican: Nit: I propose we introduce a new method `assertNotSemanticdbFileFor` that checks the non-existence of a semanticdb file, just as we do with the combo `assertIsFile` and `aasertNotFile`.

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
tgodzik: Should we create a separate CompileSpec for Scala 3? Or do you mean to add them here?

tgodzik: Or just he asserts from the test cases there?

tgodzik:
@jvican Seems we will not be able to run Dotty compilation on JDK 11 due to the [issue](https://github.com/lampepfl/dotty/blob/master/sbt-bridge/src/xsbt/CompilerClassLoader.java#L102) @romanowski found :/ 

So either we don't run tests for now or we have to fix the issue in Dotty itself.

alexarchambault:
FYI, if you're not familiar with that JDK11 issue, there's [a fix for a similar issue in metabrowse](https://github.com/scalameta/metabrowse/blob/c405b5be580e4925252390763955ad9eb9e266f8/metabrowse-tests/src/test/scala/metabrowse/tests/JsonSuite.scala#L26-L35). It basically amounts to assume the `getURLs` of a `jdk.internal.loader.ClassLoaders$AppClassLoader` is `System.getProperty("java.class.path").split(File.pathSeparator)`. It's described [there](https://blog.codefx.org/java/java-11-migration-guide/#Casting-To-URL-Class-Loader) too.

tgodzik:
@alexarchambault Thanks! I did find the issue, but still not sure exactly where to fix it.

Got something working here: https://github.com/scalacenter/zinc/compare/1.x-bloop...tgodzik:fix-classpath

tgodzik:
@jvican I updated Dotty version to a nightly with a fix, otherwise the tests will not work.

I also added a couple of checks to the test case, but couldn't find too many more. Some workflows are being tested there in `CompileSpec`, but I don't believe it's a place to add them here in `BspMetalsClientSpec`. What checks did you mean exactly?

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
```diff
@@ -331,6 +331,57 @@ class BspMetalsClientSpec(
...
+        scalaOrg = Some("ch.epfl.lamp"),
+        scalaCompiler = Some("dotty-compiler_0.23")
+      )
+      val projects = List(`A`)
+      val configDir = TestProject.populateWorkspace(workspace, projects)
+      val logger = new RecordingLogger(ansiCodesSupported = false)
+      loadBspState(workspace, projects, logger) { state =>
+        val compiledState = state.compile(`A`).toTestState
+        assertExitStatus(compiledState, ExitStatus.Ok)
+        assertValidCompilationState(compiledState, projects)
```
jvican: Re my previous comments, these two assertions are enough :smile: It's important that we assert the compilation state is valid and it's a successful compilation because there are lots of invariants that need to be met for a compilation to be correct. 👍

PR #1166, opened 2020-02-06
Add BSP endpoint to query jvm test environment

tpasternak:
Implement BSP endpoint: `buildTarget/jvmTestEnvironment` according
to the new BSP specification.

EXCERPT FROM BSP SPECIFICATION:

The JVM test environment request is sent from the client to the server in order to
gather information required to launch a Java process. This is useful when the
client wants to control the Java process execution, for example to enable custom
Java agents or launch a custom main class during unit testing or debugging

The data provided by this endpoint may change beteween compilations, so it should
not be cached in any form. The client should ask for it right before test exection,
after all the targets are compiled.

- method: `buildTarget/jvmTestEnvironment`
- params: `JvmEnvironmentParams`

```ts
export interface JvmEnvironmentParams(
    targets: BuildTargetIdentifier[],
    originId?: String
)
```

Response:

- result: `JvmEnvironmentResult`, defined as follows
- error: JSON-RPC code and message set in case an exception happens during the
  request.

```ts
export interface JvmEnvironmentEntry{
    target: BuildTargetIdentifier;
    classpath: String[];
    jvmOptions: String[];
}

export interface JvmEnvironmentResult{
    entries: JvmEnvironmentEntry[];
    workingDirectory: String;
    environmentVariables: Map<String, String>;
}
```

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -28,7 +28,7 @@ import ch.epfl.scala.bsp
 import ch.epfl.scala.bsp.ScalaBuildTarget.encodeScalaBuildTarget
 import ch.epfl.scala.bsp.{
   BuildTargetIdentifier,
-  JvmEnvironmentEntry,
+  JvmEnvironmentItem,
   MessageType,
   ShowMessageParams,
   endpoints
@@ -669,8 +669,8 @@ final class BloopBspServices(
   }
 
   def jvmTestEnvironment(
-      params: bsp.JvmEnvironmentParams
-  ): BspEndpointResponse[bsp.JvmEnvironmentResult] = {
+      params: bsp.JvmTestEnvironmentParams
+  ): BspEndpointResponse[bsp.JvmTestEnvironmentResult] = {
     ifInitialized(None) { (state: State, logger: BspServerLogger) =>
       mapToProjects(params.targets, state) match {
         case Left(error) =>
@@ -678,7 +678,7 @@ final class BloopBspServices(
           Task.now((state, Left(JsonRpcResponse.invalidRequest(error))))
 
         case Right(projects) =>
-          val environmentEntries: List[Either[String, JvmEnvironmentEntry]] = for {
+          val environmentEntries: List[Either[String, JvmEnvironmentItem]] = for {
             (id, project) <- projects.toList
             dag = state.build.getDagFor(project)
             fullClasspath = project.fullClasspath(dag, state.client).map(_.toString)
@@ -688,7 +688,7 @@ final class BloopBspServices(
             project.platform match {
               case Platform.Jvm(config, _, _) =>
                 Right(
-                  bsp.JvmEnvironmentEntry(
+                  bsp.JvmEnvironmentItem(
                     id,
                     fullClasspath.toList,
                     config.javaOptions.toList,
@@ -712,7 +712,7 @@ final class BloopBspServices(
           }
 
           val resp = sequenceListOfEithers(List.empty, environmentEntries)
-            .map(new bsp.JvmEnvironmentResult(_))
+            .map(new bsp.JvmTestEnvironmentResult(_))
           Task.now((state, resp.swap.map(JsonRpcResponse.invalidRequest(_)).swap))
       }
     }
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -683,9 +683,9 @@ final class BloopBspServices(
             fullClasspath = project.fullClasspath(dag, state.client).map(_.toString)
             environmentVariables = state.commonOptions.env.toMap
             workingDirectory = state.commonOptions.workingDirectory
-            javaOptions = project.platform match {
-              case Platform.Jvm(config, _, _) => config.javaOptions.toList
-              case _ => List()
+            javaOptions <- project.platform match {
+              case Platform.Jvm(config, _, _) => Some(config.javaOptions.toList)
+              case _ => None
             }
           } yield {
             bsp.JvmEnvironmentItem(
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -678,42 +678,28 @@ final class BloopBspServices(
           Task.now((state, Left(JsonRpcResponse.invalidRequest(error))))
 
         case Right(projects) =>
-          val environmentEntries: List[Either[String, JvmEnvironmentItem]] = for {
+          val environmentEntries = for {
             (id, project) <- projects.toList
             dag = state.build.getDagFor(project)
             fullClasspath = project.fullClasspath(dag, state.client).map(_.toString)
             environmentVariables = state.commonOptions.env.toMap
             workingDirectory = state.commonOptions.workingDirectory
+            javaOptions = project.platform match {
+              case Platform.Jvm(config, _,_) => config.javaOptions.toList
+              case _ => List()
+            }
           } yield {
-            project.platform match {
-              case Platform.Jvm(config, _, _) =>
-                Right(
                   bsp.JvmEnvironmentItem(
                     id,
                     fullClasspath.toList,
-                    config.javaOptions.toList,
+                    javaOptions,
                     workingDirectory,
-                    environmentVariables
-                  )
-                )
-              case _ => Left(s"Not a JVM project: ${project.name}")
+                    environmentVariables)
             }
-          }
 
-          def sequenceListOfEithers[L, R](
-              agg: List[R],
-              col: List[Either[L, R]]
-          ): Either[L, List[R]] = {
-            col match {
-              case Nil => Right(agg)
-              case Left(head) :: _ => Left(head)
-              case Right(head) :: tail => sequenceListOfEithers(agg :+ head, tail)
-            }
-          }
 
-          val resp = sequenceListOfEithers(List.empty, environmentEntries)
-            .map(new bsp.JvmTestEnvironmentResult(_))
-          Task.now((state, resp.swap.map(JsonRpcResponse.invalidRequest(_)).swap))
+          val resp = new bsp.JvmTestEnvironmentResult(environmentEntries)
+          Task.now((state, Right(resp)))
       }
     }
   }
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -685,18 +685,18 @@ final class BloopBspServices(
             environmentVariables = state.commonOptions.env.toMap
             workingDirectory = state.commonOptions.workingDirectory
             javaOptions = project.platform match {
-              case Platform.Jvm(config, _,_) => config.javaOptions.toList
+              case Platform.Jvm(config, _, _) => config.javaOptions.toList
               case _ => List()
             }
           } yield {
-                  bsp.JvmEnvironmentItem(
-                    id,
-                    fullClasspath.toList,
-                    javaOptions,
-                    workingDirectory,
-                    environmentVariables)
-            }
-
+            bsp.JvmEnvironmentItem(
+              id,
+              fullClasspath.toList,
+              javaOptions,
+              workingDirectory,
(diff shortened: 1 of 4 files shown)


frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -662,6 +662,45 @@ final class BloopBspServices(
...
+            project.platform match {
+              case Platform.Jvm(config, _, _) => Right(
+                bsp.JvmEnvironmentEntry(id, fullClasspath.toList,
+                  config.javaOptions.toList, workingDirectory, environmentVariables)
+              )
+              case _ => Left(s"Not a JVM project: ${project.name}")
+            }
+          }
+
+          def sequenceListOfEithers[L, R](agg: List[R], col: List[Either[L, R]]): Either[L, List[R]] = {
```
tpasternak: Maybe I should change it to `Traversable[_].sequence` from cats. But it would require `-Ypartial-unification`. Please let me know which approach you prefer.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -662,6 +668,56 @@ final class BloopBspServices(
...
+          logger.error(error)
+          Task.now((state, Left(JsonRpcResponse.invalidRequest(error))))
+
+        case Right(projects) =>
+          val environmentEntries: List[Either[String, JvmEnvironmentEntry]] = for {
+            (id, project) <- projects.toList
+            dag = state.build.getDagFor(project)
+            fullClasspath = project.fullClasspath(dag, state.client).map(_.toString)
+            environmentVariables = state.commonOptions.env.toMap
+            workingDirectory = state.commonOptions.workingDirectory
```
olafurpg: I think it would be fine to silently ignore non-JVM targets here. It's expected that non-JVM targets have no JVM execution environment.
```suggestion
            workingDirectory = state.commonOptions.workingDirectory
            config <- project.platform match {
              case Platform.Jvm(config, _, _) => Some(config)
              case _ => None
            }         
```

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -662,6 +668,56 @@ final class BloopBspServices(
...
+      mapToProjects(params.targets, state) match {
+        case Left(error) =>
+          logger.error(error)
+          Task.now((state, Left(JsonRpcResponse.invalidRequest(error))))
+
+        case Right(projects) =>
+          val environmentEntries: List[Either[String, JvmEnvironmentEntry]] = for {
+            (id, project) <- projects.toList
+            dag = state.build.getDagFor(project)
+            fullClasspath = project.fullClasspath(dag, state.client).map(_.toString)
```
olafurpg: We should update the implementation of `project.fullClasspath` to automatically use manifest jars. It would be good to move the logic from here into some reusable method, or change `JvmProcessForker` to use `project.fullClasspath`

https://github.com/scalacenter/bloop/blob/dfd03f5333c2fc292db78bfcd4205a0cdd828967/frontend/src/main/scala/bloop/exec/JvmProcessForker.scala#L158-L189

I would personally prefer if the manifest jar did not get deleted after execution finishes, I've had several cases where I have been unable to debug the java invocation because the manifest jar got deleted.

olafurpg: This will change also fix an issue where `bloop console ...` currently fails for large classpath since I see `bloop console` also uses `project.fullClasspath`

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
olafurpg: I think cats is on the classpath because of the circe dependency via lsp4s. I think it's best not to use cats APIs directly, however.

tpasternak: Not applicable if we silently ignore non-jvm configs

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
tpasternak: Done

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
tpasternak: I'm wondering what would be the right approach there. Currently we create `manifest.jar` as a tmp file, so keeping it forever would lead to a file leak. We have a few choices here:

1. Let it leak
2. Change file location from a random tmp file to a deterministic one
3. Generate manifest.jar on intellij site while running the app

olafurpg: I would be fine with 1 or 2. The manifest jars are normally small since they only contain a MANIFEST.MD text file that points to other paths on disk so I don't think it's a big concern if we use 1) with temporary files and let them leak. Option 2 might be good if we use a stable temporary directory and regularly garbage collect (for example when the directory contains more than N number of manifest jars). I would like to avoid 3 since it means that every client of `jvmTestEnvironment` will need to deal with manifest jars.

What do you think @jvican ?

olafurpg: I think it's helpful that the manifest jar doesn't get deleted immediately since that allows the user to re-run the `java` command from the terminal. For example, IntelliJ displays the full `java -classpath ...` invocation when you run tests.

tpasternak: If we are going to generate manifest.jar in server side, we probably need to extend `JvmTestEnvironmentParams` to `manifestJar:String` field

olafurpg: Following up from an offline conversation. We agreed that option 3 (generate manifest.jar on the intellij side) is the best way to go forward. The reasoning is that this approach allows intellij to delete the manifest jar after running tests. BSP clients are expected to handle too large classpaths anyways (they should not rely on the server producing a manifest jar).

jvican: Sounds good. I don't have an opinion on this. However, having an option in bloop to generate manifest jars in the future might prove handy.

One approach that hasn't been considered would be a variation of 2: have a deterministic path within `.bloop/$project` where bloop writes the manifest jar used for the last invocation. Every time a test execution happens, it gets overwritten with the manifest used to run tests. This would make it useful for debugging. The manifest jar is just a bunch of paths so I think it's benign to keep the manifest jar with the dated paths in a fixed file location.

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
```diff
@@ -95,6 +94,39 @@ class BspProtocolSpec(
...
+
+      val projects = List(`A`)
+      loadBspState(workspace, projects, logger) { state =>
+        val (stateA, result) = state.jvmTestEnvironment(`A`, None)
+        assert(stateA.status == ExitStatus.Ok)
+        val environmentItem = result.items.head
+        assert(result.items.size == 1)
+        assert(environmentItem.environmentVariables.contains("BLOOP_OWNER"))
+        assert(Paths.get(environmentItem.workingDirectory).getFileName.toString == "frontend")
+        assert(environmentItem.classpath.exists(_.contains(s"target/${`A`.config.name}")))
```
olafurpg: This is assertion is failing on Windows.

```
X bloop.bsp.TcpBspProtocolSpec.check the correct contents of jvm test environment 65ms 
  utest.AssertionError: assert(environmentItem.classpath.exists(_.contains(s"target/${`A`.config.name}")))
environmentItem: ch.epfl.scala.bsp.JvmEnvironmentItem = JvmEnvironmentItem(BuildTargetIdentifier(Uri(file:/C:/Users/RUNNER~1/AppData/Local/Temp/bloop-test-workspace9860621520280569103/a/?id=a)),List(C:\Users\runneradmin\AppData\Local\Temp\bloop-test-workspace9860621520280569103\target\a\bloop-bsp-clients-classes\classes-test-bloop-client-70Ud6cGnSdKcGnICDKsqhw==, C:\
```

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
olafurpg: I agree that option 2 is desirable. I think it's worth implementing to improve the Bloop cli testing experience (easier to debug) but it's not necessary to unblock the IntelliJ BSP testing experience.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -662,6 +668,42 @@ final class BloopBspServices(
...
+          Task.now((state, Left(JsonRpcResponse.invalidRequest(error))))
+
+        case Right(projects) =>
+          val environmentEntries = for {
+            (id, project) <- projects.toList
+            dag = state.build.getDagFor(project)
+            fullClasspath = project.fullClasspath(dag, state.client).map(_.toString)
+            environmentVariables = state.commonOptions.env.toMap
+            workingDirectory = state.commonOptions.workingDirectory
+            javaOptions = project.platform match {
```
olafurpg: This should be
```diff
-             javaOptions = project.platform match {
+             javaOptions <- project.platform match {
```
So that we don't return classpaths for non-JVM projects

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -36,20 +42,19 @@ import monix.execution.atomic.AtomicInt
...
-
+import scala.util.{Failure, Success, Try}
 import monix.execution.Cancelable
 import io.circe.{Decoder, Json}
 import bloop.engine.Feedback
 import monix.reactive.subjects.BehaviorSubject
 import bloop.engine.tasks.compilation.CompileClientStore
 import bloop.data.ClientInfo.BspClientInfo
 import bloop.logging.BloopLogger
+import cats.Traverse
```
olafurpg: unused import

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
tpasternak: Hm, I'm not sure. Please note, we have list of lists here. The return value of this endpoint has  `items: List[JvmEnvironmentItem]`, and the `JvmEnvironmentItem` contains `jvmOptions: List[String]`

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
tpasternak: Done

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
olafurpg: Right, what if we do it like this?
```scala
            javaOptions <- project.platform match {
              case Platform.Jvm(config, _, _) => Some(config.javaOptions.toList)
              case _ => None
            }
```

The endpoint explicitly requests the JVM environment so I'm not sure if we should return the classpath for Native/JS targets 🤔

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
```diff
@@ -95,6 +96,42 @@ class BspProtocolSpec(
...
+        jvmConfig = jvmConfig
+      )
+
+      val projects = List(`A`)
+      loadBspState(workspace, projects, logger) { state =>
+        val (stateA, result) = state.jvmTestEnvironment(`A`, None)
+        assert(stateA.status == ExitStatus.Ok)
+        val environmentItem = result.items.head
+        assert(result.items.size == 1)
+        assert(environmentItem.environmentVariables.contains("BLOOP_OWNER"))
```
jvican: We generally prefer to use `assertNoDiff` to check invariants when possible. This check and the jvm options check are good candidates for `assertNoDiff`, could you change it to use that?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -36,13 +42,11 @@ import monix.execution.atomic.AtomicInt
 import monix.execution.atomic.AtomicBoolean
 
 import scala.collection.concurrent.TrieMap
-import scala.collection.mutable
+import scala.collection.{immutable, mutable}
```
jvican: Might want to remove `immutable` as it's unused

PR #1173, opened 2020-02-14
Fix error checking blacklisted versions

jvican:
This pull request fixes a bug detecting when a version is blacklisted or
not. The issue was happening because 2.12.10 starts with 2.12.1, so it
was being considered blacklisted.

I would expected codebases on 2.12.10 and forward will run faster after
this change.


jvican:
Tests passed. Merged this because I want to test it works locally as well

PR #1174, opened 2020-02-14
Add default `-Xss1m` compiler option

jvican:
The stack size required by many macros such as shapeless is large.
Sometimes, compilations would fail on the bloop side because the stack
size is exceeded. This commit uses a larger default of 1m for the stack
that should remove these issues in the future.


jvican:
Let's get this merged! :smile:

PR #1175, opened 2020-02-14
Use right working directory to run app or tests from sbt

### LINKED ISSUE #1167: Bloop does not use the root of the project as working directory
[Relevant gitter conversation](https://gitter.im/scalacenter/bloop?at=5e3b3200594a0517c26c32be) 

Bloop follows SBT's behaviour when fork is enabled. SBT considers the submodule's folder as CWD when fork is enabled, but the root of the project when fork is disabled. 

My personal opinion on the matter is that this behaviour makes little sense : 

1. As @tpolecat points out, forking is disabled by default in SBT. I don't think many people are aware that enabling it has the effect of changing the working directory when running/testing programs. Therefore, bloop's default behaviour differs from SBT's default behaviour. 

2. Intuitively, I personally expect that a JVM program should consider whatever folder I run it from to be the CWD, unless specified otherwise. If my main is `println(os.pwd)` using [os-lib](https://github.com/lihaoyi/os-lib), and I run it from `~/foo`, I'd expect to see `/Users/myUserName/foo` printed in the terminal, whether I run this program from bloop or from a packaged jar . Not following this behaviour makes it weird when developing CLI tools using bloop 

3. macros that capture location information, such as the ones from [sourcecode](https://github.com/lihaoyi/sourcecode), are extremely useful for things like test-frameworks. In particular, it enables neat workflows in vscode when you click on an error in your terminal to jump directly to the relevant piece of code. However, bloop makes it harder for that workflow to work, by making it impossible to create a sensible relative path since the root's of the project isn't available from the runtime unless you are aware that you need to set it manually in your java options. As a result, bloop, which makes running tests one of its goals, gives me a worse UX than the build tool when I'm testing. 

I hope that my examples show that this behaviour breaks [the principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). Moreover, using SBT's behaviour as a justification for it goes against bloop being build-tool-agnostic, as the exact same behaviour impacts all users using alternative bloop-able build tools  😉

jvican:
Implements the approach discussed in the ticket.

Fixes #1167

### DIFF
--- a/frontend/src/test/scala/bloop/RunSpec.scala
+++ b/frontend/src/test/scala/bloop/RunSpec.scala
@@ -197,12 +197,10 @@ class RunSpec {
       val mainClassName = "hello.ShowCwd"
       val state = loadTestProject("cross-test-build-scalajs-0.6")
       val command = Commands.Run(List("test-project"), Some(mainClassName), args = List.empty)
-      val targetMsg = {
-        if (bloop.util.CrossPlatform.isWindows) "cross-test-build-scalajs-0.6\\test-project\\jvm"
-        else "cross-test-build-scalajs-0.6/test-project/jvm"
-      }
+      val targetMsg = "cross-test-build-scalajs-0.6"
 
       runAndCheck(state, command) { messages =>
+        println(messages.mkString(System.lineSeparator))
         assert(
           messages.reverse.find(_._1 == "info").exists(_._2.endsWith(targetMsg))
         )
--- a/frontend/src/test/scala/bloop/RunSpec.scala
+++ b/frontend/src/test/scala/bloop/RunSpec.scala
@@ -200,7 +200,6 @@ class RunSpec {
       val targetMsg = "cross-test-build-scalajs-0.6"
 
       runAndCheck(state, command) { messages =>
-        println(messages.mkString(System.lineSeparator))
         assert(
           messages.reverse.find(_._1 == "info").exists(_._2.endsWith(targetMsg))
         )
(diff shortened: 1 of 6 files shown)


frontend/src/test/scala/bloop/RunSpec.scala
```diff
@@ -197,12 +197,10 @@ class RunSpec {
...
       val state = loadTestProject("cross-test-build-scalajs-0.6")
       val command = Commands.Run(List("test-project"), Some(mainClassName), args = List.empty)
-      val targetMsg = {
-        if (bloop.util.CrossPlatform.isWindows) "cross-test-build-scalajs-0.6\\test-project\\jvm"
-        else "cross-test-build-scalajs-0.6/test-project/jvm"
-      }
+      val targetMsg = "cross-test-build-scalajs-0.6"
 
       runAndCheck(state, command) { messages =>
+        println(messages.mkString(System.lineSeparator))
```
olafurpg: Remove
```suggestion

```

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -784,7 +788,10 @@ object BloopDefaults {
       }
     } else {
       Def.task {
-        val config = Config.JvmConfig(Some(javaHome.toPath), javaOptions.toList)
+        val isForkedExecution = (Keys.fork in configuration in forkScopedTask).value
+        val workingDir = if (isForkedExecution) Keys.baseDirectory.value else rootBaseDirectory
+        val extraJavaOptions = List(s"-Duser.dir=${workingDir.getAbsolutePath}")
+        val config = Config.JvmConfig(Some(javaHome.toPath), (javaOptions ++ extraJavaOptions).toList)
```
olafurpg: You might want to swap the order so users can still override `-Duser.dir`
```suggestion
        val config = Config.JvmConfig(Some(javaHome.toPath), (extraJavaOptions ++ javaOptions ++).toList)
```

jvican:
All good suggestions, will address tonight

jvican:
Suggestions addressed, thanks for the review @olafurpg

PR #1176, opened 2020-02-14
Add new "sourcesGlobs" field to Bloop JSON configuration.

### LINKED ISSUE #1164: Add support for sources globs
Currently, the Bloop JSON configuration has as "sources" field which contains a list of files and directories.
```json
{
  "sources": ["/path/to/A.scala", "/path/to/src/scala"]
}
```

Some build tools like Pants and Bazel support globs like this
```
scala_library(
  name = "foo",
  sources = rglobs("Foo*.scala", excludes = ["*Test.scala"])
)
```

The sources for the "foo" target include

* all files that match the glob `Foo*.scala`
* excluding files that end with `*Test.scala`

It would be nice if the Bloop JSON config supported this somehow, for example with a new `sourcesGlobs` field
```json
{
  "sources": [ "... same as before ..." ],
  "sourcesGlobs": {
    "includes": ["/path/to/**/Foo*.scala"],
    "excludes": ["/path/to/**/*Test.scala"]
  }
}
```

olafurpg:
Fixes #1164.

Previously, Bloop targets could only have either source files or
source directories. This caused problems when integrating with builds
tools like Pants that allow targets to have declare the list of source
files via include/exclude globs.

Now, Bloop targets can specify the list of source files via file globs
using the new "sourcesGlobs" field in the Bloop JSON configuration
section. For example,
```json
{
  "sources": ["/path/to/Foo.scala"],
  "sourcesGlobs": [
      {
        "directory": "/glob/directory",
        "walkDepth": 3, // optional
        "includes": ["*.scala"],
        "excludes": ["*Test.scala"]
      }
  ]
}
```

The example above would expand to all Scala source files that are in
depth 3 from the directory "globDirectory" and don't have a filename
that ends with "*Test.scala".

BSP does not support file globs so the `buildTargets/sources` request
expands file globs at request time.

### DIFF
--- a/config/src/main/scala/bloop/config/Config.scala
+++ b/config/src/main/scala/bloop/config/Config.scala
@@ -221,11 +221,19 @@ object Config {
       modules: List[Module]
   )
 
+  case class SourcesGlobs(
+      directory: Path,
+      walkDepth: Option[Int],
+      includes: List[String],
+      excludes: List[String]
+  )
+
   case class Project(
       name: String,
       directory: Path,
       workspaceDir: Option[Path],
       sources: List[Path],
+      sourcesGlobs: Option[List[SourcesGlobs]],
       dependencies: List[String],
       classpath: List[Path],
       out: Path,
@@ -241,7 +249,7 @@ object Config {
 
   object Project {
     // FORMAT: OFF
-    private[bloop] val empty: Project = Project("", emptyPath, None, List(), List(), List(), emptyPath, emptyPath, None, None, None, None, None, None, None)
+    private[bloop] val empty: Project = Project("", emptyPath, None, List(), None, List(), List(),  emptyPath, emptyPath, None, None, None, None, None, None, None)
     // FORMAT: ON
 
     def analysisFileName(projectName: String) = s"$projectName-analysis.bin"
@@ -282,6 +290,7 @@ object Config {
         workingDirectory,
         Some(workingDirectory),
         List(sourceFile),
+        None,
         List("dummy-2"),
         List(scalaLibraryJar),
         outDir,
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -934,7 +934,7 @@ final class BloopBspServices(
       val response = bsp.SourcesResult(
         projects.iterator.map {
           case (target, project) =>
-            val sources = project.sources.map {
+            val sources = project.allSourceFilesAndDirectories().map {
               s =>
                 import bsp.SourceItemKind._
                 val uri = s.underlying.toUri()
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -930,30 +930,30 @@ final class BloopBspServices(
         projects: Seq[ProjectMapping],
         state: State
     ): BspResult[bsp.SourcesResult] = {
-
-      val response = bsp.SourcesResult(
-        projects.iterator.map {
-          case (target, project) =>
-            val sources = project.allSourceFilesAndDirectories().map {
-              s =>
-                import bsp.SourceItemKind._
-                val uri = s.underlying.toUri()
-                val (bspUri, kind) = if (s.exists) {
-                  (uri, if (s.isFile) File else Directory)
-                } else {
-                  val fileMatcher = FileSystems.getDefault.getPathMatcher("glob:*.{scala, java}")
-                  if (fileMatcher.matches(s.underlying.getFileName)) (uri, File)
-                  // If path doesn't exist and its name doesn't look like a file, assume it's a dir
-                  else (new URI(uri.toString + "/"), Directory)
-                }
-                // TODO(jvican): Don't default on false for generated, add this info to JSON fields
-                bsp.SourceItem(bsp.Uri(bspUri), kind, false)
+      val sourcesItems = projects.iterator.map {
+        case (target, project) =>
+          project.allSourceFilesAndDirectories.map { sources =>
+            val items = sources.map { s =>
+              import bsp.SourceItemKind._
+              val uri = s.underlying.toUri()
+              val (bspUri, kind) = if (s.exists) {
+                (uri, if (s.isFile) File else Directory)
+              } else {
+                val fileMatcher = FileSystems.getDefault.getPathMatcher("glob:*.{scala, java}")
+                if (fileMatcher.matches(s.underlying.getFileName)) (uri, File)
+                // If path doesn't exist and its name doesn't look like a file, assume it's a dir
+                else (new URI(uri.toString + "/"), Directory)
+              }
+              // TODO(jvican): Don't default on false for generated, add this info to JSON fields
+              bsp.SourceItem(bsp.Uri(bspUri), kind, false)
             }
-            bsp.SourcesItem(target, sources)
-        }.toList
-      )
+            bsp.SourcesItem(target, items)
+          }
+      }.toList
 
-      Task.now((state, Right(response)))
+      Task.sequence(sourcesItems).map { items =>
+        (state, Right(bsp.SourcesResult(items)))
+      }
     }
 
     ifInitialized(None) { (state: State, logger: BspServerLogger) =>
--- a/frontend/src/main/scala/bloop/data/Project.scala
+++ b/frontend/src/main/scala/bloop/data/Project.scala
@@ -28,6 +28,7 @@ import ch.epfl.scala.{bsp => Bsp}
 import xsbti.compile.{ClasspathOptions, CompileOrder}
 import bloop.config.ConfigCodecs
 import scala.util.control.NonFatal
+import monix.eval.Task
 
 final case class Project(
     name: String,
@@ -86,7 +87,7 @@ final case class Project(
   }
 
   /** Returns concatenated list of "sources" and expanded "sourcesGlobs". */
-  def allSourceFilesAndDirectories(): List[AbsolutePath] = {
+  def allSourceFilesAndDirectories: Task[List[AbsolutePath]] = Task {
     val buf = mutable.ListBuffer.empty[AbsolutePath]
     buf ++= sources
     sourcesGlobs.foreach { glob =>
--- a/frontend/src/main/scala/bloop/data/Project.scala
+++ b/frontend/src/main/scala/bloop/data/Project.scala
@@ -246,7 +246,7 @@ object Project {
       scala.map(_.options).getOrElse(Nil),
       project.java.map(_.options).getOrElse(Nil),
       project.sources.map(AbsolutePath.apply),
-      SourcesGlobs.fromConfig(project),
+      SourcesGlobs.fromConfig(project, logger),
       project.test.map(_.frameworks).getOrElse(Nil),
       project.test.map(_.options).getOrElse(Config.TestOptions.empty),
       AbsolutePath(project.out),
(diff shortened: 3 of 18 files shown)


frontend/src/main/scala/bloop/data/SourcesGlobs.scala
```diff
@@ -0,0 +1,69 @@
...
+      includes: List[String],
+      excludes: List[String]
+  ): List[SourcesGlobs] = {
+    if (includes.isEmpty) Nil
+    else {
+      val uriPath = directory.toUri().getPath()
+      val basePath =
+        if (Properties.isWin) uriPath.stripPrefix("/")
+        else uriPath
+      val prefix = s"glob:${basePath}"
```
olafurpg: It would be nice to avoid the prefix here for (hypothetically) performance reasons but I struggled to get it working correctly

Screenshot 2020-02-14 at 18 53 01

I expected `glob:*Test.scala` to match `/Users/lgeirsson/FooTest.scala` 🤔

frontend/src/main/scala/bloop/data/SourcesGlobs.scala
```diff
@@ -0,0 +1,69 @@
+package bloop.data
+
+import java.nio.file.FileSystems
+import java.nio.file.PathMatcher
+import java.nio.file.Path
+import bloop.config.Config
+import scala.util.Properties
+
+case class SourcesGlobs(
+    directory: Path,
```
olafurpg: Should this be `AbsolutePath` instead? I tried doing that refactoring but ended up only only unwrapping it at every call-site

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
```diff
@@ -344,6 +345,76 @@ class BspProtocolSpec(
...
+      val `A` = baseProject.copy(config = baseProject.config.copy(sources = Nil))
+
+      val projects = List(`A`)
+      loadBspState(workspace, projects, logger) { state =>
+        val project = state.underlying.build.loadedProjects.head
+        def assertSourcesMatches(expected: String): Unit = {
+          val obtained = for {
+            item <- state.requestSources(`A`).items
+            source <- item.sources
+          } yield globDirectory.relativize(source.uri.toPath).iterator().asScala.mkString("/")
```
olafurpg: In `scala.meta.io.RelativePath.toRelative(isDirectory: Boolean)` we do the same processing: `.iterator().asScala.mkString("/")` 

Worth adding to relative path? `nio.Path.toUri` absolutizes relative paths

frontend/src/main/scala/bloop/data/SourcesGlobs.scala
olafurpg: I figured out how to avoid the absolute globs. The trick was the relativize the path before calling `matches()`, which makes sense.

frontend/src/main/scala/bloop/data/SourcesGlobs.scala
olafurpg: Done, I refactored to use `AbsolutePath`.

frontend/src/test/scala/bloop/SourcesGlobsCompileSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+                excludes = List("glob:*Test.scala")
+              )
+            )
+          )
+        )
+      )
+      val projects = List(`A`)
+      val state = loadState(workspace, projects, logger)
+      val compiledState = state.compile(`A`)
+      assertExitStatus(compiledState, ExitStatus.Ok)
```
jvican: Should we use `assertValidCompilationState` to make sure that the compilation class files and certain file system invariants hold after compilation?

frontend/src/test/scala/bloop/SourcesGlobsCompileSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+      assertNoDiff(
+        logger.compilingInfos.mkString("\n"),
+        "Compiling a (1 Scala source)"
+      )
+
+      // Create new file and assert that we now compile two files instead of one file.
+      logger.clear()
+      Files.write(globDirectory.resolve("Foo2.scala"), Array.emptyByteArray)
+      val compiledState2 = state.clean(`A`).compile(`A`)
+      assertNoDiff(
```
jvican: Should we make sure the compilation state is valid here as well as before to avoid unexpected errors?

frontend/src/test/scala/bloop/io/SourcesGlobsSpec.scala
```diff
@@ -0,0 +1,163 @@
...
+  checkGlobs(
+    "double-asterisk-slash",
+    List("Foo.scala", "inner/Bar.scala"),
+    List("glob:**/*.scala"),
+    List(),
+    // NOTE(olafur): globs in Bloop follow `java.nio.file.PathMatcher`
+    // semantics, which may be incompatible with how other builds tools
+    // interpret file globs. For example, "Foo.scala" matches the glob "**/*" in
+    // Pants while it doesn't with Bloop. Client tools are responsible for
+    // pre-processing globs so that they match `PathMatcher` semantics.
```
jvican: This is a good point, I'll take this into account when documenting this change in release notes + website

frontend/src/test/scala/bloop/io/SourcesGlobsSpec.scala
```diff
@@ -0,0 +1,163 @@
...
+import java.nio.file.Files
+import scala.concurrent.Await
+import scala.concurrent.duration.FiniteDuration
+import bloop.util.TestProject
+import java.util.concurrent.TimeUnit
+import bloop.cli.ExitStatus
+import bloop.config.Config
+import bloop.data.SourcesGlobs
+
+object SourcesGlobsSpec extends bloop.testing.BaseSuite {
```
jvican: Great test suite to encode the semantics of the sources globs!

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
jvican: Yes, I think it makes sense to do so :+1:

frontend/src/main/scala/bloop/data/SourcesGlobs.scala
```diff
@@ -0,0 +1,80 @@
...
+            directory,
+            walkDepth.getOrElse(Int.MaxValue),
+            includes = includes.map(fs.getPathMatcher),
+            excludes = excludes.map(fs.getPathMatcher)
+          )
+        )
+      } catch {
+        case NonFatal(e) =>
+          logger.error(
+            s"ignoring invalid 'sourcesGlobs' object containing directory '$directory' in project '$projectName'"
```
jvican: Nit: capital `I`?

frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -76,6 +85,30 @@ final case class Project(
     customWorkingDirectory.getOrElse(baseDirectory)
   }
 
+  /** Returns concatenated list of "sources" and expanded "sourcesGlobs". */
+  def allSourceFilesAndDirectories(): List[AbsolutePath] = {
+    val buf = mutable.ListBuffer.empty[AbsolutePath]
+    buf ++= sources
+    sourcesGlobs.foreach { glob =>
+      if (Files.isDirectory(glob.directory.underlying)) {
+        Files.walkFileTree(
```
jvican: What would happen if this walk fails? e.g. another process could be going through the same tree and modifying it concurrently, which could cause this to throw exceptions. Would wrapping this in a `Task` make sense? I see the only call site is `BloopBspServices`.

frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -76,6 +85,30 @@ final case class Project(
     customWorkingDirectory.getOrElse(baseDirectory)
   }
 
+  /** Returns concatenated list of "sources" and expanded "sourcesGlobs". */
+  def allSourceFilesAndDirectories(): List[AbsolutePath] = {
```
jvican: Should we remove the extra `()`?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -934,7 +934,7 @@ final class BloopBspServices(
       val response = bsp.SourcesResult(
         projects.iterator.map {
           case (target, project) =>
-            val sources = project.sources.map {
+            val sources = project.allSourceFilesAndDirectories().map {
```
jvican: Nit: should we remove extra parens?

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -221,11 +221,19 @@ object Config {
       modules: List[Module]
   )
 
+  case class SourcesGlobs(
+      directory: Path,
```
jvican: Looks great! :sparkles:

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
olafurpg: Sounds good. Done, I ported the implementation from `scala.meta.io.RelativePath.toUri()` which handles trickier cases like URI-encoding individual parts of paths.

frontend/src/test/scala/bloop/SourcesGlobsCompileSpec.scala
olafurpg: Done.

frontend/src/test/scala/bloop/SourcesGlobsCompileSpec.scala
olafurpg: Good idea. Done.

frontend/src/main/scala/bloop/data/SourcesGlobs.scala
olafurpg: Done. In Metals I used the opposite convention (always start  with lowercase) and make the same nit comments 😅

frontend/src/main/scala/bloop/data/Project.scala
olafurpg: Good point. I updated this method to return a `Task`. I didn't add error handling but make the `Task` propagate to the return value in `BloopBspServices`.

frontend/src/main/scala/bloop/data/Project.scala
olafurpg: Removed. I added the parentheses to indicate it's a stateful method (it returns different values depending on I/O) on an otherwise stateless case class. Now that it returns a `Task` it's stateless :)

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
olafurpg: Removed.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -930,30 +930,30 @@ final class BloopBspServices(
...
+      val sourcesItems = projects.iterator.map {
+        case (target, project) =>
+          project.allSourceFilesAndDirectories.map { sources =>
+            val items = sources.map { s =>
+              import bsp.SourceItemKind._
+              val uri = s.underlying.toUri()
+              val (bspUri, kind) = if (s.exists) {
+                (uri, if (s.isFile) File else Directory)
+              } else {
+                val fileMatcher = FileSystems.getDefault.getPathMatcher("glob:*.{scala, java}")
```
stuhood: How are you implementing file watching for your glob expansion? Unlike literal file matches, this requires a watch on the parent directory.

olafurpg: This particular method is called for the BSP “buildTarget/sources” request. File watching for globs will need to watch the “SourcesGlobs.directory” field (not implemented in this PR)

PR #1178, opened 2020-02-21
Upgrade nuprocess version

jvican:
We need the fix in https://github.com/brettwooldridge/NuProcess/pull/107
to run bloop in systems that use Zulu. Otherwise, logic breaks with:

```
[E] Error looking up function 'Java_java_lang_ProcessImpl_init': /usr/lib/jvm/zulu-8-amd64/jre/lib/amd64/libjava.so: undefined symbol: Java_java_lang_ProcessImpl_init
```


tgodzik:
Updated in https://github.com/scalacenter/bloop/pull/1501 - everything seems to work fine. We can inline `compileSources` if any issues pop up.

PR #1183, opened 2020-02-27
Add support for packagePrefix

lukaszwawrzyk:
See https://github.com/build-server-protocol/build-server-protocol/pull/109 for details

I added packagePrefix at project/build target level. Not sure if it should be more granular, and if so, where exactly it should be located as we have sources that is a list of files or dirs but also sourceGlobs. Sources seems to usually contain specific files, attaching packagePrefix to each of these would be quite strange.

### DIFF
--- a/config/src/main/scala/bloop/config/Config.scala
+++ b/config/src/main/scala/bloop/config/Config.scala
@@ -234,6 +234,7 @@ object Config {
       workspaceDir: Option[Path],
       sources: List[Path],
       sourcesGlobs: Option[List[SourcesGlobs]],
+      sourceRoots: Option[List[String]],
       dependencies: List[String],
       classpath: List[Path],
       out: Path,
@@ -249,7 +250,7 @@ object Config {
 
   object Project {
     // FORMAT: OFF
-    private[bloop] val empty: Project = Project("", emptyPath, None, List(), None, List(), List(),  emptyPath, emptyPath, None, None, None, None, None, None, None)
+    private[bloop] val empty: Project = Project("", emptyPath, None, List(), None, None, List(), List(),  emptyPath, emptyPath, None, None, None, None, None, None, None)
     // FORMAT: ON
 
     def analysisFileName(projectName: String) = s"$projectName-analysis.bin"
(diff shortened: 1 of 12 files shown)


lukaszwawrzyk:
@jvican Could you take a look at this PR and let me know if it makes sense from bloops perspective, and if not what would be the reasonable place to put the package prefix field?

jvican:
Will have a look at this shortly @lukaszwawrzyk

config/src/main/scala/bloop/config/Config.scala
```diff
@@ -234,6 +234,7 @@ object Config {
       workspaceDir: Option[Path],
       sources: List[Path],
       sourcesGlobs: Option[List[SourcesGlobs]],
+      sourceRoots: Option[List[String]],
```
olafurpg: Should this be `Option[List[Path]]`?

olafurpg: Even if bloop doesn't need this typed as a `Path`, I believe it's more readable for clients using bloop-config that this should be a path  and not a random string. Additionally, `Path` uses structural sharing resulting in lower memory usage.

project/Dependencies.scala
```diff
@@ -13,7 +13,7 @@ object Dependencies {
   val nailgunCommit = "a2520c1e"
 
   val zincVersion = "1.3.0-M4+32-b1accb96"
-  val bspVersion = "2.0.0-M5"
+  val bspVersion = "2.0.0-M5+1-e75ba0aa" // TODO bump bsp to make it compile
```
olafurpg: The release is out now
```suggestion
  val bspVersion = "2.0.0-M7" // TODO bump bsp to make it compile
```

olafurpg:
The Windows CI failure looks unrelated
```
 (error,Server disconnected unexpectedly: [WinError 10053] An established connection was aborted by the software in your host machine)
  bloop.nailgun.NailgunTestUtils$Client.expectSuccess(NailgunTestUtils.scala:226)
```

jvican:
> I'm not sure if we should block this PR on tests however. What do you think @jvican?
  
We can go ahead and merge, no problem. I encourage @lukaszwawrzyk to add some tests for this when he's implemented it; the more tested every feature is the better.

PR #1184, opened 2020-02-27
Wip/fix hydra tests

dotta:
I'm hopeful this PR will fix the execution of tests with Hydra. Let's see.

About the PR itself, please have a look at the commit messages. In particular, we should discuss the thread-safe Reporter and possibly a better name for the IBloopAnalysisCallback :)

PR #1190, opened 2020-03-04
Fix hydra integration and make tests succeed

dotta:
Mirrors https://github.com/scalacenter/bloop/pull/1184

This PR should fix the execution of tests with Hydra. It includes two important fixes:

1) Thread-safe analysis callback, which is required by Hydra since several compilation threads may be stepping on each others.
2) Updated the bloop-hydra-bridge version as it contains an important fix to
correctly handle invalidated classfiles.

For more details, please have a look at the commit messages.

Aspects that still need to be improved:

1) Shall we rename `IBloopAnalysisCallback` to something else?
2) A thread-safe Reporter is likely need for correct execution with Hydra. But switching the mutable data structures used in the different Reporter implementations into concurrent ones might be a first good approximation.

### DIFF
--- a/backend/src/main/scala/bloop/reporter/Reporter.scala
+++ b/backend/src/main/scala/bloop/reporter/Reporter.scala
@@ -30,7 +30,7 @@ abstract class Reporter(
     val logger: Logger,
     override val cwd: AbsolutePath,
     override val config: ReporterConfig,
-    val _problems: mutable.Buffer[ProblemPerPhase] = mutable.ArrayBuffer.empty
+    val _problems: Reporter.Buffer[ProblemPerPhase]
 ) extends ZincReporter {
   private case class PositionId(sourcePath: String, offset: Int)
   private val _severities = TrieMap.empty[PositionId, Severity]
@@ -307,5 +307,68 @@ object Reporter {
         }
     }
     problemsPerFile.toMap
+  }
+
+  import java.util.concurrent.{ConcurrentLinkedQueue => JConcurrentLinkedQueue}
+  import scala.collection.JavaConverters._
+  import scala.reflect.ClassTag
+  object Buffer {
+    def apply[A](needsConcurrentBuffer: Boolean): Buffer[A] = {
+      if (needsConcurrentBuffer) new ConcurrentBuffer[A](new JConcurrentLinkedQueue())
+      else new DelegatingBuffer[A](mutable.ArrayBuffer.empty)
+    }
+  }
+
+  sealed trait Buffer[A] extends Any {
+    def +=(element: A): Unit
+    def clear(): Unit
+    def exists(predicate: A => Boolean): Boolean
+    def foreach(f: A => Unit): Unit
+    def map[B](f: A => B): Buffer[B]
+    def reverse: Buffer[A]
+    def toArray[B >: A: ClassTag]: Array[B]
+    def toList: List[A]
+  }
+
+  private class DelegatingBuffer[A](val underlying: mutable.Buffer[A])
+      extends AnyVal
+      with Buffer[A] {
+    override def +=(element: A): Unit = underlying += element
+    override def clear(): Unit = underlying.clear()
+    override def exists(predicate: A => Boolean): Boolean = underlying.exists(predicate)
+    override def foreach(f: A => Unit): Unit = underlying.foreach(f)
+    override def map[B](f: A => B): Buffer[B] = new DelegatingBuffer(underlying.map(f))
+    override def reverse: Buffer[A] = new DelegatingBuffer(underlying.reverse)
+    override def toList: List[A] = underlying.toList
+    override def toArray[B >: A: ClassTag]: Array[B] = underlying.toArray
+  }
+
+  private class ConcurrentBuffer[A](val underlying: JConcurrentLinkedQueue[A])
+      extends AnyVal
+      with Buffer[A] {
+    override def +=(element: A): Unit = { underlying.add(element); () }
+    override def clear(): Unit = underlying.clear()
+    override def exists(predicate: A => Boolean): Boolean = weakIterator.exists(predicate)
+    override def foreach(f: A => Unit): Unit = weakIterator.foreach(f)
+    override def map[B](f: A => B): Buffer[B] = {
+      val mapped = new JConcurrentLinkedQueue[B]()
+      weakIterator.foreach(element => mapped.add(f(element)))
+      new ConcurrentBuffer(mapped)
+    }
+    override def reverse: Buffer[A] = {
+      val reversed = mutable.Stack.empty[A]
+      weakIterator.foreach(reversed.push)
+      newBuffer(reversed: _*)
+    }
+    override def toList: List[A] = weakIterator.toList
+    override def toArray[B >: A: ClassTag]: Array[B] = weakIterator.toArray
+
+    private def newBuffer[B](elements: B*): ConcurrentBuffer[B] = {
+      val mapped = new JConcurrentLinkedQueue[B]()
+      elements.foreach(mapped.add)
+      new ConcurrentBuffer(mapped)
+    }
+
+    private def weakIterator: Iterator[A] = underlying.iterator().asScala
   }
 }
--- a/backend/src/main/scala/bloop/reporter/Reporter.scala
+++ b/backend/src/main/scala/bloop/reporter/Reporter.scala
@@ -14,6 +14,7 @@ import scala.collection.mutable
 import scala.util.Try
 import bloop.logging.CompilationEvent
 import scala.concurrent.Promise
+import monix.execution.atomic.AtomicInt
 
 /**
  * A flexible reporter whose configuration is provided by a `ReporterConfig`.
@@ -21,6 +22,9 @@ import scala.concurrent.Promise
  * etc.
  *
  * A reporter has internal state and must be instantiated per compilation.
+ *
+ * Note: Implementations must be thread-safe or concurrency hazards will surface
+ *       when compilation is carried out with Hydra.
  *
  * @param logger The logger that will receive the output of the reporter.
  * @param cwd    The current working directory of the user who started compilation.
@@ -36,14 +40,13 @@ abstract class Reporter(
   private val _severities = TrieMap.empty[PositionId, Severity]
   private val _messages = TrieMap.empty[PositionId, List[String]]
 
-  private var _nextID = 1
-  private def nextID(): Int = { val id = _nextID; _nextID += 1; id }
+  private val _nextID = AtomicInt(1)
+  private def nextID(): Int = _nextID.getAndIncrement()
   override def reset(): Unit = {
     _problems.clear()
     _severities.clear()
     _messages.clear()
-    _nextID = 1
-    ()
+    _nextID.set(1)
   }
 
   override def hasWarnings(): Boolean = {
--- a/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopIncremental.scala
+++ b/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopIncremental.scala
@@ -33,7 +33,8 @@ object BloopIncremental {
       options: IncOptions,
       mode: CompileMode,
       manager: ClassFileManager,
-      tracer: BraveTracer
+      tracer: BraveTracer,
+      isHydraEnabled: Boolean
   ): Task[(Boolean, Analysis)] = {
     def getExternalAPI(lookup: Lookup): (File, String) => Option[AnalyzedClass] = { (_: File, binaryClassName: String) =>
       lookup.lookupAnalysis(binaryClassName) flatMap {
@@ -51,7 +52,11 @@ object BloopIncremental {
     val internalBinaryToSourceClassName = (binaryClassName: String) => previousRelations.productClassName.reverse(binaryClassName).headOption
     val internalSourceToClassNamesMap: File => Set[String] = (f: File) => previousRelations.classNames(f)
 
-    val builder = () => new BloopAnalysisCallback(mode, internalBinaryToSourceClassName, internalSourceToClassNamesMap, externalAPI, current, output, options, manager)
+    val builder: () => IBloopAnalysisCallback = {
+      if (!isHydraEnabled)() => new BloopAnalysisCallback(mode, internalBinaryToSourceClassName, internalSourceToClassNamesMap, externalAPI, current, output, options, manager)
+      else
+        () => new ConcurrentAnalysisCallback(mode, internalBinaryToSourceClassName, internalSourceToClassNamesMap, externalAPI, current, output, options, manager)
+    }
     // We used to catch for `CompileCancelled`, but we prefer to propagate it so that Bloop catches it
     compileIncremental(sources, uniqueInputs, lookup, previous, current, compile, builder, reporter, log, output, options, manager, tracer)
   }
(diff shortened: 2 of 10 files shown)


backend/src/main/scala/bloop/reporter/Reporter.scala
```diff
@@ -30,6 +30,8 @@ abstract class Reporter(
     val logger: Logger,
     override val cwd: AbsolutePath,
     override val config: ReporterConfig,
+    // FIXME: _problems needs to be thread-safe. Shall we use a Java ConcurrentLinkedQueue or would you prefer something else?
+    //        Keeping the mutable buffer can cause Hydra to crash because of off-by-one exceptions thrown for instance by Buffer#clear
     val _problems: mutable.Buffer[ProblemPerPhase] = mutable.ArrayBuffer.empty
```
jvican: Hmmm, would it be possible to parameterize the reporter with a buffer interface of sorts so that we can switch the implementation to `ConcurrentLinkedQueue` when using Hydra?

backend/src/main/scala/bloop/reporter/Reporter.scala
```diff
@@ -38,7 +40,10 @@ abstract class Reporter(
 
   private var _nextID = 1
   private def nextID(): Int = { val id = _nextID; _nextID += 1; id }
-  override def reset(): Unit = {
+  // FIXME: Adding a synchronized here is clearly not the proper fix, but it seems to help preventing random failures of the test. As
+  //        I'd like to see the test suite succeed on the CI, I've decided to include it in the current PR, even though we will of course
+  //        remove this before merging it.
+  override def reset(): Unit = synchronized {
```
jvican: I wonder why `reset` is being called concurrently while compiling a project. I would expect Hydra doesn't call it unless all threads are done compiling a project. I'm concerned the fact we need `synchronized` might be making some other parts in Bloop not work well with Hydra, what do you think @dotta?

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopIncremental.scala
```diff
@@ -51,7 +52,11 @@ object BloopIncremental {
     val internalBinaryToSourceClassName = (binaryClassName: String) => previousRelations.productClassName.reverse(binaryClassName).headOption
     val internalSourceToClassNamesMap: File => Set[String] = (f: File) => previousRelations.classNames(f)
 
-    val builder = () => new BloopAnalysisCallback(mode, internalBinaryToSourceClassName, internalSourceToClassNamesMap, externalAPI, current, output, options, manager)
+    val builder: () => IBloopAnalysisCallback = {
```
jvican: :+1: Looks great!

dotta:
Thanks @jvican! Will work on this right away. By the way, I think you got my github handle wrong ;-)

backend/src/main/scala/bloop/reporter/Reporter.scala
dotta: Sounds good. I'm going to have an abstract Buffer interface with two implementations: one using the current Scala `mutable.Buffer` and another one using a `ConcurrentLinkedQueue`.

jvican: Yes, I think having a simple interface with only the methods we use should be enough, thanks Mirco 👍

backend/src/main/scala/bloop/reporter/Reporter.scala
dotta: You are correct that `reset` is not called concurrently. Maybe the race condition I was seeing was an artifact of the tests, but unfortunately I'm unable to reproduce the exception (I believe the exception I was seeing was an `ArrayIndexOutOfBoundException` with index -1).

Regardless, since errors may be reported concurrently, we still need to make the underlying `_problems` data structure thread-safe. That's exactly what I'm doing and the `synchronized` will disappear.

backend/src/main/scala/bloop/reporter/Reporter.scala
dotta: Yep, that's exactly what I have. The implementations will be value classes, so we should have no allocation penalty (not that I think it really matters here, but I thought you'd still like that :)).

backend/src/main/scala/bloop/reporter/Reporter.scala
dotta: Jorge, the call to `reporter.reset` is done in [`Global`](https://github.com/scala/scala/blob/2.13.x/src/compiler/scala/tools/nsc/Global.scala#L1493). That's why the race condition with the `Reporter._problems` could surface when compilation is carried out with  Hydra.

backend/src/main/scala/bloop/reporter/Reporter.scala
jvican: Great, another way we might want to implement this is by avoiding creating an interface for a buffer and just having some abstract methods to add/manage problems in the reporter interface that are implemented in a `SingleThreadedReporter` and a `ConcurrentReporter`. I almost find this solution better than creating an interface for buffers. What do you think?

backend/src/main/scala/bloop/reporter/Reporter.scala
jvican: Right, so it is indeed being called concurrently when the compiler is preparing the compilation units. Any chance you can change this in Hydra?

dotta: Why would this need to be changed in Hydra and not in the vanilla Scala compiler too? The approach I'm taking is to make the Reporters implementation thread-safe, as we discussed.

backend/src/main/scala/bloop/reporter/Reporter.scala
dotta: I've just pushed the changes that implement the solution we initially discussed. TBH, I'd actually suggest to make the concurrent implementation the default, as I don't believe users will experience any performance penalty, and it would keep the code simpler (as there is only one implementation and all `Reporter` implementations simply need to be thread-safe)

backend/src/main/scala/sbt/internal/inc/bloop/internal/ThreadSafeAnalysisCallback.scala
```diff
@@ -0,0 +1,398 @@
...
+ * This class provides a thread-safe implementation of `xsbti.AnalysisCallback` which is required to compile with the
+ * Triplequote Hydra compiler.
+ *
+ * In essence, the implementation is a merge of the `BloopAnalysisCallback` with the original sbt default `AnalysisCallback`
+ * implementation (https://github.com/sbt/zinc/blob/develop/internal/zinc-core/src/main/scala/sbt/internal/inc/Compile.scala),
+ * which is already thread-safe.
+ *
+ * IMPORTANT: All modifications made to BloopAnalysisCallback` must be replicated here.
+ */
+final class ThreadSafeAnalysisCallback(
```
dotta: A better name would be `ConcurrentAnalysisCallback`. I'm happy to make the change if you'd like so.

dotta:
@jvican I've rebased on master (as I spotted there was a commit addressing flaky tests) and essentially the last two commits contain the new code to review (addressing thread-safety for all Reporter implementations). Please, read the commit messages for details.

frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
```diff
@@ -16,22 +16,40 @@ import scala.collection.mutable
...
+
   private lazy val taskId = logger.nextTaskId
 
   /** A cycle count, initialized to 0 when it's a no-op. */
-  private var cycleCount: Int = 0
+  private val cycleCount: AtomicInt = AtomicInt(0)
 
   /** A thread-safe map with all the files under compilation. */
-  private val compilingFiles = mutable.HashMap.empty[File, Boolean]
+  private val compilingFiles = TrieMap.empty[File, Boolean]
```
jvican: Let's use `ConcurrentHashMap` instead of `TrieMap`

backend/src/main/scala/bloop/reporter/Reporter.scala
jvican: > I'd actually suggest to make the concurrent implementation the default
  
I prefer to err on the side of caution and given that we don't have the time to run benchmarks to make sure this is the case that we keep these two implementations separate

backend/src/main/scala/bloop/reporter/Reporter.scala
jvican: Even if you make the implementation thread-safe, could we not have a race condition here? If one hydra compile run finishes early before other compile runs and calls `reset` then the problems diagnosed by the other runs could be disappeared, no?

frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
dotta: I've used a `TrieMap` for coherence with the other fields defined in the same class (see [here](https://github.com/scalacenter/bloop/blob/master/frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala#L37-L40)). But if you prefer a `ConcurrentHashMap` just for this field, I'm of course happy to make the change. Let me know!

frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
```diff
@@ -16,22 +16,40 @@ import scala.collection.mutable
...
+      config,
+      reportAllPreviousProblems,
+      createBuffer[ProblemPerPhase](project)
+    )
+
   private lazy val taskId = logger.nextTaskId
 
   /** A cycle count, initialized to 0 when it's a no-op. */
-  private var cycleCount: Int = 0
+  private val cycleCount: AtomicInt = AtomicInt(0)
```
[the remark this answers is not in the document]
dotta: Yes, otherwise there can be visibility problems due to the `Reporter` instance being accessed from different threads. Also, the commit message in https://github.com/scalacenter/bloop/pull/1190/commits/ae26519b437881291880c116740e5d649dcdd63b provides context on this change (in case you missed it)

backend/src/main/scala/sbt/internal/inc/bloop/internal/ThreadSafeAnalysisCallback.scala
dotta: Alright, I'll do it if there are other changes that are needed (which I'm not sure there will be). So, this shouldn't be blocking a merge if you are happy with the PR.

backend/src/main/scala/bloop/reporter/Reporter.scala
dotta: `reset` is not called by Hydra. And because the current implementation works when used with the vanilla Scala compiler, there should be no reason for it to not work with Hydra.

dotta: To clarify, `Reporter.reset` is called in Global as mentioned earlier, but this call happens _before_ the sources are compiled. So, while it's true there is a chance of a race-condition, empirical evidence from our experience with Hydra is that this call to `Reporter.reset` is generally executed across all Hydra threads before a problem is reported.

backend/src/main/scala/bloop/reporter/Reporter.scala
dotta: Sure. I'm totally ok with keeping things as they are. Shall we mark this conversation resolved?

backend/src/main/scala/bloop/reporter/Reporter.scala
jvican: > So, while it's true there is a chance of a race-condition, empirical evidence from our experience with Hydra is that this call to Reporter.reset

Sure, so long as we're explicit about this and we acknowledge that this race condition could happen in the future or could be reported by users, I'm fie with it :smile:

frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
jvican: Yes, please, let's use `ConcurrentHashMap` here, this field is accessed quite often so I prefer to use a more efficient implementation than `TrieMap`

jvican:
> It's worth mentioning that the `Reporter.nextID` while being now thread-safe,
may actually not return a valid ID when compilation is done with Hydra.
  
I think that this might cause some important problems, `nextID` generates a unique ID per problem and IIRC this id is relied on by clients to deduplicate problems, so it's quite critical the uniqueness property is maintained. What would it take to satisfy that?
  
Otherwise, this is looking neat! :+1: Thanks for the hard work

frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
dotta: You got it. Will mark this resolved.

backend/src/main/scala/bloop/reporter/Reporter.scala
dotta: Sounds good. I've actually filed a ticket on our side as I think this is should be easy to fix (and I think we can have a fix included in the next Hydra release). I'm going to marke this discussion resolved as there is no further action to take here.

dotta:
> I think that this might cause some important problems, nextID generates a unique ID per problem and IIRC this id is relied on by clients to deduplicate problems, so it's quite critical the uniqueness property is maintained. What would it take to satisfy that?

Actually, I believe I was too negative. `nextID` exposes the same race-condition we discussed for `Reporter.reset` (which I mentioned we should be able to fix on the Hydra side). Other than that, the returned ID will be guaranteed to be unique since it's generated with an atomic increment. So, in short, I'll update the commit message to match what I said now.

frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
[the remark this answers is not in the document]
dotta: I'm going to mark this resolved.

dotta:
@jvican I believe I've addressed all comments. Here are the set of changes I've just pushed:

* Renamed `ThreadSafeAnalysisCallback` to `ConcurrentAnalysisCallback`.
* Updated [this commit message](https://github.com/scalacenter/bloop/pull/1190/commits/6e555f9817c1f83e9b44fd6c0d779b63f7bdda7c) to correct my incorrect description of concurrency problems related to `Reporter.nextID` (the message matches what I said in [this comment](https://github.com/scalacenter/bloop/pull/1190#issuecomment-601141571)). Sorry for the confusion! (I was confused myself 😅)
* [Pushed an additional commit](https://github.com/scalacenter/bloop/pull/1190/commits/ac876943e067a9f7383c7240975e8cf162610b38) to use a Java `ConcurrentHashMap` instead of a Scala `TrieMap`.

Let me know if I forgot something.

dotta:
Whohooo!! Thank YOU @jvican 🙌 

@olafurpg @tgodzik Would be fantastic if Metals could update its version of Bloop to include the changes in this PR (and, if possible, cut a release), as I think that's the easiest way to collect feedback from users. I'd for sure use it myself 😄

tgodzik:
@dotta If there is no changes to the way the launcher works, you can just change the Bloop version inside the settings to test it out.  We're working on bumping bloop version currently, we will follow up with the version with these changes later.

dotta:
@tgodzik Excellent. I've set the Bloop version in Metals to v1.4.0-RC1-142-90456e06 and compilation with Hydra is working fine. I have one question though, how can I correlate the SHA in the Bloop version (i.e., 90456e06) to a commit in Bloop? I'm asking because it's hard to know if the version of Bloop I'm using actually includes the changes I need. I'm just gussing it does because it was published on March 24th (I could find this very useful bit of information [here](https://scalacenter.github.io/bloop/docs/launcher-reference#install-the-launcher)). Thanks!

PR #1191, opened 2020-03-05
Add debug sampling flag option for Zipkin traces

wiwa:
For certain debug use cases, we would like the Zipkin traces submitted to override sampling settings on the server by setting the `debug` flag on a trace.
This change will allow users to set a `zipkin.trace.debug` property for this purpose.

### DIFF
--- a/backend/src/main/scala/bloop/tracing/BraveTracer.scala
+++ b/backend/src/main/scala/bloop/tracing/BraveTracer.scala
@@ -4,10 +4,11 @@ import brave.{Span, Tracer}
 import brave.propagation.TraceContext
 import monix.eval.Task
 import monix.execution.misc.NonFatal
-
 import scala.util.Failure
+import scala.util.Properties
 import scala.util.Success
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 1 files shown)


backend/src/main/scala/bloop/tracing/BraveTracer.scala
```diff
@@ -4,10 +4,9 @@ import brave.{Span, Tracer}
 import brave.propagation.TraceContext
 import monix.eval.Task
 import monix.execution.misc.NonFatal
-
-import scala.util.Failure
-import scala.util.Success
+import scala.util.{Failure, Properties, Success}
```
olafurpg: nit: please configure IntelliJ to not group imports by prefix

wiwa: done

wiwa:
Without the `debug` flag, traces submitted to the Zipkin server may be lost due to how the server samples traces. The `debug` flag will let the submission override the server's attempt to neglect it.

olafurpg:
Ignoring unrelated CI failure
```
[error] sbt.librarymanagement.ResolveException: Error downloading com.lihaoyi:upickle-implicits_2.12:0.7.1
```

PR #1193, opened 2020-03-06
Use Project#getPath instead of #getName

bspaulding-invitae:
Please forgive if this solution is naive, it appeared to fix my calling project though!

From the gradle java api docs:
```
String getName()
Returns the name of this project. The project's name is not necessarily unique within a project hierarchy. You should use the getPath() method for a unique identifier for the project.
```

This _does_ generate possibly "weird" looking project names with the colons in them, ie `:one:two:three`. I don't _think_ this is a problem, but we could def transform this if that's desired to look nicer. /shrug

### DIFF
--- a/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
+++ b/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
@@ -414,10 +414,28 @@ class BloopConverter(parameters: BloopParameters) {
   }
 
   def getProjectName(project: Project, sourceSet: SourceSet): String = {
+    val projectsWithName = project.getRootProject
+      .getAllprojects()
+      .asScala
+      .filter(p => p.getName == project.getName)
+    val rawProjectName = if (projectsWithName.size == 1) project.getName else project.getPath
+    val sanitizedProjectName = rawProjectName.zipWithIndex
+      .map {
+        case (c, i) =>
+          if (i == 0 && c == ':') {
+            None
+          } else if (c == ':') {
+            Some('-')
+          } else {
+            Some(c)
+          }
+      }
+      .flatten
+      .mkString
     if (sourceSet.getName == SourceSet.MAIN_SOURCE_SET_NAME) {
-      project.getName
+      sanitizedProjectName
     } else {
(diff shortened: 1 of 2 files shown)


Arthurm1:
Hi @bspaulding-invitae 

Thanks for the contribution

This was considered before in #611.  The thoughts at the time were to switch to `getPath` only if there was a clash in project names when using `getName`.  This meant that the simpler short project names were preserved the majority of the time.  The work was never finished.

Could you change your PR to check for duplicate names before switching from `getName` to `getPath`?  All projects are available in `project.getRootProject.getAllprojects` so this should be possible.
For consistency you'd have to apply this to non-main sourcesets as well - see #611 

Colon is a reserved filename character in Windows - so you'd probably have to change that to hyphen or somesuch - again you can see this in #611 

Finally - it might be an idea to add a test to check that `getPath` is only used when hitting duplicate project names.

bspaulding-invitae:
@Arthurm1 Thanks for the review! I _did_ look for this PR, as it was mentioned on the website, but I guess I didn't look hard enough. I'll take a stab at these updates. 👍

jvican:
Thanks for the PR @bspaulding-invitae! Let me know when these updates are done and I'll review the PR.

bspaulding-invitae:
Will do @jvican, I haven't abandoned this, just been busy.

jvican:
:+1: Looks like the CI failures are legit. Let me know when they're addressed and I'll have a look at the PR.

bspaulding-invitae:
@jvican I’ve addressed the test situation, but looks like there’s a fail on windows. I could use some eyes on this. Is this related? 🧐

jvican:
Could you rebase on top of latest master?

bspaulding-invitae:
@jvican rebased, but now I also am reminded of this from @Arthurm1 above:

> Colon is a reserved filename character in Windows - so you'd probably have to change that to hyphen or somesuch - again you can see this in #611

So I think this needs another tweak there.

jvican:
> So I think this needs another tweak there.

Do you wanna do it in this PR?

bspaulding-invitae:
@jvican I'm happy to do it in this PR.

bspaulding-invitae:
@jvican We are green! 🥳 I think this is ready for a review? 🙏

Issue #1182, opened 2020-02-25
Add new "refreshProjectsCommand" workspace setting 


olafurpg:
Currently, `.bloop/bloop.settings.json` includes workspace settings like the SemanticDB version, which is only used by Metals to simplify the installation of SemanticDB.

This issue is a proposal to add a new `refreshProjectsCommand: Option[List[String]]` field with the goals to improve the IntelliJ BSP experience. The contract of the `refreshProjectCommand` field is that Bloop invokes this system command to re-export the build when users press the "Reimport All BSP Projects" button in IntelliJ.

<img width="221" alt="Screenshot 2020-02-25 at 18 02 59" src="https://user-images.githubusercontent.com/1408093/75273597-150f0300-57f9-11ea-83f8-557e2a42b7ca.png">

This allows for example the Pants/Bloop integration to export the Pants build when users press the "refresh BSP projects" button.

I have a WIP branch to Bloop (https://github.com/olafurpg/bloop/compare/refresh-project?expand=1) that implements this functionality. Demo:

![2020-02-25 18 26 26](https://user-images.githubusercontent.com/1408093/75276809-c49aa400-57fe-11ea-80c3-f2cd11b84ab3.gif)

To manually test out the changes with Pants, you will need the following change in Metals https://github.com/scalameta/metals/pull/1467. I recommend checking out that branch of Metals and run `sbt publishLocal` first.

Then create a file `bloop.settings.json`
```json
❯ cat bloop.settings.json
{
  "refreshProjectsCommand": [
    "/usr/local/bin/coursier", // EDIT THIS
    "launch",
    "org.scalameta:metals_2.12:0.8.1-SNAPSHOT",
    "--main",
    "scala.meta.internal.pantsbuild.BloopPants",
    "--",
    "refresh",
    "--workspace",
    "/Users/lgeirsson/dev/bsp-pants", // EDIT THIS
    "package-filename"
  ]
}
```

Next, open a Pants IntelliJ BSP project with the following instructions

```
git clone -b refresh-projects https://github.com/olafurpg/bsp-pants
cd intellij-bsp-pants
./setup-bsp.sh
```

Make sure to use the latest IntelliJ 2020.1 EAP, the version I've been using

```
IntelliJ IDEA 2020.1 EAP (Community Edition)
Build #IC-201.5616.10, built on February 19, 2020
Runtime version: 11.0.6+8-b722.15 x86_64
[...]

PR #1195, opened 2020-03-10
Allow regenerating bloop project data with external tool

lukaszwawrzyk:
This is my work done on top of this issue: https://github.com/scalacenter/bloop/issues/1182
In short ot adds `refreshProjectsCommand` to `bloop.settings.json` so that it is possible to regenerate bloop project data from external model on request for buildTargets. All the details are in the mentioned issue. 
I did some improvements, like ensure that project is properly reloaded after refresh, fixed tests, etc.


olafurpg:
Ideally, the integration test uses a meaningful refresh command that generates new Bloop projects and then assert that "workspace/buildTargets" returns the newly exported targets. I'm not exactly sure how to best implement such a test, maybe we can do it via the existing build tools integrations (maven, gradle or sbt) or roll something custom 🤔

lukaszwawrzyk:
> Ideally, the integration test uses a meaningful refresh command that generates new Bloop projects and then assert that "workspace/buildTargets" returns the newly exported targets. I'm not exactly sure how to best implement such a test, maybe we can do it via the existing build tools integrations (maven, gradle or sbt) or roll something custom

I think I will try with SBT

olafurpg:
@lukaszwawrzyk Did you manage to take a look at how to test this change? Please let me know if you need help with the tests.

frontend/src/test/scala/bloop/bsp/BspIntellijClientSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+  test("refresh project data on buildTargets request") {
+    TestUtil.withinWorkspace { workspace =>
+      val logger = new RecordingLogger(ansiCodesSupported = false)
+      val `A` = TestProject(workspace, "a", Nil)
+      val `B` = TestProject(workspace, "b", Nil)
+      val initialProjects = List(`A`)
+      val configDir = TestProject.populateWorkspace(workspace, initialProjects)
+
+      val refreshProjectsCommand = {
+        val content = sanitizeForBash(`B`.toJson)
```
olafurpg: I think we can avoid the `sanitizeForBash` method by writing the contents to a temporary file and then `cp` that file over. What do you think?

frontend/src/test/scala/bloop/bsp/BspIntellijClientSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+        WorkspaceSettings(None, None, refreshProjectsCommand = Some(refreshProjectsCommand))
+      WorkspaceSettings.writeToFile(configDir, workspaceSettings, logger)
+
+      loadBspState(workspace, initialProjects, logger, bspClientName = "IntelliJ") { state =>
+        def projectNames(w: WorkspaceBuildTargetsResult) = w.targets.flatMap(_.displayName)
+
+        val initialTargets = state.workspaceTargets
+        assertEquals(projectNames(initialTargets), List("a"))
+        val finalTargets = state.workspaceTargets
+        assertEquals(projectNames(finalTargets), List("a", "b"))
```
olafurpg: Amazing! Very nice to have this assertion 👍

frontend/src/test/scala/bloop/bsp/BspIntellijClientSpec.scala
lukaszwawrzyk: Yeah, the sanitizing might not be reliable. I updated the PR with `mv` on tmp file.

frontend/src/test/scala/bloop/bsp/BspIntellijClientSpec.scala
olafurpg: The Windows CI failure looks legitimate 

```diff
X bloop.bsp.TcpBspIntellijClientSpec.refresh project data on buildTargets request 274ms 
  bloop.testing.DiffAssertions$TestFailedException: failed assertion at D:\a\bloop\bloop\frontend\src\test\scala\bloop\testing\BaseSuite.scala:82
failed assertion at D:\a\bloop\bloop\frontend\src\test\scala\bloop\testing\DiffAssertions.scala:82
===========
=> Obtained
===========
List(a)

=======
=> Diff
=======
@@ -1,1 +1,1 @@
-List(a)
+List(a, b)
```

I wonder if the bash command is working correctly on Windows? It might be worth include the BSP logging statements to the assertion error.

olafurpg:
The new test is passing now on Windows, the CI failure is unrelated 
```
[error] 	bloop.DeduplicationSpec
```

frontend/src/main/scala/bloop/bsp/RefreshProjects.scala
```diff
@@ -0,0 +1,19 @@
...
+import bloop.io.AbsolutePath
+import bloop.logging.Logger
+
+import scala.sys.process.ProcessLogger
+import scala.sys.process.Process
+
+object RefreshProjects {
+  def run(workspaceDir: AbsolutePath, command: List[String], logger: Logger): Unit = {
+    val exit = Process(command, cwd = Some(workspaceDir.toFile))
+      .!(ProcessLogger(logger.info, logger.info))
```
jvican: Can we use nuprocess instead of std's `Process`? This is the standard throughout the codebase to do any kind of system process

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -870,7 +885,15 @@ final class BloopBspServices(
   def buildTargets(
       request: bsp.WorkspaceBuildTargetsRequest
   ): BspEndpointResponse[bsp.WorkspaceBuildTargetsResult] = {
-    ifInitialized(None) { (state: State, logger: BspServerLogger) =>
+    // need a separate block so so that state is refreshed after regenerating project data
+    val refreshTask = ifInitialized(None) { (state: State, logger: BspServerLogger) =>
+      state.client.refreshProjectsCommand.foreach { command =>
```
jvican: We should be running this within a `Task` so that if it ever throws (I know it doesn't now) we catch it within the `Task` and we can report the error to the client. Also, what's the contract if this command fails? I see you're right now ignoring the error, shouldn't the client know that this side effect failed?

frontend/src/main/scala/bloop/data/WorkspaceSettings.scala
```diff
@@ -40,12 +36,27 @@ import com.github.plokhotnyuk.jsoniter_scala.macros.{JsonCodecMaker, CodecMakerC
  * we should skip the resolution of the plugin.
  */
 case class WorkspaceSettings(
-    semanticDBVersion: String,
-    supportedScalaVersions: List[String]
-)
+    semanticDBVersion: Option[String],
+    supportedScalaVersions: Option[List[String]],
+    refreshProjectsCommand: Option[List[String]]
```
jvican: Nit: can we have a top-level field in the workspace settings json called `commands` and then define an object within it with the key `refreshProjects` and its value? I prefer this as it leaves the door open to new commands in the future.

frontend/src/main/scala/bloop/bsp/RefreshProjects.scala
```diff
@@ -0,0 +1,19 @@
+package bloop.bsp
+
+import bloop.io.AbsolutePath
+import bloop.logging.Logger
+
+import scala.sys.process.ProcessLogger
+import scala.sys.process.Process
+
+object RefreshProjects {
+  def run(workspaceDir: AbsolutePath, command: List[String], logger: Logger): Unit = {
```
jvican: Just to guarantee the call site this never throws, we should be returning a `Task`, what do you think? 🤔

frontend/src/main/scala/bloop/data/WorkspaceSettings.scala
jvican: I've given this some thought, I retract what I said, there's no need to do this

frontend/src/main/scala/bloop/bsp/RefreshProjects.scala
lukaszwawrzyk: I just replaced this object with call to Forker.run which returns Task

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
lukaszwawrzyk: now the behavior is defined by Forker.run logic which also returns a Task

frontend/src/main/scala/bloop/bsp/RefreshProjects.scala
lukaszwawrzyk: I used Forker.run api, which uses nuprocess under the hood

olafurpg:
@jvican stdout/stdout is logged via BSP so it's the responsibility of the client to display the failure to the user. I'm not sure if we should fail the `workspace/buildTargets` request when this command fails. We can collect some user feedback on this scenario and update later.

jvican:
>  I'm not sure if we should fail the workspace/buildTargets request when this command fails. We can collect some user feedback on this scenario and update later.
  
I'm not sure either, I just want to figure out a sensible contract. I think we will see many refresh commands failing in the future, so we should think about how we're going to communicate to the user that failure and how we're going to make it discoverable.

olafurpg:
The CI is green now. I have a feeling we will end up failing `workspace/buildTargets` when the refresh command fails, but we can followup with that change later after consulting with users

frontend/src/test/scala/bloop/bsp/BspIntellijClientSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+import ch.epfl.scala.bsp.WorkspaceBuildTargetsResult
+
+object LocalBspIntellijClientSpec extends BspIntellijClientSpec(BspProtocol.Local)
+object TcpBspIntellijClientSpec extends BspIntellijClientSpec(BspProtocol.Tcp)
+
+class BspIntellijClientSpec(
+    override val protocol: BspProtocol
+) extends BspBaseSuite {
+
+  test("refresh project data on buildTargets request") {
```
jvican: Great test! :+1:

Issue #1088, opened 2019-11-11
Emacs interlock files break Bloop compile


marcsaegesser:
When a buffer is modified, Emacs creates an [interlock](https://www.gnu.org/software/emacs/manual/html_node/emacs/Interlocking.html#Interlocking) file in order to detect and prevent another user (or another Emacs instance) from editing the same file and possibly overwriting changes. These interlock files have the form .#filename and (at least on Linux) are symlinks to a non-existent file whose names encodes the user and process that has unsaved modifications to the file.

>  ls -al .\#*
lrwxrwxrwx 1 marc marc 33 Nov 11 09:11 .#Day10.scala -> marc@marc-desktop.7691:1573423575

The existence of one of these interlock files breaks Bloop compiles:

> [E] Failed to compile 'advent2016'
> Watching 1 directories... (press Ctrl-C to interrupt)
> [E] File watching threw an exception: 
> /home/marc/work/AdventOfCode2016/src/main/scala/.#Day10.scala

The interlock files are deleted when a file is saved so there may be timing issues as well depending on when Bloop sees the modified files and when Emacs deletes the interlock file. SBT is not affected at all by the existence of these interlock files.

I encountered this issue via Metals and it is being discussed as issue [956](https://github.com/scalameta/metals/issues/956) there. Within Emacs/Metals, this error manifests as compile errors not being updated as code changes. New errors do not get highlighted and fixed errors don't get removed. Forcing a Bloop compile when there are no interlock files appears to clear the problem until another interlock is created.

The only work-around is to disable the creation of interlock files by setting the Emacs variable ```create-lockfiles``` to ```nil```.

jvican:
Hi @marcsaegesser, thanks for reporting! It looks this could be fixed in `SourceWatcher.scala` by intentionally skipping these interlock files. I would also re-run the command with `--verbose` to have more information about the type of exception thrown by the file watcher, that should tell us what's at fault here.
  
Is this issue something you'd be interested in working on?

marcsaegesser:
Running with --verbose I get the following (with the very annoying ANSI escape codes removed):

> Loading workspace settings from bloop.settings.json
> 
> Computing sources and classpath hashes for advent2016
> Failed to compile 'advent2016'
> Watching 1 directories... (press Ctrl-C to interrupt)
> /home/marc/work/AdventOfCode2016/src/main/scala
> Assuming ExtendedWatchEventModifier.FILE_TREE is not supported
[...]

PR #1202, opened 2020-03-17
Don't match on hidden source files

jvican:
To avoid conflicts when leading with Emacs Interlock files, we
 whitelist hidden source files from the `SourceWatcher` so that fired
 modifications on these files are ignored and from the `SourceHasher` so
 that these files are never discovered and hashed during the
 `CompileInputs` bundling.

 Fixes https://github.com/scalacenter/bloop/issues/1088


jvican:
Merging as CI error looks unrelated and published bloop version will be used for testing

PR #1203, opened 2020-03-18
Respect global configuration in `$HOME/.bloop/bloop.json`

olafurpg:
Previously, it was only possible to configure JVM options for the Bloop
server via the `$HOME/.bloop/.jvmopts` file. This commit adds a new way
to globally configure Bloop via `$HOME/.bloop/bloop.json`.
The JSON file supports two fields:

* `javaOptions`: which is equivalent to `$HOME/.bloop/.jvmopts`. For
  backwards compatibility, Bloop continues to respect `.jvmopts` but for
  new usage we should start recommending `bloop.json` instead.
* `javaHome`: to control what Java version the Bloop server should use.
  Currently, it's difficult for users to control exactly what Java
  version is picked when launching Bloop. For example, Java 11 may get
  automatically used by IntelliJ 2020.1 calls the Bloop launcher
  because IntelliJ runs on Java 11 itself.

Parse errors in `~/.bloop/bloop.json` cause the Bloopgun/Launcher
processes to fail.

To simplify manual testing of this feature, Bloopgun now accepts a new
--bloop-version flag to override the default Bloop version. This makes
it possible to `bloopgun/run` from sbt shell without a `publishLocal`
step.

### DIFF
--- a/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
+++ b/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
@@ -443,6 +443,10 @@ class BloopgunCli(
         if (globalOptions.nonEmpty) {
           logger.info(s"Picked up custom Java options $globalOptions")
         }
+        // NOTE: global options take precedence because `bloop.json` is the
+        // recommended way to customize the JVM options and users should be able
+        // use `bloop.json` without having to learn about caveats when it
+        // doesn't work.
         baseOptions ++ globalOptions
       }
 
@@ -509,7 +513,11 @@ class BloopgunCli(
         StatusCommand(code, output)
       } catch {
         case e: IOException =>
-          StatusCommand(1, e.getMessage())
+          logger.trace(e)
+          StatusCommand(
+            1,
+            s"Failed to start process '${cmd.mkString(" ")}'\nError: ${e.getMessage()}"
+          )
       }
     }
 
--- a/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
+++ b/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
@@ -29,7 +29,6 @@ import scala.collection.mutable
 
 import scopt.OParser
 
-import scala.tools.nsc.Global
 import scala.sys.process.ProcessIO
 import java.lang.ProcessBuilder.Redirect
 import scala.util.Try
@@ -64,7 +63,7 @@ import java.io.IOException
  * because it can affect already connected clients to the server instance.
  */
 class BloopgunCli(
-    bloopVersion: String,
+    baseBloopVersion: String,
     in: InputStream,
     out: PrintStream,
     err: PrintStream,
@@ -72,12 +71,12 @@ class BloopgunCli(
     cwd: Path
 ) {
   def this(
-      bloopVersion: String,
+      baseBloopVersion: String,
       in: InputStream,
       out: PrintStream,
       err: PrintStream,
       shell: Shell
-  ) = this(bloopVersion, in, out, err, shell, Environment.cwd)
+  ) = this(baseBloopVersion, in, out, err, shell, Environment.cwd)
   def run(args: Array[String]): Int = {
     var setServer: Boolean = false
     var setPort: Boolean = false
@@ -200,7 +199,7 @@ class BloopgunCli(
 
     val (cliArgsToParse, extraArgsForServer) =
       if (args.contains("--")) args.span(_ == "--") else (args, Array.empty[String])
-    OParser.parse(cliParser, cliArgsToParse, BloopgunParams(bloopVersion), setup) match {
+    OParser.parse(cliParser, cliArgsToParse, BloopgunParams(baseBloopVersion), setup) match {
       case None => 1
       case Some(params0) =>
         val params = params0.copy(args = additionalCmdArgs ++ extraArgsForServer)
@@ -442,7 +441,7 @@ class BloopgunCli(
           }
         val globalOptions = globalSettings.javaOptions.getOrElse(Nil)
         if (globalOptions.nonEmpty) {
-          logger.info(s"Picked up global Java options $globalOptions")
+          logger.info(s"Picked up custom Java options $globalOptions")
         }
         baseOptions ++ globalOptions
       }
--- a/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
+++ b/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
@@ -87,9 +87,14 @@ class BloopgunCli(
     val cliParser = {
       val builder = OParser.builder[BloopgunParams]
       val bloopVersionOpt = builder
-        .opt[String]("bloop-version")
-        .action((bloopVersion, params) => { params.copy(bloopVersion = bloopVersion) })
-        .text("Specify the Bloop version to use for launching a new Bloop server")
+        .opt[String]("fallback-bloop-version")
+        .action((fallbackBloopVersion, params) => {
+          params.copy(bloopVersion = fallbackBloopVersion)
+        })
+        .text(
+          "Specify the Bloop version to use for launching a new Bloop server when there is no running server. " +
+            "This option is ignored when there is a running server in the background."
+        )
       val nailgunServerOpt = builder
         .opt[String]("nailgun-server")
         .action((server, params) => { setServer = true; params.copy(nailgunServer = server) })
@@ -436,6 +441,9 @@ class BloopgunCli(
             jvmOpts ++ extraJvmOpts
           }
         val globalOptions = globalSettings.javaOptions.getOrElse(Nil)
+        if (globalOptions.nonEmpty) {
+          logger.info(s"Picked up global Java options $globalOptions")
+        }
         baseOptions ++ globalOptions
       }
 
--- a/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
+++ b/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
@@ -4,6 +4,7 @@ import bloop.bloopgun.core.Shell
 import bloop.bloopgun.core.DependencyResolution
 import bloop.bloopgun.util.Environment
 import bloop.bloopgun.util.Feedback
+import bloop.bloopgun.util.GlobalSettings
 import bloop.bloopgun.core.AvailableAtPath
 import bloop.bloopgun.core.AvailableWithCommand
 import bloop.bloopgun.core.ListeningAndAvailableAt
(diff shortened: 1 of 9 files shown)


olafurpg:
Creating this as a draft since there are no tests yet. I have so far only manually checked that it works as expected. @jvican any advice on testing integration like these?

olafurpg:
Also, where would be a good place to document this? I was thinking of adding two section to https://scalacenter.github.io/bloop/docs/launcher-reference.html#usage  for "Customize Java version for the Bloop server" and "Customize JVM options for the Bloop server"

jvican:
Hi @olafurpg, thanks for sending this PR! I have a few questions before I jump to reviewing code:
  
> For backward compatibility, Bloop continues to respect .jvmopts but for
new usage we should start recommending bloop.json instead.
  
How do we handle conflicts in the interim? If a user has both `.jvmopts` and the new `.bloop/bloop.json`, which JVM options wins over? Do we merge them together?
  
> To simplify manual testing of this feature, Bloopgun now accepts a new
--bloop-version flag to override the default Bloop version.
  
I'm concerned about adding such a flag. If the user runs bloopgun with the expectation of starting a concrete bloop version, we get into version management territory. For example, what should we do if a bloop server with a higher version is running? What if it's lower?
  
Right now, bloopgun is released at the same cadence as the bloop server and its job is simple: it starts a bloop server only if there's no server running. At the moment bloopgun accepts bloop versions via `--bloop-version`, the user expectation is that `bloopgun` will comply with that version requirement.
  
Maybe the way to avoid those expectations is by changing the name of the argument to something like `--fallback-bloop-version`.
  
> I have so far only manually checked that it works as expected. @jvican any advice on testing integration like these?
  
Regarding integration testing, have you had a look at `LauncherSpec`? It creates a virtual environment of sorts where you can add this file and then run the launcher to check how it behaves.
  
> Also, where would be a good place to document this?  I was thinking of adding two section to scalacenter.github.io/bloop/docs/launcher-reference.html#usage for "Customize Java version for the Bloop server" and "Customize JVM options for the Bloop server"
  
We should document this in the build server reference https://scalacenter.github.io/bloop/docs/server-reference We can add some information to the launcher reference and then link out to the build server reference to better explain how users can tweak the server behavior. Does this make sense?

olafurpg:
> How do we handle conflicts in the interim? If a user has both .jvmopts and the new .bloop/bloop.json, which JVM options wins over? Do we merge them together?

I this PR, I concatenate `.jvmopts` and `bloop.json` options so that the JSON configuration takes precedence. We could also error if both are non-empty if you think that would be better, the same way the launcher fails if `bloop.json` has a parse error.

> Maybe the way to avoid those expectations is by changing the name of the argument to something like --fallback-bloop-version.

This is a good idea, I will apply the change. The flag is indeed only used as a fallback when there is no running server. I can imagine using it with the native-image I have installed on my computer to start a bloop server with a custom version (currently I use a verbose coursier launch command for that).

----


I'll check it the tests and docs, thanks!

olafurpg:
Any idea what might cause the following error? It happens when running launcher tests on GraalVM 19.2
```
[error] java.lang.IllegalStateException: Classes to be shaded have to be zero because namespaces should cover all
[error] 	at build.Shading$.toShadeClasses(Shading.scala:93)
[error] 	at build.BloopShadingPlugin$.$anonfun$projectSettings$2(ShadingPlugin.scala:80)
```

jvican:
Can you add `"com.github.plokhotnyuk.jsoniter_scala"` to the `shadeNamespaces` setting inside the function `shadeSettingsForModule` in `build.sbt`?

olafurpg:
The CI is green now. The Windows failure is unrelated
```
- * SuiteSelector - sbt.ForkMain$ForkError: java.lang.OutOfMemoryError: Metaspace
```

olafurpg:
>  Can you add "com.github.plokhotnyuk.jsoniter_scala" to the shadeNamespaces setting inside the function shadeSettingsForModule in build.sbt?

This fixed the GraalVM issues!

jvican:
Thanks for the work here @olafurpg! I will review this tomorrow :+1:

bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
```diff
@@ -84,6 +85,15 @@ class BloopgunCli(
 
     val cliParser = {
       val builder = OParser.builder[BloopgunParams]
+      val bloopVersionOpt = builder
+        .opt[String]("fallback-bloop-version")
```
jvican: Great, I think this is a good name for this argument :+1:

bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
```diff
@@ -417,15 +428,22 @@ class BloopgunCli(
...
+          if (extraJvmOpts.forall(opt => jvmOpts.contains(opt))) jvmOpts
+          else {
+            usedExtraJvmOpts = true
+            jvmOpts ++ extraJvmOpts
+          }
+        val globalOptions = globalSettings.javaOptions.getOrElse(Nil)
+        if (globalOptions.nonEmpty) {
+          logger.info(s"Picked up custom Java options $globalOptions")
         }
+        baseOptions ++ globalOptions
```
jvican: Global options have precedence over base jvm options, is that something intended?

bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
```diff
@@ -483,19 +501,26 @@ class BloopgunCli(
...
-      StatusCommand(code, output)
+      try {
+        val started = process.start()
+        val is = started.getInputStream()
+        val code = started.waitFor()
+        val output = scala.io.Source.fromInputStream(is).mkString
+        StatusCommand(code, output)
+      } catch {
+        case e: IOException =>
+          StatusCommand(1, e.getMessage())
```
jvican: We might want to give some context to this message, I imagine that if the error message is too generic we might have a hard time tracking where the error originated from. What do you think?

docs/server.md
```diff
@@ -55,6 +55,41 @@ If you are integrating your tool with bloop and want to install and start the se
   <dd><p>Use the server jar or script in the given path</p></dd>
 </dl>
 
+## Global settings for the server
```
jvican: I wonder if we should change all references to `.jvmopts` in favor of this approach... The build server references mentions `.jvmopts` often.

launcher/src/test/scala/bloop/launcher/GlobalSettingsSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  def runBspLauncherWithGlobalJsonSettings(json: String)(fn: BspLauncherResult => Unit): Unit = {
+    writeJsonSettings(json)
+    val result = runBspLauncherWithEnvironment(launcherArguments, shellWithPython)
+    fn(result)
+  }
+
+  test("fail to start server when bloop.json has syntax error") {
```
jvican: Great tests :sparkles:

launcher/src/test/scala/bloop/launcher/GlobalSettingsSpec.scala
```diff
@@ -0,0 +1,74 @@
+package bloop.launcher
+
+import java.nio.file.Files
+import scala.concurrent.Promise
+import bloop.bloopgun.util.Environment
+import bloop.internal.build.BuildInfo
+
+object GlobalSettingsSpec extends LauncherBaseSuite("1.3.2", BuildInfo.bspVersion, 9014) {
```
jvican: Should we make this suite run and install the latest bloop version (`BuildInfo.version`) instead of `1.3.2`?
  
When you select the launcher with the latest version, it doesn't run by default unless you `install` it first in sbt. This step happens automatically on CI.
  
Testing the launcher works this way because we need to use the python script to install bloop and resolve its artifacts, and these resources are only generated during `install`.
  
We could probably update the `1.3.2` reference in `LauncherSpec` too. I imagine you copied this from there, we were testing the launcher spec for 1.3.2 because the launcher was released in 1.3.3 and I wanted to make sure that the launcher would work even if people used 1.3.2. We no longer need to guarantee that.

olafurpg: I made the changes in `GlobalSettingsSpec` but left `LauncherSpec` unchanged since there is already a spec covering `BuildInfo.version`
```scala
object LatestStableLauncherSpec extends LauncherSpec("1.3.2")
object LatestMasterLauncherSpec extends LauncherSpec(BuildInfo.version)
```

bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
olafurpg: I added a note on why global settings take precedence.

olafurpg:
@jvican looking at the logs, can you please confirm that `GlobalSettingsSpec` is running as expected?

```
 -------------------------------- Running Tests --------------------------------
java.lang.RuntimeException: The server did not start, got FailedToConnectToServer
	at bloop.launcher.LauncherMain.failPromise$1(Launcher.scala:95)
	at bloop.launcher.LauncherMain.runLauncher(Launcher.scala:122)
	at bloop.launcher.LauncherMain.cli(Launcher.scala:82)
	at bloop.launcher.LauncherBaseSuite.$anonfun$runBspLauncherWithEnvironment$2(LauncherBaseSuite.scala:338)
	at bloop.launcher.LauncherBaseSuite.$anonfun$runBspLauncherWithEnvironment$2$adapted(LauncherBaseSuite.scala:336)
	at bloop.launcher.LauncherBaseSuite.setUpLauncher(LauncherBaseSuite.scala:232)
	at bloop.launcher.LauncherBaseSuite.$anonfun$runBspLauncherWithEnvironment$1(LauncherBaseSuite.scala:336)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at monix.eval.internal.TaskRunLoop$.monix$eval$internal$TaskRunLoop$$loop$1(TaskRunLoop.scala:187)
	at monix.eval.internal.TaskRunLoop$.startWithCallback(TaskRunLoop.scala:260)
	at monix.eval.internal.TaskRunLoop$.$anonfun$restartAsync$1(TaskRunLoop.scala:59)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
+ bloop.launcher.GlobalSettingsSpec.fail to start server when bloop.json has syntax error 1101ms  
```

olafurpg:
Nevermind, I just realized that the tests are stressing that the server does not start successfully. The stack traces are expected. This PR is ready to go!

PR #1207, opened 2020-03-21
Don't create the hydra.license file for PRs external to the repo

dotta:
The Hydra tests are skipped if the hydra.license doesn't exist. If a PR is
opened from an external repo, repositories' secrets cannot be decripted and
hence the hydra.license file should not be created because its content would be
wrong.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,9 +34,7 @@ jobs:
           curl -Lo coursier https://git.io/coursier-cli && chmod +x coursier && ./coursier --help
           yarn --help
           java -version
-          mkdir -p $HOME/.triplequote && echo "$HYDRA_LICENSE" > "$HOME/.triplequote/hydra.license"
-          echo "$HOME"
-          echo "$HOME/.triplequote/hydra.license" | wc -c
+          [[ $HYDRA_LICENSE == floating-key=* ]] && mkdir -p $HOME/.triplequote && echo "$HYDRA_LICENSE" > "$HOME/.triplequote/hydra.license" || echo "Hydra license file was not created"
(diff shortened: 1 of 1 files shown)


dotta:
@olafurpg Very sorry for the disruption I've created on your PR https://github.com/scalacenter/bloop/pull/1203. But I'm confident things will work as expected once this PR is merged.

olafurpg:
No worries! It’s great that the Hydra tests can now run in CI 😄

dotta:
@olafurpg Thanks for merging!

Unfortunately, the current solution is still suboptimal as Hydra tests will run only when the PR branch lives in the main repo (because secrets are not accessible from forks). However, I'm thinking we could at least make sure they are run during the [release step](https://github.com/scalacenter/bloop/blob/master/.github/workflows/ci.yml#L87). This way at least we can be aware of problems and it will be easier to pinpoint what change caused a Hydra test failure. WDYT?

olafurpg:
I think it's OK to only test this integration on master for now, assuming it's possible to manually configure the license when testing locally. The release step should ideally only release so that the artifacts get published as fast as possible. We're often consuming the latest CI release from other projects like Metals, adding a 30 minute delay cuts time off the day for us to work on these projects.

Before stable releases, we can manually inspect the CI logs to ensure that the Hydra integration is not broken.

jvican:
Was planning to look at this today, thanks for submitting this PR to unblock the open PRs @dotta 😄

dotta:
@jvican Sure thing! It was just an oversight on my part, as I completely forgot I had to get back to it. Anyway, I'm glad I didn't cause too much disruption :)

Wish you all a great weekend!

PR #1208, opened 2020-03-22
Add sbt scripted test for hydra sbt integration

dotta:
The test checks that the generated bloop configuration includes the main Hydra
artifacts.


dotta:
The failure is spurious and it's unrelated from the changes in this PR.

Here is the failing test:

> X bloop.dap.DebugServerSpec.does not accept a connection unless the previous session requests a restart 5622ms

PR #1211, opened 2020-03-24
Hide fine-grained Zipkin traces behind system property

wiwa:
This change introduces a way to hide some verbose fine-grained zipkin traces with a `bloop.trace.verbose` system property which defaults to `false`. This reduces the number of spans in a given trace around the neighborhood of 80% by culling some traces when the amount of time spent in those traced operations is very low, think microseconds or a few milliseconds in a compile taking a few seconds.

@olafurpg

### DIFF
--- a/backend/src/main/scala/bloop/tracing/BraveTracer.scala
+++ b/backend/src/main/scala/bloop/tracing/BraveTracer.scala
@@ -26,34 +26,46 @@ final class BraveTracer private (
   }
 
   def trace[T](name: String, tags: (String, String)*)(
-      thunk: BraveTracer => T
+      thunk: BraveTracer => T,
+      verbose: Boolean = false
   ): T = {
-    val newTracer = startNewChildTracer(name, tags: _*)
-    try thunk(newTracer) // Don't catch and report errors in spans
-    catch {
-      case NonFatal(t) =>
-        newTracer.currentSpan.error(t)
-        throw t
-    } finally {
-      try newTracer.terminate()
-      catch { case NonFatal(t) => () }
+    if (!verbose || BraveTracer.verboseTrace) {
+      val newTracer = startNewChildTracer(name, tags: _*)
+      try thunk(newTracer) // Don't catch and report errors in spans
+      catch {
+        case NonFatal(t) =>
+          newTracer.currentSpan.error(t)
+          throw t
+      } finally {
+        try newTracer.terminate()
+        catch {
+          case NonFatal(t) => ()
+        }
+      }
+    } else {
+      thunk(this)
     }
   }
 
   def traceTask[T](name: String, tags: (String, String)*)(
-      thunk: BraveTracer => Task[T]
+      thunk: BraveTracer => Task[T],
+      verbose: Boolean = false
   ): Task[T] = {
-    val newTracer = startNewChildTracer(name, tags: _*)
-    thunk(newTracer)
-      .doOnCancel(Task(newTracer.terminate()))
-      .doOnFinish {
-        case None => Task.eval(newTracer.terminate())
-        case Some(value) =>
-          Task.eval {
-            newTracer.currentSpan.error(value)
-            newTracer.terminate()
-          }
-      }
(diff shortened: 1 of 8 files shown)


olafurpg:
I think it's an acceptable compromise to disable these traces when a system property is enabled. We can enable this system property when users export Pants builds.

jvican:
Note that I'm proposing to make the system property disable these traces by default. So it's not up to Pants to enable this system property.
  
I just want an escape hatch for me and people doing performance work following our Performance Guide to get this fine-grained data if they want to. It's been proven useful in the past and I'm sure it'll prove useful in the future.

wiwa:
Thanks for feedback, updated with proposed changes.

backend/src/main/scala/bloop/tracing/BraveTracer.scala
```diff
@@ -26,34 +26,46 @@ final class BraveTracer private (
...
+        catch {
+          case NonFatal(t) => ()
+        }
+      }
+    } else {
+      thunk(this)
     }
   }
 
   def traceTask[T](name: String, tags: (String, String)*)(
```
olafurpg: nit: should we add a `def traceVerboseTask`? It doesn't read nicely with the `thunk` lambda on the same parameter list as the `verbose` flag. We could have a `private def traceTaskInternal` that takes all the parameters, which the public `traceTask` and `traceVerboseTask` forward to

backend/src/main/scala/bloop/tracing/BraveTracer.scala
```diff
@@ -78,6 +90,7 @@ object BraveTracer {
     "http://127.0.0.1:9411/api/v2/spans"
   )
   val debugTrace = Properties.propOrFalse("zipkin.trace.debug")
+  val verboseTrace = Properties.propOrFalse("zipkin.trace.verbose")
```
olafurpg: ?
```suggestion
  val isDebugEnabled = Properties.propOrFalse("zipkin.trace.debug")
  val isVerboseEnabled = Properties.propOrFalse("zipkin.trace.verbose")
```

backend/src/main/scala/bloop/tracing/BraveTracer.scala
jvican: :+1: Could be something like `private def traceTaskInternal(name: String, verbose: Boolean, tags: (String, String*))(thunk: BraveTracer => Task[T]): Task[T]`

backend/src/main/scala/bloop/tracing/BraveTracer.scala
jvican: Sounds good to me. Should we use the `bloop` prefix instead of `zipkin`?

olafurpg: Using a bloop prefix would make sense.

backend/src/main/scala/bloop/tracing/BraveTracer.scala
```diff
@@ -77,7 +111,8 @@ object BraveTracer {
   val zipkinServerUrl = Option(System.getProperty("zipkin.server.url")).getOrElse(
     "http://127.0.0.1:9411/api/v2/spans"
   )
-  val debugTrace = Properties.propOrFalse("zipkin.trace.debug")
+  val isDebugTrace = Properties.propOrFalse("zipkin.trace.debug")
```
olafurpg: Although this line is unrelated to the PR I think it's good to also use a `bloop.` prefix for consistency with `bloop.trace.verbose`
```suggestion
  val isDebugTrace = Properties.propOrFalse("bloop.trace.debug")
```

jvican:
@wiwa Any chance you can have a look at the latest feedback? Should be easy to address, we can merge right after it's ready!

backend/src/main/scala/bloop/tracing/BraveTracer.scala
wiwa: done!

PR #1212, opened 2020-03-24
Add `tags` to Bloop configuration

Duhemm:
This field is used to add tags to the build target described by a given
configuration file. The tags of this target will then be exposed via the
`tags` property of BSP.

Currently, the existing integrations set the `test` tag for test
targets, and the `library` tag for all other targets.

### DIFF
--- a/docs/assets/bloop-schema.json
+++ b/docs/assets/bloop-schema.json
@@ -580,6 +580,25 @@
               "$ref": "#/definitions/nativePlatform"
             }
           ]
+        },
+        "tags": {
+          "$id": "/properties/project/properties/tags",
+          "type": "array",
+          "title": "The Tags Schema",
+          "description": "The tags associated with that project",
+          "examples": [
+              ["test"],
+              ["library", "cross-platform"]
+          ],
+          "items": {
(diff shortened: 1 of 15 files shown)


docs/assets/bloop-schema.json
```diff
@@ -580,6 +580,25 @@
               "$ref": "#/definitions/nativePlatform"
             }
           ]
+        },
+        "tags": {
```
jvican: :sparkles:

integrations/sbt-bloop/src/sbt-test/sbt-bloop/cross-compile-test-configurations/build.sbt
```diff
@@ -69,18 +73,22 @@ checkBloopFile in ThisBuild := {
     fooTestConfigContents.project.sources.exists(_.toString.contains("StraySourceFile.scala")),
     "Source file is missing in foo."
   )
+  assert(hasTags(fooTestConfigContents, List(bloop.config.Tag.Test)))
```
jvican: Great tests 👍

PR #1214, opened 2020-03-25
Add buildTargets/jvmEnvironment endpoint

lukaszwawrzyk:
Depends on https://github.com/build-server-protocol/build-server-protocol/pull/113
Just implements the new endpoint, to currently be exactly the same as jvmTestEnvironment.
Fills in new optional parameters javaHome and javaVersion, currently based on the jvm that bloop is running on. In future we may allow to customize it through Config.Project

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -864,13 +864,10 @@ final class BloopBspServices(
   private def toBuildTargetId(project: Project): bsp.BuildTargetIdentifier =
     bsp.BuildTargetIdentifier(project.bspUri)
 
-  private def toJvmBuildTarget(project: Project): bsp.JvmBuildTarget = {
-    project.jdkConfig match {
-      case Some(jdk) =>
-        val javaHome = bsp.Uri(jdk.javaHome.toBspUri)
-        bsp.JvmBuildTarget(Some(javaHome), None)
-      case None =>
-        bsp.JvmBuildTarget(None, None)
+  private def toJvmBuildTarget(project: Project): Option[bsp.JvmBuildTarget] = {
+    project.jdkConfig.map { jdk =>
+      val javaHome = bsp.Uri(jdk.javaHome.toBspUri)
+      bsp.JvmBuildTarget(Some(javaHome), None)
     }
   }
 
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -2,6 +2,7 @@ package bloop.bsp
 
 import java.io.InputStream
 import java.net.URI
+import java.nio.file.Paths
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicInteger
 import java.nio.file.{FileSystems, Files, Path}
@@ -863,6 +864,13 @@ final class BloopBspServices(
   private def toBuildTargetId(project: Project): bsp.BuildTargetIdentifier =
     bsp.BuildTargetIdentifier(project.bspUri)
 
+  private def toJvmBuildTarget(project: Project): bsp.JvmBuildTarget = {
+    val javaHome = Option(System.getProperty("java.home")).map(s => bsp.Uri(Paths.get(s).toUri))
+    val javaVersion = Option(System.getProperty("java.vm.specification.version"))
+
+    bsp.JvmBuildTarget(javaHome, javaVersion)
+  }
+
   def toScalaBuildTarget(project: Project, instance: ScalaInstance): bsp.ScalaBuildTarget = {
     def toBinaryScalaVersion(version: String): String = {
       version.split('.').take(2).mkString(".")
@@ -880,7 +888,8 @@ final class BloopBspServices(
       scalaVersion = instance.version,
       scalaBinaryVersion = toBinaryScalaVersion(instance.version),
       platform = platform,
-      jars = jars
+      jars = jars,
+      jvmBuildTarget = toJvmBuildTarget(project)
     )
   }
 
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -865,10 +865,13 @@ final class BloopBspServices(
     bsp.BuildTargetIdentifier(project.bspUri)
 
   private def toJvmBuildTarget(project: Project): bsp.JvmBuildTarget = {
-    val javaHome = Option(System.getProperty("java.home")).map(s => bsp.Uri(Paths.get(s).toUri))
-    val javaVersion = Option(System.getProperty("java.vm.specification.version"))
-
-    bsp.JvmBuildTarget(javaHome, javaVersion)
+    project.jdkConfig match {
+      case Some(jdk) =>
+        val javaHome = bsp.Uri(jdk.javaHome.toBspUri)
+        bsp.JvmBuildTarget(Some(javaHome), None)
+      case None =>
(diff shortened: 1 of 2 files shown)


frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -863,6 +864,13 @@ final class BloopBspServices(
   private def toBuildTargetId(project: Project): bsp.BuildTargetIdentifier =
     bsp.BuildTargetIdentifier(project.bspUri)
 
+  private def toJvmBuildTarget(project: Project): bsp.JvmBuildTarget = {
+    val javaHome = sys.props.get("java.home").map(s => bsp.Uri(Paths.get(s).toUri))
```
olafurpg: nit: `sys.props` allocates a new immutable map on every method call so I personally use `Option(System.getProperty(..))` instead

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -863,6 +864,13 @@ final class BloopBspServices(
   private def toBuildTargetId(project: Project): bsp.BuildTargetIdentifier =
     bsp.BuildTargetIdentifier(project.bspUri)
 
+  private def toJvmBuildTarget(project: Project): bsp.JvmBuildTarget = {
+    val javaHome = Option(System.getProperty("java.home")).map(s => bsp.Uri(Paths.get(s).toUri))
```
olafurpg: I just learned from an offline discussion with @jvican that Bloop supports configuring custom Java home per-project. See `.platform` here https://github.com/scalacenter/bloop/blob/906110dc384dbebb8d7d1b2f62bd226231a4f3bf/frontend/src/main/scala/bloop/data/Project.scala#L52

We should use `JdkConfig.javaHome` instead of the `java.home` system property.

build.sbt
```diff
@@ -27,6 +27,7 @@ val benchmarkBridge = project
 lazy val bloopShared = (project in file("shared"))
   .settings(
     name := "bloop-shared",
+    resolvers += Resolver.sonatypeRepo("snapshots"),
```
olafurpg: We should cut a new stable BSP release so we don't use SNAPSHOTs

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -863,6 +864,20 @@ final class BloopBspServices(
   private def toBuildTargetId(project: Project): bsp.BuildTargetIdentifier =
     bsp.BuildTargetIdentifier(project.bspUri)
 
+  private def toJvmBuildTarget(project: Project): bsp.JvmBuildTarget = {
+    Option(project.platform)
```
jvican: We have some of this logic already in a method called `project.jdkConfig`. What do you think about using that instead of this? We can also replace the values of `java.home` and the java version and use those that are available in `JavaRuntime`.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -863,6 +864,20 @@ final class BloopBspServices(
...
+      .collect {
+        case jvm: Platform.Jvm =>
+          val javaHome = bsp.Uri(jvm.config.javaHome.toBspUri)
+          bsp.JvmBuildTarget(Some(javaHome), None)
+      }
+      .getOrElse {
+        val javaHome = Option(System.getProperty("java.home")).map(s => bsp.Uri(Paths.get(s).toUri))
+        val javaVersion = Option(System.getProperty("java.vm.specification.version"))
+        bsp.JvmBuildTarget(javaHome, javaVersion)
+      }
```
jvican: Isn't this method returning a jvm build target always, regardless of the platform?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -863,6 +864,20 @@ final class BloopBspServices(
...
+  private def toJvmBuildTarget(project: Project): bsp.JvmBuildTarget = {
+    Option(project.platform)
+      .collect {
+        case jvm: Platform.Jvm =>
+          val javaHome = bsp.Uri(jvm.config.javaHome.toBspUri)
+          bsp.JvmBuildTarget(Some(javaHome), None)
+      }
+      .getOrElse {
+        val javaHome = Option(System.getProperty("java.home")).map(s => bsp.Uri(Paths.get(s).toUri))
+        val javaVersion = Option(System.getProperty("java.vm.specification.version"))
```
jvican: Are you really use you want `java.vm.specification.version` here instead of `java.version`?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
lukaszwawrzyk: yeah, I guess it is enough to use project.jdkConfig, anyway it sets the correct jvm default config if not provided, and if it returns none, JvmBuildTarget can be empty as it means native or js.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
jvican: Why return a `JvmBuildTarget` here for projects that are not JVM?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -863,6 +864,17 @@ final class BloopBspServices(
   private def toBuildTargetId(project: Project): bsp.BuildTargetIdentifier =
     bsp.BuildTargetIdentifier(project.bspUri)
 
+  private def toJvmBuildTarget(project: Project): bsp.JvmBuildTarget = {
+    project.jdkConfig
```
jvican: Nit: could we make all of this as concise as possible by using pattern matching here ?

lukaszwawrzyk: ok

lukaszwawrzyk:
> What happened to the use of `java.specification.vm.version`?

So I addressed the comment about not filling in JvmBuildTarget in case when target is not jvm.
If project.jdkConfig is None, it means it is Native or Js, so in such case I put `JvmBuildTarget(None, None)` rather than bloop runtime jvm. I actually never fill in javaVersion, as it is not available in `project.platform`, and I can't assume that `sys.props("java.version")` is valid for given `project.jdkConfig.javaHome`

jvican:
@lukaszwawrzyk Great, thanks for the changes!
  
>  so in such case I put JvmBuildTarget(None, None) rather than bloop runtime jvm. 
  
Why do we return `JvmBuildTarget(None, None)` instead of `None`? This looks like a limitation in the BSP design of `JvmBuildTarget`, we should not assume every `ScalaBuildTarget` has a `JvmBuildTarget`, it should accept `Option[JvmBuildTarget]` instead.
  
@olafurpg Any thoughts?

olafurpg:
Good catch! The `ScalaBuildTarget.jvmBuildTarget` field needs to be typed as `Option[BuildTarget]`, otherwise new clients won't be able to parse `ScalaBuildTarget` from old servers.

olafurpg:
We will need to do the same change for bsp4j, this field is optional.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -2,6 +2,7 @@ package bloop.bsp
 
 import java.io.InputStream
 import java.net.URI
+import java.nio.file.Paths
```
olafurpg, line 5: Unused import

jvican:
Spoke with @olafurpg offline and we agreed on merging this so that we don't need to wait for another review cycle to merge this :smile: We'll remove the unused import whenever we set up scalafix automatically in the bloop build.

PR #1215, opened 2020-03-26
Report error when refreshProjectsCommand fails

lukaszwawrzyk:
If refresh project fails it looks like this:
fail

Not sure what is the ideal error reporting mechanism. Exception seems to work reasonably well. Most of endpoints in case of error still return success with empty result and just log the error.
Some return `JsonRpcResponse.invalidRequest`, but it doesn't seem to work, the request shows as successful. JsonRpcResponse.error requires RequestId which is not available.

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -23,7 +23,6 @@ import bloop.io.{AbsolutePath, RelativePath}
 import bloop.logging.{BspServerLogger, DebugFilter, Logger}
 import bloop.reporter.{BspProjectReporter, ProblemPerPhase, ReporterConfig, ReporterInputs}
 import bloop.testing.{BspLoggingEventHandler, TestInternals}
-
 import ch.epfl.scala.bsp
 import ch.epfl.scala.bsp.ScalaBuildTarget.encodeScalaBuildTarget
 import ch.epfl.scala.bsp.{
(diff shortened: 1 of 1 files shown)


frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -889,12 +889,21 @@ final class BloopBspServices(
...
         .map { command =>
-          Forker.run(cwd = state.build.origin, command, logger, state.commonOptions)
+          val cwd = state.build.origin
+          Forker.run(cwd = cwd, command, logger, state.commonOptions).map { exitCode =>
+            val status = Forker.exitStatus(exitCode)
+            if (!status.isOk) {
+              val message =
+                s"""Refresh projects command failed with exit code $exitCode.
+                   |Command run in $cwd: ${command.mkString(" ")}""".stripMargin
+              throw new RuntimeException(message)
```
jvican: I think you can avoid throwing the exception and return a `Left` with the message

lukaszwawrzyk: Right. It wasn't working before because I was actually ignoring this left from refresh and always proceeded with lodaing targets. Now it's ok.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -889,12 +888,20 @@ final class BloopBspServices(
...
-          Forker.run(cwd = state.build.origin, command, logger, state.commonOptions)
+          val cwd = state.build.origin
+          Forker.run(cwd = cwd, command, logger, state.commonOptions).map { exitCode =>
+            val status = Forker.exitStatus(exitCode)
+            if (!status.isOk) {
+              val message =
+                s"""Refresh projects command failed with exit code $exitCode.
+                   |Command run in $cwd: ${command.mkString(" ")}""".stripMargin
+              (state, Left(JsonRpcResponse.invalidRequest(message)))
+            } else (state, Right(()))
```
olafurpg: nit: I think the code reads nicer when both branches of an if/else use consistent braces
```suggestion
            } else {
              (state, Right(()))
            }
```

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -889,12 +888,22 @@ final class BloopBspServices(
...
         .map { command =>
-          Forker.run(cwd = state.build.origin, command, logger, state.commonOptions)
+          val cwd = state.build.origin
+          Forker.run(cwd = cwd, command, logger, state.commonOptions).map { exitCode =>
+            val status = Forker.exitStatus(exitCode)
+            if (!status.isOk) {
+              val message =
+                s"""Refresh projects command failed with exit code $exitCode.
+                   |Command run in $cwd: ${command.mkString(" ")}""".stripMargin
+              (state, Left(JsonRpcResponse.invalidRequest(message)))
```
jvican: Using `JsonRpcResponse.internalError` instead of `invalidRequest` is more descriptive of what happens :smile:

PR #1219, opened 2020-03-28
Export sbt meta builds via BSP

### LINKED ISSUE #1218: Show sbt meta projects through BSP server
It seems that it needs just to fix it by adding meta-build handling [here](https://github.com/scalacenter/bloop/blob/master/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala#L945)

Required for scalameta/metals-feature-requests#48

dos65:


tgodzik:
@dos65 What's the status for this change? Are we planning on merging it or are we going another way?

dos65:
@tgodzik 
I discussed it with @jvican.

He thinks that it would be better to add full sbt-meta-builds support into bloop. Here `full` means that in addition to scala build targets bloop will return sbt build targets associated with workspace by default in a single connection.

The current plan is that I'm going to implement this feature in bloop and I hope there will be some updates in the next week.

tgodzik:
> @tgodzik
> I discussed it with @jvican.
> 
> He thinks that it would be better to add full sbt-meta-builds support into bloop. Here `full` means that in addition to scala build targets bloop will return sbt build targets associated with workspace by default in a single connection.
> 
> The current plan is that I'm going to implement this feature in bloop and I hope there will be some updates in the next week.

Ok cool! That would solve the issue with the additional connection and I think there wouldn't be that many changes to the Metals PR. Let me know if you need any help, I will try to help as much as I can, since sbt support would be a really awesome new feature.

dos65:
I moved these changes to #1297

tgodzik:
The PR in Metals looks almost ready. @jvican what do you think about merging this change?

jvican:
Does the metals PR connect to the sbt metabuild as an independent build?

tgodzik:
> Does the metals PR connect to the sbt metabuild as an independent build?

Yes, @dos65 went with the multiple connection approach and from what I tested it worked pretty well.

jvican:
Great, I think that's the right approach to support sbt meta builds. Thank you all

PR #1209, opened 2020-03-22
Improve docs and landing page

jvican:
I have seen a lot of misunderstandings with Bloop, how it's installed,
how to set it up, how it works, etc that I've tried to address some of
these misunderstanding by improving the docs: the landing page, the
installation page, and better build tool docs.

The following changes are heavily inspired by the Bazel, Metals and
Babel website, so kudos to all the people that contributed there, they've
been a really important source of inspiration.

Fixes https://github.com/scalacenter/bloop/issues/1180
Fixes https://github.com/scalacenter/bloop/issues/1163

PR #1227, opened 2020-04-01
Fix build export installation step

### LINKED ISSUE #1220: Broken build tool sections in installation page
See this 7 second video. http://i.jsatk.us/pOLpng

Duhemm:
The recent refactoring of the documentation (see #1209) moved the
installation instructions for every individual build tool supported, so
that the directory structure was different than the one for package
managers.

Unfortunately, this broke the JS which displays the installation step
(the documentation for every build tool was no longer found).

This commit restores the previous directory structure, which is
unfortunately not as flat.

Fixes #1220

### DIFF
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -9,22 +9,13 @@
         "title": "Gradle",
         "sidebar_label": "Gradle"
       },
-      "build-tools/gradle/export": {
-        "title": "build-tools/gradle/export"
-      },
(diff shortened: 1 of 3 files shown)


Duhemm:
This seemed to me to be a lot easier than changing the JS... What do you think @jvican ?

Duhemm:
It turns out that fixing the JS directly wasn't much harder. Please take another look 😄

PR #1230, opened 2020-04-01
Report test projects without test frameworks

Duhemm:
Bloop used to report an error when running the `test` task against a
project that has an empty list of configured test frameworks. This
behavior is not convenient for targets that do not have any test
framework configured, and for which Bloop cannot infer what project (if
any) is the test variant.

This commit changes the test task, so that an error is reported if it is
run against a project that has the test or integration-test tags, and
that has no test frameworks configured. When run against a project that
is not a test project, and that has no test frameworks configured, the
test task will report a warning indicating that no test frameworks were
found.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
@@ -46,9 +46,8 @@ object TestTask {
       mode: RunMode
   ): Task[Int] = {
     import state.logger
-    val isTestProject = project.tags.contains(Tag.Test) || project.tags.contains(
-      Tag.IntegrationTest
-    )
+    val isTestProject = project.tags.contains(Tag.Test) ||
+      project.tags.contains(Tag.IntegrationTest)
     def handleEmptyTestFrameworks: Task[Int] = {
       if (isTestProject) {
         logger.error(s"Missing configured test frameworks in ${project.name}")
--- a/frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
@@ -1,7 +1,7 @@
 package bloop.engine.tasks
 
 import bloop.cli.ExitStatus
-import bloop.config.Config
+import bloop.config.{Config, Tag}
 import bloop.data.{Platform, Project}
 import bloop.engine.{Dag, Feedback, State}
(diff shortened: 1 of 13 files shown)


frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
```diff
@@ -43,12 +43,14 @@ object TestTask {
       rawTestOptions: List[String],
       testFilter: String => Boolean,
       handler: LoggingEventHandler,
-      failIfNoTestFrameworks: Boolean,
       mode: RunMode
   ): Task[Int] = {
     import state.logger
+    val isTestProject = project.tags.contains(Tag.Test) || project.tags.contains(
```
olafurpg, line 49: nit: this reads cleaner IMO
```suggestion
    val isTestProject = project.tags.contains(Tag.Test) ||
      project.tags.contains(Tag.IntegrationTest)

Comment (olafurpg) [reply frontend/src/main/scala/bloop/engine/tasks/TestTask.scala:49]: manual line breaks around infix operators are needed by scalafmt

File: frontend/src/test/scala/bloop/TestSpec.scala [hunk 2]
```diff
@@ -392,3 +392,25 @@ object JvmTestSpec extends BaseTestSpec("test-project-test", "cross-test-build-s
     }
   }
 }
+
+object NoTestFrameworksSpec extends ProjectBaseSuite("no-test-frameworks") {
```
olafurpg, line 396: Nice test suite! It was educational for me to see how to test interactions like these

Duhemm:
> Would it make sense to have bloop also fail if test frameworks are defined for a project with the test tag but no test is found? I was wondering if we could remove the "test frameworks defined" variable and just make it so that "bloop fails if a test or integration-test project defines no tests". WDYT?

I'm afraid this wouldn't play so nice with `--cascade` and `--include-dependencies` at the moment. We'd probably need to re-introduce the bailout mechanism that was in place before this PR. At this point, maybe it'd be better to add a flag?

jvican:
> I'm afraid this wouldn't play so nice with --cascade and --include-dependencies at the moment. 

If a project has the test/integration-test tag but defines no tests, shouldn't that be either a configuration error of the build exporter or an error that the user should see?

Duhemm:
Not necessarily. For instance, the sbt projects that we export at the moment (almost) always have a corresponding test project. Many of those projects have no tests though.

PR #1231, opened 2020-04-02
Add additional system properties for customizing zipkin traces

wiwa:
This change will allow users to set three properties:
1) `bloop.trace.localServiceName` with the default being `"bloop"` like before
2) `bloop.trace.traceStartAnnotation` if this is set, it will be annotated on every span right after the span is started.
3) `bloop.trace.traceEndAnnotation` if this is set, it will be annotated on every span right before the span is closed.

These additions will make it easier for consumers of zipkin trace data to label the spans as needed for existing data pipelines.

@olafurpg

### DIFF
--- a/backend/src/main/scala/bloop/tracing/BraveTracer.scala
+++ b/backend/src/main/scala/bloop/tracing/BraveTracer.scala
@@ -1,5 +1,7 @@
 package bloop.tracing
 
+import bloop.tracing.BraveTracer.traceEndAnnotation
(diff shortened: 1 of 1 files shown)


olafurpg:
The CI failures look unrelated. 

```
##[error]fatal: reference is not a tree: 2d38b15dabad625414f75a0c8668b4e56fe2d0a7
```

I tried restarting the jobs but the same error happened again. @wiwa can you amending the commit and force-pushing?

PR #1237, opened 2020-04-07
Remove work-in-progress in relation to Metals

ckipp01:
A while back we made an explicit choice to remove the "work in progress" that we had labeled Metals as across the project. Would it be possible to remove that from here as well to match the documentation across metals and avoid the assumption that Metals is still experimental?

https://github.com/scalameta/metals/blob/f60048577e5da4e442af95f397273f7cac32ce37/website/blog/2019-06-28-thorium.md#no-more-work-in-progress


ckipp01:
Closing this. Apologies 😆 I'll finish my read through and submit one pr rather than a bunch of patches.

PR #1239, opened 2020-04-08
Implement jvmRunEnvironment endpoint

tpasternak:
This endpoint is a mirror to jvmRunEnvironment. In current bloop
implementation they return exactly the same data. Accpording to this,
I extracted the code to to common functions, both in tests and
BloopBspServices

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -699,31 +699,6 @@ final class BloopBspServices(
   def jvmEnvironment(
       targets: Seq[BuildTargetIdentifier]
   ): BspEndpointResponse[List[bsp.JvmEnvironmentItem]] = {
-    def jvmEnvironment(
-        state: State,
-        projects: Seq[(BuildTargetIdentifier, Project)]
-    ): Seq[JvmEnvironmentItem] = {
-      for {
-        (id, project) <- projects.toList
-        dag = state.build.getDagFor(project)
-        fullClasspath = project.fullClasspath(dag, state.client).map(_.toBspUri.toString)
-        environmentVariables = state.commonOptions.env.toMap
-        workingDirectory = project.workingDirectory.toString
-        javaOptions <- project.platform match {
-          case Platform.Jvm(config, _, _) => Some(config.javaOptions.toList)
-          case _ => None
-        }
-      } yield {
-        bsp.JvmEnvironmentItem(
-          id,
-          fullClasspath.toList,
-          javaOptions,
-          workingDirectory,
-          environmentVariables
-        )
-      }
-    }
-
     ifInitialized(None) { (state: State, logger: BspServerLogger) =>
       mapToProjects(targets, state) match {
         case Left(error) =>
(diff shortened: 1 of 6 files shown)


tpasternak:
Please let mi know if I missed something. If I interpreted the `run` and `test` services' code well, it seems that now they both use the same environment.

I think at some point I have to refactor the `test` and `jvmTestEnvironment` services to ensure the `test` is using exactly the same environment as `jvmTestEnvironment` is returning. Same applies to `run` and `jvmRunEnvironment`. But this seems to be a quite complex change.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -114,6 +108,7 @@ final class BloopBspServices(
     .requestAsync(endpoints.BuildTarget.dependencySources)(p => schedule(dependencySources(p)))
     .requestAsync(endpoints.DebugSession.start)(p => schedule(startDebugSession(p)))
     .requestAsync(endpoints.BuildTarget.jvmTestEnvironment)(p => schedule(jvmTestEnvironment(p)))
+    .requestAsync(endpoints.BuildTarget.jvmRunEnvironment)(p => schedule(jvmRunEnvironment(p)))
```
olafurpg, line 111: Would it be possible reuse the same method here?
```suggestion
     .requestAsync(endpoints.BuildTarget.jvmTestEnvironment)(p => schedule(jvmEnvironment(p)))
    .requestAsync(endpoints.BuildTarget.jvmRunEnvironment)(p => schedule(jvmEnvironment(p)))
```

tpasternak: They have different sets of input/output types: JvmTestEnvironmentParams, JvmTestEnvironmentResult etc.

olafurpg: I see. Still looks like there is room to reuse more code between `jvmRunEnvironment` and `jvmTestEnvironment`. Something like
```scala
def jvmEnvironment[T](targets: Seq[BuildTargetIdentifier])(fn: Seq[JvmEnvironmentItem] => T): T = ???
```

olafurpg: It's a bit unsatisfying to see the duplicate structure for `ifInitialized(None)`/error handling/logging/...

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -684,40 +686,53 @@ final class BloopBspServices(
...
   def jvmTestEnvironment(
       params: bsp.JvmTestEnvironmentParams
-  ): BspEndpointResponse[bsp.JvmTestEnvironmentResult] = {
+  ): BspEndpointResponse[bsp.JvmTestEnvironmentResult] =
+    jvmEnvironment(params.targets).map(_.map(new bsp.JvmTestEnvironmentResult(_)))
+
+  def jvmEnvironment(
+      targets: Seq[BuildTargetIdentifier]
+  ): BspEndpointResponse[List[bsp.JvmEnvironmentItem]] = {
+    def jvmEnvironment(
```
olafurpg, line 702: nit: should this method be inlined like it was implemented before? It's a bit difficult to read a nested method with the same name as the enclosing method

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
```diff
@@ -98,50 +97,84 @@ class BspProtocolSpec(
     }
   }
 
-  test("check the correct contents of jvm test environment") {
-    TestUtil.withinWorkspace { workspace =>
-      object Sources {
-        val `A.scala` =
-          """/A.scala
-            |object A
+  private def testEnivronmentFetching(
```
Duhemm, line 100: Typo: environment

tpasternak: Done

tpasternak:
I didn't squash changes in order to ease issues resolution tracking. I can squash and force-push the changes if there are no more isssues

Duhemm:
Thanks @tpasternak ! I've squashed the commits and merged the result.

tpasternak:
Thanks @jvican . If you see see any issues related to this PR - please let me know, I can open another one to fix them.

BTW I thought about doing some rework in order to make it clear that the same values are used in both `jvmRunEnvironment` and `run` endpoints, but it seems to be more complex thing.

PR #1241, opened 2020-04-11
Change zipkin URL in docs

henricook:
When reading the docs i noticed this link no longer worked


henricook:
I don't think I have access to retry these jobs, but I doubt this doc change is the cause of a breakage? Can someone help out with CI?

PR #1243, opened 2020-04-13
Make sbt docs clearer for multiple projects on initial setup

onesupercoder:
It took me a while to figure this out and I'm hoping it might save other people some time.

### DIFF
--- a/docs/build-tools/sbt.md
+++ b/docs/build-tools/sbt.md
@@ -27,6 +27,8 @@ After that, start up sbt or reload your sbt shell to load the plugin.
 
 The sbt command `bloopInstall` exports your sbt build to bloop.
 
+If you have multiple projects, you should run bloopInstall in the projects that aren't dependant on other projects first. For example:  acme-core-server references acme-shared-lib.  So in the acme-shared-lib project run sbt bloopInstall first and then in the acme-core-server run sbt bloopInstall second.
+
 In bloop, an sbt project is represented as a pair of `(sbt project, sbt configuration)` and it's
 written to a configuration directory. The default location of this directory in your workspace is
 `.bloop/` (you may want to add `.bloop/` to your `.gitignore` file).
--- a/docs/build-tools/sbt.md
+++ b/docs/build-tools/sbt.md
@@ -27,13 +27,13 @@ After that, start up sbt or reload your sbt shell to load the plugin.
 
 The sbt command `bloopInstall` exports your sbt build to bloop.
 
-If you have multiple projects, you should run `bloopInstall` in the projects that aren't dependent on other projects first. For example:  acme-core-server references acme-shared-lib.  So in the acme-shared-lib project run `sbt bloopInstall` first and then in the acme-core-server run sbt bloopInstall second.
-
 In bloop, an sbt project is represented as a pair of `(sbt project, sbt configuration)` and it's
 written to a configuration directory. The default location of this directory in your workspace is
 `.bloop/` (you may want to add `.bloop/` to your `.gitignore` file).
 
-For example, a build with a single project `foo` generates two configuration files by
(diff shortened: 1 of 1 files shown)


docs/build-tools/sbt.md
```diff
@@ -27,6 +27,8 @@ After that, start up sbt or reload your sbt shell to load the plugin.
 
 The sbt command `bloopInstall` exports your sbt build to bloop.
 
+If you have multiple projects, you should run `bloopInstall` in the projects that aren't dependent on other projects first. For example:  acme-core-server references acme-shared-lib.  So in the acme-shared-lib project run `sbt bloopInstall` first and then in the acme-core-server run sbt bloopInstall second.
```
jvican, line 30: Could we move this paragraph after the following paragraph, where we explain how an sbt project is represented?

docs/build-tools/sbt.md
```diff
@@ -31,7 +31,9 @@ In bloop, an sbt project is represented as a pair of `(sbt project, sbt configur
 written to a configuration directory. The default location of this directory in your workspace is
 `.bloop/` (you may want to add `.bloop/` to your `.gitignore` file).
 
-For example, a build with a single project `foo` generates two configuration files by
+If you have multiple projects, you should run `bloopInstall` in the projects that aren't dependent on other projects first. For example:  acme-core-server references acme-shared-lib.  So in the acme-shared-lib project run `sbt bloopInstall` first and then in the acme-core-server run sbt bloopInstall second.
```
jvican, line 34: Actually I've just read this and this clarification seems out of place in the flow of the section. Additionally, I do not understand what you mean here. Do you mean that you should run `sbt bloopInstall` for every build that a developer has? `sbt bloopInstall` exports all of the build target within a build. If a developer defines independent sbt builds, they naturally need to be exported manually, but that doesn't require it to do it in order necessarily.
  
Also, keep in mind that defining several sbt build targets should be done with care and is somewhat an anti-pattern borrowed from the gradle and maven world.

onesupercoder: @jvican I may be configuring our sbt projects wrong.  I think its multi project, but is it multi build targets?

```scala
import sbt.project

lazy val acmeSharedLibVersion = "4.1.10.8"
lazy val `acme-shared-lib-dir` = file("..") / "acme-shared-lib"

lazy val `acme-shared-lib` = RootProject(`acme-shared-lib-dir`)

lazy val `acme-core-server` = (project in file(".")).configure { project =>
  if (sys.env.contains("BUILD_NUMBER") || !`acme-shared-lib-dir`.exists()) {
    project
      .settings(
        libraryDependencies ++= Seq(
          "com.dealermade" %% "acme-shared-lib" % acmeSharedLibVersion,
          "com.dealermade" %% "acme-shared-lib" % acmeSharedLibVersion % Test classifier "tests"
        )
      )
  } else {
    project
      .settings(migrationsDirectory := Some((resourceDirectory in (`acme-shared-lib`, Compile)).value))
      .dependsOn(`acme-shared-lib` % "compile->compile;test->test")
  }
}
```

jvican: A build is a collection of build targets, and a build target is the unit that defines a collection of sources with some meaning. A build target in sbt is created per (project, configuration), the sbt-bloop docs go into more detail about that.
  
Thanks for sharing your build. `RootProject`s dependencies are not supported in bloop. The best way you can set up source dependencies in sbt is via `ProjectRef` dependencies because it specifies the project dependencies more explicitly and `sbt-bloop` is able to identify them and extract them.
  
Keep in mind though, you're getting into uncharted territory. Source dependencies are not well supported in sbt, beware when using them, and many would consider what you're doing in your build hacky.

jvican: When you use `ProjectRef`s in your build, make sure you enable source dependencies support in sbt-bloop via https://scalacenter.github.io/bloop/docs/build-tools/sbt#enable-sbt-project-references

onesupercoder: Thank you for the information and feedback!  😊

jvican:
Alright, closing this ticket. If you want to follow up with other doc improvements, feel free to submit another PR @onesupercoder

PR #1245, opened 2020-04-15
Fixed providing jvm options in DAP

alekseiAlefirov:
Issue: `jvmOptions` in `ScalaMainClass` have been ignored while running a debug session.

Note: not really sure, if the testing is done elegant enough, or can even be put into already existing `test`.

### DIFF
--- a/frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
+++ b/frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
@@ -46,7 +46,7 @@ private final class MainClassDebugAdapter(
       env,
       workingDir,
       mainClass.`class`,
-      mainClass.arguments.toArray,
+      (mainClass.arguments ++ mainClass.jvmOptions).toArray,
       skipJargs = false,
       RunMode.Debug
     )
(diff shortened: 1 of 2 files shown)


frontend/src/test/scala/bloop/dap/DebugServerSpec.scala
```diff
@@ -129,6 +129,56 @@ object DebugServerSpec extends DebugBspBaseSuite {
     }
   }
 
+  test("accepts argumetns and jvm options") {
```
jvican, line 132:
```suggestion
  test("accepts arguments and jvm options") {
```

frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
```diff
@@ -46,7 +46,7 @@ private final class MainClassDebugAdapter(
       env,
       workingDir,
       mainClass.`class`,
-      mainClass.arguments.toArray,
+      (mainClass.arguments ++ mainClass.jvmOptions).toArray,
```
jvican, line 49: Hmmm, interesting. I thought we were passing the JVM options. Did you double-check that the test passes without this fix?

alekseiAlefirov: yep, it fails without it.

Issue #1200, opened 2020-03-16
Add `coursier install bloop` as the recommended installation


jvican:
The next bloop release will be mainly installed via `coursier`, instead of forcing clients to either run the generic `curl` script or install bloop via a dependency manager. The brew, Scoop and Linux distros installation formulas/recipes will be changed to use `coursier install` to install bloop. The only reason why somebody would install bloop this way instead of via a manual `coursier install` would be to install also bash/zsh autocompletions.
  
Right now I'm investigating what changes we need, so all of this is still being considered and there's no final decision. I've also made a comment in https://github.com/coursier/apps/pull/8 to ask if `coursier install` can deal with the fact that we don't have native binaries for Windows.

jvican:
To close this ticket and the 1.4.0 release we need the following:
  
1. Native binaries need to be created in CI and published to the release.
2. Our dependency on `install.py` needs to go away.
    - We will still keep the installer for a while but we will no longer use it in our official installation methods.
    - We will mark the "universal" installation method on the website as deprecated.
    - The launcher implementation needs to change to not use the `install.py` anymore.
3. Our formulas will install completions and the bloop CLI. The bloop server will be launched automatically based on the version of the CLI.
4. Formulas will no longer install systemd and brew services. We will keep them around in case people feel really strongly that they want to use them, but they will no longer be installed automatically and recommended.

jvican:
Made good progress porting the brew formula to install our CLI directly and changing the launcher to use coursier as the only way of starting the bloop server. I will follow up with more progress updates here soon.

jvican:
Brew installation is now out of the way, I'm focusing on porting the archlinux installation. Latest master already has code to publish GraalVM binaries for macOS and Linux as well, so we'll reuse that to attach them to the created GitHub release automatically.

agilesteel:
The button which expands the universal installation section on the webpage now says `coursier` however the installation information still describes the python based installation.

agilesteel:
Never mind. It's fixed now.

PR #1246, opened 2020-04-15
Use `coursier install bloop` in installation methods

jvican:
And remove installation code that relies on the server being installed
by our installation script, as well.

Fixes https://github.com/scalacenter/bloop/issues/1200

Needs still changes in the docs, will come soon.


jvican:
The brew formula recipe looks quite straight-forward. After implementing this, I'm quite confident it could be auto-generated by coursier itself, as well as those of other package managers. This option could be supported via `coursier install --generate-formula brew ..$ARGS`. This command would need to support additional resources such as command autocompletion but it would greatly simplify the process of installing a jar in many environments proper. Just some ideas based on my experience /cc @alexarchambault

alexarchambault:
I'm probably going to add support for easier ways to pass the application descriptor at least… Like maybe accepting it inline, like
```bash
$ cs install 'bloop:{"repositories": […], …}'
```
so that it doesn't have to be written in a distinct directory, and `--default-channel` and `--channel` don't need to be specified.

I'll have a closer look at the brew / scoop / … stuff that you're generating.

jvican:
@alexarchambault That's a good idea. Adding the option to disable compiling to native in certain platforms if a pre-built version is not found would make sense, as we discussed. In Windows 10, you need to install special drivers in order to be able to build GraalVM binaries.

PR #1252, opened 2020-04-16
Make `bloop about` and `bloop help` always succeed

### LINKED ISSUE #1249: Update build logic to ignore loading builds from `$HOME/.bloop`
When running from `$HOME`

`coursier launch ch.epfl.scala:bloopgun_2.12:latest.release -- about`

the server correctly starts with the config values from `$HOME/.bloop/bloop.json` but it also gives this error:

`[E] Failed to load project from $HOME/.bloop/bloop.json`

This happens because you run `bloop about` in your `$HOME`,  but the change that introduced `$HOME/.bloop/bloop.json` has the shape of a bloop build so bloop tries to load `bloop.json` from there.

olafurpg:
Previously, running `bloop help` or `bloop about` could fail when executed in a workspace that contained invalid `.bloop/*.json` config files. This can for example happen in the `$HOME` directory with `.bloop/bloop.json` global settings.  This commit refactors the `bloop help` and `bloop about` commands so that they print out a message and don't attempt to load the workspace JSON files.

Fixes #1249

### DIFF
--- a/frontend/src/main/scala/bloop/engine/caches/StateCache.scala
+++ b/frontend/src/main/scala/bloop/engine/caches/StateCache.scala
@@ -9,6 +9,7 @@ import bloop.cli.CommonOptions
 import bloop.engine.{Build, BuildLoader, State, ClientPool}
 import bloop.io.AbsolutePath
 import bloop.cli.ExitStatus
+import bloop.util.Environment
 
 import monix.eval.Task
 
@@ -91,6 +92,7 @@ final class StateCache(cache: ConcurrentHashMap[AbsolutePath, StateCache.CachedS
       clientSettings: Option[WorkspaceSettings]
   ): Task[State] = {
     val empty = Build(from, Nil)
+    if (from == Environment.defaultBloopDirectory) return Task.now(createState(empty))
     val previousState = getStateFor(from, client, pool, commonOptions, logger)
     val build = previousState.map(_.build).getOrElse(empty)
 
--- a/frontend/src/main/scala/bloop/Cli.scala
+++ b/frontend/src/main/scala/bloop/Cli.scala
@@ -124,6 +124,29 @@ object Cli {
      """.stripMargin
   }
 
+  private def aboutAsked: String = {
+    val bloopName = bloop.internal.build.BuildInfo.bloopName
+    val bloopVersion = bloop.internal.build.BuildInfo.version
+    val scalaVersion = bloop.internal.build.BuildInfo.scalaVersion
+    val zincVersion = bloop.internal.build.BuildInfo.zincVersion
+    val developers = bloop.internal.build.BuildInfo.developers.mkString(", ")
+    val javaVersion = JavaRuntime.version
+    val javaHome = JavaRuntime.home
+    val jdiStatus = {
+      if (JavaRuntime.loadJavaDebugInterface.isSuccess)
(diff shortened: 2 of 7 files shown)


olafurpg:
Would it be OK to make `StateCache.loadState()` skip loading `$HOME` in addition to the changes in this PR? I have hit on cases where `bloop about` and `bloop help` has failed in non-`$HOME` workspaces and the current diff in this PR also fixes those issues without adding a special-case for `$HOME`.

Currently, if you run `bloop compile` in `$HOME` the error message is not too bad IMO

```sh
❯ bloop compile foo
[E] Failed to load project from /Users/lgeirsson/.bloop/bloop.json
```

Would it make sense to customize the error message into something like this instead?

```sh
❯ bloop compile foo
[E] Failed to load project from /Users/lgeirsson/.bloop/bloop.json (did you intend to run Bloop in another directory? Your current working directory is $HOME and this file contains global Bloop configuration, which is not valid build configuration)
```

jvican:
Yes, that makes sense, we can keep these changes as they are and then skip loading the build from `$HOME` in `StateCache.loadState` so that it's never considered a valid build.
  
> Would it make sense to customize the error message into something like this instead?
  
When you run `bloop compile foo` in a directory that has no `.bloop` today, you get the following error:

```
[E] Missing configuration directory in /Users/jvicentecantero/Code/.bloop.
[E]
[E]   1. Did you run bloop outside of the working directory of your build?
[E]      If so, change your current directory or point directly to your `.bloop` directory via `--config-dir`.
[E]
[E]   2. Did you forget to generate configuration files for your build?
[E]      Check the installation instructions https://scalacenter.github.io/bloop/setup
[E]
```
  
My expectation would be that we output the same error in the case of `$HOME` just as what would happen in old versions.

I think maybe we can improve the initial part of the message to make the error clearer. For example, replacing "Missing configuration directory in" by "Missing valid Bloop build in".

olafurpg:
Sounds good. I updated the error message for `bloop compile` in `$HOME` to be like this now
```
❯ bloop compile a
[E] Missing valid Bloop build in /Users/lgeirsson/.bloop.
[E]
[E]   1. Did you run bloop outside of the working directory of your build?
[E]      If so, change your current directory or point directly to your `.bloop` directory via `--config-dir`.
[E]
[E]   2. Did you forget to generate configuration files for your build?
[E]      Check the installation instructions https://scalacenter.github.io/bloop/setup
[E]
```

frontend/src/main/scala/bloop/engine/caches/StateCache.scala
```diff
@@ -91,6 +92,7 @@ final class StateCache(cache: ConcurrentHashMap[AbsolutePath, StateCache.CachedS
       clientSettings: Option[WorkspaceSettings]
   ): Task[State] = {
     val empty = Build(from, Nil)
+    if (from == Environment.defaultBloopDirectory) return Task.now(createState(empty))
```
olafurpg, line 95: I can refactor to avoid `return` here but this was the smallest diff I could come up with. I personally use `return` sometimes in cases like this when it's undesirable to indent the rest of the method body

jvican: I sometimes use this, but rarely so. I think it's fine in this case :+1:

PR #1253, opened 2020-04-17
Make sure we use all the settings that already exist in the project

tgodzik:
Previously, we would replace the scalac options when Metals was the client.
Now, we make sure the base settings are used.


frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -364,7 +364,7 @@ object Project {
       val ysemanticdb = if (!options.contains("-Ysemanticdb")) List("-Ysemanticdb") else Nil
       val sourceRoot =
         if (!options.contains("-sourceroot")) List("-sourceroot", workspaceDir.toString()) else Nil
-      ysemanticdb ++ sourceRoot
+      options ++ ysemanticdb ++ sourceRoot
```
tgodzik, line 367: This is the main reason we've been having issue and a 100% my fault :sweat_smile:

PR #1254, opened 2020-04-17
Register Scala generated class files before compiling Java

Duhemm:
Classfiles originating from Scala recompiled sources would be marked as
invalidated until after the Analysis object is created. This caused
issues with the `BloopForkedJavaCompiler`, which creates invalid
classfiles, since these invalid classfiles would be chosen over the new
ones, thus aborting compilation.

This commit changes the analysis callbacks used by Bloop so that the
generated classfiles are pushed to the classfile manager as soon as they
are known. Thanks to this change, no invalid classfiles will be created
for the recompiled sources.

### DIFF
--- a/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
+++ b/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
@@ -354,7 +354,14 @@ final class BloopAnalysisCallback(
     }
   }
 
-  override def apiPhaseCompleted(): Unit = ()
+  override def apiPhaseCompleted(): Unit = {
+    /*
+     * Inform the class file manager of the generated Scala classes as soon as
+     * the Zinc API phase has run and collected them so that the class file
+     * invalidation registers these files before compiling Java files incrementally.
+     */
+    manager.generated(classToSource.keysIterator.toArray)
+  }
   override def dependencyPhaseCompleted(): Unit = ()
   override def classesInOutputJar(): java.util.Set[String] = ju.Collections.emptySet()
 
--- a/frontend/src/test/scala/bloop/BaseCompileSpec.scala
+++ b/frontend/src/test/scala/bloop/BaseCompileSpec.scala
@@ -1543,6 +1543,78 @@ abstract class BaseCompileSpec extends bloop.testing.BaseSuite {
     }
   }
 
+  test("recompile entire mixed project") {
+    TestUtil.withinWorkspace { workspace =>
+      object Sources {
+        val `Dummy1.scala` =
+          """/foo/Dummy1.scala
+            |class Dummy1""".stripMargin
+        val `Dummy2.scala` =
+          """/foo/Dummy2.scala
+            |class Dummy2""".stripMargin
+        val `Dummy3.scala` =
+          """/foo/Dummy3.scala
+            |class Dummy3""".stripMargin
+        val `Dummy4.scala` =
+          """/foo/Dummy4.scala
+            |class Dummy4""".stripMargin
+        val `Dummy1-changed.scala` =
+          """/foo/Dummy1.scala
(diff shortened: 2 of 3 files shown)


frontend/src/test/scala/bloop/BaseCompileSpec.scala
```diff
@@ -1543,6 +1543,78 @@ abstract class BaseCompileSpec extends bloop.testing.BaseSuite {
     }
   }
 
+  test("Recompile entire mixed project") {
```
jvican, line 1546: Great test! Can we use remove the capital case from the test though? We use lowercase in the whole codebase.

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
```diff
@@ -354,7 +354,14 @@ final class BloopAnalysisCallback(
     }
   }
 
-  override def apiPhaseCompleted(): Unit = ()
+  override def apiPhaseCompleted(): Unit = {
+    // We collect all the generated classes during `API`. When this phase completes, we can pass to
+    // the classfiles manager all the classes that were generated by scalac. This will mark them as
+    // invalidated, so that javac (if it runs after) can be notified.
+    // NOTE: When the compilation order is `JavaThenScala`, scalac will be notified about the
```
jvican, line 361: Hmm, I don't think this is true. `CallbackGlobal` does not call `generated` and none of the generation logic depends on the compiler order. What did you mean?

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
```diff
@@ -354,7 +354,14 @@ final class BloopAnalysisCallback(
...
+    // invalidated classfiles in `CallbackGlobal`.
+    manager.generated(classToSource.keysIterator.toArray)
```
jvican, line 363: So, together with this spot, it means that we will call `generated` three times in the case where we use java compilation, given that we have not removed the call to `generated` in `IncrementalCommon`.
  
I think this is not ideal but it's fine. When I get the time, I will move stuff around to guarantee that "`generated` is called by the incremental compiler as soon as class files are known from either of the local java compiler, the forked java compiler or the scala compiler" without repeated calls.

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
```diff
@@ -354,7 +354,14 @@ final class BloopAnalysisCallback(
     }
   }
 
-  override def apiPhaseCompleted(): Unit = ()
+  override def apiPhaseCompleted(): Unit = {
+    // We collect all the generated classes during `API`. When this phase completes, we can pass to
```
jvican, line 358: What about we replace all of this comment by this:
  
```
  /*
   * Inform the class file manager of the generated Scala classes as soon as
   * the Zinc API phase has run and collected them so that the class file 
   * invalidation registers these files before compiling Java files incrementally.
   */
```

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
Duhemm: I meant that when we start by compiling the Java sources, we'll need to tell scalac which classfiles have been invalidated by this compilation step. This information is passed to scalac from `CallbackGlobal`.

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopAnalysisCallback.scala
Duhemm: Sounds better :+1:

PR #1258, opened 2020-04-21
Support links in path to javac

Duhemm:
We were using a forked javac process to compile Java sources in cases
where the `java.home` properties and the JVM platform of a project were
different, but pointed to the same file. Instead of comparing the two
paths to decide whether to fork, this commit compares the two files that
are pointed to by the paths.

### DIFF
--- a/backend/src/main/scala/bloop/CompilerCache.scala
+++ b/backend/src/main/scala/bloop/CompilerCache.scala
@@ -63,7 +63,7 @@ final class CompilerCache(
       getScalaCompiler(_, componentProvider)
     )
 
-    val javaCompiler = javaCompilerCache.computeIfAbsent(javacBin, getJavaCompiler(_))
+    val javaCompiler = javaCompilerCache.computeIfAbsent(javacBin, getJavaCompiler(logger, _))
 
     val javaDoc = Javadoc.local.getOrElse(Javadoc.fork())
     val javaTools = JavaTools(javaCompiler, javaDoc)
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/bloop/io/AbsolutePath.scala
```diff
@@ -21,6 +21,7 @@ final class AbsolutePath private (val underlying: Path) extends AnyVal {
   def exists: Boolean = Files.exists(underlying)
   def isFile: Boolean = Files.isRegularFile(underlying)
   def isDirectory: Boolean = Files.isDirectory(underlying)
+  def isSameFile(other: AbsolutePath): Boolean = Files.isSameFile(underlying, other.underlying)
```
jvican, line 24: What happens if this throws? Before we were only comparing strings so we didn’t need to worry about this. It would be handy if we report this error explicitly to the user with a runtime exception

Duhemm: I added `isSameFile` in `CompilerCache`. WDYT?

backend/src/main/scala/bloop/CompilerCache.scala
```diff
@@ -327,4 +327,17 @@ final class CompilerCache(
       }
     }
   }
+
+  /**
+   * Equivalent to `path1.isSameFile(path2)`, but will log exceptions and return `false`.
+   */
+  private def isSameFile(logger: Logger, path1: AbsolutePath, path2: AbsolutePath): Boolean = {
```
jvican, line 334: What about changing the name of this method and make it more specific to java path? We can then provide a custom error in case java home doesn’t exist

jvican: @Duhemm Feel free to merge when you address this!

Duhemm: Done!

PR #1261, opened 2020-04-22
Force source generation for cached 'sbt bloopGenerate'

### LINKED ISSUE #1257: Force source generation for cached `sbt bloopGenerate`
Running `bloopGenerate` task in sbt should trigger the source generation even if the sbt settings have not changed since the last source generation.

adpi2:
Closes #1257 

I think the result is really nice, because:
- I run `~bloopInstall` in the background
- Any time I change a source file, `bloopInstall` is not triggered (those files are not watched by the source generation nor by `bloopInstall`)
- Any time I change a template file,  `bloopInstall` is triggered, and the generated sources are updated, the bloop config is not updated because it is already cached
- If I change the sbt settings, Metals propose me to import the build, and it overwrites the bloop config
- Most of the time I dont even need to reload my sbt shell in the background

### DIFF
--- a/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
+++ b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
@@ -877,8 +877,11 @@ object BloopDefaults {
     else if (!hasConfigSettings) inlinedTask[Option[File]](None)
     else if (generated.isDefined && generated.get.fromSbtUniverseId == currentSbtUniverse) {
       Def.task {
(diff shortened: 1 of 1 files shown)


Duhemm:
The goal of this change is to run the source generators in the refresh / fast track branch, where we run `bloopInstall` a second time. `managedSources` is already called in the normal branch here: https://github.com/adpi2/bloop/blob/218406d5ba218d57aac029a92060e233df5279f9/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala#L1014-L1015

PR #1262, opened 2020-04-22
Add workpace settings for Zipkin traces

wiwa:
Previously, it was only possible to customize the Zipkin integration globally via system properties.
This PR will make it possible to configure Zipkin integration properties per-workspace via `bloop.settings.json`. The motivation for this change is to make it possible to track compile/test metrics for Bloop only when working in a specific codebase.

### DIFF
--- a/backend/src/main/scala/bloop/tracing/BraveTracer.scala
+++ b/backend/src/main/scala/bloop/tracing/BraveTracer.scala
@@ -1,36 +1,33 @@
 package bloop.tracing
 
-import bloop.tracing.BraveTracer.traceEndAnnotation
-import bloop.tracing.BraveTracer.traceStartAnnotation
-import brave.{Span, Tracer}
 import brave.propagation.SamplingFlags
 import brave.propagation.TraceContext
 import brave.propagation.TraceContextOrSamplingFlags
+import brave.Span
+import brave.Tracer
 import monix.eval.Task
 import monix.execution.misc.NonFatal
-import scala.util.Failure
 import scala.util.Properties
-import scala.util.Success
-import java.util.concurrent.TimeUnit
-import zipkin2.codec.SpanBytesEncoder.{JSON_V1, JSON_V2}
+import zipkin2.codec.SpanBytesEncoder.JSON_V1
+import zipkin2.codec.SpanBytesEncoder.JSON_V2
 
 final class BraveTracer private (
     tracer: Tracer,
     val currentSpan: Span,
-    closeCurrentSpan: () => Unit
+    closeCurrentSpan: () => Unit,
+    traceProperties: TraceProperties
 ) {
   def startNewChildTracer(name: String, tags: (String, String)*): BraveTracer = {
-    import brave.propagation.TraceContext
     val span = tags.foldLeft(tracer.newChild(currentSpan.context).name(name)) {
       case (span, (tagKey, tagValue)) => span.tag(tagKey, tagValue)
     }
 
     span.start()
-    traceStartAnnotation.foreach(span.annotate)
+    TraceProperties.TraceStartAnnotation.getOrGlobal(traceProperties).foreach(span.annotate)
     new BraveTracer(tracer, span, () => {
-      traceEndAnnotation.foreach(span.annotate)
+      TraceProperties.TraceEndAnnotation.getOrGlobal(traceProperties).foreach(span.annotate)
       span.finish()
-    })
+    }, traceProperties)
   }
 
   def trace[T](name: String, tags: (String, String)*)(
@@ -60,7 +57,7 @@ final class BraveTracer private (
   private def traceInternal[T](name: String, verbose: Boolean, tags: (String, String)*)(
       thunk: BraveTracer => T
   ): T = {
-    if (!verbose || BraveTracer.isVerboseTrace) {
+    if (!verbose || TraceProperties.Verbose.getOrGlobal(traceProperties)) {
       val newTracer = startNewChildTracer(name, tags: _*)
       try thunk(newTracer) // Don't catch and report errors in spans
       catch {
@@ -81,7 +78,7 @@ final class BraveTracer private (
   private def traceTaskInternal[T](name: String, verbose: Boolean, tags: (String, String)*)(
       thunk: BraveTracer => Task[T]
   ): Task[T] = {
-    if (!verbose || BraveTracer.isVerboseTrace) {
+    if (!verbose || TraceProperties.Verbose.getOrGlobal(traceProperties)) {
       val newTracer = startNewChildTracer(name, tags: _*)
       thunk(newTracer)
         .doOnCancel(Task(newTracer.terminate()))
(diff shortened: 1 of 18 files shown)


wiwa:
Seems like I forgot to add one of my changes. I would like feedback on my approach of adding `WorkspaceSettings` to `Build`, but I also want to include some tests in `BuildLoaderSpec` regarding different `TraceProperties` in different workspaces. Is there another test suite where I might want to test my changes?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -390,6 +377,7 @@ final class BloopBspServices(
       originId: Option[String],
       logger: BspServerLogger
   ): BspResult[bsp.CompileResult] = {
+    val workspaceSettings = WorkspaceSettings.readFromFile(state.build.origin, logger)
```
Duhemm, line 380: I think this is unused.

frontend/src/main/scala/bloop/data/WorkspaceSettings.scala
```diff
@@ -38,7 +36,8 @@ import com.github.plokhotnyuk.jsoniter_scala.macros.{JsonCodecMaker, CodecMakerC
 case class WorkspaceSettings(
```
Duhemm, line 36: Reading the comment above, I'm starting to think that maybe this is not the best place to put the `traceProperties` 🤔 

I agree both are some kind of workspace settings, but the trace properties will typically be user defined, whereas the semantic DB settings are apparently a concern of the build server only.

I'd like to see some sort of workspace settings in Bloop. These settings would contain the tracing properties and other settings such as aggregates (so that we can support them without encoding them as dependencies), default reporter, etc.

Unfortunately it looks like this is not the place. Should we introduce a different class holding these settings? I'm afraid of the unintended consequences of exposing this config to the users.

I'd suggest renaming this to `InternalServerSettings` (better name more than welcome), and introducing a new `WorkspaceSettings` to hold these user defined settings, which would be read from a different file. Hopefully renaming the current `bloop.settings.json` to something else shouldn't be a big deal since it's only an implementation detail?

What do you think @jvican?

backend/src/main/scala/bloop/tracing/TraceProperties.scala
```diff
@@ -0,0 +1,79 @@
+package bloop.tracing
+
+import scala.util.Properties
+
+case class TraceProperties(
+    zipkinServerUrl: Option[String],
+    debug: Option[Boolean],
+    verbose: Option[Boolean],
+    localServiceName: Option[String],
```
jvican, line 9: Should we call this `serviceName`? Not sure if it makes sense to distinguish between "local" service names and just normal service names.

backend/src/main/scala/bloop/tracing/TraceProperties.scala
```diff
@@ -0,0 +1,79 @@
+package bloop.tracing
+
+import scala.util.Properties
+
+case class TraceProperties(
+    zipkinServerUrl: Option[String],
+    debug: Option[Boolean],
```
jvican, line 7: What is the difference between `verbose` and `debug`?

backend/src/main/scala/bloop/tracing/BraveTracer.scala
```diff
@@ -60,7 +57,7 @@ final class BraveTracer private (
   private def traceInternal[T](name: String, verbose: Boolean, tags: (String, String)*)(
       thunk: BraveTracer => T
   ): T = {
-    if (!verbose || BraveTracer.isVerboseTrace) {
+    if (!verbose || TraceProperties.Verbose.getOrGlobal(traceProperties)) {
```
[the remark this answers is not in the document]
jvican: I think the commit I made in the linked branch already addresses this

frontend/src/test/scala/bloop/BuildLoaderSpec.scala
```diff
@@ -247,6 +246,33 @@ object BuildLoaderSpec extends BaseSuite {
     }
   }
 
+  test("build respects different trace properties in separate workspaces") {
```
jvican, line 249: This is a good test to make sure we don't make any mistake when loading settings from the workspace settings file. Re your previous question, I don't think it's necessary to add more tests than this, I'm happy with what we have

wiwa:
I agree with your changes to the PR, what do you think about having a separate `UserManagedSettings` to house `refreshProjectsCommand` and `traceSettings`? It might make it clearer.

jvican:
If we can do it in a way that doesn’t affect the underlying json format, it’s fine by me 👍

frontend/src/main/scala/bloop/data/WorkspaceSettings.scala
```diff
@@ -38,7 +36,8 @@ import com.github.plokhotnyuk.jsoniter_scala.macros.{JsonCodecMaker, CodecMakerC
...
     semanticDBVersion: Option[String],
     supportedScalaVersions: Option[List[String]],
-    refreshProjectsCommand: Option[List[String]]
+    refreshProjectsCommand: Option[List[String]],
+    traceProperties: TraceProperties
```
olafurpg, line 40: This field needs to be optional so that it doesn't have to appear in `bloop.settings.json`

frontend/src/main/scala/bloop/data/WorkspaceSettings.scala
olafurpg: I don't think we need to separate user/bloop managed settings in this PR. I agree it might be a nice change as a follow PR, but this PR is using the same consistent pattern as the `refreshProjectCommands` setting, which is only user-managed.

jvican:
@wiwa Any update on this? :smile: Would be great to merge it in master soon

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
Duhemm: Can you please remove this line?

backend/src/main/scala/bloop/tracing/TraceProperties.scala
```diff
@@ -0,0 +1,38 @@
+package bloop.tracing
+
+import scala.util.Properties
+
+case class TraceProperties(
+    serverUrl: String,
+    // Note: `debug` corresponds to a sampling flag sent to Zipkin server,
+    // while `verbose` determines if we send certain spans marked as verbose spans
+    debug: Boolean,
```
jvican, line 9: Should we call it `enableZipkinDebug` to try to make the distinction clearer for users that don't have access to the source?

frontend/src/main/scala/bloop/bsp/BloopBspDefinitions.scala
```diff
@@ -1,11 +1,10 @@
 package bloop.bsp
```
jvican, line 2: Thanks for removing the wildcard imports, I know it's a pain to get the previous, explicit imports right :+1:

frontend/src/main/scala/bloop/data/WorkspaceSettings.scala
jvican: Summarizing the discussion in this PR: after the proposed changes, it should be fine to have both user/bloop managed settings to be hosted in the same `bloop.settings.json` file. We will only reload the build if the bloop-managed settings change value

integrations/gradle-bloop/src/test/scala/bloop/integrations/gradle/ConfigGenerationSuite.scala
```diff
@@ -1586,7 +1585,8 @@ abstract class ConfigGenerationSuite {
     assert(Files.exists(configDir.toPath), "Does not exist: " + configDir)
     val configDirectory = AbsolutePath(configDir)
     val loadedProjects = BuildLoader.loadSynchronously(configDirectory, logger)
-    val build = Build(configDirectory, loadedProjects)
+    val workspaceSettings = WorkspaceSettings.readFromFile(configDirectory, logger)
+    val build = Build(configDirectory, loadedProjects, workspaceSettings)
     State.forTests(build, TestUtil.getCompilerCache(logger), logger)
```
jvican, line 1590: You might need to rebase/push merge commit @wiwa because the zinc submodule diff below should not show up

backend/src/main/scala/bloop/tracing/TraceProperties.scala
wiwa: I like that, let's leave that to a separate PR perhaps?

wiwa:
The Windows CI failures look unrelated?
```
X bloop.bsp.TcpBspCompileSpec.create orphan client classes directory and make sure loading a BSP session cleans it up 3642ms 
```

jvican:
Yup, they are unrelated. I've just changed system properties to start with `bloop.tracing.*` instead of `bloop.trace.*` and I've also renamed the debug field so that nobody can use the latest version, I wanted to fix that in this PR because it touches the JSON representation. As soon as CI passes, I'll be merging! Thanks @wiwa

PR #1266, opened 2020-04-30
Don't consider CLI classes directory an orphan directory

jvican:
This commit fixes a bug in the logic deleting the orphan bsp classes
directories that was not updated when the CLI client classes dir was
moved to behave like just a normal bsp client classes dir.

This bug could have produced important problems such as
`ClassNotFoundException`s when starting a new connection while a bloop
CLI command is running.

I don't think this issue is directly related with the metals connection
issues but it might.

### DIFF
--- a/backend/src/main/scala/bloop/io/ParallelOps.scala
+++ b/backend/src/main/scala/bloop/io/ParallelOps.scala
@@ -74,7 +74,7 @@ object ParallelOps {
       scheduler: Scheduler,
       logger: Logger,
       enableCancellation: Boolean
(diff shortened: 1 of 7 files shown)


tgodzik:
> I don't think this issue is directly related with the metals connection
> issues but it might.

Here's hoping! Thanks for the awesome investigation.

We can get it into the next release and let people report back if it helped. In the meantime, I will proactively look into any more exceptions or other problems.

jvican:
@tgodzik Great, let's merge this so that you can give it a try asap

Issue #1264, opened 2020-04-24
Support environment variables parameters for debug session to start


alekseiAlefirov:
It would be useful to have an option to start a debug session (via `debugSession/start` request method) with environment variables configured.

As a context:
recently [Metals](https://github.com/scalameta) have supported [user configuration for starting a debug session](https://github.com/scalameta/metals-vscode/issues/217), and now are being asked, if enviranment variables can be supported as well.
Metals are using `bloop` to instantiate a debug adapter.

PR #1268, opened 2020-05-04
Start debug session with user-provided environment variables

alekseiAlefirov:
Issue: https://github.com/scalacenter/bloop/issues/1264

PR to BSP: https://github.com/build-server-protocol/build-server-protocol/pull/118

### DIFF
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -571,6 +571,7 @@ object Interpreter {
                   mainClass,
                   cmd.args.toArray,
                   cmd.skipJargs,
+                  envVars = Nil,
                   RunMode.Normal
                 )
             }
--- a/frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
+++ b/frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
@@ -50,6 +50,7 @@ trait JvmProcessForker {
       mainClass: String,
       args0: Array[String],
       skipJargs: Boolean,
+      envVars: List[String],
       logger: Logger,
       opts: CommonOptions,
       extraClasspath: Array[AbsolutePath] = Array.empty
@@ -58,7 +59,7 @@ trait JvmProcessForker {
       if (skipJargs) (Array.empty[String], args0)
       else args0.partition(_.startsWith("-J"))
 
-    runMain(cwd, mainClass, userArgs, userJvmOptions, logger, opts, extraClasspath)
+    runMain(cwd, mainClass, userArgs, userJvmOptions, envVars, logger, opts, extraClasspath)
   }
 
   def runMain(
@@ -66,6 +67,7 @@ trait JvmProcessForker {
       mainClass: String,
       args: Array[String],
       jargs: Array[String],
+      envVars: List[String],
       logger: Logger,
       opts: CommonOptions,
       extraClasspath: Array[AbsolutePath]
@@ -110,6 +112,7 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
       mainClass: String,
       args: Array[String],
       jargs: Array[String],
+      envVars: List[String],
       logger: Logger,
       opts: CommonOptions,
       extraClasspath: Array[AbsolutePath]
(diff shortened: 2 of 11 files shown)


frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -571,6 +571,7 @@ object Interpreter {
                   mainClass,
                   cmd.args.toArray,
                   cmd.skipJargs,
+                  envVars = Nil,
```
tgodzik, line 574: There is no way currently to add env variables inside the CLI and I am not sure about how to add it with a proper user experience. We could have additiona `-E` as in case  of properties.
`-E-TEST=Hello`

But I am not sure. @jvican What do you think?

jvican: Environment variables are propagated from the terminal where the bloop CLI is used thanks to Nailgun, so there's no need to pass them explicitly through a flag. Users can just export an environment variable and access it from a project run with the CLI.

tgodzik: Ach, you are right. This works under bash, but not in fish. Most likely not related then.

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
```diff
@@ -118,6 +121,11 @@ final class JvmForker(config: JdkConfig, classpath: Array[AbsolutePath]) extends
     val jvmOptions = jargs.map(_.stripPrefix("-J")) ++ config.javaOptions
     val fullClasspath = classpath ++ extraClasspath
     val fullClasspathStr = fullClasspath.map(_.syntax).mkString(File.pathSeparator)
+    val runOpts = if (envVars.nonEmpty) opts.copy(env = CommonOptions.currentEnv) else opts
```
tgodzik, line 124: So turns out if we don't copy the env variable will also be set for the next run, which popped up as a failure in an unrelated test.

tgodzik:
Changes are minimal here, is anyone against merging it?

@jvican @Duhemm  ?

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
jvican: This comment looks a bit weird, why is this caused? Environment variables are set on a session based basis, so different clients should have different environment variables.

Also, what variables are picked up in a BSP session?

tgodzik: I can't remember why this solved the issue, but the problem was that setting environment variable to opts cause those variable to be available on the next run too. I will dig in to see why this fixed the tests.

tgodzik: So `opts.env` was actually shared between tests (not sure where it's defined for them) and if I added an environment variable in one test, it was available in another.

`opts.copy(env = CommonOptions.currentEnv) ` helped, although this might not be the full solution. We should most likely have something like `opts.copy(env = CommonOptions.newEnv)`

All this could be an issue only in tests, but is JVMProcessForker invoked in a separate JVM for each session? If not then we have to copy the current environment instead o relying on lazy val `CommonOptions.currentEnv` What do you think @jvican  ?

giuscri:
Maintainers, please note this unlocks the ability to seriously debug services in VSCode :)

frontend/src/main/scala/bloop/exec/JvmProcessForker.scala
jvican: Thanks for the explanation and sorry for my delay in getting back to you @tgodzik `JvmProcessForker` is invoked in the same class loader so we should not be copying the environment for it, it will pick up whatever the environment is at a given moment. I would try to fix the issue with environment variables be shared in tests and remove this change. What do you think?

tgodzik: Good idea! I was actually thinking about the same, but didn't manage to get back to it yet.  I will try to do it as soon as possible.

tgodzik:
@jvican Fixed the tests and removed the unnecessary change, should be good to merge.

jvican (approved):
I think this approach with the environment variables should work, LGTM :+1:

PR #1269, opened 2020-05-06
Support globs in file watching

Duhemm:
File watching was not supported for projects that define their source
list as globs, because the `SourceWatcher` was directly collecting the
sources from the eponymous field in the configuration files.

The source watcher now respects globs when collecting sources and
directories to watch.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -110,20 +110,29 @@ object Interpreter {
       f: State => Task[State]
   ): Task[State] = {
     val reachable = Dag.dfs(getProjectsDag(projects, state))
-    val allSources = reachable.iterator.flatMap(_.sources.toList).map(_.underlying)
-    val watcher = SourceWatcher(projects.map(_.name), allSources.toList, state.logger)
-    val fg = (state: State) => {
-      val newState = State.stateCache.getUpdatedStateFrom(state).getOrElse(state)
-      f(newState).map { state =>
-        watcher.notifyWatch()
-        State.stateCache.updateBuild(state)
-      }
-    }
+    val projectsSourcesAndDirs = reachable.map(_.allSourceFilesAndDirectories)
+    val groupTasks =
+      projectsSourcesAndDirs.grouped(8).map(group => Task.gatherUnordered(group)).toList
+    Task
+      .sequence(groupTasks)
+      .map(fp => fp.flatten.flatten.map(_.underlying))
+      .flatMap { allSources =>
+        val watcher = SourceWatcher(projects.map(_.name), allSources, state.logger)
+        val fg = (state: State) => {
+          val newState = State.stateCache.getUpdatedStateFrom(state).getOrElse(state)
+          f(newState).map { state =>
+            watcher.notifyWatch()
+            State.stateCache.updateBuild(state)
+          }
+        }
 
-    if (!bloop.util.CrossPlatform.isWindows)
-      state.logger.info("\u001b[H\u001b[2J")
-    // Force the first execution before relying on the file watching task
-    fg(state).flatMap(newState => watcher.watch(newState, fg))
+        if (!bloop.util.CrossPlatform.isWindows)
+          state.logger.info("\u001b[H\u001b[2J")
+
+        // Force the first execution before relying on the file watching task
+        fg(state).flatMap(newState => watcher.watch(newState, fg))
+
+      }
   }
 
   private def runCompile(
--- a/frontend/src/test/scala/bloop/FileWatchingSpec.scala
+++ b/frontend/src/test/scala/bloop/FileWatchingSpec.scala
@@ -18,6 +18,7 @@ import java.util.concurrent.TimeUnit
 
 import scala.concurrent.duration.FiniteDuration
 import monix.execution.misc.NonFatal
+import bloop.data.SourcesGlobs
 
 object FileWatchingSpec extends BaseSuite {
   System.setProperty("file-watcher-batch-window-ms", "100")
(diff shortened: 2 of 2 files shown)


frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -110,20 +110,27 @@ object Interpreter {
...
-    val allSources = reachable.iterator.flatMap(_.sources.toList).map(_.underlying)
-    val watcher = SourceWatcher(projects.map(_.name), allSources.toList, state.logger)
-    val fg = (state: State) => {
-      val newState = State.stateCache.getUpdatedStateFrom(state).getOrElse(state)
-      f(newState).map { state =>
-        watcher.notifyWatch()
-        State.stateCache.updateBuild(state)
-      }
-    }
+    val projectsSourcesAndDirs = reachable.map(_.allSourceFilesAndDirectories)
```
jvican, line 113: Is it intentional that we're only resolving sources once, before file watching?

frontend/src/test/scala/bloop/FileWatchingSpec.scala
```diff
@@ -188,6 +189,78 @@ object FileWatchingSpec extends BaseSuite {
...
+        waitUntilWatchIteration(logsObservable, totalIterations, HasIterationStoppedMsg, duration)
+
+      def testValidLatestState: TestState = {
+        val state = compiledState.getLatestSavedStateGlobally()
+        assert(state.status == ExitStatus.Ok)
+        assertValidCompilationState(state, projects)
+        state
+      }
+
+      TestUtil.await(FiniteDuration(20, TimeUnit.SECONDS), ExecutionContext.ioScheduler) {
```
jvican, line 249: 20 seconds, jeez 😅 this is a slow test!

frontend/src/test/scala/bloop/FileWatchingSpec.scala
```diff
@@ -188,6 +189,78 @@ object FileWatchingSpec extends BaseSuite {
     }
   }
 
+  flakyTest("support globs", 3) {
```
jvican, line 192: Great :+1:

frontend/src/test/scala/bloop/FileWatchingSpec.scala
Duhemm: I copied everything over from the other, similar test in this file. On my machine it takes ~11s to complete 😞

frontend/src/main/scala/bloop/engine/Interpreter.scala
Duhemm: Yes. New files created after the file watcher has started will still be picked up and watched.

jvican: Are you sure about that? What we're doing here is resolving sources at the beginning of source watching, but the watch iterations don't force a source resolution?

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -110,20 +110,27 @@ object Interpreter {
...
+    Task
+      .gatherUnordered(projectsSourcesAndDirs)
```
jvican, line 115: Can we use `Task.gatherN` (assuming it's availble in our Monix version)? I'd like to avoid submitting to the IO pool a lot of tasks that cause the pool to allocate a lot of threads at once

frontend/src/main/scala/bloop/engine/Interpreter.scala
Duhemm: That's correct. There's a single call (per project) to `allSourceFilesAndDirectories`, and new files are picked up and watched.

jvican: I'm not sure we're talking about the same thing. If I file watch, but then I create a new file that would pick up by a regex, that would not be caught by this implementation. Is that right?

frontend/src/main/scala/bloop/engine/Interpreter.scala
Duhemm: We use monix 2.3.3, where it's not available.

frontend/src/main/scala/bloop/engine/Interpreter.scala
Duhemm: No, the new file will be watched. It'd be interesting to verify that's true on all platforms actually. It's probably watched only because you can only watch directories on OSX, or some detail like this 🤔

jvican: As discussed offline, I think there are some cases where the file changes will not be picked up (directory hosting the regex not existing before the source resolution and with regexes that are more than two level deep) but I think handling those cases will become more complicated, so I'm fine merging as is, just wanted to point out some limitations with this approach. :+1:

frontend/src/main/scala/bloop/engine/Interpreter.scala
jvican: Can we use something like this?

```
      val groupTasks = projectSourcesAndDirs.grouped(8).map(group => Task.gatherUnordered(group)).toList
      Task.sequence(groupTasks).map(fp => fp.flatten)
              ...
```

It's the normal pattern we use in the codebase today, in absences of `Task.gatherN`

The 8 number is totally arbitrary but given that this is IO it's fine?

frontend/src/main/scala/bloop/engine/Interpreter.scala
olafurpg: I think it should be fine to recursively watch `SourcesGlobs.directory` instead of expanding globs and watching the individual files. This would correctly catch new files in nested directories. There's a risk we would watch too large directories, but that's already the case with normal source directories.

Issue #1276, opened 2020-05-11
Unexpected error when compiling root after upgrading to 1.4.0


danking:
I killed the bloop process listed by `ps aux | grep bloop`. I removed the `.bloop` folder. I ran `sbt bloopInstall` on a project that otherwise works with `sbt` and then I asked bloop to compile the project and got this error message.

```
[E] Unexpected error when compiling root: 'Unexpected exception when computing compile inputs for root'
[E] Failed to compile 'root'
```

How can I get more detailed information? I don't know what `--debug` does but I couldn't get it to work. `--verbose` doesn't seem to help. The project itself is rather larger and constructing a small reproducible example will be difficult. I'd prefer to learn how to tell bloop to show me what the issue is.

```
# rm -rf .bloop      
# sbt bloopInstall
[info] Loading settings for project global-plugins from plugins.sbt ...
[info] Loading global plugins from /Users/dking/.sbt/1.0/plugins
[info] Loading settings for project hail-build from plugins.sbt,metals.sbt ...
[info] Loading project definition from /Users/dking/projects/hail/hail/project
[info] Loading settings for project root from build.sbt ...
[info] Set current project to hail (in build file:/Users/dking/projects/hail/hail/)
[success] Generated .bloop/root.json
[success] Generated .bloop/root-test.json
[success] Total time: 1 s, completed May 11, 2020 6:44:00 PM
# bloop compile --verbose root 
No server running at 127.0.0.1:8212, let's fire one...
Resolving ch.epfl.scala:bloop-frontend_2.12:1.4.0...
Starting bloop server at 127.0.0.1:8212...
Attempting a connection to the server...
Attempting a connection to the server...
Attempting a connection to the server...
Attempting a connection to the server...
[D] Loading 2 projects from '/Users/dking/projects/hail/hail/.bloop'...
[D] Loading project from '/Users/dking/projects/hail/hail/.bloop/root.json'
[D] Loading project from '/Users/dking/projects/hail/hail/.bloop/root-test.json'
[D] Cache miss for scala instance org.scala-lang:scala-compiler:2.11.12.
[D]   => /Users/dking/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/jline/jline/2.14.3/jline-2.14.3.jar
[...]

PR #1277, opened 2020-05-12
Check if coursier upgrade is responsible of 1.4.0 failure

jvican:


launcher/src/main/scala/bloop/launcher/BloopServerVersion.scala
```diff
@@ -12,13 +12,13 @@ case class BloopServerVersion(
 object BloopServerVersion {
   def apply(serverVersion: String, out: PrintStream): Option[BloopServerVersion] = {
     Version(serverVersion).items.toList match {
-      case (major: Version.Number) :: (minor: Version.Number) :: rest =>
+      case Version.Number(major) :: Version.Number(minor) :: rest =>
```
jvican, line 15: @alexarchambault This is a revert of the changes I did for bloop to upgrade to the latest coursier. Was this a good upgrade to the new version API?

jvican:
@tgodzik and I have confirmed that this causes the issue some people have experienced in 1.4.0 https://github.com/scalacenter/bloop/issues/1276 I will cut a release tonight

launcher/src/main/scala/bloop/launcher/BloopServerVersion.scala
alexarchambault: What problem are you running into with these changes?

jvican: We don't know yet, we have bisected the commit history and gotten to these changes. I think the upgrade is somehow affecting the automatic SemanticDB resolution.

ivanopagano:
Can you confirm if this fix will allow the user to get back bash-completions?

Was that eventually included in the 1.4.0 cut?

jvican:
Bash completions are part of the installer but you need to uninstall and install back again to get them to work because there was a bug in the home brew recipe.

olafurpg:
I was experimenting with bloop-launcher 1.4.0 and hit on the following error
```
[error] Exception in thread "bsp-client-to-server" Exception in thread "main" java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
[error] 	at bloop.launcher.bsp.BspBridge.copyContents(BspBridge.scala:290)
[error] 	at bloop.launcher.bsp.BspBridge.forwardStreamContents(BspBridge.scala:275)
[error] 	at bloop.launcher.bsp.BspBridge.$anonfun$wireBspConnectionStreams$1(BspBridge.scala:197)
[error] 	at bloop.bloopgun.core.Shell$$anon$2.run(Shell.scala:155)
```
Downgrading to v1.4.0-RC1-235-3231567a fixes the problem. I wonder if this could be the root cause of the 1.4.0 errors?

tgodzik:
> I was experimenting with bloop-launcher 1.4.0 and hit on the following error
> 
> ```
> [error] Exception in thread "bsp-client-to-server" Exception in thread "main" java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
> [error] 	at bloop.launcher.bsp.BspBridge.copyContents(BspBridge.scala:290)
> [error] 	at bloop.launcher.bsp.BspBridge.forwardStreamContents(BspBridge.scala:275)
> [error] 	at bloop.launcher.bsp.BspBridge.$anonfun$wireBspConnectionStreams$1(BspBridge.scala:197)
> [error] 	at bloop.bloopgun.core.Shell$$anon$2.run(Shell.scala:155)
> ```
> 
> Downgrading to v1.4.0-RC1-235-3231567a fixes the problem. I wonder if this could be the root cause of the 1.4.0 errors?

That's weird, Ihaven't seen that issue and `flip` should be avlaiable both in JDK 8 and 11

tgodzik:
So I did some digging and it seems that 1.4.0 works fine on JDK 11, but it doesn't on JDK 8. I can reproduce 100% with JDK 8.

jvican:
This attempt at downgrading bloop is a red herring, real fix in https://github.com/scalacenter/bloop/pull/1278

PR #1284, opened 2020-05-17
Fix ArchLinux package for v1.4+

TheElectronWill:
It turns out that the "single binary" generated by `coursier install` is a launcher script `bloop` + the graalvm-compiled binary `.bloop.aux`, that's why we need a directory for the Bloop installation.

Question :upside_down_face:
Some users found the systemd service useful, and it doesn't cost much to add it back (even if its use is questionable). Should I add the service file to `ReleaseUtils#installationArtifacts()`? (with some logic so that it only ends up in the arch package)

### DIFF
--- a/project/ReleaseUtils.scala
+++ b/project/ReleaseUtils.scala
@@ -309,14 +309,23 @@ object ReleaseUtils {
        |build() {
        |  mkdir channel
        |  mv "$coursierChannelName" "channel/bloop.json"
-       |  coursier install --install-dir "$$srcdir" --default-channels=false --channel channel bloop
+       |  coursier install --install-dir "$$srcdir" --default-channels=false --channel channel --only-prebuilt=true bloop
        |}
        |
        |package() {
        |  cd "$$srcdir"
        |
-       |  ## binaries
-       |  install -Dm755 bloop "$$pkgdir"/usr/bin/bloop
+       |  # patch the bloop launcher so that it works when symlinked from /usr/bin
+       |  sed 's|$$(dirname "$$0")|/usr/lib/bloop|' -i bloop
+       |
+       |  # install to /usr/lib/bloop
+       |  # NOTE: bloop is just a launcher, the actual program is .bloop.aux
+       |  install -Dm755 bloop "$$pkgdir"/usr/lib/bloop/bloop
+       |  install -Dm755 .bloop.aux "$$pkgdir"/usr/lib/bloop/.bloop.aux
+       |
+       |  # add link to /usr/bin
+       |  mkdir -p "$$pkgdir"/usr/bin
+       |  ln -s /usr/lib/bloop/bloop "$$pkgdir"/usr/bin/bloop
        |
        |  # shell completion
        |  install -Dm644 $bashResourceName "$$pkgdir"/etc/bash_completion.d/bloop


jvican:
Thanks for doing this @TheElectronWill I'm surprised `coursier install` generates a bloop.aux 🤔 Will review the PR in a bit.

> Some users found the systemd service useful, and it doesn't cost much to add it back
  
The cost is that we lose uniformity of how the bloop server is supposed to work. The version of the server should be set by the client that spawns bloop, not by the version installed in the machine. The systemd service goes against this new guarantee of bloop so I discourage its use.
  
As mentioned in the AUR website, there are good uses for systemd services. For example, controlling how many resources are used in one's machine. But these are use cases of power users. Whomever wants a systemd service, they can install it from the repo, but I don't want to add it to the recipe -- it should be opt-in.

project/ReleaseUtils.scala
```diff
@@ -309,14 +309,23 @@ object ReleaseUtils {
...
        |package() {
        |  cd "$$srcdir"
        |
-       |  ## binaries
-       |  install -Dm755 bloop "$$pkgdir"/usr/bin/bloop
+       |  # patch the bloop launcher so that it works when symlinked from /usr/bin
+       |  sed 's|$$(dirname "$$0")|/usr/lib/bloop|' -i bloop
+       |
+       |  # install to /usr/lib/bloop
+       |  # NOTE: bloop is just a launcher, the actual program is .bloop.aux
```
jvican, line 322: @alexarchambault Are you aware of coursier install doing this in Linux machines? This is not how `coursier install` works in Windows and macOS 🤔

TheElectronWill: For information, the "launcher" looks like this:
```
#!/usr/bin/env sh
export CS_NATIVE_LAUNCHER="true"
export IS_CS_INSTALLED_LAUNCHER="true"
exec "$(cd "$(dirname "$0")"; pwd)/.bloop.aux" "$@"
PK...
```
(then some binary data)

kurnevsky:
> The version of the server should be set by the client that spawns bloop, not by the version installed in the machine. The systemd service goes against this new guarantee of bloop so I discourage its use.

Could you please elaborate on this? How different clients (metals, sbt, cli, etc) are supposed to select bloop version? Should they spawn different versions in the same time? That'd use a LOT of memory. Or should they negotiate version somehow?

marcin-rzeznicki:
> Whomever wants a systemd service, they can install it from the repo, but I don't want to add it to the recipe -- it should be opt-in.

You're of course aware that the act of installing a systemd unit file doesn't mean that it is automatically run anywhere, right? It's  always been opt-in.

project/ReleaseUtils.scala
alexarchambault: > This is not how `coursier install` works in Windows and macOS

It should work this way there too. It's mostly there so that the script here can also act as a zip file, containing metadata about the installed app (how to update it, etc.), just like for JVM launchers.

marcin-rzeznicki:
@kurnevsky Nope, they just try to connect blindly to whatever happens to be running. If it's running then it's fine - if not then they'll try to install it via coursier and run (which implies that it's not that 'clients select version' - it's rather: the first client gets to select a version :smiley: ). You can potentially run different versions via different ports. So, yeah - running a service is fine.

TheElectronWill:
> As mentioned in the AUR website, there are good uses for systemd services. For example, controlling how many resources are used in one's machine. But these are use cases of power users.

I think it’s a bit weird to consider that Arch users aren’t _power users_ enough to handle a service they have to manually activate 🤔 If you insist, the service could be in a separate package to achive opt-in yet easy installation.

jvican:
@TheElectronWill the AUR installer does not install the bloop server, it only starts the CLI. 

Because of that, it doesn’t make sense and I don’t want the majority of bloop users on Linux to rely on an installed service that restarts the server all the time and doesn’t let clients specify the version that should be started in the background. The mere existence of this systemd service can break the integration of some clients, which rely on their ability to kill the server and start it with a new version.

If the Arch community wants to create a recipe for the systemd service, I’m fine with it — but it won’t be advertised as official. Everyone in Arch is a power user but they don’t necessarily need to install a systemd service if it’s not for resource management.

project/ReleaseUtils.scala
jvican: Thanks for the clarification @alexarchambault

TheElectronWill:
> the AUR installer does not install the bloop server, it only installs the CLI.
> [...] The mere existence of this systemd service can break the integration of some clients, which rely on their ability to kill the server and start it with a new version.

Oh, right! I didn't know that.
In that case I'll let my branch as it is :slightly_smiling_face:  Thanks for the explanations.

PR #1285, opened 2020-05-18
Fix local brew auto-completions installation

eanderson13:
The local file names didn't match the remote file names, so the `mv` commands at https://github.com/scalacenter/bloop/blob/master/project/ReleaseUtils.scala#L176-L186 couldn't find the files.


jvican:
@eanderson13 Thanks for the contribution. `install.py` is deprecated, that's why it was not working. Why do you need to use `install.py` to install bloop?

eanderson13:
@jvican This fix is for Homebrew local installation, as I was discussing on Gitter. I only updated `install.py` to keep the file references correct. But I can back out that part of the change if you want.

PR #1297, opened 2020-05-31
Detect sbt meta builds as a part of workspace

dos65:


jvican:
Thanks for contributing this change @dos65 :smile:  
  
FTR @dos65 and I discussed several approaches offline to improve the approach to adding sbt metaproject support in bloop proper. Those changes will be added to this PR.

dos65:
@jvican Updated, but it's a little bit different from what we discussed.

I noticed that `State` has `ResultCache` field and sharing it between different build might lead to problems because of caching.

So, instead of extending it, I made a different structure - `BspState`. It keeps a state for a usual build and states for meta-builds and it's used only for BSP.

tgodzik:
@jvican any updates on what we plan to do with the PR? It would be awesome to finally support sbt in Metals.

jvican:
As I've discussed in private with @dos65, I think this PR adds a lot of complexity to both the bsp implementation and the build loader, and I'm concerned with the performance implications of it. So I don't see a way we can merge it as is. I've discussed with Vadim that I believe the best way to support sbt is either to implement Metals support for several builds (which it's something that needs to happen sooner or later) or wait for BSP support to land in sbt.

jvican:
Closing because this is no longer needed after Metals supports multi BSP connections and can establish a connection with the meta build on its own. Thanks for your work @dos65 I appreciate the effort you went into to prototype this 🙏

PR #1299, opened 2020-06-01
Port bloop-config to Scala.js

MaximeKjaer:
This PR adds cross-compilation of the `bloop-config` module to Scala.js.

`jsoniter-scala` has support for Scala.js since [v2.2.2](https://github.com/plokhotnyuk/jsoniter-scala/releases/tag/v2.2.2), and for Scala.js 1.1.0 since [v2.2.4](https://github.com/plokhotnyuk/jsoniter-scala/releases/tag/v2.2.4) so this PR updates the dependency from 2.0.4 to 2.2.4 for all modules.

Note that the tests require Node.js 12.10.0 ([released 2019-09-04](https://nodejs.org/en/download/releases/)) or later, because they use [`fs.rmdirSync`](https://nodejs.org/api/fs.html#fs_fs_rmdirsync_path_options) with the `recursive` option.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -88,37 +88,64 @@ val publishJsonModuleSettings = List(
   publishLocal := publishLocal.dependsOn(publishM2).value
 )
 
-lazy val jsonConfig210 = project
+val testJSSettings = List(
+  testOptions += Tests.Argument(TestFrameworks.JUnit, "-v", "-a"),
+  scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.CommonJSModule)),
+  libraryDependencies ++= List(
+    "com.github.plokhotnyuk.jsoniter-scala" %%% "jsoniter-scala-core" % Dependencies.jsoniterVersion,
+    "com.github.plokhotnyuk.jsoniter-scala" %%% "jsoniter-scala-macros" % Dependencies.jsoniterVersion % Provided
+  )
+)
+
+val testResourceSettings = {
+  // FIXME: Shared resource directory is ignored, see https://github.com/portable-scala/sbt-crossproject/issues/74
+  Seq(Test).flatMap(inConfig(_) {
+    unmanagedResourceDirectories ++= {
+      unmanagedSourceDirectories.value
+        .map(src => (src / ".." / "resources").getCanonicalFile)
+        .filterNot(unmanagedResourceDirectories.value.contains)
+        .distinct
+    }
+  })
+}
+
+lazy val jsonConfig210 = crossProject(JVMPlatform)
+  .crossType(CrossType.Pure)
   .in(file("config"))
   .disablePlugins(ScriptedPlugin)
-  .settings(testSettings)
   .settings(publishJsonModuleSettings)
   .settings(
     name := "bloop-config",
-    target := (file("config") / "target" / "json-config-2.10").getAbsoluteFile,
     scalaVersion := Scala210Version,
-    libraryDependencies ++= {
-      List(
-        Dependencies.circeParser,
-        Dependencies.circeCore,
-        Dependencies.circeGeneric,
-        compilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full),
-        Dependencies.scalacheck % Test
-      )
-    }
+    libraryDependencies +=
+      compilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full),
+    testResourceSettings
+  )
+  .jvmSettings(
+    testSettings,
+    target := (file("config") / "target" / "json-config-2.10" / "jvm").getAbsoluteFile,
+    libraryDependencies ++= Seq(
+      Dependencies.circeParser,
+      Dependencies.circeCore,
+      Dependencies.circeGeneric
+    )
   )
 
-lazy val jsonConfig211 = project
+lazy val jsonConfig211 = crossProject(JSPlatform, JVMPlatform)
+  .crossType(CrossType.Pure)
   .in(file("config"))
   .disablePlugins(ScriptedPlugin)
-  .settings(testSettings)
   .settings(publishJsonModuleSettings)
   .settings(
     name := "bloop-config",
-    target := (file("config") / "target" / "json-config-2.11").getAbsoluteFile,
     scalaVersion := Scala211Version,
     unmanagedSourceDirectories in Compile +=
-      Keys.baseDirectory.value./("src")./("main")./("scala-2.11-13"),
+      Keys.baseDirectory.value / ".." / "src" / "main" / "scala-2.11-13",
+    testResourceSettings
+  )
+  .jvmSettings(
+    testSettings,
+    target := (file("config") / "target" / "json-config-2.11" / "jvm").getAbsoluteFile,
     libraryDependencies ++= {
       List(
         Dependencies.jsoniterCore,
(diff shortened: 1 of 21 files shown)


MaximeKjaer:
The tests failed because I had forgotten to update the build configuration for 2.10 and 2.11. Scala.js 1.x does not support 2.10, so the 2.10 build will continue to be JVM-only; I've added cross-compilation to Scala.js for 2.11 in a643f47708effe21d4159ee6c04d7ffdcd833120.

> Note that the tests require Node.js 12.10.0

I see that the CI runs on 10.20.1, but this shouldn't be a problem. The only thing that breaks when Node is <12.10.0 is that some test artifacts aren't deleted correctly, so this doesn't really matter on the CI.

MaximeKjaer:
It seems like the tests consistently pass on the JVM, but fail on GraalVM... but for different reasons on each platform. On Ubuntu it encounters a fatal error in the VM, on Windows it runs out of memory, and on MacOS a macro complains :man_shrugging:

config/.jvm/src/main/scala/bloop/config/TargetPlatform.scala
```diff
@@ -0,0 +1,28 @@
+package bloop.config
+
+import java.nio.file.{Files, Paths}
+
+object TargetPlatform {
```
jvican, line 5: What do you think of calling this `PlatformFiles` instead? Looks clearer at call-sites

config/src/test/scala/bloop/config/ConfigCodecsSpec.scala
```diff
@@ -1,18 +1,27 @@
...
-import org.junit.{Assert, Test}
-import java.nio.file.Paths
-import java.nio.file.Files
+import org.junit.{AfterClass, Assert, Test}
 import java.nio.charset.StandardCharsets
-import java.io.BufferedReader
-import java.io.InputStreamReader
-import scala.io.Source
+
+object ConfigCodecsSpec {
```
jvican, line 7: When you define an object with the same name as a class, does the execution of the object take precedence over the class one? Or do both run?

build.sbt
```diff
@@ -88,37 +88,63 @@ val publishJsonModuleSettings = List(
...
+val testJSSettings = List(
+  testOptions += Tests.Argument(TestFrameworks.JUnit, "-v", "-a"),
+  scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.CommonJSModule)),
+  libraryDependencies ++= List(
+    "com.github.plokhotnyuk.jsoniter-scala" %%% "jsoniter-scala-core" % Dependencies.jsoniterVersion,
+    "com.github.plokhotnyuk.jsoniter-scala" %%% "jsoniter-scala-macros" % Dependencies.jsoniterVersion % Provided
+  )
+)
+
+val testResourceSettings =
```
jvican, line 100: Nit: can we enclose this whole expression with curly braces?

config/.jvm/src/main/scala/bloop/config/TargetPlatform.scala
MaximeKjaer: I'm all for it, fixed in aafbcae8c5d856f7045adc11114613e4c01ad0d6

build.sbt
MaximeKjaer: Yep, fixed in 0a6fd08da1725e0b5aa3a39e51d5f42e00764152

config/src/test/scala/bloop/config/ConfigCodecsSpec.scala
MaximeKjaer: I'm pretty sure the object would run first. It's definitely the case in JUnit, where it may need to run `@BeforeClass` methods on the object before instantiating the class.

jvican: Hmmm, but why do we need to define an object instead of a class? Can’t we have the after all method in the class? My expectation is that junit will run it fine

sjrd: JUnit's `@BeforeClass` and `@AfterClass` must be *static* methods. JUnit will look for static methods annotated with those. In Scala you can't directly write static methods, but methods in a companion object that do not clash with instance methods in the class will receive static forwarders, with the same annotations, in the class. So the only correct way to specify `@BeforeClass` and `@AfterClass` is indeed to define them in the companion object of your test class.

That doesn't make the object a direct target test for JUnit. JUnit is still only concerned with the `class`. But it will find the static forwarders in the class, and handle appropriately.

jvican: Yeah... I know how static methods are encoded in Scala 😀 I was missing that they have to be static methods at all. Looks like this is fixed in junit 5

MaximeKjaer:
I've rebased on top op `master` so that the PR can be merged without conflicts. The rebase cleans up the commit history a little, and also removes a line I had added to the contributing guidelines about doing the necessary setup for Scala Native (that shouldn't be bundled into an unrelated commit about Scala.js).

jvican:
I am rerunning all jobs to see if failures are spurious. If they are gone by tomorrow I'll merge, great work here :+1:

MaximeKjaer:
I'm rebasing to make it possible to merge without conflicts, which will launch a new test run. Here's what happened in the previous one:

- **Ubuntu and MacOS GraalVM**: a `jsoniter-scala` macro fails when compiling `bloop-config` for Scala 2.11-2.13, with the following error message:
  
  ```scala
  [error] /Users/runner/runners/2.263.0/work/bloop/bloop/config/.js/../src/main/scala-2.11-13/bloop/config/ConfigCodecs.scala:118:42: Cannot evaluate a parameter of the 'make' macro call for type 'bloop.config.Config.JvmConfig'. It should not depend on code from the same compilation module where the 'make' macro is called. Use a separated submodule of the project to compile all such dependencies before their usage for generation of codecs. Cause:
  [error] java.lang.AssertionError: assertion failed: com package com <none>
  [error]     JsonCodecMaker.make[Config.JvmConfig](
  [error]                                          ^
  ```

  This error message is one of the [known issues of jsoniter-scala](https://github.com/plokhotnyuk/jsoniter-scala#known-issues). I don't know why this error only happens on GraalVM, and I cannot reproduce it on the JVM on my machine.
  
- **Windows GraalVM**: runs out of memory
  
  ```
  #
  # There is insufficient memory for the Java Runtime Environment to continue.
  # Native memory allocation (malloc) failed to allocate 32744 bytes for ChunkPool::allocate
  # An error report file with more information is saved as:
  # D:\a\bloop\bloop\frontend\hs_err_pid1488.log
  ```

- **MacOS JVM**: It seems that GitHub Actions encountered an internal error; the compilation/test action did not produce any output.

Also note that `sbt jsonConfig210/test` (now redubbed `sbt jsonConfig210JVM/test`) fails on my machine, both on master and on this branch. I can open an issue for this, but I suggest that it be solved separately from this PR.

jvican:
Re the macro error, can you make sure the new sbt projects have the scala doc step in the build disabled? In my experience this issue happens during the scaladoc generation.

Re Windows, we might want to consider disabling new tests in windows as it’s memory budget it’s limited.

Re macOS, I believe rerunning should make the error go again.

plokhotnyuk:
I see that usage of the `Eval.eval` call is hitting more and more your project.

I can release a new version of jsoniter-scala with an additional `makeWithRequiredCollectionFields` call which will not invoke `Eval.eval` for evaluation of configuration parameters that are the same here in the project:
```
CodecMakerConfig.withTransientEmpty(false).withRequireCollectionFields(true)
```
So, all
```
JsonCodecMaker.make(
      CodecMakerConfig.withTransientEmpty(false).withRequireCollectionFields(true)
)
```
can be easily replaced by:
```
JsonCodecMaker.makeWithRequiredCollectionFields
```

A better option would be refactoring of the jsoniter-scala-macros module to avoid `Eval.eval` calls at all (and be ready for adding the Scala 3 support).

MaximeKjaer:
I've disabled Scaladoc for the new Scala.js projects in 53c9e2a8a3abb62066e9b02d26bdc79358509a24. Hopefully that will do the trick.

For Windows, it seems the new tests aren't being run before Windows crashes, so which tests can I disable instead? Is the way to disable some tests on Windows to add a new Windows-only task in `ci.yml`?

For MacOS, indeed, re-running worked.

plokhotnyuk:
@MaximeKjaer could you please update the GraalVM version to [20.1.0](https://github.com/shyiko/jabba/blob/87109f8e7d1ddce03d1d3fca0756e4a09402f141/index.json#L183)? Probably that issue with Windows is already fixed in the latest release.

plokhotnyuk:
I've released jsoniter-scala v2.3.0 with a workaround proposed above. 

It is already available on the Maven Central for download.

MaximeKjaer:
There is one call to the `make` macro (in `ConfigCodecs.scala:172`) that I was not able to replace with `makeWithRequiredCollectionFields` as it has an extra config setting about the discriminator. It seems that this call is the one that makes builds fail:

```scala
JsonCodecMaker.make[JsoniterPlatform](
  CodecMakerConfig
    .withDiscriminatorFieldName(Some("name"))
    .withTransientEmpty(false)
    .withRequireCollectionFields(true)
)
```

How can I proceed?

MaximeKjaer:
Rebased to drop commits that did not fix the failing builds.

jvican:
Excellent will review in a few hours and merge. Thank you for that catch @plokhotnyuk

plokhotnyuk:
Please peek [the latest version](https://repo1.maven.org/maven2/com/github/plokhotnyuk/jsoniter-scala/jsoniter-scala-macros_2.13/2.4.0/) with new `makeWithRequiredCollectionFieldsAndNameAsDiscriminatorFieldName` macro for the remaining case.

MaximeKjaer:
Nice, the tests are passing! Thank you very much for the help with this @plokhotnyuk !

PR #1303, opened 2020-06-04
Update Maven Bloop Scala version to 2.12

tgodzik:
It seems that under 2.10 `Config.Platform.Jvm(Config.JvmConfig(javaHome, launcher.getJvmArgs().toList), mainClass, None, None)` is serialized as:

```json
        "platform" : {
            "name" : "jvm",
            "config" : {
                "home" : "/usr/lib/jvm/java-11-openjdk",
                "options" : [
                ]
            },
            "mainClass" : [
            ],
            "classpath" : [
            ],
            "resources" : [
            ]
        },
```

and under 2.12 it is:
```json
        "platform": {
            "name": "jvm",
            "config": {
                "home": "/usr/lib/jvm/java-11-openjdk",
                "options": [
                    
                ]
            },
            "mainClass": [
                
            ]
        },
```

With the 2.10 version the classpath does not contain the scala library and because of that anything we try fails to run:
https://github.com/scalameta/metals/issues/1805

I don't know the reason why we are keeping the maven plugin at 2.10, but this seemed like a good opportunity to upgrade.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -584,8 +584,8 @@ lazy val sbtBloop013Shaded =
 lazy val mavenBloop = project
   .in(integrations / "maven-bloop")
   .disablePlugins(ScriptedPlugin)
-  .dependsOn(jsonConfig210)
-  .settings(name := "maven-bloop", scalaVersion := Scala210Version)
+  .dependsOn(jsonConfig212)
+  .settings(name := "maven-bloop", scalaVersion := Scala212Version)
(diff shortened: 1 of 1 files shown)


tgodzik:
We could come up with alternative solutions like:
- always add scala library to classpath
- investigate exactly why the behaviour is different

Though, this is much faster to do and I suspect 2.10 is just too old a version and might contain unexpected bugs.

PR #1306, opened 2020-06-09
Fix binaries release on Windows

tgodzik:


tgodzik:
I was hoping that this would solve the issues with the releases that we were having, but not sure anymore :/ 

It seems that we are being hit by https://github.com/sbt/sbt/issues/5587 despite using a totally different version of sbt.
But it might be that we are using the default sbt version from the launcher in that particular job. I will need to investigate a bit.

tgodzik:
@jvican Not 100% sure if it's the case but I am thinking that when we run :
```
          "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat" &&
          "%CD%"\bin\sbt-ci-release.bat
```
the current directory is not that of the project, so we are not detecting the `/project/build.properties` file and using the default 1.3.12, which would explain the issue.

I added `-sbt-version` that should help out if that is the case

jvican:
Let’s give it a try thanks for investigating @tgodzik !

tgodzik:
It doesn't seem to have helped, I will investigate further today.

PR #1310, opened 2020-06-10
Make sure we try to revert to JAVA_HOME if java is not on PATH

tgodzik:
This a fix for the regression in https://github.com/scalacenter/bloop/pull/1203

The additional catch in sysproc caused the exception not to be thrown further.


tgodzik:
> LGTM  I'm sorry for introducing the regression! Thank you for fixing it.

It was my bad, I should have handled the exception there instead of throwing it further. And there were no tests, since it's really hard to simulate java no on PATH on CI :sweat_smile:

Issue #1201, opened 2020-03-17
Add support for Android projects in Gradle


Elephant454:
I set up the Gradle plugin and did a `./gradlew bloopInstall -Si` and was told

`Ignoring 'bloopInstall' on non-Scala and non-Java project 'RadiusMessage'`

I looked at the build.gradle for my app, and it is indeed an Android project as opposed to a Java project. How difficult would it be to add minimal Android support in order to allow compilation of Java code?

jvican:
> I looked at the build.gradle for my app, and it is indeed an Android project as opposed to a Java project. How difficult would it be to add minimal Android support in order to allow compilation of Java code?
  
I don't think it should be difficult. f we can indeed export an Android project safely it'd be enough to whitelist it in the sources of the gradle plugin. Mind looking into it? @Elephant454

jvican:
I've never worked on Android projects in Gradle, but which JDK does the Android project require for compilation? Can we assume you won't run the android project with bloop and that you'll keep using Gradle?

Elephant454:
@jvican I would love to look into it. 🙂

I think the project can use whichever JDK version, as the Gradle config doesn't seem to specify one. Android Studio is currently set to use its own internal JDK version, which is OpenJDK 1.8.

Yes, I would like to continue using Gradle. My plan is to be switching between using Android Studio and Emacs (with Metals).

jvican:
> Yes, I would like to continue using Gradle. My plan is to be switching between using Android Studio and Emacs (with Metals).
  
Cool, so long as you don't run your android project with bloop that should be fine. I would expect that running your android project with the JVM bloop uses would make it crash because IIRC Android projects don't work well with JDK >= 8.
  
I encourage you to have a look at the gradle plugin sources (the entrypoint is `bloop.integrations.gradle.BloopPlugin`) and change the code to make it work with android projects. The best way to go around this is to add a test that includes some minimal android configuration, run the test and see how it behaves. Let me know if I can help out with something and I'll try to chime in as my time allows. Thanks

Elephant454:
I've never worked on a codebase this large before, and I've never worked in Scala either, but this has certainly piqued my interest. I'll let you know what I find. :slightly_smiling_face:

jvican:
Let me know whenever you pick this up @Elephant454

Arthurm1:
This can probably be closed because of #1464

tgodzik:
Thanks!

PR #1313, opened 2020-06-15
Improve gradle dependency detection

### LINKED ISSUE #1311: Undetected Gradle test dependencies are ignored
Given a Gradle project setup with subprojects:

- `a`
- `b`, which depends on `a.test` artefacts

The `b-test.json` bloop file generated from `bloopInstall` does not take the `a.test`dependency  into account (`a-test` not a dependency and its output dir is not on the classpath). This means `bloop test b` does not work, instead failing with 

```shell
❯ bloop test b
Compiling b-test (1 Scala source)
[E] [E4] b/src/test/scala/com/beachape/BSpec.scala:9:3
[E]      not found: value it
[E]      L9:   it("should lol") {
[E]            ^
[E] [E3] b/src/test/scala/com/beachape/BSpec.scala:6:12
[E]      not found: value yo
[E]      L6:     assert(yo == "yo")
[E]                     ^
[E] [E2] b/src/test/scala/com/beachape/BSpec.scala:5:3
[E]      not found: value it
[E]      L5:   it("should yo") {
[E]            ^
[E] [E1] b/src/test/scala/com/beachape/BSpec.scala:3:21
[E]      not found: type ASpecLike
[E]      L3: class BSpec extends ASpecLike {
[E]                              ^
[E] b/src/test/scala/com/beachape/BSpec.scala: L3 [E1], L5 [E2], L6 [E3], L9 [E4]
Compiled b-test (45ms)
[E] Failed to compile 'b-test'
```

I've put up a basic repro here with more details in the README https://github.com/lloydmeta/bloop-gradle-subprojects

I've looked at https://github.com/scalacenter/bloop/issues/800, https://github.com/scalacenter/bloop/issues/808, and https://github.com/scalacenter/bloop/issues/818, which seem to handle this, but it seems to be broken.

### LINKED ISSUE #1312: Gradle plugin doesn't handle `scalaCompilerPlugins`
When you use `scalaCompilerPlugins` to add a scalac plugin as per [the docs](https://docs.gradle.org/current/userguide/scala_plugin.html#sec:scala_compiler_plugins), it doesn't appear to work or get picked up by the bloop definition from `bloopInstall`.

For instance, I have a project where I have `-Xfatal-warnings` enabled and added the [`silencer`](https://github.com/ghik/silencer) plugin to selectively silence deprecation errors. While it works in Gradle, the generated bloop project errors out despite adding `silent` annotations.

Example min repro project: https://github.com/lloydmeta/bloop-with-gradle-and-silencer

Arthurm1:
Project dependency discovery and classpath creation reworked to handle classpath ordering issues and project sourceset identification issues.

Add [TestFixtures](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures) functionality.
Add [ScalaCompilerPlugins](https://docs.gradle.org/current/userguide/scala_plugin.html#sec:scala_compiler_plugins) functionality. Fixes #1312
Cope with lazily defined archives. Fixes #1311
Remove irrelevant artifacts populating resolution modules
Bump max version compatibility to Gradle 6.5

### DIFF
--- a/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
+++ b/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
@@ -14,13 +14,15 @@ import org.gradle.api.{GradleException, Project}
 import org.gradle.api.artifacts._
 import org.gradle.api.artifacts.result.{ComponentArtifactsResult, ResolvedArtifactResult}
 import org.gradle.api.component.Artifact
-import org.gradle.api.internal.artifacts.publish.ArchivePublishArtifact
+import org.gradle.api.file.FileCollection
+import org.gradle.api.internal.artifacts.dsl.LazyPublishArtifact
+import org.gradle.api.internal.artifacts.publish.{ArchivePublishArtifact, DecoratingPublishArtifact}
 import org.gradle.api.internal.file.copy.DefaultCopySpec
-import org.gradle.api.internal.tasks.DefaultSourceSetOutput
 import org.gradle.api.internal.tasks.compile.{DefaultJavaCompileSpec, JavaCompilerArgumentsBuilder}
 import org.gradle.api.plugins.{ApplicationPluginConvention, JavaPluginConvention}
 import org.gradle.api.specs.Specs
-import org.gradle.api.tasks.{AbstractCopyTask, SourceSet, SourceSetOutput}
+import org.gradle.api.tasks.bundling.AbstractArchiveTask
+import org.gradle.api.tasks.{AbstractCopyTask, SourceSet}
 import org.gradle.api.tasks.compile.{CompileOptions, JavaCompile}
 import org.gradle.api.tasks.scala.{ScalaCompile, ScalaCompileOptions}
 import org.gradle.api.tasks.testing.Test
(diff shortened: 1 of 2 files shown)


Arthurm1:
Spurious test failure (I assume): timeout in TcpBspProtocolSpec on Graal Windows

jvican:
As I said, I'm merging, feel free to answer those questions whenever you have some time @Arthurm1 :smile:

Arthurm1:
I see what you mean - all projects are resolved every time a single project is resolved - I'm unsure if Gradle itself caches this - I'll look into it.

I'll also have a look an making project dependencies non-transitive again - I didn't think this would matter and it was easier for me :-)

The only thing to improve the integration that springs to mind at the moment is to add `--parallel` to the call Metals makes to `gradlew bloopInstall`.  This speeds up the install in multi-module exports that don't have `org.gradle.parallel=true` in their `gradle.properties`
In my 130 module/buildTarget project it goes from 11 seconds down to 4 seconds.
I was going to have a look at how easy #1201 is to implement but my Android knowledge is thin.
Otherwise I think the Gradle export is fairly comprehensive now.  Do you know of any projects on GitHub that don't export correctly?  I can have a look at them.

jvican:
> Otherwise I think the Gradle export is fairly comprehensive now. Do you know of any projects on GitHub that don't export correctly? I can have a look at them.
  
I do not know of more gradle projects that don't work well with Bloop. So far my gradle builds at work are working fine, module the performance annoyance of gradle not caching resolution.

> I see what you mean - all projects are resolved every time a single project is resolved - I'm unsure if Gradle itself caches this - I'll look into it.
  
Great! This would be a super important life-of-quality improvement.
  
> The only thing to improve the integration that springs to mind at the moment is to add --parallel to the call Metals makes to `gradlew bloopInstall`. This speeds up the install in multi-module exports that don't have org.gradle.parallel=true in their gradle.properties
  
Awesome, I didn't know about this, it's good to know, thanks :smile:
  
One thing I've been thinking: imagine gradle-bloop would take over gradle Scala integration. How could we implement cross-building projects in a nice way? Is there some state of the art here? I would like to have the opportunity to define cross-built Scala projects (for other platforms as well) so that we can use gradle to do what's possible in other builds. Do you have any thoughts here?

PR #1314, opened 2020-06-16
Don't expand resource directory entries in BSP server

olafurpg:
Previously, the Bloop BSP server recursively walked all directory
resource entries and returned individual files. Now, the BSP server
returns the URIs of the directories unchanged and leaves it up to the
client to recursively walk the directories, if necessary.

There's nothing in the BSP spec that says that resource URIs cannot be
directories. The current behavior looses information because client
won't be able to relativize resource items by their enclosing directory.

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -1032,13 +1032,8 @@ final class BloopBspServices(
       val response = bsp.ResourcesResult(
         projects.iterator.map {
           case (target, project) =>
-            val resources = project.runtimeResources.flatMap { s =>
-              if (s.exists) {
-                val resources = Files.walk(s.underlying).collect(Collectors.toList[Path]).asScala
-                resources.map(r => bsp.Uri(r.toUri()))
-              } else {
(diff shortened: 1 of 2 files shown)


Duhemm:
@satabin Can you please comment on why does Bloop walk over all the resources? See in #946

olafurpg:
Unrelated test failure in CI on Windows
```
X bloop.bsp.TcpBspProtocolSpec.find test classes 32697ms 
```

> Does BSP always assume that the server and the client are running on the same machine, and that the client has direct access to the files?

The protocol itself doesn't make that assumption but the Bloop/Metals/IntelliJ implementation assume that `file:` URIs are available on local disk.

> Should this instead be fixed by having the protocol return the "resource root" of each resources along with the resource path?

I don't think that's necessary. The new behavior is consistent with the BSP `buildTarget/sources` request

jvican:
LGTM 👍

satabin:
> @satabin Can you please comment on why does Bloop walk over all the resources? See in #946

Hi,

This is a good question, and IIRC I had the idea to potentially support completion on some resources (like configuration keys). But there is not good reason that comes to my mind right now.

PR #1316, opened 2020-06-17
Workaround Windows publishing issue

Duhemm:
The native image binaries publishing on Windows is currently broken
because of an issue where sbt fails to find
org.scala-lang.modules#scala-xml_2.12;1.2.0!scala-xml_2.12.jar when sbt
is started with `cmd` rather than `bash`.

To workaround this issue, we launch sbt a first time with `bash`, and
then do the publishing with `cmd`.

Why it fails with `cmd` is a mystery to me.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,6 +162,14 @@ jobs:
           echo $JAVA_HOME
           "$JAVA_HOME\bin\gu.cmd" install native-image
         shell: bash
+      - name: Pre-load sbt with bash
+        # FIXME: This step shouldn't be necessary, but sbt fails to find
+        # org.scala-lang.modules#scala-xml_2.12;1.2.0!scala-xml_2.12.jar when launched with
+        # cmd.
+        # Keep the sbt version in sync with `sbt-ci-release.bat`.
+        run: |
+          sbt -sbt-version 1.3.3 version
+        shell: bash
       - name: Publish GraalVM Native artifacts
         run: >-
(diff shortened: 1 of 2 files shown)


Duhemm:
It looks like the publishing on Windows was working consistenly until #1300 was merged. After that, it failed consistently. I tried reverting #1300, but it's still failing. I have no idea why. All I know is that opening sbt with bash first works 🤷 

@jvican Do you know what's going on?

tgodzik:
@Duhemm I've been looking into this, but I haven't had much luck. Everything seem to work correctly when testing on a desktop Windows.

The error looks like: https://github.com/sbt/sbt/issues/5587 But it's a totally different sbt version. Maybe we could change the way we run sbt to test it out?

olafurpg:
What happens if we `rm -rf ~/.sbt/preloaded/` before launching sbt?

jvican:
> @jvican Do you know what's going on?
  
No idea, looking into this now

Duhemm:
> What happens if we `rm -rf ~/.sbt/preloaded/` before launching sbt?

I tried here, that didn't help: https://github.com/scalacenter/bloop/runs/780756019?check_suite_focus=true

olafurpg:
The Windows job failed with a different error now
```
 [error] java.io.IOException: Cannot run program "C:/Users/runneradmin/.jabba/jdk/graalvm@20.0.0/bin/native-image.cmd" (in directory "D:\a\bloop\bloop\bloopgun\target\graalvm-native-image"): CreateProcess error=2, The system cannot find the file specified
```

Duhemm:
That's the culprit this time: https://github.com/scalacenter/bloop/blob/bdae2be2e42b6440eb492d63a2a42561ddf14bb2/build.sbt#L331-L335

PR #1321, opened 2020-06-19
Workaround releasing issue

### LINKED ISSUE #1320: Releases are failing in CI
From https://github.com/scalacenter/bloop/runs/787304809?check_suite_focus=true#step:7:988
```
[error] sbt.internal.util.Init$RuntimeUndefined: References to undefined settings at runtime.
[error] setting(ScopedKey(Scope(Select(ProjectRef(file:/home/runner/work/bloop/bloop/,bloop)), Zero, Zero, Zero),releaseSonatypeBundle)) at LinePosition(/home/runner/work/bloop/bloop/build.sbt,904) referenced from ScopedKey(Scope(Select(ProjectRef(file:/home/runner/work/bloop/bloop/,bloop)), Zero, Zero, Zero),sonatypeBundleRelease)
[error] 	at sbt.internal.util.Init.sbt$internal$util$Init$$handleUndefined(Settings.scala:693)
[error] 	at sbt.internal.util.Init$$anon$9.apply(Settings.scala:697)
```

@jvican Any idea what might be causing this issue?

Duhemm:
The release was calling the task `sonatypeBundleRelease`, which doesn't
exist (a command with the same name exists, however). To fix it, we call
the command `sonatypeBundleRelease`.

Fixes #1320

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -939,5 +939,5 @@ addCommandAlias(
     .mkString(";", ";", "")
 )
 
-val allReleaseActions = List("releaseEarlyAllModules", "releaseSonatypeBundle")
+val allReleaseActions = List("releaseEarlyAllModules", "sonatypeBundleRelease")
 addCommandAlias("releaseBloop", allReleaseActions.mkString(";", ";", ""))


jvican:
That task name does indeed exist, it’s defined in build.sbt, the thing it does is to avoid running sonatypeReleaseBundle if there haven’t been any published artifact (which can happen when retrying a CI run for a release tag).

Duhemm:
> That task name does indeed exist, it’s defined in build.sbt

I'm not following. The one that is defined in `build.sbt` is `releaseSonatypeBundle`, not `sonatypeBundleRelease`. Can you point me to the place where `sonatypeReleaseBundle` is assigned?

jvican:
I’m on the phone but I think what you’re getting at is that the build was failing because we were calling the command `sonatypeReleaseBundle` as if it was a task in `releaseSonatypeBundle`.

I’m good with merging this but what I wanted to point out is that by calling the sonatype command directly then we’ll fail if we retrigger a failed CI run for a release.

Duhemm:
Thank you for confirming! Let's merge as is and I'll open a ticket to investigate support for re-running releases.

jvican:
No need to open a ticket, let’s just not support that for now and if we neee it then we’ll do it :)

PR #1327, opened 2020-06-24
Carry underlying error in `GlobalError`

Duhemm:
Previously, `GlobalError` contained only an error message, which means
that the exception causing a `GlobalError`, if any, was lost along with
the information it contains.

This commit adds an optional field holding the underlying exception,
which will be logged when present.


Duhemm:
Reformated 🙈 

Thanks for the review!

PR #1331, opened 2020-06-30
Add support for Scala 2.13.3 and prepare 1.4.3

jvican:


backend/src/main/scala/sbt/internal/inc/BloopComponentCompiler.scala
```diff
@@ -109,7 +109,17 @@ object BloopComponentCompiler {
      * Note that this method cannot be defined in [[CompilerBridgeProvider]] because [[ModuleID]]
      * is a Scala-defined class to which the compiler bridge cannot depend on.
      */
-    private def compiledBridge(bridgeSources: ModuleID, scalaInstance: ScalaInstance): File = {
+    private def compiledBridge(bridgeSources0: ModuleID, scalaInstance: ScalaInstance): File = {
+      val scalaVersion = scalaInstance.version()
+      val requiresPrevZincVersion = bridgeSources0.revision == "1.3.0-M4+45-d4354be3" && {
```
Duhemm, line 114: Can we move `"1.3.0-M4+45-d4354be3"` and `"1.3.0-M4+42-5daa8ed7"` to `Dependencies.scala`, and export it via sbt-buildinfo?

```scala
object Dependencies { 
  ...
  // Scala 2.13.3 broke some APIs used by the compiler bridge, so people using
  // Scala < 2.13.3 will keep using the old bridge.
  val pre2133BridgeVersion = "1.3.0-M4+42-5daa8ed7"
  val zincVersion = "1.3.0-M4+45-d4354be3"
  ...
}
```

Otherwise I'm sure we'll forget to bump this file at some point when upgrading Zinc.

jvican: Is it OK if I instead use an assert? I think it'll be better than depending on these values from the build, I think we already have a lot of build dependencies into our project and would like to avoid depending on more so that we can refactor the build in the future more easily. What do you think?

Duhemm: If we decouple this from the build, let's do this and add a big fat comment in `Dependencies.scala` saying to keep this file up-to-date with the Zinc version 👍

jvican: Will be following to your feedback in another PR, thanks for the review @Duhemm 🙏

Duhemm: Hehe I just got bitten by this in #1416 😏

PR #1341, opened 2020-07-15
Destroy process group alongside process

tpasternak:
Background
In Unix systems, the `kill` syscall kills only a single process. This
means, it's subprocesses are not stopped. Instead, they are
re-parented and they continue to run.

A way to manage lifetime of a set of processes is to form a Process
Group and then kill the whole group. Unfortunatelly, the
`NuProcess.start` method currently does not support Process Groups.
A way to workaround this is to delegate group creation to the process
itself and handle both cases: if it creates a group, and if it does not.

Nevertheless, we may utilize the fact that, when a process creates
a Process Group, the group's ID is equal to the PID. So the Forker may
always try to kill the group of that id. If the process created a group, then it would
be killed in total. If not, then nothing would happen. It is still
better than just killing the root process and allow children to be re-parented.

References
https://en.wikipedia.org/wiki/Orphan_process
https://man7.org/linux/man-pages/man2/kill.2.html

### DIFF
--- a/frontend/src/main/scala/bloop/exec/Forker.scala
+++ b/frontend/src/main/scala/bloop/exec/Forker.scala
@@ -3,6 +3,7 @@ package bloop.exec
 import java.io.{FileNotFoundException, IOException}
 import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
+import java.nio.file.FileSystems
 import java.util.concurrent.TimeUnit
 
 import bloop.cli.{CommonOptions, ExitStatus}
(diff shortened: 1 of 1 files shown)


jvican:
Hey @tpasternak, this is an interesting PR, any reason why you killed it? Would like to understand more the use case behind this.

tpasternak:
Hi, we did have a problem with killing fastpass, but we already solved this in a different way (just insisde fastpass). The problem with this PR was that it assumed that the child process will start a new *process group* itself. This means we end up with a split responsibility (child responsible for starting a group, bloop responsible to kill it) which I consider as a bad design.

To do it right we would have to make bloop start a new `process group`, as well as kill it during cancellation. Unfortunately we can't achieve it with `NuProcess`. One way to overcome it would be to prefix the command with `setsid`, but we can't assume it's available on all systems.

PR #1343, opened 2020-07-17
Recognise Gradle integration tests as test projects

### LINKED ISSUE #1335: Missing test frameworks for integration test projects
### Problem

Gradle supports having integration-tests under a different source set with an accompanying task ([docs](https://docs.gradle.org/current/userguide/java_testing.html#sec:configuring_java_integration_tests)), but it seems like the [bloop Gradle plugin](https://scalacenter.github.io/bloop/setup#gradle) doesn't work with this, because the resulting bloop project definition lacks the `project.tests` stanza, causing attempts to run tests to do nothing and log a warning as follows:

```shell
[W] Missing configured test frameworks in bloop-on-gradle-with-its-it
Total duration: 0ms
No test suites were run.
```

### Probable cause

I think this is because all the decisions in [`BloopConverter`](https://github.com/scalacenter/bloop/blob/master/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala) to check whether a source set is test-related or not is driven by testing the source set name to see if it's `test`

https://github.com/scalacenter/bloop/blob/b96374d7c30684cbf8afd5dbac6cd4979fd60747/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala#L69

In particular, I think this is the bit that causes the bloop test config (with its frameworks) to not be defined when the source set name is not `test`:

https://github.com/scalacenter/bloop/blob/b96374d7c30684cbf8afd5dbac6cd4979fd60747/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala#L407-L414

### Possible fix

I think we can naively "fix" this by simply always returning `Some(Config.Test.defaultConfiguration)` from 

https://github.com/scalacenter/bloop/blob/b96374d7c30684cbf8afd5dbac6cd4979fd60747/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala#L407

but I'm not sure whether always having the test config defined for projects would break anything.

Alternatively, we'd need an improved way of detecting whether or not a source set is test related, one that looks at more than just the name (what though?)

### Min repo

I have a min repro [here](https://github.com/lloydmeta/bloop-on-gradle-with-its); the only differences between it and the official guide are:
* the integration-test-hook has been added as a plugin to keep things DRY in case we want to test with subprojects
* the integration tests are located in `src/it`, sbt-style

### Workaround

One can make this work by copying-and-pasting `project.tests` from `$projectname-test.json`, then:

```shell
❯ bloop test bloop-on-gradle-with-its-it
MehItSpec:
- should w/e
Execution took 16ms
1 tests, 1 passed
All tests in com.beachape.MehItSpec passed

Total duration: 16ms
All 1 test suites passed.
```

Arthurm1:
Exports any Gradle source set, that has a Test task relating to it, as a test project with test tag and test frameworks

Fixes #1335

### DIFF
--- a/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
+++ b/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
@@ -20,7 +20,6 @@ import org.gradle.api.internal.artifacts.publish.{ArchivePublishArtifact, Decora
 import org.gradle.api.internal.file.copy.DefaultCopySpec
 import org.gradle.api.internal.tasks.compile.{DefaultJavaCompileSpec, JavaCompilerArgumentsBuilder}
 import org.gradle.api.plugins.{ApplicationPluginConvention, JavaPluginConvention}
-import org.gradle.api.specs.Specs
 import org.gradle.api.tasks.bundling.AbstractArchiveTask
 import org.gradle.api.tasks.{AbstractCopyTask, SourceSet}
 import org.gradle.api.tasks.compile.{CompileOptions, JavaCompile}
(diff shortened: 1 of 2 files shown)


Arthurm1:
I don't think the CI failures are relevant to this PR as it looks as though the Gradle tests run in the `Run build tool + native tests` section and that hasn't, or doesn't, run for Windows (the failing builds).

I'm not very familiar with Github Actions - is there an easier way to read the test results without just scrolling through all the logs?  E.g. is there a ScalaTest result web page published for the run?

jvican:
Actually, nevermind that comment, we're exporting all source sets so integration test source sets should have the config too. Let me know if my intuition here is correct

Arthurm1:
We export all source sets as separate projects and if _any_ of those have a Test task against them then we'll mark it with a Test Tag and add the frameworks.  So integration test source sets should be tagged with Test and have the test frameworks associated.

PR #1348, opened 2020-07-21
Ignore empty directory layout in external classes dir

jvican:
Because the proces owning external classes directories might force the 
 creation of certain empty directories, the process responsible for copying
 directories would consider the external client classes directory was not
 empty and would not copy class files from the internal classes directory.

 This commit fixes that and ignores empty directories in the external 
 classes dir, resulting in the copying process working as expected.


tgodzik:
I was thinking we could add https://github.com/scalacenter/bloop/pull/1268
Bsp already got released and I rebased the branch, but I am wondering how we should add the environment variable when they are run via console? A separate parameter to the CLI?

PR #1349, opened 2020-07-23
Strip ANSI color escapes when colors are disabled

### LINKED ISSUE #1119: Unexpected color when --no-color flag is used
bloop 1.3.4
```
bloop compile root --no-color
Compiling root (8 Scala sources)
[W]  [E-1] there was one feature warning; re-run with -feature for details
[W]  [E-1] there was one deprecation warning (since 2.11.0); re-run with -deprecation for details
Compiled root (905ms)
Compiling root (5 Scala sources)
[E] [E1] root/src/main/scala/.../File.scala:23:99
...
[E]      L23: [some error]
[E]                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[W]  [E-1] there was one feature warning; re-run with -feature for details
[W]  [E-1] there was one deprecation warning (since 2.11.0); re-run with -deprecation for details
[E] root/src/main/scala/.../File.scala: L23 [E1]
```
E-1, E1 are blue
L23 is red

Duhemm:
Previously, the `BloopLogger` would remove the colored prefix of the log
messages when colors were disabled via `--color` or `NO_COLOR`. However,
color that may be present in the log output (inserted by the running
application or a test framework for instance) would not be removed.

This commit preprocesses the log messages so that the ANSI colors are
removed from the log messages.

Fixes #1119

### DIFF
--- a/backend/src/main/scala/bloop/logging/BloopLogger.scala
+++ b/backend/src/main/scala/bloop/logging/BloopLogger.scala
@@ -3,6 +3,7 @@ package bloop.logging
 import java.io.PrintStream
 
 import scala.Console.{CYAN, GREEN, RED, RESET, YELLOW}
+import scala.util.matching.Regex
 
 /**
  * Creates a logger that writes to the given streams.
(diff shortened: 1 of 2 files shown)


Duhemm:
I'm mostly interested in making the test frameworks respect NO_COLOR. Unfortunately, they use different flags to disable colors in their output. An alternative would be to maintain a mapping from test framework to flags required to disable the colors. I prefer this solution.

PR #1355, opened 2020-08-17
Use swoval file tree views for source monitoring

eatkins:
The directory watcher library used by bloop is not well optimized for
mac os. When I use metals via coc.nvim, I find that the `~` command
often stops working in my sbt projects. The reason is likely that
directory watcher is exhausting the file watching resources provided by
the OS. The swoval file tree views library originally started out as a
fork of directory watcher to fix its resource management issues. It can
be used almost as a drop in replacement for directory watcher though the
api is somewhat different.

The problem is that the directory watcher is written to
implement the java nio watchservice which is itself based on inotify but
the apple file event system is very different from inotify so there is
an impedence mismatch. The inotify api only works on a particular path.
In mac os, the main file watching api watches the root of a file system
tree. This means that if you want to watch two subdirectories foo/bar
and foo/bar/baz, there is no need to create separate event streams for
foo/bar and foo/bar/baz, the events for foo/bar/baz will be
automatically detected by the stream for foo/bar. The swoval library
verifies that each path that is registered is not covered by an existing
stream. It also removes redundant streams if a new stream is opened that
covers an existing stream. Filtering of unwanted subdirectories of a
root directory is handled at the jvm application level.This stream
management is critical because mac os does not allow unlimited numbers
of streams. When adding a recursive directory watch to a
DirectoryWatcher, it creates a watch stream for every subdirectory which
can quickly cause os to stop accepting new watches. It also creates a
new thread for every subdirectory. Swoval will use one thread to handle
events for all of the directories that it is watching.

To minimize any potential downstream impact of this change, I made the
swoval file tree watcher still emit directory watcher
DirectoryChangeEvents. There isn't a 1:1 mapping from a
DirectoryChangeEvent to a swoval data structure anyway so it was
convenient to continue using it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -71,6 +71,7 @@ lazy val backend = project
       Dependencies.sourcecode,
       Dependencies.monix,
       Dependencies.directoryWatcher,
+      Dependencies.swovalFiles,
       Dependencies.zt,
       Dependencies.brave,
       Dependencies.zipkinSender,
--- a/frontend/src/main/scala/bloop/io/SourceWatcher.scala
+++ b/frontend/src/main/scala/bloop/io/SourceWatcher.scala
@@ -9,6 +9,12 @@ import bloop.util.monix.FoldLeftAsyncConsumer
 
 import scala.collection.JavaConverters._
 import scala.concurrent.duration.FiniteDuration
+import scala.util.Try
+
+import com.swoval.files.{FileTreeRepositories, FileTreeRepository, TypedPath}
+import com.swoval.logging.{Logger => SLogger, Loggers}
+import com.swoval.files.FileTreeDataViews.{CacheObserver, Entry}
+import com.swoval.files.FileTreeViews.Observer
 
 import io.methvin.watcher.DirectoryChangeEvent.EventType
 import io.methvin.watcher.{DirectoryChangeEvent, DirectoryChangeListener, DirectoryWatcher}
@@ -20,6 +26,7 @@ import monix.reactive.{MulticastStrategy, Observable}
 import monix.execution.misc.NonFatal
 import monix.execution.atomic.AtomicBoolean
 import monix.reactive.OverflowStrategy
+import java.io.IOException
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.TimeUnit
 import java.nio.file.attribute.BasicFileAttributes
@@ -51,23 +58,32 @@ final class SourceWatcher private (
       )
 
     var watchingEnabled: Boolean = true
-    val listener = new DirectoryChangeListener {
-      override def isWatching: Boolean = watchingEnabled
+    val swovalObserver = new CacheObserver[TypedPath] {
 
       // Make sure that errors on the file watcher are reported back
-      override def onException(e: Exception): Unit = {
+      override def onError(e: IOException): Unit = {
         slf4jLogger.debug(s"File watching threw an exception: ${e.getMessage}")
         // Enable tracing when https://github.com/scalacenter/bloop/issues/433 is done
         //logger.trace(e)
       }
 
       private[this] val scheduledResubmissions = new ConcurrentHashMap[Path, Cancelable]()
-      override def onEvent(event: DirectoryChangeEvent): Unit = {
-        val targetFile = event.path()
-        val attrs = Files.readAttributes(targetFile, classOf[BasicFileAttributes])
+      override def onCreate(entry: Entry[TypedPath]): Unit =
+        if (watchingEnabled) onEvent(entry, DirectoryChangeEvent.EventType.CREATE)
+      override def onDelete(entry: Entry[TypedPath]): Unit =
+        if (watchingEnabled) onEvent(entry, DirectoryChangeEvent.EventType.DELETE)
+      override def onUpdate(previous: Entry[TypedPath], current: Entry[TypedPath]): Unit =
+        if (watchingEnabled) onEvent(current, DirectoryChangeEvent.EventType.MODIFY)
+      def onEvent(entry: Entry[TypedPath], kind: DirectoryChangeEvent.EventType): Unit = {
+        val targetFile = entry.getTypedPath.getPath()
+        val event = new DirectoryChangeEvent(kind, targetFile, 1)
 
-        if (attrs.isRegularFile && SourceHasher.matchSourceFile(targetFile)) {
-          if (attrs.size != 0L) {
+        if (entry.getTypedPath.isFile && SourceHasher.matchSourceFile(targetFile)) {
+          val defer =
+            if (kind == DirectoryChangeEvent.EventType.MODIFY)
+              Try(Files.size(targetFile)).getOrElse(0L) == 0L
+            else kind == DirectoryChangeEvent.EventType.DELETE
+          if (!defer) {
             val resubmission = scheduledResubmissions.remove(targetFile)
             if (resubmission != null) resubmission.cancel()
             observer.onNext(event)
@@ -100,7 +116,7 @@ final class SourceWatcher private (
               new Runnable {
                 def run(): Unit = {
                   scheduledResubmissions.remove(targetFile)
-                  if (Files.size(targetFile) == 0) {
+                  if (Try(Files.size(targetFile)).getOrElse(0L) == 0L) {
                     observer.onNext(event)
                   }
                   ()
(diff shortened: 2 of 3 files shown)


eatkins:
I have seen a number of failures in tests that I don't think could have been affected by this change like `NailgunSpec`.

jvican:
Hey @eatkins, thanks for the contribution. I will need to evaluate what it means to depend on swoval so I'd likely need some time to look at this proposed change in depth.

eatkins:
Sure. Every dependency is a liability so you should be cautious. How can I help you evaluate swoval? I provided more context as to what the specific problem with directory-watcher is on osx and why I developed swoval in response to its limitations in the thread in this PR: https://github.com/scalameta/metals/pull/2014. As long as sbt depends on swoval, the entire scala ecosystem is dependent upon it working well so I think the risk of adopting it is relatively low. Swoval is close to a drop in replacement for directory-watcher, but the reverse is not true. This PR should have no effect outside of SourceWatcher.scala and the changes there could easily be reverted if there are unforeseen problems.

In addition to the comments I made in that thread, I'd add that swoval's testing suite is far more extensive than directory-watcher. It's a crude measure but just measured by source files and lines of code swoval has 20 test source files and 3600 lines of test code. Directory-watcher has 5 test source files and 933 lines of code. The directory watcher tests also take 90 seconds to run on my mac and a number of the tests fail (there is no mac testing in directory-watcher's CI). Swoval runs 260 test cases in 3 seconds on my mac laptop. The vast majority of them are doing io, modifying files and asynchronously detecting events. They also run concurrently because they are mostly utest async tests. No corners were cut to get a fast test execution time.

Code coverage is by no means a perfect metric, but swoval also requires 84% line coverage and 74% branch coverage to pass CI which I suspect is on the high side for an OSS project. What is uncovered is also largely unimportant. All of the tests are run a minimum of 50 times in CI to ensure that there are no flaky tests. In addition to swoval's tests, sbt io and sbt main also have extensive testing of watch including 16 scripted tests which all rely upon swoval working correctly.

Swoval is cross platform and works on scala.js as well as the jvm so the internal logic has been tested with two completely different concurrency models. There were a number of jvm bugs that were only brought to light by js testing. The majority of the scala.js code is generated by automatically converting java source code to scala. This required a high degree of discipline within the project because I could only use java constructs that could be easily converted to scala and I had to create concurrency constructs that could be used by both platforms that didn't rely on scala Future.

I also went out of my way to document the entire public api:
* javadoc https://swoval.github.io/docs/swoval/2.1.5/jvm/api
* scaladoc https://swoval.github.io/docs/swoval/2.1.5/js

The library also is now quite widely used since it is used for file monitoring in sbt 1.3.x:
Screen Shot 2020-08-25 at 8 22 44 AM
An older version of the library, apple-file-events, was introduced in sbt 1.1.5 so it has gotten a lot of battle testing over a multi year period. At the moment, there are no major outstanding issues related to file monitoring open in the sbt project (there is one minor issue for which the user hasn't yet provided a repro where sbt triggers twice on some updates to symlinked files). 

Obviously I am going to be an advocate for my own library but that is because I have put an enormous amount of effort into ensuring that it is a high quality library that just works. I don't have an axe to grind with directory-watcher. It was instrumental in helping me learn how to write a file monitoring library. It's just that I built on its ideas to, in my opinion, make something better. To me the fatal flaw with directory-watcher is that it can break functionality in _other_ programs. This is really apple's fault for poorly implementing fsevents, but I really worked hard to workaround the limitations of apple's apis in a way that directory-watcher has not.

It seems likely that it has been metals (which has already accepted the switch from directory-watcher to swoval), not bloop, that has impacted my personal workflow so it probably won't affect me much either way whether or not this PR is accepted. I had assumed it was a bloop problem because I knew that bloop used directory-watcher and that metals used bloop. I didn't realize when I opened this PR that metals also used directory-watcher internally to make goto def work. Even though bloop may not have been responsible for my issue, my thinking is that it will help other people who may be using bloop's watch mode in large repositories with enough subdirectories to get directory-watcher to exhaust osx's limits for watch streams which is **system global** not local to a particular process.

What is insidious about this is that a user who hits the problem of directory watcher exhausting the available watch streams is that the failure mode is completely silent. A different program that tries to watch files with fsevents will just silently fail to detect any events. In older versions of osx, you at least got a noisy error printed to System.err, but in Catalina no errors are displayed. When this happens to me personally, I am often confused as to whether or not my editor saved the file. Once I know that watch isn't working, I have to kill all of my java processes and restart sbt in order to get watch to work again. I was only able to figure out that that was what was happening because I had already seen this behavior when I first stated using directory-watcher almost three years ago.

tgodzik:
Seems like people are getting issues on MacOS like:
`2020-09-22 10:40 java[45159] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)`

@eatkins might this be coming from the issue you mention?

swoval seems stable both in sbt and Metals, maybe we should merge?

jvican:
This is not a high-priority PR for me so I can't provide timely feedback now. I appreciate the contribution but I'm skeptical this is the way to go, even if Metals is already using it. I will need more time to carefully consider if I want to add this dependency and depend on it for the time to come.

tgodzik:
> This is not a high-priority PR for me so I can't provide timely feedback now. I appreciate the contribution but I'm skeptical this is the way to go, even if Metals is already using it. I will need more time to carefully consider if I want to add this dependency and depend on it for the time to come.

Sure, I understand. I just wanted to mention some issues users are experiencing. And with both sbt and Metals I was thinking it is a safer bet to try it out.

eatkins:
> Seems like people are getting issues on MacOS like:
2020-09-22 10:40 java[45159] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)

Yes, this is exactly what this PR is intended to address. I tried my best in the metals thread to explain why the design of directory watcher makes it far more likely that users would experience this issue than swoval. Here is the specific comment where I explain what the problem is: https://github.com/scalameta/metals/pull/2014#issuecomment-677942540

> I appreciate the contribution but I'm skeptical this is the way to go, even if Metals is already using it. 

I would appreciate some feedback about what you are skeptical about. These two libraries provide very similar functionality but swoval was specifically designed to address this issue that is real on osx and that users do experience as @tgodzik's comment demonstrates. The worst case scenario in merging this is that I did a horrible job writing swoval and as a result watch breaks for some users. That seems unlikely given that it passes bloop's watch related tests and the success of the library in sbt but I'll allow for the possibility. If that worst case scenario came to pass, you could just revert this commit. It's not as though this is a part of the code base that changes often.

eatkins:
@jvican it seems clear to me that you do not want to accept this contribution but are unwilling to say so. Life is too short to waste time trying to contribute to a project that clearly does not want my contribution. You are obviously under no obligation to accept or even evaluate any PR from anyone but your dismissive attitude towards my work leaves a very sour taste in my mouth.

tgodzik:
@eatkins Thanks for the contribution! I do not know the codebase that well to merge without @jvican 's approval, but I believe you did a lot of great work. Sorry if you were discouraged by the replies.

I followed with an issue here: https://github.com/scalacenter/bloop/issues/1376 

This seems to be a real issue we are experiencing and honestly unless we are able to fix it in the directory watcher itself I am not sure this PR should be dismissed.

@jvican Do you have any other idea how we should fix the issue? This was actually reported by a number of people on gitter

darjutak:
@eatkins we at the Scala Center appreciate your time, work, and contributions. 

We are sorry for this unfortunate experience - we are currently internally working on how to prevent this from happening again in one of the projects under our organization. I hope once we find a solution, you will find some time to continue contributing to our projects.

Even though we also appreciate Jorge's work and investment, it is worth noting that @jvican no longer represents the Scala Center organization, due to change of jobs, and does this in his spare time, which is why I assume he has different priorities these days.

kammoh:
I'm running into multiple issues with the current file watcher on macOS Big Sur. `swoval` seems like a solid choice and @eatkins contribution seems really nice (though I was not able to build bloop locally to try it out). 

`\begin{rant}`
I also see other contributions from @eatkins and other people who have been pretty much ignored by the bloop team. This is going to be very discouraging for others thinking about putting time and effort into these fixes. IMHO this doesn't paint a too bright picture about bloop's future in general. Hope things are going to improve with Scala Center taking control.
`\end{rant}`

@eatkins is there a chance you could consider resurrecting this PR or if someone else can PR a derivative?

tgodzik:
@kammoh I haven't yet been able to dig deeply into Bloop yet, but I do plan at some point to revisit the PR. Sorry, it's taking me so long, but there is a lot of work surrounding tooling for Scala 3.

PR #1356, opened 2020-08-17
Upgrade bloop build to 1.4.0-M2

eatkins:
These changes make bloop work with sbt 1.4.0-M2. There were some small
api changes in zinc that needed to be addressed. sbt's classpath changed
slightly and now includes scala.xml so that had to be excluded from the
shading namespace (assuming that bloop doesn't want/need to shade scala
xml).


eatkins:
You might want to wait until 1.4.0-RC1 or 1.4.0 but I figured I'd share this to save someone else the trouble.

eatkins:
Actually this probably needs to wait for an M3, RC1 or an update of bloop's fork of zinc due to https://github.com/sbt/zinc/issues/884

jvican:
Thanks for the PR! We are not likely to catch up with the latest sbt release at least for some months, so I'm closing this PR and I'll revisit it in the future when we want to update.

PR #1359, opened 2020-08-24
Fix #1358: Run test classes only once

### LINKED ISSUE #1358: Bloop runs the same test class multiple times
Test classes that match more than a single fingerprint are run more than once:

`build.sbt`:
```scala
lazy val framework = project
  .in(file("framework"))
  .settings(
    libraryDependencies += "org.scala-sbt" % "test-interface" % "1.0"
  )

lazy val test = project
  .in(file("test"))
  .dependsOn(framework)
  .settings(
    testFrameworks += TestFramework("foo.Framework")
  )
```

`framework/src/main/scala/foo/Framework.scala`:
```scala
package foo

import sbt.testing._

trait Test
trait Test2
case class SCFingerprint(superclassName: String) extends SubclassFingerprint {
  def isModule: Boolean = false
  def requireNoArgConstructor: Boolean = false
}

class Framework extends sbt.testing.Framework {
  def name(): String = "My awesome framework"
  def runner(args: Array[String], remoteArgs: Array[String], testClassLoader: ClassLoader): Runner = new Runner(args, remoteArgs)
  def fingerprints(): Array[Fingerprint] =
    Array(SCFingerprint("foo.Test"), SCFingerprint("foo.Test2"))
}

class Runner(val args: Array[String], val remoteArgs: Array[String]) extends sbt.testing.Runner {
  override def done(): String = ""
  override def tasks(tasks: Array[TaskDef]): Array[Task] = tasks.map { t =>
    new Task {
      def execute(handler: EventHandler, loggers: Array[Logger]): Array[Task] = {
        loggers.foreach(l => l.info("Running task: " + t.fullyQualifiedName()))
        Array.empty
      }
      def tags(): Array[String] = Array.empty
      def taskDef(): TaskDef = t
    }
  }
}
```

`test/src/test/scala/foo/MyTest.scala`:
```scala
package foo
class MyTest extends Test with Test2
```

sbt output:
```
sbt:multi-run> test
[info] Running task: foo.MyTest
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
```

bloop output:
```
Running task: foo.MyTest
Execution took 0ms
No test suite was run

Running task: foo.MyTest
Execution took 0ms
No test suite was run

Total duration: 0ms

```

The test should have been run only once in Bloop.

Duhemm:
Bloop was collecting all the test fingerprint, and creating a test task
for every matching fingerprint. This meant that test classes would be
run as many times as they have matching fingerprints.

Bloop now follows the same behavior as sbt, which will run the test with
the first matching framework.

Fixes #1358

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
@@ -277,14 +277,23 @@ object TestTask {
     import scala.collection.mutable
     val (subclassPrints, annotatedPrints) = TestInternals.getFingerprints(frameworks)
     val definitions = TestInternals.potentialTests(analysis)
-    val discovered = Discovery(subclassPrints.map(_._1), annotatedPrints.map(_._1))(definitions)
+    val discovered =
+      Discovery(subclassPrints.map(_._1).toSet, annotatedPrints.map(_._1).toSet)(definitions)
     val tasks = mutable.Map.empty[Framework, mutable.Buffer[TaskDef]]
+    val seen = mutable.Set.empty[String]
     frameworks.foreach(tasks(_) = mutable.Buffer.empty)
     discovered.foreach {
       case (defn, discovered) =>
         TestInternals.matchingFingerprints(subclassPrints, annotatedPrints, discovered).foreach {
           case (_, _, framework, fingerprint) =>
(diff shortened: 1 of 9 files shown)


jvican (approved):
LGTM :+1:

PR #1361, opened 2020-08-27
Add trailing slash only when necessary in manifest

Duhemm:
When creating a manifest JAR, `JvmProcessForker` would add a trailing
slash to all the directory entries. While this trailing slash is
apparently necessary, it would be added also to entries that already
have a trailing slash. As a result, the path to resources resolved from
a directory entry ending with a double slash would contain this double
slash.


Duhemm:
@wiwa Can you please take a look as well?

wiwa:
lgtm

jvican (approved):
LGTM 👍

PR #1363, opened 2020-09-01
Use our fork of test-agent

Duhemm:
The sbt test-agent recently received a fix for duplicated tasks being
run (see https://github.com/sbt/sbt/pull/5800). We would like to start
using this fix in Bloop, but don't want to wait for the next release.

This commit adds a new artifact, called `test-agent`, which contains the
fixed version of the test-agent, so that we can use it without waiting
for sbt to publish a new version.

/cc @tpasternak

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -242,7 +242,8 @@ lazy val frontend: Project = project
     backend,
     backend % "test->test",
     jsonConfig212.jvm,
-    buildpressConfig % "it->compile"
+    buildpressConfig % "it->compile",
+    testAgent
   )
   .disablePlugins(ScriptedPlugin)
   .enablePlugins(BuildInfoPlugin)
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -776,6 +777,20 @@ lazy val nativeBridge04 = project
     fork in Test := true
   )
 
+// Temporaryily publish a fork of sbt test-agent.
+// This version includes https://github.com/sbt/sbt/pull/5800
+// while we wait for the next release of sbt.
+lazy val testAgent = project
```
olafurpg, line 783: Would it be simpler to manually publish the artifact from the `sbt` repo under a different orgzaniation (ch.epfl.scala) instead of creating a new project in the bloop build? When upstream sbt publishes a new version we only have to change the organization back. I can help publish the fork if you like, I got it working from my local computer the other day.

Duhemm: Good point. Moreover, it appears there are more mysteries that I'd have to solve in the build, and it's unlikely we'll have more changes to make to the test agent. Thanks!

PR #1364, opened 2020-09-01
Use forked test agent

Duhemm:
A fix that prevents test suites to be run twice or more was recently
merged to sbt/sbt, but it hasn't been released yet (see
https://github.com/sbt/sbt/pull/5800).

To avoid waiting for the next sbt release, we manually published the
fixed artifact. This commit makes Bloop use this artifact.


Duhemm:
Previous attempt was in #1363, but @olafurpg rightly suggested that it'd be easier to simply publish the new artifact manually, and use it there.

jvican (approved):
LGTM :+1:

PR #1368, opened 2020-09-03
Select last `user.dir` when inferring cwd

Duhemm:
Bloop tries to guess the current working directory to use from the java
options. Before, it'd incorrectly use the first `user.dir` and not look
for further occurrences. It turns out that the JVM will use the last
occurrence of this property instead.


olafurpg (approved):
Nice catch! LGTM 👍

PR #1369, opened 2020-09-05
Handle issues in Gradle artifact resolution

Arthurm1:
This changes the way artifacts (jars) are resolved so that it copes with unresolvable configurations which can pop-up in complex Gradle projects.

This is mainly because I was getting issues with [Gradle Crossbuild Scala Plugin](https://github.com/prokod/gradle-crossbuild-scala)

The strict dependencies on main sourceset are no longer needed because the classpath handling is more comprehensive.

### DIFF
--- a/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
+++ b/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
@@ -11,8 +11,10 @@ import bloop.integrations.gradle.BloopParameters
 import bloop.integrations.gradle.syntax._
 import org.gradle.api.{GradleException, Project}
 import org.gradle.api.artifacts._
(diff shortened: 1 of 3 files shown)


Arthurm1:
I think this failure is unrelated - `bloop.bsp.TcpBspProtocolSpec`

Earlier failures were due to 2.11.  I'd forgotten that the Gradle plugin is released for 2.11 and 2.12.  Is there any need for this?  We could replace `gradle-bloop_2.11` and `gradle-bloop_2.12` with a single `gradle-bloop` and just compile against the latest supported Scala version in Bloop.

tgodzik (approved):
LGTM! Not 100%  familiar with the gradle plugin, but the changes seem fine and we should be good especially with the tests.

tgodzik:
The test failure is not related for sure.

jvican (approved):
LGTM 👍 This indeed looks good, thanks @Arthurm1

Issue #1322, opened 2020-06-19
--add-exports doesn't seem to work


mnd999:
When I try to compile a module with some dubious module usage, I get errors in bloom:

```
markspc% bloop compile --verbose annotations         
[D] Loading workspace settings from bloop.settings.json
[D] Computing sources and classpath hashes for annotations
[D] Scheduling compilation for annotations...
[D] Ignoring analysis for annotations, directory /home/mark/neo4j/neo4j-4.2/public/annotations/target/classes-empty-annotations is missing
[D] Increasing counter for /home/mark/neo4j/neo4j-4.2/public/annotations/target/classes-empty-annotations to 1
[D] External classes directory /home/mark/neo4j/neo4j-4.2/public/annotations/target/bloop-bsp-clients-classes/classes-bloop-cli
[D] Read-only classes directory /home/mark/neo4j/neo4j-4.2/public/annotations/target/classes-empty-annotations
[D] New rw classes directory /home/mark/neo4j/neo4j-4.2/public/annotations/target/bloop-internal-classes/classes-bloop-cli-6br8Q2p_QPGbdSBfk9e1_A==
[D] No previous setup found, invalidating everything.
[D] Full compilation, no sources in previous analysis.
[D] All sources are invalidated.
[D] Initial set of included nodes: 
[D] Recompiling all sources: number of invalidated sources > 50.0% of all sources
Compiling annotations (12 Java sources)
[D] Attempting to call com.sun.tools.javac.api.JavacTool@75554fa9 directly...
[...]

PR #1370, opened 2020-09-07
Support for Java Modules

### LINKED ISSUE #519: Detect java projects in the Maven plugin
Currently, bloops Maven plugin ignores non-scala projects, but it shouldn't.

See also this repo: https://github.com/lefou/bloop-it-maven-non-scala  for a minimal test case.

(This is a follow up to #507)

mnd999:
Currently we only generate config for a maven module if we detect the Scala compiler plugin. With this change, we always generate it, but only add the scala libraries for scala modules. Also try to take into account the compile order from the .pom and any Java compiler args.

I don't think this is perfect by any means (it doesn't compile Neo4j for example due to #1322) , but I think it's better than what we currently have. Might be enough to close #519 though.


jvican (approved):
Hey @mnd999, thanks for the PR! I agree this is not complete but if it's good enough for your use case I'm happy to merge it :+1:

PR #1374, opened 2020-09-16
Print logs in case of bsp connection issues

tgodzik:
Turns out `status.output` was always empty, so we never really got diagnostics about why the connection failed.


jvican (approved):
LGTM 👍

PR #1378, opened 2020-09-27
Support remote debugging

pvid:
A first draft of adding support of remote debugging in bloop.

It seems that a new value for `DebugSessionParamsDataKind ` should be addend to BSP. I'd love your input if this is the way to go and what else needs to be changed or updated.

I have added a "happy path" test so far and I have also done some manual tests using a version of `metals` with necessary changes

test-workspace-1601236890032

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -586,6 +586,9 @@ final class BloopBspServices(
           convert[bsp.ScalaMainClass](main => DebuggeeRunner.forMainClass(projects, main, state))
         case bsp.DebugSessionParamsDataKind.ScalaTestSuites =>
           convert[List[String]](filters => DebuggeeRunner.forTestSuite(projects, filters, state))
+        // TODO incorporate into BSP, maybe?
+        case "scala-attach-remote" =>
+          Right(DebuggeeRunner.forAttachRemote(state))
         case dataKind => Left(JsonRpcResponse.invalidRequest(s"Unsupported data kind: $dataKind"))
       }
     }
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -280,7 +280,8 @@ final class BloopBspServices(
               resourcesProvider = Some(true),
               buildTargetChangedProvider = Some(false),
               jvmTestEnvironmentProvider = Some(true),
-              jvmRunEnvironmentProvider = Some(true)
+              jvmRunEnvironmentProvider = Some(true),
+              canReload = Some(false)
             ),
             None
           )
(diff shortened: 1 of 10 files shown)


frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -586,6 +586,9 @@ final class BloopBspServices(
           convert[bsp.ScalaMainClass](main => DebuggeeRunner.forMainClass(projects, main, state))
         case bsp.DebugSessionParamsDataKind.ScalaTestSuites =>
           convert[List[String]](filters => DebuggeeRunner.forTestSuite(projects, filters, state))
+        // TODO incorporate into BSP, maybe?
+        case "scala-attach-remote" =>
```
tgodzik, line 590: I think that's a good call.

frontend/src/main/scala/bloop/dap/DebugSession.scala
```diff
@@ -178,6 +178,9 @@ final class DebugSession(
         response.command = Command.LAUNCH.getName
         attachedPromise.success(())
         super.sendResponse(response)
+      case "attach" =>
```
marek1840, line 181: Please, add comment that it is a response to an actual "attach" request and not our synthetic one

tgodzik (changes requested):
Thanks for the contribution! This looks really good! I think we should indeed try to add attach to bsp. Let me know if you need help with the PR there.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -586,6 +586,9 @@ final class BloopBspServices(
           convert[bsp.ScalaMainClass](main => DebuggeeRunner.forMainClass(projects, main, state))
         case bsp.DebugSessionParamsDataKind.ScalaTestSuites =>
           convert[List[String]](filters => DebuggeeRunner.forTestSuite(projects, filters, state))
+        // TODO incorporate into BSP, maybe?
```
marek1840, line 589: Agreed, looks like a natural progression for the debug API

marek1840 (approved):
Looks good!

pvid:
> Thanks for the contribution! This looks really good! I think we should indeed try to add attach to bsp. Let me know if you need help with the PR there.

I opened a PR in https://github.com/build-server-protocol/build-server-protocol/pull/139 to add the needed constants there

jvican:
Thanks for the contribution @pvid! Let me know when this PR has caught up with the BSP release and I'll have a look at it.

tgodzik:
I think the changes should we available under the snapshot `2.0.0-M12+45-f04150db-SNAPSHOT`

frontend/src/main/scala/bloop/dap/DebugSession.scala
pvid: I attend some more comment to both `sendResponse` and also `dispatchRequest` function, could you please take a look if it's better?

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -280,7 +280,8 @@ final class BloopBspServices(
               resourcesProvider = Some(true),
               buildTargetChangedProvider = Some(false),
               jvmTestEnvironmentProvider = Some(true),
-              jvmRunEnvironmentProvider = Some(true)
+              jvmRunEnvironmentProvider = Some(true),
+              canReload = Some(false)
```
pvid, line 284: @jvican not sure about right thing to do here - `canReload = Some(false)` or `canReload = None`?

pvid:
> Thanks for the contribution @pvid! Let me know when this PR has caught up with the BSP release and I'll have a look at it.

Hi! Sorry for the longer pause, I have been enjoying some time off. I rebased my branch onto `scalacenter:master` and updated the code to use the latest BSP release

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
tgodzik: I think `false` is correct here, since that issue was introduced for sbt and is not used in Bloop

tgodzik (approved):
LGTM! Thanks for the great contribution!

tgodzik:
@jvican I think it's good to merge, what do you think?

Issue #1362, opened 2020-08-28
Stale zipkin threads increase CPU time


tgodzik:
It seems to happen to quite few people that at some point Bloop start using all available processor power and the only solution is to kill it.

Took a stack trace:

[jstack.txt](https://github.com/scalacenter/bloop/files/5143391/jstack.txt)

what struck me the most is that it looks like there are 932 zipkin threads and I am not doing anything with zipkin at all. Shouldn't it be disabled normally?

Any ideas what is happening?

tgodzik:
It also seems to happen more if I compile Scala 3, although in this case I was working on Metals itself.

jvican:
Hey @tgodzik thanks for the report! I'll give this some priority and investigate what's going on. The zipkin thread detail is a good lead.

ornicar:
Hi and thank you so much for bloop.

I'm using it to develop lichess.org ([source code](https://github.com/ornicar/lila)) and running into this issue all the time. After using bloop for a few minutes, it will start hogging a CPU at 100%.
Killing the bloop process then running `bloop compile lila` immediately gets a CPU indefinitely back to 100%.

Removing all bloop files from the project will get rid of the problem for a few minutes.

Here's what it looks like in the CPU sampler of visualVM:
![](https://i.imgur.com/yaIjXuh.png)

```
java -version
java version "11.0.8.0.2" 2020-07-09 LTS
Java(TM) SE Runtime Environment GraalVM EE 20.2.0 (build 11.0.8.0.2+1-LTS-jvmci-20.2-b03)
Java HotSpot(TM) 64-Bit Server VM GraalVM EE 20.2.0 (build 11.0.8.0.2+1-LTS-jvmci-20.2-b03, mixed mode, sharing)
```
scala: 2.13.3
OS: Linux 5.8.10-arch1-1

If there is anything I could do to help figuring this out, please let me know.

jvican:
Hi @ornicar, thanks for the detailed comment! We've used lichess in our community build to test for compilation correctness and speed in the past, it's great you're using bloop.
  
I'm going to investigate what's going on by first trying to reproduce the issue in my macOS machine. Will get back to you to let you know if I've succeeded. If you could share a yourkit cpu profile with the methods that hog the CPU that will simplify this investigation much further.

jvican:
[...]

PR #1386, opened 2020-10-21
Reuse Zipkin reporter between tasks

### LINKED ISSUE #1385: bloop randomly starts burning CPU when idle
I find that from time to time bloop consumes 25-30% of my CPU even when it's idle and no clients are connected. A thread dump indicated 2066 `AsyncReporter` threads, which seems suspicious (see [dump.txt](https://github.com/scalacenter/bloop/files/5411110/dump.txt)).

Duhemm:
Previously, Bloop would create a new instance of `AsyncReporter` for
every task. Each and every of these reporters would create new threads,
which would never be stopped.

This commit caches the `AsyncReporter`s so that they can be reused and
we keep a single instance per reporting server.

Fixes #1385

cc/ @wiwa

### DIFF
--- a/backend/src/main/scala/bloop/tracing/BraveTracer.scala
+++ b/backend/src/main/scala/bloop/tracing/BraveTracer.scala
@@ -10,6 +10,7 @@ import monix.execution.misc.NonFatal
 import scala.util.Properties
 import zipkin2.codec.SpanBytesEncoder.JSON_V1
 import zipkin2.codec.SpanBytesEncoder.JSON_V2
+import java.util.concurrent.ConcurrentHashMap
 
 final class BraveTracer private (
     tracer: Tracer,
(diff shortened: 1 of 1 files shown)


olafurpg (approved):
LGTM 👍

tgodzik (commented):
LGTM! This should also fix https://github.com/scalacenter/bloop/issues/1362

tgodzik:
@Duhemm it seems that the recent releases fail with:
```
[info] Found a staging repository [chepflscala-2004] status:open, profile:ch.epfl.scala(17f10a3293d844) description: [sbt-sonatype] bloop 1.4.4-23-dbacf644
[info] Dropping staging repository [chepflscala-2004] status:open, profile:ch.epfl.scala(17f10a3293d844) description: [sbt-sonatype] bloop 1.4.4-23-dbacf644
[info] Dropped successfully: chepflscala-2004
[info] Created successfully: chepflscala-2005
Error:  java.io.IOException: Supplied file /home/runner/work/bloop/bloop/target/sonatype-staging/1.4.4-23-dbacf644 is a not an existing directory!
Error:  Use 'last' for the full log.
Error: Process completed with exit code 1.
```

any idea what is going on?

PR #1387, opened 2020-10-22
Support runtime JDK config

Duhemm:
Previously, Bloop would support only a single JDK configuration per
project, which would apply to both compile-time and runtime. However,
some tools (e.g. Pants) support defining a different JDK configuration
for compile-time and runtime.

To be able to export faithfully builds from these tools, Bloop needs to
support this kind of configuration as well. This commit adds a new
field, `runtimeConfig`, to JVM projects' `platform` configuration, which
has the same schema as the existing `config`.

The `config` is used to configure the compile-time configuration,
whereas `runtimeConfig` is used for runtime. When `runtimeConfig`
doesn't exist, Bloop falls back to `config`.

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -715,10 +715,7 @@ final class BloopBspServices(
             fullClasspath = project.fullRuntimeClasspath(dag, state.client).map(_.toBspUri.toString)
             environmentVariables = state.commonOptions.env.toMap
             workingDirectory = project.workingDirectory.toString
-            javaOptions <- project.platform match {
-              case Platform.Jvm(config, _, _, _, _) => Some(config.javaOptions.toList)
-              case _ => None
-            }
+            javaOptions <- project.runtimeJdkConfig.map(_.javaOptions.toList)
           } yield {
             bsp.JvmEnvironmentItem(
               id,
@@ -795,7 +792,8 @@ final class BloopBspServices(
           Task.now(sys.error(s"Failed to run main class in $project due to: ${error.getMessage}"))
         case Right(mainClass) =>
           project.platform match {
-            case Platform.Jvm(config0, _, _, _, _) =>
+            case Platform.Jvm(compileConfig0, _, _, _, _, _) =>
+              val config0 = project.runtimeJdkConfig.getOrElse(compileConfig0)
               val mainArgs = mainClass.arguments.toArray
               val config = JdkConfig(config0.javaHome, config0.javaOptions ++ mainClass.jvmOptions)
               Tasks.runJVM(
@@ -878,7 +876,7 @@ final class BloopBspServices(
     bsp.BuildTargetIdentifier(project.bspUri)
 
   private def toJvmBuildTarget(project: Project): Option[bsp.JvmBuildTarget] = {
-    project.jdkConfig.map { jdk =>
+    project.compileJdkConfig.map { jdk =>
       val javaHome = bsp.Uri(jdk.javaHome.toBspUri)
       bsp.JvmBuildTarget(Some(javaHome), None)
     }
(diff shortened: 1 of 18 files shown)


frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -795,7 +797,8 @@ final class BloopBspServices(
           Task.now(sys.error(s"Failed to run main class in $project due to: ${error.getMessage}"))
         case Right(mainClass) =>
           project.platform match {
-            case Platform.Jvm(config0, _, _, _, _) =>
+            case Platform.Jvm(compileConfig0, _, _, runtimeConfig0, _, _) =>
+              val config0 = runtimeConfig0.getOrElse(compileConfig0)
```
olafurpg, line 801: I think it would be good to move this "runtimeConfig.getOrElse(compileConfig)" logic into a reusable method on `Platform.Jvm` to ensure it's used consistently in all places that need the "runtime config"

frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -186,7 +186,7 @@ final case class Project(
 
   def jdkConfig: Option[JdkConfig] = {
     platform match {
-      case Platform.Jvm(config, _, _, _, _) => Some(config)
+      case Platform.Jvm(config, _, _, _, _, _) => Some(config)
```
olafurpg, line 189: nit: would it make sense to refactor into `case jvm: Platform.Jvm => Some(jvm.config)`?

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -562,7 +562,8 @@ object Interpreter {
                     if (!state.status.isOk) Task.now(state)
                     else Tasks.runNativeOrJs(state, project, cwd, mainClass, args)
                 }
-              case Platform.Jvm(javaEnv, _, _, _, _) =>
+              case Platform.Jvm(compileEnv, _, _, runtimeEnv, _, _) =>
+                val javaEnv = runtimeEnv.getOrElse(compileEnv)
```
olafurpg, line 566: Duplicated logic from comment above.

frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
```diff
@@ -158,7 +158,8 @@ object TestTask {
     import state.logger
     implicit val logContext: DebugFilter = DebugFilter.Test
     project.platform match {
-      case Platform.Jvm(env, _, _, _, _) =>
+      case Platform.Jvm(compileConfig, _, _, runtimeConfig, _, _) =>
+        val env = runtimeConfig.getOrElse(compileConfig)
```
olafurpg, line 162: Duplicated logic.

frontend/src/test/scala/bloop/RunSpec.scala
```diff
@@ -469,4 +469,32 @@ class RunSpec extends BloopHelpers {
...
 
+  @Test
+  def runUsesRuntimeEnvironment(): Unit = {
+    TestUtil.withinWorkspace { workspace =>
+      object Sources {
+        val `a/A.scala` =
+          """/a/A.scala
+            |object A {
+            |  def main(args: Array[String]): Unit = {
+            |    assert("goodbye" == sys.props("test"))
```
olafurpg, line 480: It might be tricky to implement, but how hard would it be to assert that the runtime property is not available at compile-time?

olafurpg (commented):
Only minor code style comments. Otherwise looks great! Thank you for implementing this new feature 😄

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
Duhemm: Absolutely, I forgot that I wanted to do this and rename `jdkConfig`, which is easy to misuse at the moment 👍

frontend/src/test/scala/bloop/RunSpec.scala
Duhemm: It would be difficult to test, because we fork compilation only for Java, and under certain conditions only (ie. the Java home is different from the current Java). I guess we'd need a Java annotation processor that would check for the property?

tgodzik (approved):
LGTM!

PR #1393, opened 2020-11-02
Update Dotty to Scala 3

tgodzik:
Fixes https://github.com/scalacenter/bloop/issues/1394

There was recent change, which moved Dotty back to Scala organisation and renamed artifacts to use `scala3` instead of Dotty.

I see there is also one small issue in the zinc fork: https://github.com/scalacenter/zinc/blob/topic/bloop-rebase-next-2.0/internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala#L123

I have no idea how to properly update the zinc used in Bloop, I don't see any main branch or how things can be released.

### DIFF
--- a/backend/src/main/scala/bloop/ScalaInstance.scala
+++ b/backend/src/main/scala/bloop/ScalaInstance.scala
@@ -42,7 +42,9 @@ final class ScalaInstance private (
   }
 
   /** Is this `ScalaInstance` using Dotty? */
-  def isDotty: Boolean = organization == "ch.epfl.lamp" && version.startsWith("0.")
+  def isDotty: Boolean =
+    (organization == "ch.epfl.lamp" && version.startsWith("0.")) ||
+      (organization == "org.scala-lang" && version.startsWith("3."))
 
   override lazy val loaderLibraryOnly: ClassLoader =
     new URLClassLoader(Array(libraryJar.toURI.toURL), ScalaInstance.bootClassLoader)
@@ -55,7 +57,8 @@ final class ScalaInstance private (
   import ScalaInstance.ScalacCompilerName
   private def isJar(filename: String): Boolean = filename.endsWith(".jar")
   private def hasScalaCompilerName(filename: String): Boolean =
-    filename.startsWith(ScalacCompilerName) || (isDotty && filename.startsWith("dotty-compiler"))
+    filename.startsWith(ScalacCompilerName) ||
+      (isDotty && (filename.startsWith("dotty-compiler") || filename.startsWith("scala3-compiler")))
   private def hasScalaLibraryName(filename: String): Boolean =
     filename.startsWith("scala-library")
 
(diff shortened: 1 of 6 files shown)


tgodzik:
Tests seem largely unstable under grallvm, but it doesn't seem to be caused by particular issue :/

Arthurm1:
I don't think the Gradle part works (and tests fail) because of the change in format of the Dotty version number in Scala 3 (more full stops).  I have a PR for this which is lifted from the SBT dotty plugin which I can submit after this is merged if that's easiest.  I didn't submit this before as I couldn't get Scala3 compiling under Bloop but it looks like you've fixed that.

tgodzik:
> I don't think the Gradle part works (and tests fail) because of the change in format of the Dotty version number in Scala 3 (more full stops). I have a PR for this which is lifted from the SBT dotty plugin which I can submit after this is merged if that's easiest. I didn't submit this before as I couldn't get Scala3 compiling under Bloop but it looks like you've fixed that.

Are the gradle tests not running in CI? I haven't seen them failing there. I will check them out again, but it looked like it was working. For sure I can revert the gradle parts and we can merge your fix later on.

Arthurm1:
I think the build tool tests only run under the Graal tests - which fail earlier and then don't run the build tool tests.

The [Mac](https://github.com/scalacenter/bloop/pull/1393/checks?check_run_id=1342056670) one does fail the build tool tests though:

tgodzik:
> I think the build tool tests only run under the Graal tests - which fail earlier and then don't run the build tool tests.
> 
> The [Mac](https://github.com/scalacenter/bloop/pull/1393/checks?check_run_id=1342056670) one does fail the build tool tests though:

I removed the changes to gradle, I will try to merge this one and we can follow up with your fix.

Alternatively, you can just push the changes to this PR and we can update it all together.

tgodzik:
@Arthurm1 the tests seems to be failing  due to the fact that the endpoint seems to be pointing to `3.0.0-M2`, which doesn't exist. We are getting it from https://dotty.epfl.ch/versions/latest-nightly-base

Arthurm1:
The Ubuntu/GraalVM tests ran successfully and they include the Gradle tests.  Can't see any results for them though - just that the Gradle tests were started.  Is there an easy way to view unit test results from GitHub actions or do you just have to search the logs?

tgodzik:
> The Ubuntu/GraalVM tests ran successfully and they include the Gradle tests. Can't see any results for them though - just that the Gradle tests were started. Is there an easy way to view unit test results from GitHub actions or do you just have to search the logs?

Usually the only way to find any tests is to dig through the logs unfortunately.

Btw. I updated the Scala 3 version in some of the tests to the newly released `3.0.0-M1`.

tgodzik:
@Duhemm This seems to be the last blocker to release Metals with 3.0.0-M1 support. Should we merge it together with https://github.com/scalacenter/bloop/pull/1387 and do a release?

Duhemm (approved):
How exciting! LGTM 👍

PR #1397, opened 2020-11-04
Implement javacOptions endpoint

tpasternak:


Arthurm1:
Hi @tpasternak 

I'm glad you're doing this.  Is the idea to support Java only build targets?  And then is the plan for Intellij to support this or are you aiming for a different client?  Also - what is the status?

Arthurm1:
Hi @tpasternak 

Could you rebase this and convert from "Draft"?  I tested against a Javac client and it looks complete.

tgodzik (approved):
LGTM!

PR #1398, opened 2020-11-05
Fix issues config issues on sbt 0.13

tgodzik:
This should fix issues caused by https://github.com/scalacenter/bloop/pull/1387

It seems circe has an error in which it will print all empty Option values and that is especially problematic with things that should be printed as objects. So currently the generated jsons will contain:
```
        "platform" : {
            "name" : "jvm",
            "config" : {
                "home" : "/usr/lib/jvm/java-8-openjdk-amd64/jre",
                "options" : [
                    "-Duser.dir=/home/tgodzik/Documents/metals/tests/slow/target/e2e/sbt-bloop-import/min-sbt-version"
                ]
            },
            "mainClass" : [
            ],
            "runtimeConfig" : [
            ],
            "classpath" : [
            ],
            "resources" : [
            ]
        },
```

And runtime config cannot be parsed back as it's supposed to be an object and should not really be defined unless it contains `home` and `options`. Tested this locally and it seems to work, however we never run any tests for sbt `0.13...`, which we should most likely start doing.

### DIFF
--- a/config/src/main/scala-2.10/bloop/config/ConfigCodecs.scala
+++ b/config/src/main/scala-2.10/bloop/config/ConfigCodecs.scala
@@ -109,24 +109,24 @@ object ConfigCodecs {
 
   val OptionStringEncoder = implicitly[RootEncoder[Option[String]]]
   val OptionListPathEncoder = implicitly[RootEncoder[Option[List[Path]]]]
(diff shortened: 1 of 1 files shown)


Duhemm (approved):
LGTM, good catch @tgodzik :+1:

PR #1399, opened 2020-11-08
Fix broken URL in $id of bloop-schema.json

dsabanin:
The new URL is apparently `https://scalacenter.github.io/bloop/docs/assets/bloop-schema.json`


tgodzik (approved):
LGTM! Thanks for catching this!

PR #1400, opened 2020-11-11
Documentation: update maven plugin version number

jqno:
It looks like Bloop is no longer published for Scala 2.10, but it published for Scala 2.12 instead. This edit reflects that change.


tgodzik (approved):
LGTM Good catch! I missed that when updating.

PR #1404, opened 2020-11-20
Proposed changes to add troubleshooting section.

dannyhlee:
I had difficulty with my install, and ended up uninstall scoop and bloop before finding the post on gitter: https://gitter.im/scalacenter/bloop?at=5efe02e1bb149531ede5494c

### DIFF
--- a/docs/tools/scoop/install.md
+++ b/docs/tools/scoop/install.md
@@ -22,3 +22,32 @@ scoop install bloop
 
 1. **Java 8 or higher**
 1. **Python 2 or 3** (installation script and CLI client are written in Python)
+
+##### Troubleshooting
+
+If you encounter `coursier.error.ResolutionError$CantDownloadModule: Error downloading` when installing.  
+
+Edit `~\scoop\buckets\bloop-stable\bloop.json` to read:  
+```
+{
+  "version": "1.4.5",
+  "url": "https://github.com/scalacenter/bloop/releases/download/v1.4.5/bloop-coursier.json",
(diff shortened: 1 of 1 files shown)


docs/tools/scoop/install.md
```diff
@@ -21,4 +21,33 @@ scoop install bloop
...
-1. **Python 2 or 3** (installation script and CLI client are written in Python)
\ No newline at end of file
+1. **Python 2 or 3** (installation script and CLI client are written in Python)
+
+##### Troubleshooting
+
+If you encounter `coursier.error.ResolutionError$CantDownloadModule: Error downloading` when installing.  
+
+Edit `~\scoop\buckets\bloop-stable\bloop.json` to read:  
+```
```
tgodzik, line 31:
```suggestion
```json
```

tgodzik (changes requested):
Thanks for the contribution! I wonder what we would need to do to fix this in order to not need the additional information? Or is it possible only with the workaround?

Aside from that just one small  nitpick. If possible, could you also rebase the PR? The CI was broken for a while.

dannyhlee:
When you asked what we might need to not need more information on the page, it turned on a light bulb for me and I went looking at scoop/coursier for a solution.  I ended up at scalacenter/scoop-bloop and this [pull request](https://github.com/scalacenter/scoop-bloop/pull/7).  I commented there and if they resolve the issue with scoop-bloop, then there is no need for these extraneous notes.  Closing.  Thank you for your attention.

PR #1406, opened 2020-11-23
Update setup-scala action to v10 and setup dependabot

tgodzik:
Curently the CI is broken,  the change was similar in the scalameta repositories.


Duhemm (approved):
LGTM, thank you 👍  !

PR #1413, opened 2020-11-25
Fix incompatibility for actions/download-artifact

tgodzik:
Based on https://github.com/actions/download-artifact#compatibility-between-v1-and-v2


mzarnowski (approved):
LGTM

PR #1414, opened 2020-11-25
Fix brew formula for future releases

tgodzik:
Follow up from https://github.com/scalacenter/homebrew-bloop/pull/11

homebrew-bloop is being regenerated with the new releases


mzarnowski (approved):
LGTM

PR #1416, opened 2020-11-30
Bump to latest Zinc

Duhemm:
The new version matches this commit:

  https://github.com/scalacenter/zinc/commit/edbe573e75f7da060f9bd64b12d86fcd59c2c811

This commit fixes an issue where Zinc tries to initialize symbols that
are missing from the classpath. There are legitimate situations where
symbols can be missing from the classpath, such as when Bloop's
configuration files originate from Pants with `strict_deps` enabled, for
instance.

See sbt/zinc#949 and sbt/zinc#950

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -11,7 +11,10 @@ object Dependencies {
   val nailgunVersion = "ee3c4343"
   // Used to download the python client instead of resolving
   val nailgunCommit = "a2520c1e"
-  val zincVersion = "1.3.0-M4+45-d4354be3"
+
+  // Keep in sync in BloopComponentCompiler
+  val zincVersion = "1.3.0-M4+46-edbe573e"
+
   val bspVersion = "2.0.0-M13"
   val javaDebugVersion = "0.21.0+1-7f1080f1"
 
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -11,7 +11,7 @@ object Dependencies {
   val nailgunVersion = "ee3c4343"
   // Used to download the python client instead of resolving
   val nailgunCommit = "a2520c1e"
-  val zincVersion = "1.3.0-M4+45-d4354be3"
+  val zincVersion = "1.3.0-M4+46-edbe573e"
```
tgodzik, line 14: The tests seem no to be able to fetch this version I think? Is it published or maybe not published yet?

Duhemm: I published it just a few minutes before opening this PR. I'll restart the build, as I think it's working fine on my machine 🤔

Duhemm:
I think the failure is spurious:

```
X bloop.bsp.TcpBspProtocolSpec.find test classes 32788ms
```

@tgodzik Can you please take another look?

tgodzik (approved):
LGTM! That failure is 100% spurious, I should probably spend some time fixing the CI at some point.

PR #1417, opened 2020-11-30
Fetch tags when cloning

Duhemm:
Newer versions of the GitHub action `actions/checkout` do not fetch all
tags when cloning, preventing us from computing the current version
number correctly. This can be seen from the most recent releases to
Maven Central, that have as version number `0.0.0-1-<sha>`.

This commit fixes the issue by performing a deep clone.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
       - uses: actions/checkout@v2.3.4
         with:
           submodules: true
+          fetch-depth: 0
       - uses: olafurpg/setup-scala@v10
         with:
           java-version: ${{ matrix.jdk }}
@@ -110,6 +111,7 @@ jobs:
       - uses: actions/checkout@v2.3.4
         with:
           submodules: true
+          fetch-depth: 0
       - uses: olafurpg/setup-scala@v10
         with:
           java-version: ${{ matrix.jdk }}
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
Thanks for fixing this! I just noticed it and we had a similar issue before in Metals, so I should have probably known :sweat:

ckipp01:
If you want I believe you can even make this easier and instead use the built-in way to do it with the checkout action.

https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches

This is what we did with Metals.

https://github.com/scalameta/metals/blob/13d21f776325d8d9dabed80423f540c24f450b4d/.github/workflows/release.yml#L10-L12

tgodzik:
> If you want I believe you can even make this easier and instead use the built-in way to do it with the checkout action.
> 
> https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches
> 
> This is what we did with Metals.
> 
> https://github.com/scalameta/metals/blob/13d21f776325d8d9dabed80423f540c24f450b4d/.github/workflows/release.yml#L10-L12

I think 0 fetches everything, so maybe this is better here?

ckipp01:
> I think 0 fetches everything, so maybe this is better here?
ahh, possibly yea,  I missed the _still shallow_ part.

tgodzik:
> > I think 0 fetches everything, so maybe this is better here?
> 
> ahh, possibly yea, I missed the _still shallow_ part.

Of course, I have no idea shallow will not break at some point :sweat_smile:

Duhemm:
It turns out that this method doesn't work either! (my guess is that even though the tag is there, there's no path from the current commit to the tag?)

I think it's best to stick with the same solution you used in Metals. I'll update the PR!

Thanks for the suggestion @ckipp01 @tgodzik 🙏

PR #1418, opened 2020-12-01
Go back to sbt test-agent artifact

Duhemm:
While waiting for the release of sbt 1.4.0, we published a fork of sbt's
test-agent. Now that sbt 1.4 has been released with the fix that we
needed, we can safely go back to using the mainline artifact.

See #1364
See sbt/sbt#5800, sbt/sbt#5801


tgodzik (approved):
LGTM! The errors look spurious, I will rerun them just in case.

PR #1419, opened 2020-12-02
Add release notes for 1.4.6

Duhemm:


tgodzik (approved):
LGTM!

Duhemm:
All the failures are caused by the Hydra license that is about to expire. I'll move forward with the release and look into this later.

tgodzik:
@TheElectronWill are you able to update the arch package for 1.4.6? I wonder if it's possible to get that step back into the CI in order not to bother you every time? Any idea why it's not there?

TheElectronWill:
> @TheElectronWill are you able to update the arch package for 1.4.6? I wonder if it's possible to get that step back into the CI in order not to bother you every time? Any idea why it's not there?

I've just updated the package. If you have the credentials of the [bloopoid AUR account](https://aur.archlinux.org/account/bloopoid) it should be easy to re-enable the CI. Has Jorge shared them with you? TBH I don't remember well why the AUR publication was disabled at some point.

tgodzik:
Thanks! I don't have those credentials unfortunately. I will try to get them.

tgodzik:
@jvican Do you remember why publishing to AUR was disabled? Also, would be possible to store the credentials somewhere for Scala Center to use?

PR #1423, opened 2020-12-05
Support Scala.js 1.x where x > 0

lolgab:
- Add test for Scala.js 1.x (latest) leaving the 1.0 test to ensure
  we don't break backward compatibility with older Scala.js versions


tgodzik (approved):
LGTM from my side, thanks for the contribution! @sjrd would it be possible for you to confirm?

project/Dependencies.scala
```diff
@@ -46,7 +46,9 @@ object Dependencies {
   val scalaNative03Version = "0.3.9"
   val scalaNative04Version = "0.4.0-M2"
   val scalaJs06Version = "0.6.32"
-  val scalaJs10Version = "1.0.1"
+  val scalaJs1Version = "1.3.1"
+  val scalaJsEnvsVersion = "1.1.1"
+  val scalaJsEnvJsdomNodejsVersion = "1.1.1"
```
sjrd, line 51: This is unused. (And if were used, it should be 1.1.0.)

sjrd (approved):

project/Dependencies.scala
tgodzik: @lolgab could you remove `scalaJsEnvJsdomNodejsVersion` ?

lolgab: Ops, I was adding it and then I went for a string literal, since there was only a single usage.
Removed ✅

tgodzik: Awesome! Thanks!

PR #1428, opened 2020-12-07
Remove wrong user settings and an unsupported dependency for homebrew

tgodzik:


project/ReleaseUtils.scala
```diff
@@ -146,7 +143,7 @@ object ReleaseUtils {
        |      mkdir "channel"
        |
        |      mv "${artifacts.bloopCoursier.name}", "channel/bloop.json"
-       |      system "coursier", "install", "--install-dir", "bin", "--default-channels=false", "--channel", "channel", "bloop", "-J-Divy.home=$ivyHome"
```
tgodzik, line 149: This would add ivy home for gihub runner :O

project/ReleaseUtils.scala
```diff
@@ -124,7 +121,8 @@ object ReleaseUtils {
        |
        |  depends_on "bash-completion"
        |  depends_on "coursier/formulas/coursier"
-       |  depends_on "openjdk@8+"
+       |  depends_on java: "1.8+" if MacOS.version < :big_sur
```
tgodzik, line 124: Taken from https://github.com/coursier/homebrew-formulas/blob/master/coursier.rb

tgodzik: Actually, it was pointed out to me that it's actually wrong.

project/ReleaseUtils.scala
Duhemm: How strange 🤔

Duhemm (approved):
LGTM, thank you 👍

PR #1430, opened 2020-12-10
Bump Scala 2.12.8 to 2.12.12 and remove mill-bloop.

ckipp01:
This pr bumps the Scala 2.12 version up to 2.12.12. This also removes
the mill-bloop plugin since it's been deprecated for a while, mill
provides contrib module for this, and it was also blocking the upgrade
to 2.12.12.

This is the follow-up to this conversation: https://github.com/scalacenter/bloop/discussions/1427#discussioncomment-149047


frontend/src/test/scala/bloop/bsp/BspCompileSpec.scala
```diff
@@ -426,9 +426,9 @@ class BspCompileSpec(
       }
 
       // Change the semanticdb jar every time we upgrade Scala version
-      assert(BuildInfo.scalaVersion == "2.12.8")
+      assert(BuildInfo.scalaVersion == "2.12.12")
       val sourceDir = workspace.resolve("a").resolve("src")
-      val semanticdbJar = unsafeGetResource("semanticdb_2.12.8-4.1.11.jar")
+      val semanticdbJar = unsafeGetResource("semanticdb_2.12.12-4.1.11.jar")
```
tgodzik, line 431: Does it even exist? I think that version might not be supported.

ckipp01: yea exactly, I'm playing around with that now and trying to bump it, but trying to ensure we don't have any references to other versions of semantic db elsewhere. I don't want a mismatch.

ckipp01:
So I'm stuck a bit on the graal stuff. Any hints about what's going on here or how to address this?

tgodzik:
> So I'm stuck a bit on the graal stuff. Any hints about what's going on here or how to address this?

Looks like maybe graal version is wrong? I have no idea about it, but it's saying that some features are not supported, so maybe 2.12.12 started using something new?

tgodzik:
Digging a bit into errors it seems that maybe Scala is using something now that Graal does not yet support? https://github.com/oracle/graal/issues/2761 I wonder how other people worked around - or hopefully the fix works :crossed_fingers:

ckipp01:
> Digging a bit into errors it seems that maybe Scala is using something now that Graal does not yet support? [oracle/graal#2761](https://github.com/oracle/graal/issues/2761) I wonder how other people worked around - or hopefully the fix works 🤞

Yea, https://github.com/scalameta/sbt-native-image defaults to 20.2.0 and works with 2.12.12, so I'm _hoping_ that just bumping that will do the trick. If not I'll just keep digging.

tgodzik:
Looks like the tests are running, so I think this helped!

tgodzik:
> Looks like the tests are running, so I think this helped!

Damn, that was premature :/

ckipp01:
> > Looks like the tests are running, so I think this helped!
> 
> Damn, that was premature :/

Yea :/ It looks like this was hit on here https://github.com/scala/bug/issues/12116 and I'm trying the svm-subs solution to see. Longer term, I'm curious of this would all be easier to manage with using scalameta/sbt-native-image to not have to worry about any of this.

tgodzik:
> > > Looks like the tests are running, so I think this helped!
> > 
> > 
> > Damn, that was premature :/
> 
> Yea :/ It looks like this was hit on here [scala/bug#12116](https://github.com/scala/bug/issues/12116) and I'm trying the svm-subs solution to see. Longer term, I'm curious of this would all be easier to manage with using scalameta/sbt-native-image to not have to worry about any of this.

Might be good to do at some point. I am also thinking of using sbt-ci-release isntead of the current solution. Release seems to fail pretty often, but not yet sure what is the reason for that.

build.sbt
```diff
@@ -323,6 +323,7 @@ lazy val bloopgun: Project = project
       Dependencies.coursierCache,
       Dependencies.jsoniterCore,
       Dependencies.jsoniterMacros % Provided,
+      "org.scalameta" %% "svm-subs" % "20.1.0" % "compile-internal", // Just giving this a try....
```
tgodzik, line 326: Seems to work in Windows only (!). Is the subs versions coupled with graal version? Maybe it should be 20.2.0?

ckipp01: :/ nope, same result.

ckipp01: ah ha! teaches me to just see the three failures and _assume_ it's the same. 🤔

tgodzik: I would try to print the classes that are problematic and then maybe add it to :
` shadeIgnoredNamespaces := Set("com.google.gson", "org.slf4j"),`

?

ckipp01:
Closing this. I'm going to extract parts of this out into individual prs. 

https://github.com/scalacenter/bloop/pull/1465

PR #1432, opened 2020-12-17
Change the semanticdb flag to -Xsemanticdb starting from 3.0.0-M3

tgodzik:
-Ysemanticdb flag should be still avialable, but it does no longer work, which is why we needed to change it now.

### DIFF
--- a/frontend/src/main/scala/bloop/data/Project.scala
+++ b/frontend/src/main/scala/bloop/data/Project.scala
@@ -360,6 +360,12 @@ object Project {
     val workspaceDir = project.workspaceDirectory.getOrElse(configDir.getParent)
     val isDotty = project.scalaInstance.exists(_.isDotty)
 
+    def isAtLeastScala3M3(version: String) = {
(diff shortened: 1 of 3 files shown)


frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -360,6 +360,12 @@ object Project {
     val workspaceDir = project.workspaceDirectory.getOrElse(configDir.getParent)
     val isDotty = project.scalaInstance.exists(_.isDotty)
 
+    def isAfterScala3M3(version: String) = {
```
sjrd, line 363: Perhaps call this
```suggestion
    def isAtLeastScala3M3(version: String) = {
```
?

sjrd (approved):

frontend/src/main/scala/bloop/data/Project.scala
tgodzik: Good idea! Fixed.

PR #1433, opened 2020-12-18
Fix issues with meta builds not resolving sbt artifacts

tgodzik:
It seems that due to https://github.com/sbt/sbt/pull/6085/files we will no longer resolve sbt artifacts when running with coursier, which is the default.
sbt will instead try to use artifacts from the launcher, which does not seem to work in case of plugins.
There is an existing issue https://github.com/sbt/sbt/issues/6203, but in the meanitime we can force resolution using:

```scala
Keys.reresolveSbtArtifacts := true
```


tgodzik:
@ckipp01 This should fix the issue, when testing locally it seems to help. I think the change you found was indeed responsible for this.

ckipp01:
> @ckipp01 This should fix the issue, when testing locally it seems to help. I think the change you found was indeed responsible for this.

Cool, glad that it was actually the issue.

tgodzik:
This is needed for the Metals release so I will go ahead with the merge. Let me know if anyone finds anything worrisome here.

PR #1438, opened 2021-01-05
Don't show message about unsupported Scala version for Scala 3

tgodzik:
Scala 3 in contrast to Scala 2 does not need to download the semanticdb plugin separately, but only to enable the -Xsemanticdb flag.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/BuildLoader.scala
+++ b/frontend/src/main/scala/bloop/engine/BuildLoader.scala
@@ -217,7 +217,8 @@ object BuildLoader {
     // Recognize 2.12.8-abdcddd as supported if 2.12.8 exists in supported versions
     val isUnsupportedVersion = !supportedScalaVersions.exists(scalaVersion.startsWith(_))
     if (isUnsupportedVersion) {
-      logger.debug(Feedback.skippedUnsupportedScalaMetals(scalaVersion))(DebugFilter.All)
+      if (!scalaVersion.startsWith("3."))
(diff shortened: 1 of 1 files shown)


frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -217,7 +217,8 @@ object BuildLoader {
     // Recognize 2.12.8-abdcddd as supported if 2.12.8 exists in supported versions
     val isUnsupportedVersion = !supportedScalaVersions.exists(scalaVersion.startsWith(_))
     if (isUnsupportedVersion) {
-      logger.debug(Feedback.skippedUnsupportedScalaMetals(scalaVersion))(DebugFilter.All)
+      if (!scalaVersion.startsWith("3."))
+        logger.debug(Feedback.skippedUnsupportedScalaMetals(scalaVersion))(DebugFilter.All)
```
Duhemm, line 221: So we still want to enable it in that case? Aha I see that down the line, it is perfectly fine when using Dotty / Scala.

The flag in Bloop is `-Ysemanticdb`, has it been changed to `-Xsemanticdb` in recent versions?

Duhemm (approved):

frontend/src/main/scala/bloop/engine/BuildLoader.scala
tgodzik: It did change, I had to do a quick fix for 3.0.0-M3 :sweat:

PR #1439, opened 2021-01-06
Add ScalaInstance top class loader

adpi2:
Because the Scala 3 compiler depends directly on `org.scala-sbt:compiler-interface`, it must load the `xsbti.*` classes from the same class loader than Bloop. Otherwise we have `LinkageError` exception.

The current Scala 3 compiler bridge hacks its own class loader to avoid `LinkageError` but this solution is fragile and it won't work with Zinc 1.4.x. This PR makes it possible to remove that hack from the compiler: https://github.com/lampepfl/dotty/pull/10960.

This PR introduces the `ScalaInstanceTopLoader`: the parent of the `ScalaInstance.loader` that loads the `xsbti.*` classes from the Bloop loader.

Similar solutions are already used in:
- https://github.com/sbt/sbt/pull/6199
- https://github.com/lihaoyi/mill/pull/1019

### DIFF
--- a/backend/src/main/scala/bloop/ScalaInstance.scala
+++ b/backend/src/main/scala/bloop/ScalaInstance.scala
@@ -47,7 +47,7 @@ final class ScalaInstance private (
       (organization == "org.scala-lang" && version.startsWith("3."))
 
   override lazy val loaderLibraryOnly: ClassLoader =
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM! Tests seem perfectly happy and I see similar solutions in Mill and sbt, so should be ok. Anyone else wants to take a look at it?

PR #1443, opened 2021-01-19
fix for autocompletion carriage-return problem in Windows

philwalk:
I manually tested this in the following environment:

- Windows 10
- CYGWIN_NT-10.0 d5 3.1.7(0.340/5/3) 2020-08-22 17:48 x86_64 Cygwin
- java version "11.0.9" 2020-10-20 LTS
- sbt-launch.jar version 1.4.5

I had to do the following in a pristine sandbox (before launching sbt) to avoid coursier null directory problem:
```
$ touch null
$ sbt
...
sbt:bloop> frontend/runMain bloop.Server
```
Then, in a client session in a bloop-enable project directory:
```
$ bloop abou<TAB>
$ bloop about 
bloop v1.4.6-29-72b258b0-20210119-1411

Using Scala v2.12.8 and Zinc v1.3.0-M4+46-edbe573e
Running on Java JDK v11.0.9 (C:\opt\jdk)
  -> Supports debugging user code, Java Debug Interface (JDI) is available.
Maintained by the Scala Center (Jorge Vicente Cantero, Martin Duhem)
```
Here's what the same experiment result without this fix:
```
$ bloop abou<TAB>
$ bloop about^M
Command not found: about
```


shared/src/main/scala/bloop/io/OsEnv.scala
```diff
@@ -0,0 +1,41 @@
+package bloop.io
+
+object OsEnv {
```
tgodzik, line 3: Should we maybe add the methods to `SystemProperties` or `Environment` object? Probably the latter one would better.

shared/src/main/scala/bloop/io/OsEnv.scala
```diff
@@ -0,0 +1,41 @@
...
+   *    '\r\n' on Windows, unless SHELL is a unix-like shell environment
+   *    '\n'   everywhere else
+   */
+  lazy val EOL: String = eol
+
+  /*
+   * SHELL is defined in CYGWIN, MinGW, msys, Git Bash, etc.
+   */
+  def eol: String = getenvOpt("SHELL") match {
+    case Some(str) if str.matches(".*/(bash|csh|zsh|sh)$") =>
```
tgodzik, line 15: What are potential values this matches? `/` looks like it should be escaped.

Maybe it would be simple instead of a regex to use:
```scala
val possibleShells =  Seq("bash" ,"csh","zsh","sh")
...
case Some(str) if possibleShells.exists(shell => str.endsWith(shell))
```

?

shared/src/main/scala/bloop/io/OsEnv.scala
```diff
@@ -0,0 +1,41 @@
...
+  def getenvOpt(varname: String): Option[String] = Option(System.getenv(varname))
+
+  def lineSplit(str: String): Array[String] =
+    str.split(END_OF_LINE_MATCHER, -1)
+
+  /*
+   * reliably split strings to lines, regardless of where they originated
+   * borrowed from zinc DiagnosticsReporter.scala
+   */
+  def END_OF_LINE_MATCHER = "(\r\n)|[\r]|[\n]"
```
tgodzik, line 40:
```suggestion
  def END_OF_LINE_MATCHER = "\r\n|\r|\n"
```
is it not enough?

tgodzik (changes requested):
Thanks a lot for the fix! I left some comments. Let me know what you think.

shared/src/main/scala/bloop/io/OsEnv.scala
philwalk: Looks good.

shared/src/main/scala/bloop/io/OsEnv.scala
philwalk: I think it needs to be in the shared project, since there are issues in the frontend and the backend.
I tried to add it to frontend bloop.util.Environment, but it's then not visible in the backend.
I maybe should rename it to Environment, though.

shared/src/main/scala/bloop/io/OsEnv.scala
philwalk: I like your approach.   Here's the list of 18 shells in my /etc/shells:

```
/bin/sh
/bin/ash
/bin/bash
/bin/dash
/bin/mksh
/bin/pdksh
/bin/posh
/bin/tcsh
/bin/zsh
/usr/bin/sh
/usr/bin/ash
/usr/bin/bash
/usr/bin/dash
/usr/bin/mksh
/usr/bin/pdksh
/usr/bin/posh
/usr/bin/tcsh
/usr/bin/zsh
```
The common feature is that they all end is "sh", so maybe rather than iterate over a standard list, we can just assume that anytime SHELL is defined and ends with "sh", we can assume standard line endings.

shared/src/main/scala/bloop/io/OsEnv.scala
tgodzik: I haven't noticed it's in shared :sweat: 

Maybe let's just rename it to `OsEnvironment` ?

shared/src/main/scala/bloop/io/OsEnv.scala
tgodzik: I think this should be fine. Unless there is a windows shell that works with the windows line endings?

shared/src/main/scala/bloop/io/OsEnv.scala
philwalk: Agreed (rename to OsEnvironment).   I'm working on this, but only got Intellij working a few minutes ago.

shared/src/main/scala/bloop/io/OsEnv.scala
philwalk: There are multiple workarounds for special cases, based on how the JVM is launched.
```
   unset SHELL                      # revert to default System.lineSeparator
   java -J-Dline.separator=$'\r\n'  # specify required value for lineSeparator
```

philwalk: BTW, the 2nd half of the entries in my `/etc/shells` are identical to the first half, but prefixed with `/usr`.

Since the code looks for an entry in `validShells` that is a substring of the actual `SHELL` setting,  `SHELL=/usr/bin/bash` will be matched with `/bin/bash`.

philwalk:
I'm trying to diagnose why `bloop.bsp.TcpBspProtocolSpec` is failing in the CI environment, but I'm not seeing the failure on my system.


```
sbt:bloop> frontend/testOnly bloop.bsp.TcpBspProtocolSpec
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
-------------------------------- Running Tests --------------------------------
+ bloop.bsp.TcpBspProtocolSpec.check the correct contents of scalac options 2349ms
+ bloop.bsp.TcpBspProtocolSpec.check the correct contents of jvm test environment 77ms
+ bloop.bsp.TcpBspProtocolSpec.check the correct contents of jvm run environment 35ms
+ bloop.bsp.TcpBspProtocolSpec.use client root classes directory and make sure project directories are stable 53ms
+ bloop.bsp.TcpBspProtocolSpec.find main classes 4953ms
+ bloop.bsp.TcpBspProtocolSpec.find test classes 3407ms
+ bloop.bsp.TcpBspProtocolSpec.build targets request works on complicated build 128ms
+ bloop.bsp.TcpBspProtocolSpec.build targets should be empty in build with recursive dependencies 21ms
+ bloop.bsp.TcpBspProtocolSpec.sources request works 125ms
+ bloop.bsp.TcpBspProtocolSpec.sources globs are expanded in sources request 36ms
+ bloop.bsp.TcpBspProtocolSpec.resources request works 101ms
+ bloop.bsp.TcpBspProtocolSpec.dependency sources request works 108ms
[info] utest
[info]
[info] Passed: Total 12, Failed 0, Errors 0, Passed 12
[success] Total time: 14 s, completed Jan 25, 2021, 7:24:29 AM
```

The stackdump for the failure of `test("find test classes")`:
```
-------------------------------- Running Tests --------------------------------
+ bloop.bsp.TcpBspProtocolSpec.check the correct contents of scalac options 43ms  
+ bloop.bsp.TcpBspProtocolSpec.check the correct contents of jvm test environment 59ms  
+ bloop.bsp.TcpBspProtocolSpec.check the correct contents of jvm run environment 21ms  
+ bloop.bsp.TcpBspProtocolSpec.use client root classes directory and make sure project directories are stable 30ms  
+ bloop.bsp.TcpBspProtocolSpec.find main classes 1328ms  
X bloop.bsp.TcpBspProtocolSpec.find test classes 33144ms 
  java.util.concurrent.TimeoutException: Futures timed out after [30 seconds]
  scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:259)
  scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:263)
  monix.execution.CancelableFuture$Implementation.result(CancelableFuture.scala:148)
  scala.concurrent.Await$.$anonfun$result$1(package.scala:219)
  scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:57)
  scala.concurrent.Await$.result(package.scala:146)
  bloop.util.TestUtil$.await(TestUtil.scala:125)
  bloop.util.TestUtil$.await(TestUtil.scala:144)
  bloop.bsp.BspBaseSuite$ManagedBspTestState.compile(BspBaseSuite.scala:183)
  bloop.bsp.BspProtocolSpec.$anonfun$new$37(BspProtocolSpec.scala:281)
  bloop.bsp.BspProtocolSpec.$anonfun$new$37$adapted(BspProtocolSpec.scala:279)
  bloop.bsp.BspBaseSuite.$anonfun$loadBspBuildFromResources$1(BspBaseSuite.scala:531)
  bloop.bsp.BspBaseSuite.$anonfun$loadBspBuildFromResources$1$adapted(BspBaseSuite.scala:529)
  bloop.bsp.BspBaseSuite$UnmanagedBspTestState.withinSession(BspBaseSuite.scala:67)
  bloop.bsp.BspBaseSuite.loadBspBuildFromResources(BspBaseSuite.scala:529)
  bloop.bsp.BspProtocolSpec.$anonfun$new$36(BspProtocolSpec.scala:279)
  bloop.bsp.BspProtocolSpec.$anonfun$new$36$adapted(BspProtocolSpec.scala:277)
  bloop.util.TestUtil$.withinWorkspace(TestUtil.scala:457)
  bloop.bsp.BspProtocolSpec.$anonfun$new$35(BspProtocolSpec.scala:277)
  scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
  bloop.testing.BaseSuite.$anonfun$test$1(BaseSuite.scala:567)
  bloop.testing.BaseSuite.$anonfun$tests$4(BaseSuite.scala:600)
```

Apparently the test-project compile times out after 30 seconds.
On my system the test-project compile succeeds in a few seconds:
```
sbt:cross-test-build-scalajs-0-6> compile
[info] Compiling 5 Scala sources to C:\Users\philwalk\workspace\bloopNewlineFix\frontend\src\test\resources\cross-test-build-scalajs-0.6\test-project\jvm\target\scala-2.11\classes ...
[info] Compiling 4 Scala sources to C:\Users\philwalk\workspace\bloopNewlineFix\frontend\src\test\resources\cross-test-build-scalajs-0.6\test-project\js\target\scala-2.11\classes ...
[info] Non-compiled module 'compiler-bridge_2.11' for Scala 2.11.12. Compiling...
[info]   Compilation completed in 7.335s.
[warn] C:\Users\philwalk\workspace\bloopNewlineFix\frontend\src\test\resources\cross-test-build-scalajs-0.6\test-project\shared\src\main\scala\hello\App.scala:6:9: local val in method main is neve
[warn]     val a = 1 // This is unused and will cause a warning
[warn]         ^
[warn] C:\Users\philwalk\workspace\bloopNewlineFix\frontend\src\test\resources\cross-test-build-scalajs-0.6\test-project\shared\src\main\scala\hello\App.scala:6:9: local val in method main is neve
[warn]     val a = 1 // This is unused and will cause a warning
[warn]         ^
[warn] one warning found
[warn] one warning found
[success] Total time: 9 s, completed Jan 25, 2021, 8:08:33 AM
```

Any ideas how to diagnose this?   Is there a way to `testQuick` in the failing CI environment, or to limit retesting to the `windows-latest` environment?

tgodzik:
@philwalk That test is flaky unfortunately. I need to take a closer look at some point, but that's part of a pretty large backlog :sweat_smile:

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
...
+  def getenvOpt(varname: String): Option[String] = Option(System.getenv(varname))
+
+  /*
+   * The safe general way to split Strings to lines.   Makes minimal assumptions
+   * about the source of the String.
+   *
+   * Does not depend on editor end-of-line configuration.
+   * Correctly splits strings from any of the common OSTYPES.
+   */
+  def lineSplit(str: String): Array[String] = str.split(END_OF_LINE_MATCHER, -1)
```
tgodzik, line 98: I think this one is only used in one place, we can remove it and just use `LineSplitter`

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
...
+   * The default ending in Windows is '\r\n', but either of two conditions can
+   * change the value of [[shellLineSeparator]]:
+   *    1. if [[validShell]] is nonEmpty
+   *    2. if system property `"line.separator"` is modified
+   *
+   * Unlike [[lineSeparator]], only depends on the value of SHELL.
+   *
+   * @return preferred lineSeparator for SHELL environments, or `System.lineSeparator` otherwise.
+   */
+  def shellLineSeparator: String = if (validShell.nonEmpty) "\n" else sysLineSeparator
```
tgodzik, line 25:
```suggestion
 private def shellLineSeparator: String = if (validShell.nonEmpty) "\n" else sysLineSeparator
```
?

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
...
+   * Preferred line ending for SHELL environments.
+   *
+   * An alias for `shellLineSeparator`.
+   *
+   * @example
+   * {{{
+   *   print(someText + EOL)
+   * }}}
+   */
+  lazy val EOL: String = shellLineSeparator // alias for shellLineSeparator
```
tgodzik, line 37:
```suggestion
  lazy val EOL: String = shellLineSeparator
```
it's already in the scaladoc above.

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
...
+  def validShell: String = validShells.find { _.endsWith(shellLC) } match {
+    case Some(sh) => sh
+    case _ => ""
+  }
+
+  /*
+   * Return lowercase shell with possible .exe extension removed.
+   */
+  private[this] def shellLC = shell.toLowerCase(java.util.Locale.ENGLISH) match {
+    case sh if sh.endsWith(".exe") => sh.replaceFirst(".exe$", "")
```
tgodzik, line 58: Is this in case of something like `/bin/bash.exe` ?

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
...
+   * Return the canonical lowercased shell name, or empty string.
+   *
+   * `shell` is valid if when lowercased, it has a `validShells` entry as substring.
+   *
+   * @return the matching `validShells` entry, if there is a match; empty string otherwise.
+   *
+   * @example if {{{SHELL=/usr/bin/bash}}}, returns {{{/bin/bash}}}.
+   *
+   */
+  def validShell: String = validShells.find { _.endsWith(shellLC) } match {
```
tgodzik, line 49:
```suggestion
  private def validShell: String = validShells.find { _.endsWith(shellLC) } match {
```

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
...
+
+  /**
+   * Regular expression that reliably splits Strings into lines.
+   *
+   * Effective for all commonly encountered line endings:
+   *  - "\n"
+   *  - "\r\n"
+   *  - "\r"
+   */
+  def END_OF_LINE_MATCHER = "\n|\r\n|\r"
```
tgodzik, line 121:
```suggestion
  private def END_OF_LINE_MATCHER = "\n|\r\n|\r"
```

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
...
+
+  /*
+   * A list of valid shell paths.  The SHELL environment variable must
+   * have one of these entries as a substring to be considered a valid shell.
+   */
+  lazy val validShells = Seq(
```
tgodzik, line 127:
```suggestion
  private lazy val validShells = Seq(
```

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
...
+    "/bin/pdksh",
+    "/bin/posh",
+    "/bin/tcsh",
+    "/bin/zsh"
+  )
+
+  /*
+   * Returns the empty String unless OSTYPE is defined.
+   */
+  lazy val ostype: String = getenvOpt("OSTYPE").getOrElse("")
```
tgodzik, line 142:
```suggestion
  private lazy val ostype: String = getenvOpt("OSTYPE").getOrElse("")
```

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
...
+
+  /*
+   * Returns validated, lowercase abbreviated OSTYPE value, or empty string.
+   *
+   * Replaced by the entry of [[validOsTypes]] having [[ostype]] as a substring.
+   *
+   */
+  def validOstype: String = {
```
tgodzik, line 150:
```suggestion
  private def validOstype: String = {
```

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
...
+      case os =>
+        validOstypes.find { os.contains(_) }.getOrElse("")
+    }
+  }
+
+  /*
+   * A list of valid ostype substrings.  The lower-case OSTYPE environment
+   * variable must contain one of these substrings to be supported.
+   */
+  lazy val validOstypes = Seq(
```
tgodzik, line 162:
```suggestion
  private lazy val validOstypes = Seq(
```

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,174 @@
+package bloop.io
+
+object Environment {
+
+  /**
+   * Returns "\n" if both OSTYPE and SHELL are nonEmpty, returns `System.lineSeparator` otherwise.
+   * @see [[shellLineSeparator]] which is similar.
+   */
+  def lineSeparator: String = if (validOstype.nonEmpty) shellLineSeparator else System.lineSeparator
```
tgodzik, line 9: This way we are essentially checking both `OSTYPE` and `SHELL` - is both needed to be set?

tgodzik (changes requested):
Thanks for the updates to the rest of the usages of separators. It became a bit more complicated so I left a few questions and comments.

I think you can remove the comments in all the imports (there is enough comments in the class itself) and maybe make some of the fields private, since we only really need `EOL` and `splitLines` .

shared/src/main/scala/bloop/io/Environment.scala
philwalk: The difference is that `EOL` and `lineSeparator` might have different values, depending on whether both `OSTYPE` and `SHELL` are validated. 

My intent was that `EOL` should be `'\n'` in a valid `SHELL` environment, whether `OSTYPE` is valid or not.
That fixes the autocompletion bug,  since `EOL` is used as the line ending of `BloopLogger#printInfo`.

If `OSTYPE` is undefined (or not recognized as a valid target OS), then `lineSeparator` will equal `System.lineSeparator`, so most output line endings, other than `BloopLogger#printInfo` will, be unchanged.

philwalk:
@tgodzik do I click the `[Resolve conversation]` button when I have incorporated your suggestions?  Or is that something you do when you verify the changes?

philwalk:
WRT the merge conflict with `.gitignore`, I'm not sure how to resolve it, please suggest a `git` command line.
I'm happy to revert things to the original version.

Also, perhaps part of the problem is that I accidentally pushed my initial checkin to `origin  https://github.com/philwalk/bloop.git` before switching to the `newline-fix` branch.    I suspect there's a way to undo that error, but my `git` foo is not sufficient for me to feel comfortable doing experiments, since they can trigger builds.

tgodzik:
> @tgodzik do I click the `[Resolve conversation]` button when I have incorporated your suggestions? Or is that something you do when you verify the changes?

It's probably best to apply the suggestions yourself in one commit, otherwise it will trigger CI for each change applied.

philwalk:
I already applied all your suggestions to the code, which I pushed 2 hours ago.
I was referring to the `[Resolve Conversation]` buttons on this page (I'm not familiar with that github feature).

tgodzik:
> I already applied all your suggestions to the code, which I pushed 2 hours ago.
> I was referring to the `[Resolve Conversation]` buttons on this page (I'm not familiar with that github feature).

Ach, sorry! I quickly looked into the PR and not given enough thought to my answer.

I usually resolve anything that doesn't need a discussion, so that it doesn't pollute the PR.

backend/src/test/scala/bloop/logging/RecordingLogger.scala
```diff
@@ -2,6 +2,8 @@ package bloop.logging
 
 import java.io.PrintStream
 import java.util.concurrent.ConcurrentLinkedQueue
+// Environment.lineSeparator == System.lineSeparator, but can be defined by client at JVM startup
```
tgodzik, line 5:
```suggestion

```

backend/src/test/scala/bloop/logging/RecordingLogger.scala
```diff
@@ -2,6 +2,8 @@ package bloop.logging
...
+import bloop.io.Environment.lineSeparator // allows -Dline.separator='\n' at JVM startup
```
tgodzik, line 6:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/it/scala/bloop/CommunityBuild.scala
```diff
@@ -3,6 +3,7 @@ package bloop
 import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Path}
 import java.nio.file.attribute.FileTime
+import bloop.io.Environment.LineSplitter // define String#splitLines
```
tgodzik, line 6:
```suggestion
import bloop.io.Environment.LineSplitter
```

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -9,6 +9,7 @@ import java.nio.file.{FileSystems, Files, Path}
 import java.util.concurrent.ConcurrentHashMap
 import java.util.stream.Collectors
 
+import bloop.io.Environment.lineSeparator // System.lineSeparator unless set at JVM startup
```
tgodzik, line 12:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/main/scala/bloop/dap/DebugSessionLogger.scala
```diff
@@ -3,6 +3,7 @@ package bloop.dap
 import java.net.InetSocketAddress
 
 import bloop.logging.{DebugFilter, Logger}
+import bloop.io.Environment.lineSeparator // System.lineSeparator unless set at JVM startup
```
tgodzik, line 6:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/main/scala/bloop/engine/Feedback.scala
```diff
@@ -5,9 +5,10 @@ import bloop.data.Project
 import bloop.engine.Dag.RecursiveTrace
 import bloop.io.AbsolutePath
 import bloop.util.JavaRuntime
+import bloop.io.Environment.lineSeparator // System.lineSeparator unless set at JVM startup
```
tgodzik, line 8:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -16,6 +16,7 @@ import bloop.data.{ClientInfo, Platform, Project, JdkConfig}
 import bloop.engine.Feedback.XMessageString
 import bloop.engine.tasks.toolchains.{ScalaJsToolchain, ScalaNativeToolchain}
 import bloop.reporter.{LogReporter, ReporterInputs}
+import bloop.io.Environment.lineSeparator // System.lineSeparator unless set at JVM startup
```
tgodzik, line 19:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/test/scala/bloop/DeduplicationSpec.scala
```diff
@@ -2,6 +2,7 @@ package bloop
 
 import bloop.config.Config
 import bloop.io.RelativePath
+import bloop.io.Environment.{lineSeparator, LineSplitter} // reliable line endings
```
tgodzik, line 5:
```suggestion
import bloop.io.Environment.{lineSeparator, LineSplitter}
```

launcher/src/test/scala/bloop/launcher/LauncherBaseSuite.scala
```diff
@@ -2,6 +2,7 @@ package bloop.launcher
 
 import bloop.io.Paths
 import bloop.io.AbsolutePath
+import bloop.io.Environment.{lineSeparator, LineSplitter} // reliable line endings
```
tgodzik, line 5:
```suggestion
import bloop.io.Environment.LineSplitter
```

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,172 @@
...
+   * Preferred line ending for SHELL environments.
+   *
+   * An alias for `shellLineSeparator`.
+   *
+   * @example
+   * {{{
+   *   print(someText + EOL)
+   * }}}
+   */
+  lazy val EOL: String = shellLineSeparator
```
tgodzik, line 37: This is actually only used in one place, so maybe we can just remove it?

shared/src/main/scala/bloop/io/Environment.scala
tgodzik: Maybe just:
```suggestion
  def lineSeparator: String = if (validOstype.nonEmpty && validShell.nonEmpty) shellLineSeparator else sysLineSeparator
```
shellLineSeparator is only used for EOL, which itself is only used in one place.

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,172 @@
...
+    case sh if sh.endsWith(".exe") => sh.replaceFirst(".exe$", "")
+    case sh => sh
+  }
+
+  /*
+   * Return the name of the SHELL environment variable, or empty string.
+   *
+   * Normally set in CYGWIN, MinGW, msys, Git Bash shell environments.
+   */
+  lazy val shell: String = getenvOpt("SHELL").getOrElse("")
```
tgodzik, line 67:
```suggestion
  private lazy val shell: String = getenvOpt("SHELL").getOrElse("")
```

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,172 @@
...
+   * @see [[https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#lineSeparator--]]
+   *
+   * System properties may be redefined on JVM startup, e.g., from a shell environment:
+   * @example
+   *  {{{
+   *  $ java -J-Dline.separator=$'\n' ...
+   *  }}}
+   *
+   */
+  def sysLineSeparator: String = propsOpt("line.separator").getOrElse("\n")
```
tgodzik, line 85:
```suggestion
  private def sysLineSeparator: String = propsOpt("line.separator").getOrElse("\n")
```
we should only use `lineSeparator` if we want to make sure that we address the current issues everwhere.

tgodzik (changes requested):

backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -56,13 +57,13 @@ final class BloopLogger(
...
   private def print(msg: String, fn: String => Unit): Unit = {
-    val lines = msg.split("\\r?\\n", -1)
+    val lines = msg.splitLines // reliable line splitter
     val printFn = if (colorOutput) fn else (str: String) => fn(BloopLogger.stripColors(str))
     lines.foreach(printFn)
   }
 
   private def printInfo(line: String): Unit = {
-    out.println(line)
+    out.print(line + EOL) // end-of-line matches shell environments.
```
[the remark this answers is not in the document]
philwalk: I just pushed changes incorporating all your suggestions.  
I simplified `io.Environment` by removing `EOL`, the tests for `OSTYPE`, and some other simplifications, reducing code by about 50 lines.
I also found some additional references to `System.lineSeparator`, including two places that don't link with `io.Environment`, so I had to add local definitions for `lineSeparator` and the `splitLines` regex.

```diff
diff --git a/frontend/src/test/scala/bloop/bsp/BspClientTest.scala b/frontend/src/test/scala/bloop/bsp/BspClientTest.scala
-                    .replace(System.lineSeparator, " ")
+                    .replace("\n|\r\n|\r", " ")

diff --git a/launcher/src/main/scala/bloop/launcher/package.scala b/launcher/src/main/scala/bloop/launcher/package.scala
-        .split(System.lineSeparator())
+        .split("\n|\r\n|\r", -1)
-        .mkString(System.lineSeparator()),
+        .mkString(lineSeparator),
+
+  /**
+   * SHELL path implies preference for '\n' instead of Windows default.
+   * @return '\n' if SHELL path recognized, system line.separator otherwise.
+   */
+  lazy val lineSeparator = Option(System.getenv("SHELL")) match {
+    case Some(sh) if sh.toLowerCase.matches("/.*/bin/[a-z]*sh(.exe)") => "\n"
+    case _ => System.getProperty("line.separator", "\n")
+  }
```

shared/src/main/scala/bloop/io/Environment.scala
philwalk: I removed it, and also removed the test for `OSTYPE`.

philwalk:
Looking at `DeduplicationSpec.scala` at line 51:
```Scala
      assertNoDiff(
        logger.compilingInfos.mkString(lineSeparator),
        s"""
           |Compiling macros (1 Scala source)
         """.stripMargin
      )
```
Although the test didn't fail at this point (AFAICT), the assertion is a bit fragile, in the sense that the line endings of the literal String being compared depends on the line endings of the test source file.    Since the source file (presumably) has '\n' line separator (regardless of the host OS), perhaps this test should be a line-by-line comparison?

Several similar hidden dependencies exist (there are 38 of them in `DeduplicationSpec`).

For the failed `bloop.DeduplicationSpec`, it seems that there is an `(Access is denied)` error.   Could that be the result of changes in this PR?  Or is this a random thing on Windows, maybe an access race-condition? 

-------------------------------- Running Tests --------------------------------
+ bloop.DeduplicationSpec.three concurrent clients deduplicate compilation 7618ms  
+ bloop.DeduplicationSpec.deduplication removes invalidated class files from all external classes dirs 8645ms  
+ bloop.DeduplicationSpec.deduplication doesn't work if project definition changes 7900ms  
X bloop.DeduplicationSpec.three concurrent clients receive error diagnostics appropriately 9228ms 
  java.io.FileNotFoundException: C:\Users\RUNNER~1\AppData\Local\Temp\bloop-test-**_workspace16405083693998338675\target\b\b-analysis.bin (Access is denied)_**
  java.io.FileOutputStream.open0(FileOutputStream.java:-2)
  java.io.FileOutputStream.open(FileOutputStream.java:298)
  java.io.FileOutputStream.<init>(FileOutputStream.java:237)
  java.io.FileOutputStream.<init>(FileOutputStream.java:187)
  sbt.io.Using$.$anonfun$fileOutputChannel$1(Using.scala:93)
  sbt.io.Using$$anon$2.openImpl(Using.scala:76)
  sbt.io.OpenFile.open(Using.scala:43)
  sbt.io.OpenFile.open$(Using.scala:39)
  sbt.io.Using$$anon$2.open(Using.scala:75)
  sbt.io.Using$$anon$2.open(Using.scala:75)
  sbt.io.Using.apply(Using.scala:21)
  sbt.io.IO$.$anonfun$copyFile$3(IO.scala:714)
  sbt.io.IO$.$anonfun$copyFile$3$adapted(IO.scala:713)
  sbt.io.Using.apply(Using.scala:22)
  sbt.io.IO$.copyFile(IO.scala:713)
  sbt.io.IO$.move(IO.scala:927)
  sbt.internal.inc.FileAnalysisStore$BinaryFileStore.set(FileAnalysisStore.scala:98)
  bloop.Compiler$.$anonfun$persist$1(Compiler.scala:779)
  bloop.Compiler$.$anonfun$persist$1$adapted(Compiler.scala:774)
  bloop.tracing.BraveTracer.traceInternal(BraveTracer.scala:65)
  bloop.tracing.BraveTracer.trace(BraveTracer.scala:39)
  bloop.Compiler$.persist(Compiler.scala:774)
  bloop.Compiler$.$anonfun$compile$14(Compiler.scala:443)
  scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
  monix.eval.internal.TaskRunLoop$.monix$eval$internal$TaskRunLoop$$loop$1(TaskRunLoop.scala:187)
  monix.eval.internal.TaskRunLoop$RestartCallback$1.onSuccess(TaskRunLoop.scala:119)
  monix.eval.Task$.$anonfun$forkedUnit$2(Task.scala:1463)
  java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
  java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
  java.lang.Thread.run(Thread.java:834)
Ignoring failed cancellation with deduplication on Windows
+ bloop.DeduplicationSpec.cancel deduplicated compilation finishes all clients 3766ms  
+ bloop.DeduplicationSpec.cancel deduplication on blocked compilation 8843ms  
+ bloop.DeduplicationSpec.IGNORED - two concurrent clients deduplicate compilation and run at the same time 0ms  
[info] utest
[info] 
Error:  Failed: Total 322, Failed 1, Errors 0, Passed 321
Error:  Failed tests:
Error:  	bloop.DeduplicationSpec
Error:  (frontend / Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
Error:  Total time: 623 s (10:23), completed Jan 27, 2021, 6:30:45 PM
Error: Process completed with exit code 1.
```

File: shared/src/main/scala/bloop/io/Environment.scala [hunk 28]
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Regular expression that reliably splits Strings into lines.
+   *
+   * Effective for all commonly encountered line endings:
+   *  - "\n"
+   *  - "\r\n"
+   *  - "\r"
+   */
+  private def END_OF_LINE_MATCHER = "\n|\r\n|\r"
```
tgodzik, line 100: I wonder if it shouldn't actually be `"\r\n|\n"`, otherwise it will match `\n` first and leave `\r`

shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -0,0 +1,117 @@
...
+   * Return the canonical lowercased shell name, or empty string.
+   *
+   * `shell` is valid if when lowercased, it has a `validShells` entry as substring.
+   *
+   * @return the matching `validShells` entry, if there is a match; empty string otherwise.
+   *
+   * @example if {{{SHELL=/usr/bin/bash}}}, returns {{{/bin/bash}}}.
+   *
+   */
+  private def validShell: String = validShells.find { _.endsWith(shellLC) } match {
```
tgodzik, line 34: If `SHELL` is not defined, shellLC will be `""` and this will always return true. Maybe let's just do:
```suggestion
  val lineSeparator: String = Option(System.getenv("SHELL")) match {
    case Some(currentShell) if validShells.exists( sh => currentShell.contains(sh)) => "\n"
    case None => System.getProperty("line.separator", "\n")
  }
```

I am getting a feeling that it's getting a bit overcomplicated with some many different methods, that are basically being used in only one.

launcher/src/main/scala/bloop/launcher/package.scala
```diff
@@ -10,10 +10,19 @@ package object launcher {
   def printQuoted(msg: String, out: PrintStream): Unit = {
     println(
       msg
-        .split(System.lineSeparator())
+        .split("\n|\r\n|\r", -1)
```
tgodzik, line 13:
```suggestion
        .split("\r\n|\n", -1)
```
?

frontend/src/test/scala/bloop/nailgun/NailgunSpec.scala
```diff
@@ -1,6 +1,7 @@
 package bloop.nailgun
 
 import bloop.io.{AbsolutePath, RelativePath}
+import bloop.io.Environment.{lineSeparator, LineSplitter}
```
tgodzik, line 4:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/test/scala/bloop/io/SourcesGlobsSpec.scala
```diff
@@ -1,5 +1,6 @@
 package bloop.io
 
+import bloop.io.Environment.{lineSeparator, LineSplitter}
```
tgodzik, line 3:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/test/scala/bloop/io/SourceHasherSpec.scala
```diff
@@ -1,5 +1,6 @@
 package bloop.io
 
+import bloop.io.Environment.{lineSeparator, LineSplitter}
```
tgodzik, line 3:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/test/scala/bloop/dap/DebugServerSpec.scala
```diff
@@ -13,6 +13,7 @@ import scala.collection.mutable
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{Promise, TimeoutException}
 import bloop.engine.ExecutionContext
+import bloop.io.Environment.{lineSeparator, LineSplitter}
```
tgodzik, line 16:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
```diff
@@ -6,6 +6,7 @@ import java.net.URI
 import bloop.engine.State
 import bloop.config.Config
 import bloop.io.AbsolutePath
+import bloop.io.Environment.{lineSeparator, LineSplitter}
```
tgodzik, line 9:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/test/scala/bloop/bsp/BspConnectionSpec.scala
```diff
@@ -1,5 +1,6 @@
 package bloop.bsp
 
+import bloop.io.Environment.{lineSeparator, LineSplitter}
```
tgodzik, line 3:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/test/scala/bloop/bsp/BspCompileSpec.scala
```diff
@@ -8,6 +8,7 @@ import bloop.logging.RecordingLogger
 import bloop.internal.build.BuildInfo
 import java.nio.file.attribute.FileTime
 import bloop.io.AbsolutePath
+import bloop.io.Environment.{lineSeparator, LineSplitter}
```
tgodzik, line 11:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/test/scala/bloop/bsp/BspClientTest.scala
```diff
@@ -318,7 +318,7 @@ trait BspClientTest {
                 val report = diagnostics.map(
                   _.copy(source = Some("_"), code = Some("_")).toString
                     .replace("\n", " ")
-                    .replace(System.lineSeparator, " ")
+                    .replace("\n|\r\n|\r", " ")
```
tgodzik, line 321: Best probably to make `END_OF_LINE_MATCHER` public and use it here.

frontend/src/test/scala/bloop/TestSpec.scala
```diff
@@ -2,6 +2,7 @@ package bloop
 
 import bloop.config.Config
 import bloop.io.{AbsolutePath, RelativePath, Paths => BloopPaths}
+import bloop.io.Environment.{lineSeparator, LineSplitter}
```
tgodzik, line 5:
```suggestion
import bloop.io.Environment.lineSeparator
```

frontend/src/test/scala/bloop/GenericTestSpec.scala
```diff
@@ -2,6 +2,7 @@ package bloop
 
 import bloop.cli.Commands
 import bloop.io.AbsolutePath
+import bloop.io.Environment.{lineSeparator, LineSplitter}
```
tgodzik, line 5:
```suggestion
import bloop.io.Environment.lineSeparator
```

backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -56,13 +57,13 @@ final class BloopLogger(
   }
 
   private def print(msg: String, fn: String => Unit): Unit = {
-    val lines = msg.split("\\r?\\n", -1)
+    val lines = msg.splitLines // reliable line splitter
```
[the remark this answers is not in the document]
tgodzik: Let's remove this also.

backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -56,13 +57,13 @@ final class BloopLogger(
...
     val printFn = if (colorOutput) fn else (str: String) => fn(BloopLogger.stripColors(str))
     lines.foreach(printFn)
   }
 
   private def printInfo(line: String): Unit = {
-    out.println(line)
+    out.print(line + lineSeparator) // end-of-line matches shell environments.
```
tgodzik, line 66:
```suggestion
    out.print(line + lineSeparator)
```

tgodzik (changes requested):
Thanks for addressing the feedback! I added some more comments, but I am worried that this PR is getting much more complicated than it might be neccessary. Especially the changes in tests might not be neccessary. Did they fix issues with testing on windows? We already run tests on windows and everything seemed to work fine there.

tgodzik:
Btw. tests are probably flaky, so no need to worry to much about it. I will confirm everything is working in the end.

shared/src/main/scala/bloop/io/Environment.scala
philwalk: > We already run tests on windows and everything seemed to work fine there.

I agree we shouldn't modify tests that are passing with our other changes.

> the changes in tests might not be neccessary

My test environment is intended to be vanilla Windows.   I `undef SHELL` before launching `sbt`, for example.

```Scala
String#split(System.lineSeparator) // needlessly brittle with a high chance of errors on Windows
String += lineSeparator            // sometimes only cosmetic, occasionally can break things
```

Even in test code, it seems like fixing brittle line-splitting is more likely to fix bugs than to introduce them.

In a non-SHELL environment, `System.lineSeparator == Environment.lineSeparator`, so the changes to tests are unlikely to fix test failures.

I agree, we should remove most or all test code changes, since they didn't fix any tests.

philwalk: WRT '\r\n|\n", I came to a similar conclusion.

launcher/src/main/scala/bloop/launcher/package.scala
philwalk: I got rid of the -1 as well, since that possibly generates a trailing empty line, whereas the original "string".split(lineSeparator) did not.

frontend/src/test/scala/bloop/bsp/BspClientTest.scala
philwalk: ok

backend/src/main/scala/bloop/logging/BloopLogger.scala
philwalk: Are you saying we should revert this:
```Scala
  private def print(msg: String, fn: String => Unit): Unit = {
    val lines = msg.splitLines // reliable line splitter
```
to this?
```Scala
  private def print(msg: String, fn: String => Unit): Unit = {
    val lines = msg.split(System.lineSeparator)
```

shared/src/main/scala/bloop/io/Environment.scala
philwalk: Although this means SHELL must not have '.exe' or '.Exe' suffix, although it seems like a good simplification.

frontend/src/test/scala/bloop/bsp/BspClientTest.scala
philwalk: ok

philwalk:
I'm still doing a deep review of why some tests are failing on my system.
It would help if I knew how to install the new launcher script on my system.
The instructions in the `Contributing to bloop` page refer to `frontend/target/install.py`, but I'm not sure which sbt target generates this file (it's not on my system after `publishLocal`.)

```
Generic installation (most common)
If you have installed bloop via the CURL script in the website, you can use a local Python script to override your installation in $HOME/.bloop:

Give executable permission to the install script: chmod +x $BLOOP_REPO/frontend/target/install.py.
Run script with ./$BLOOP_REPO/frontend/target/install.py.
You will roughly see something like this:

Downloading Bloop's coursier version, this may take some seconds...
Installed bloop server in '/home/user/.bloop/blp-server'
Installed bloop client in '/home/user/.bloop/bloop'
HTTP Error 404: Not Found
Couldn't download https://raw.githubusercontent.com/scalacenter/bloop/v1.2.5+464-0159c1c3+20190529-1423/etc/zsh/_bloop, please try again.
```

As an aside, I supplemented `colorsRegex` in `BloopLogger` to cleanup some additional ANSI codes that are polluting the output:

```Scala
  private lazy val colorsRegex = "(\u001b|\u2190)\\[[0-9;]*m|←".r
```
When I run a failing `DeduplicationSpec` (output is cleaned up with new colorsRegex):

```
philwalk@d5:~/workspace/bloopNewlineFix
$ bloop test frontend -o bloop.DeduplicationSpec -- "bloop.DeduplicationSpec.three concurrent clients deduplicate compilation"
[E] Could not resolve platform artifacts: coursier.error.ResolutionError$CantDownloadModule: Error downloading ch.epfl.scala:bloop-js-bridge-1_2.12:1.4.6-39-66e00d62-20210129-0437
[E]   not found: C:\Users\philwalk\.ivy2\local\ch.epfl.scala\bloop-js-bridge-1_2.12\1.4.6-39-66e00d62-20210129-0437\ivys\ivy.xml
[E]   not found: https://repo1.maven.org/maven2/ch/epfl/scala/bloop-js-bridge-1_2.12/1.4.6-39-66e00d62-20210129-0437/bloop-js-bridge-1_2.12-1.4.6-39-66e00d62-20210129-0437.pom
[E]   not found: https://dl.bintray.com/scalacenter/releases/ch/epfl/scala/bloop-js-bridge-1_2.12/1.4.6-39-66e00d62-20210129-0437/bloop-js-bridge-1_2.12-1.4.6-39-66e00d62-20210129-0437.po
m
 Running Tests bloop.DeduplicationSpec.three concurrent clients deduplicate compilation
X bloop.DeduplicationSpec.three concurrent clients deduplicate compilation 15297ms
  java.util.concurrent.TimeoutException: Futures timed out after [2 seconds]

[snipped stack]

Execution took 15s
1 tests, 1 failed
```

The artifact file exists below `~/.ivy2`, btw:
```
C:/Users/philwalk/.ivy2/local/ch.epfl.scala/bloop-js-bridge-1_2.12/1.4.6-39-66e00d62-20210129-0437/ivys/ivy.xml
```
Not sure if this is a red herring, or if I need to install or reconfigure something ...

backend/src/main/scala/bloop/logging/BloopLogger.scala
tgodzik: It's enough to remove the comment, we should leave `msg.splitLines`.

shared/src/main/scala/bloop/io/Environment.scala
tgodzik: Sure, we should leave the changes to the tests if they will help out people running them on windows. I was just wondering aloud.

tgodzik:
> It would help if I knew how to install the new launcher script on my system

I haven't worked with it locally for a while but I think it might be possible to run it via coursier. If do `sbt launcherShaded/publishLocal` and then try to invoke that version with Coursier it should work https://get-coursier.io/docs/cli-launch

But I am guessing here, we need to a update a number of things when it comes to docs, but I haven't had the time unfortunately.

philwalk:
@tgodzik , I think it might be ready to go now.
All tests are now passing on my Windows and Ubuntu servers, with:

- `jdk-11.0.9` 
-  `GraalVM CE 21.0.0 (build 11.0.10+8-jvmci-21.0-b06)`.

All tests also passed here:
[https://github.com/lampepfl/dotty/pull/11180](https://github.com/lampepfl/dotty/pull/11180)

Not sure why some tests are intermittent...

backend/src/main/scala/bloop/logging/BloopLogger.scala
```diff
@@ -147,7 +148,7 @@ object BloopLogger {
     sw.toString()
   }
 
-  private lazy val colorsRegex = "\u001b\\[[0-9;]*m".r
+  private lazy val colorsRegex = "(\u001b|\u2190)\\[[0-9;]*m".r
```
tgodzik, line 151: Why is this change needed?

frontend/src/main/scala/bloop/exec/Forker.scala
```diff
@@ -212,7 +212,10 @@ object Forker {
       @tailrec
       def traverseLines(start: Int): Unit = {
         if (start < msg.length) {
-          val lineEnd = msg.indexOf(System.lineSeparator(), start)
+          val lineEnd = msg.indexOf("\r\n", start) match {
```
tgodzik, line 215:
```suggestion
          val (lineEnd, lineEndLength) = msg.indexOf("\r\n", start) match {
            case -1 => (msg.indexOf("\n", start), 1)
            case idx => (idx, 2)
          }
```
?

frontend/src/test/scala/bloop/bsp/BspClientTest.scala
```diff
@@ -11,6 +11,7 @@ import bloop.engine.ExecutionContext
 import bloop.engine.caches.ResultsCache
 import bloop.internal.build.BuildInfo
 import bloop.io.{AbsolutePath, RelativePath}
+import bloop.io.Environment.{lineSeparator, END_OF_LINE_MATCHER}
```
tgodzik, line 14:
```suggestion
import bloop.io.Environment.END_OF_LINE_MATCHER
```

tgodzik (changes requested):
Just some small nitpicks, but otherwise it seems good to go! Thanks for the great work!

backend/src/main/scala/bloop/logging/BloopLogger.scala
philwalk: I added it to remove unwanted ansi codes, which often contain '\u2190' rather than '\u001b' (Escape).  I wasn't able to figure out where they were coming from.
My command line tool to strip color codes required this regex to remove some of it.
However, perhaps it should be part of a fix for removing unwanted ansi codes, instead of putting it here.

frontend/src/main/scala/bloop/exec/Forker.scala
philwalk: I like this change, although I assume you also intended to remove the recalculation of lineEndLength on line 217.

frontend/src/test/scala/bloop/bsp/BspClientTest.scala
philwalk: looks good.

tgodzik (approved):
Looks good! Thanks a lot for the investigation and detailed changes.

PR #1445, opened 2021-01-22
Update Scala Native to 0.4.0

lolgab:
- Drop support for Scala Native 0.3


tgodzik (approved):
Looks good! Though I have a very limited knowledge of native - changes look ok to me though.

@WojciechMazur could take a quick look at the changes if they make sense?

bridges/scala-native-0.4/src/main/scala/bloop/scalanative/NativeBridge.scala
```diff
@@ -22,35 +25,40 @@ object NativeBridge {
...
 
-    val nativeLogger = NativeLogger(logger.debug _, logger.info _, logger.warn _, logger.error _)
+    val nativeLogger =
+      build.Logger(logger.trace _, logger.debug _, logger.info _, logger.warn _, logger.error _)
     val config = setUpNativeConfig(project, classpath, config0)
     val nativeMode = config.mode match {
-      case LinkerMode.Debug => Mode.debug
-      case LinkerMode.Release => Mode.release
+      case LinkerMode.Debug => build.Mode.debug
+      case LinkerMode.Release => build.Mode.releaseFast
```
WojciechMazur, line 33: What about `build.Mode.releaseFull`? It's a bit aspirational and in most cases `releaseFast` is better, but user probably should still be able to use it.

WojciechMazur (commented):
LGTM, but I'm wondering about limited number of `build.Mode` options. If this limitations comes from bloop itself then it's probably best possible solution

bridges/scala-native-0.4/src/main/scala/bloop/scalanative/NativeBridge.scala
lolgab: Yes, it is shared between Scala.js and Scala Native.
I chose `releaseFast` because it gives good results without taking forever to link.

lolgab: Then I guess this is ready to be merged @tgodzik ?

tgodzik: Sure! I will start working on the release, but most likely it will not happen today - my best estimate is Wednesday.

PR #1446, opened 2021-01-27
Update coursier to 2.0.9

tgodzik:


build.sbt
```diff
@@ -506,7 +507,7 @@ def shadeSbtSettingsForModule(
       }.value
     },
     shadeOwnNamespaces := Set("bloop"),
-    shadeIgnoredNamespaces := Set("com.google.gson"),
+    shadeIgnoredNamespaces := Set("com.google.gson", "scala"),
```
adpi2, line 510: Why is `scala` added here?

tgodzik: Due to the `collection-compat` library that is brought in by coursier. It should be binary compatible and has a nice Mima job, so we should be fine, so we cna ignore it.

adpi2: Ok, but `collection-compat` is not the only artifact that use the `scala` namespace. What about the scala library?

tgodzik: So that did not pop up as being a problem and it should also be binary compatible for any major version.

Besides, we are already doing it for example here: https://github.com/scalacenter/bloop/blob/master/build.sbt#L394

adpi2 (approved):
LGTM!

PR #1447, opened 2021-01-27
Bump Scala, Coursier and remove mill

tgodzik:
Follow up from https://github.com/scalacenter/bloop/pull/1430/files


tgodzik:
ok, that might be too much to update at the same time :sweat_smile:

PR #1454, opened 2021-02-08
Use scala-debug-adapter as a dependency

adpi2:
As orginally discussed [here](https://github.com/scalameta/metals-feature-requests/issues/168), the plan is to extract the Bloop `DebugServer` in a separate repostiory, [scala-debug-adapter](https://github.com/scalacenter/scala-debug-adapter), so that we can reuse it for implementing the `debugSession/start` endpoint in other build tools. We'll start with sbt, but it could also be implemented in Mill.

In this PR I just want to test back that Bloop integrates well with the `scala-debug-adapter` library. It is not meant to be merged until the `scala-debug-adapter` is released.


adpi2:
The scala-debug-adapter v1.0.0 is released. So this PR is ready to be reviewed.

frontend/src/main/scala/bloop/dap/BloopDebuggeeRunner.scala
```diff
@@ -55,76 +71,57 @@ private final class MainClassDebugAdapter(
...
-  def classFilesMappedTo(
-      origin: Path,
-      lines: Array[Int],
-      columns: Array[Int]
-  ): List[Path] = {
-    DebuggeeRunner.classFilesMappedTo(origin, lines, columns, allAnalysis)
+    initialState: State,
+    ioScheduler: Scheduler
+) extends BloopDebuggeeRunner(initialState, ioScheduler) {
+  override def name: String = {
```
tgodzik, line 84: What is the name needed for?

tgodzik (approved):
Sorry for getting so late to it. I thought that I would work on zinc first, but it makes sense to maybe work a bit on fixing some current Bloop issues, so let's merge this one and release a new version before the zinc migration.

frontend/src/main/scala/bloop/dap/BloopDebuggeeRunner.scala
adpi2: The name is used by the `DebugServer` for logging when there is a problem:
- a timeout waiting for the `DebuggeeRunner` to start
- a timeout waiting for the `Terminated` event

PR #1455, opened 2021-02-09
Handle jvmOptions without -J prefix in debugSession/start request

adpi2:
To unblock the issue https://github.com/scalameta/metals/issues/2489 in Metals, Bloop must handle `jvmOptions` without the `-J` prefix in the `debugSession/start` request.

For instance the following `scalaMainClass` is correct and Bloop should be able to execute it:
```
"data": {
    "class": "ch.epfl.scala.index.server.Server",
    "arguments": [],
    "jvmOptions": [
      "-Djna.nosys\u003dtrue"
    ],
    "environmentVariables": []
  }
```

### DIFF
--- a/frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
+++ b/frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
@@ -40,14 +40,18 @@ private final class MainClassDebugAdapter(
 
   def run(debugLogger: DebugSessionLogger): Task[ExitStatus] = {
     val workingDir = state.commonOptions.workingPath
+    // TODO: https://github.com/scalacenter/bloop/issues/1456
+    // Metals used to add the `-J` prefix but it is not needed anymore
+    // So we cautiously strip it off
+    val jvmOptions = mainClass.jvmOptions.map(_.stripPrefix("-J"))
     val runState = Tasks.runJVM(
       state.copy(logger = debugLogger),
       project,
       env,
       workingDir,
       mainClass.`class`,
-      (mainClass.arguments ++ mainClass.jvmOptions).toArray,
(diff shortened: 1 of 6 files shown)


frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
```diff
@@ -40,14 +40,17 @@ private final class MainClassDebugAdapter(
 
   def run(debugLogger: DebugSessionLogger): Task[ExitStatus] = {
     val workingDir = state.commonOptions.workingPath
+    // Metals used to add the `-J` prefix but it is not needed anymore
+    // So we cautiously strip it off
```
tgodzik, line 44: Could we add TODO with an issue to follow up? I think once we release Metals with the next Bloop version we should be good to remove it.  Created the issue here: https://github.com/scalacenter/bloop/issues/1456 

Could you also add it to any other place that we might want to remove the workaround from?

tgodzik (changes requested):
Thanks a lot @adpi2 ! Could we just add a TODO so that we don't forget to remove it in the future?

frontend/src/main/scala/bloop/dap/DebuggeeRunner.scala
adpi2: > Could you also add it to any other place that we might want to remove the workaround from?
What other places do you have in mind?

adpi2: I think the other places you are mentioning are not workarounds. In their cases the options are coming from the Bloop command so the `-J` prefix is expected.

tgodzik (approved):
LGTM!

Issue #1460, opened 2021-02-18
Unexpected Error on MacOS


ckipp01:
With a fresh install of Bloop via `cs install bloop`, I now get the following when attempting to run any command:

```
❯ bloop about
No server running at 127.0.0.1:8212, let's fire one...
Resolving ch.epfl.scala:bloop-frontend_2.12:1.4.7...
Starting bloop server at 127.0.0.1:8212...
Attempting a connection to the server...
Attempting a connection to the server...
Attempting a connection to the server...
Attempting a connection to the server...
error: Unexpected error forces client exit!
java.io.EOFException
        at java.io.DataInputStream.readInt(DataInputStream.java:397)
        at snailgun.protocol.Protocol.$anonfun$processChunkFromServer$1(Protocol.scala:181)
        at scala.util.Try$.apply(Try.scala:213)
        at snailgun.protocol.Protocol.processChunkFromServer(Protocol.scala:180)
        at snailgun.protocol.Protocol.sendCommand(Protocol.scala:108)
        at snailgun.TcpClient.run(TcpClient.scala:34)
        at bloop.bloopgun.BloopgunCli.executeCmd$1(Bloopgun.scala:268)
        at bloop.bloopgun.BloopgunCli.fireCommand(Bloopgun.scala:286)
        at bloop.bloopgun.BloopgunCli.run(Bloopgun.scala:230)
        at bloop.bloopgun.Bloopgun$.main(Bloopgun.scala:638)
        at bloop.bloopgun.Bloopgun.main(Bloopgun.scala)
```

Going back in the versions this started for me with `1.4.6-38-2f6fdcba`.

OS: MacOS
Install Method: Coursier
Java Version: openjdk version "11.0.9.1"

Creating this as a follow-up to my comment in here: https://github.com/scalameta/metals/pull/2524

tgodzik:
It seems that the issue might be coming from https://github.com/scalacenter/bloop/pull/1443/files :thinking:

PR #1461, opened 2021-02-18
Fix issue when unknown SHELL was defined

tgodzik:
Previously line separator would be a non-exhaustive match and could throw an exception if an unknown shell was found. Now it defaults to `System.getProperty("line.separator", "\n")`

Fixes https://github.com/scalacenter/bloop/issues/1460


shared/src/main/scala/bloop/io/Environment.scala
```diff
@@ -67,6 +67,7 @@ object Environment {
     "/bin/pdksh",
     "/bin/posh",
     "/bin/tcsh",
-    "/bin/zsh"
+    "/bin/zsh",
+    "/bin/fish"
```
ckipp01, line 71: 🐟

ckipp01 (approved):
LGTM!

PR #1464, opened 2021-02-21
Add Gradle Android plugin support

Arthurm1:
This is an attempt at supporting the Gradle Android plugin.

The Android plugin appears to use its own way of managing sourcesets/variants/flavors etc. so there's not huge amounts of commonality between it and the Java/Scala plugins that can be exploited.  Instead I've duplicated a lot of code from the current handling of Java/Scala plugin.  I think this is actually a good thing as just in the few Android SDK versions I've tested, the API has changed and I think it'll be easier to support if the code paths are separate.

There are a few caveats here...
1) Android uses VariantTransforms to handle dependencies on java only (i.e. none Android) Gradle projects.  These transforms have to be applied when Gradle creates the classpath and the transforms require the project's output jar file to exist.  To be clear - this is only to do with dependencies from Android projects onto java-only projects (i.e. where you have the source present), not general maven dependencies i.e. junit.  Because of this, any java-only projects that are a part of the Android projects must be assembled first using Gradle `./gradlew nonAndroidProjectName::assemble` or `BloopInstall` will throw an exception.  Once they are assembled, the jar will exist, bloopInstall can be run and then Bloop will happily ignore this jar and manage compilation across all the projects: android and java-only.  I'm still looking into whether there is some way round this as I don't think assembling should be required to query Gradle for the classpath.

2) Android manages UI layout/setup using XML files stored in the project's `res` sub directory.  It "compiles" these XML files into java class files using the [AAPT2](https://developer.android.com/studio/command-line/aapt2) tool and packages them into an `R.jar`.  This allows these UI resources to be accessed as java code by the rest of the project.  I don't see a way round this and I imagine most Android projects contain UI.  It means that `./gradlew assemble` has to be run on all Android projects containing Android resources before Bloop can use the `R.jar`.  The export to Bloop will work fine but compilation will fail if the project has UI and `./gradlew assemble` has not been run.  Then whenever these resource files are changed, the user has to run `./gradlew assemble` to update the `R.jar` and be able to use them in Metals/Bloop.  I guess ideally Bloop would would directly support AAPT2 which I think is just a question of watching the `res` dir for `xml` files and calling AAPT2 on the file that's changed.

3) A BuildConfig record is often defined within the project's `build.gradle` file. e.g.
```groovy
    defaultConfig {
        buildConfigField "long", "TIMESTAMP", System.currentTimeMillis() + "L"
    }
```
Gradle will compile this into a `build/generated/source/buildConfig/variant` directory as `BuildConfig.java`.  This class is then accessible from source code (e.g. `BuildConfig.TIMESTAMP` is available from java code).  Again `./gradlew assemble` is required to create `BuildConfig.java` initially and then Bloop will happily make use of it.  I don't see a way round this.

4) All variants are exported - release & debug.  I'm not sure if this is wanted?

5) I've only tested on v4.0.0 and v4.1.0.

Because of points 1, 2 and 3, I'm not sure how useful this PR is at this stage.  If (2) could be solved then it would make this more viable.  I don't know if Bloop actually needs to support AAPT2 - it might be possible to use something like the [filewatcher](https://marketplace.visualstudio.com/items?itemName=appulate.filewatcher) VSCode extension to call `./gradlew assemble` or `AAPT2` on file change.

### DIFF
--- a/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
+++ b/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
@@ -11,16 +11,19 @@ import bloop.integrations.gradle.BloopParameters
 import bloop.integrations.gradle.syntax._
 import bloop.integrations.gradle.tasks.PluginUtils
 import org.gradle.api.{Action, GradleException, Project}
-import org.gradle.api.artifacts._
+import org.gradle.api.artifacts.{Configuration, PublishArtifact}
 import org.gradle.api.artifacts.ArtifactView.ViewConfiguration
+import org.gradle.api.artifacts.dsl.DependencyHandler
 import org.gradle.api.artifacts.result.{ComponentArtifactsResult, ResolvedArtifactResult}
 import org.gradle.api.attributes.Attribute
 import org.gradle.api.attributes.AttributeContainer
 import org.gradle.api.component.Artifact
 import org.gradle.api.file.FileCollection
+import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.internal.artifacts.dsl.LazyPublishArtifact
 import org.gradle.api.internal.artifacts.publish.{ArchivePublishArtifact, DecoratingPublishArtifact}
 import org.gradle.api.internal.file.copy.DefaultCopySpec
+import org.gradle.api.internal.file.CompositeFileCollection
 import org.gradle.api.internal.tasks.compile.{DefaultJavaCompileSpec, JavaCompilerArgumentsBuilder}
 import org.gradle.api.tasks.bundling.AbstractArchiveTask
 import org.gradle.api.tasks.{AbstractCopyTask, SourceSet}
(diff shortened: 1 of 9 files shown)


tgodzik:
I think points 1, 2, 3 are actually connected to a larger issue at hand - Bloop handles compilation, but it doesn't really handle and custom things like source generation etc. And an additional complication here is that these functionalities might be added by some plugins we've never even heard about.

We solve it in sbt by making sure that all those additional artifact generating actions are run together with `bloopInstall`, which means that we need to reimport the build on some of the changes to the build configuration. The options we have are:
1. Make Gradle make classpath querying depend on `assemble` - this might be a short term solution.
2. Add a new mechanism for Bloop to easily add new integrations, that will regenerate any artifacts needed.
3. Add BSP support for Gradle.

I would opt for 2. (not neccesarily in this PR), but we would need to work out some easily extensible solution and I am not sure if it's possible to do it correctly. Maybe something like additional config:
```json
  watching:
   [
     {toolPath : "", command: "", glob: "" }
   ]
```
And we could just add new integrations per build tool.

I think that even with the need for having to `assemble` project previously, the PR is pretty useful. I will try to take a closer look at it next week.

pierzchalski:
Hi, this is a feature which I'm interested in. What can I do to help get this merged?

Arthurm1:
@pierzchalski  We could merge this - I'm just not sure how useful it is considering the points made above.

I think using Bloop+Metals would give a bad experience without source code generation and with the issues on the Gradle export.

Also, currently, there is no specific Scala support.  How do you use Scala code in your project - do you use [gradle-android-scala-plugin](https://github.com/saturday06/gradle-android-scala-plugin) or a different Android/Scala Gradle plugin or use separate non-Android Scala projects and import them?

pierzchalski:
I'm already used to `assemble`ing more than I should, so if that was all it took for Bloop+Metals to get me Intellisense then I'd be pretty happy :smiley: 

Re. Scala: my workplace has their own Scala/Android plugin which I think was largely inspired by the one you linked. I don't know enough about how either plugin works to know what would need to be added on top of this PR for Scala support.

Arthurm1:
Random test failure: `LatestStableLauncherSpec` only on GraalVM

@tgodzik Guess this could be merged - see what the feedback is and gradually build up the Android support.

tgodzik:
> Random test failure: `LatestStableLauncherSpec` only on GraalVM
> 
> @tgodzik Guess this could be merged - see what the feedback is and gradually build up the Android support.

Cool! I will take a look later this week.

tgodzik:
Haven't managed to take a look this week, sorry about that. I will get back to it next week after short holidays. Sorry about that!

project/BuildPlugin.scala
```diff
@@ -533,12 +533,16 @@ object BuildImplementation {
       sbtbuildinfo.BuildInfoPlugin.buildInfoScopedSettings(Test) ++ List(
         Keys.fork in Test := true,
         Keys.resolvers ++= List(
-          MavenRepository("Gradle releases", "https://repo.gradle.org/gradle/libs-releases-local/")
+          MavenRepository("Gradle releases", "https://repo.gradle.org/gradle/libs-releases-local/"),
+          MavenRepository("Android plugin", "https://maven.google.com/"),
+          MavenRepository("trove4j dependency", "https://jcenter.bintray.com/"),
```
tgodzik, line 538: I don't think `bintray` is working still, so this might not work. Also looks like there is nothing about `trove4j` in the code. Maybe leftover from previous approach?

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
```diff
@@ -399,6 +613,152 @@ class BloopConverter(parameters: BloopParameters) {
...
+            val a = getOutputJarMethod.invoke(task)
+            val outputJar = a match {
+              case f: File => Set(f)
+              case f: RegularFileProperty => Option(f.getOrNull).map(_.getAsFile).toSet
+            }
+            classes ++ outputJar
+          })
+
+        // uses reflection to support multiple API versions...
+        val libs =
```
tgodzik, line 677: Are we able to query the exact API version and base the implementation on that fact (use exact method etc.) instead of using an exception?

tgodzik: Something like:
```scala
val bundleLibraryClass = if (SemVer.isBefore(apiVersion, "4.1.0")) Class
                .forName("com.android.build.gradle.internal.tasks.BundleLibraryClasses")
                .asInstanceOf[Class[Task]]
else Class
              .forName("com.android.build.gradle.internal.tasks.BundleLibraryClassesJar")
              .asInstanceOf[Class[Task]]
```

Are we even able to get that information and does it make sens to try it?

It's really hard to read it otherwise.

integrations/gradle-bloop/src/test/scala/bloop/integrations/gradle/ConfigGenerationSuite.scala
```diff
@@ -47,6 +47,234 @@ abstract class ConfigGenerationSuite extends BaseConfigSuite {
...
+      val buildFileB = new File(buildDirB, "build.gradle")
+      val buildFileC = new File(buildDirC, "build.gradle")
+
+      writeBuildScript(
+        buildFileA,
+        s"""
+           |buildscript {
+           |  repositories {
+           |    google()
+           |    jcenter()
```
tgodzik, line 77:
```suggestion
           | 
```
Is the repo still available? Probably best to remove it.

tgodzik (changes requested):
Thanks for the contribution! It looks good! I added some smaller comments and if we could improve the reflection a bit, it would be awesome, but I do understand if that is not sensible to do.

This a pretty awesome first step and it's a pity we will still need to run gradle assemble side by side, but we can think about making it work better later on.

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
Arthurm1: I can see an implementation of `SemVer` in `metals/mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala`

Is there a common library with that class in?  I guess I can just duplicate it for now.

tgodzik: I haven't seen anything, only in sbt, where it can be used from plugins.

integrations/gradle-bloop/src/test/scala/bloop/integrations/gradle/ConfigGenerationSuite.scala
Arthurm1: Removed

integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
Arthurm1: I've copied it and I've simplified the reflection based on version

project/BuildPlugin.scala
Arthurm1: Trove4J is a transitive dependency required by the android plugin but it's now on Maven Central so I've removed

Arthurm1:
When you review - is it easier for you to have more commits after changes are requested or do you prefer it kept to a single commit and force pushed?  In this case I've added a second commit rather than squashing into previous

tgodzik:
> When you review - is it easier for you to have more commits after changes are requested or do you prefer it kept to a single commit and force pushed? In this case I've added a second commit rather than squashing into previous

It's fine anyway, thanks!

tgodzik (approved):
LGTM! Thanks a lot for the fixes!

PR #1465, opened 2021-02-21
Remove old mill-bloop.

ckipp01:
This is no longer being used anymore and also the old version of mill
that this is relying on isn't published for the newer Scala versions
we'd probably like to target anyways.

I originally included this in #1430 but since I was stuck on other stuff
this never crossed the finish line so I extracted it out.


tgodzik (approved):
LGTM!

Issue #1463, opened 2021-02-19
No longer able to run with Coursier launch - regression from 1.4.6 to 1.4.7


przemek-pokrywka:
Hi,
```shell
cs launch ch.epfl.scala:bloopgun_2.12:1.4.6 -- about
```
runs fine, but
```shell
cs launch ch.epfl.scala:bloopgun_2.12:1.4.7 -- about
```
fails with error:
```
No server running at 127.0.0.1:8212, let's fire one...
Resolving ch.epfl.scala:bloop-frontend_2.12:1.4.7...
Exception in thread "main" bloop.shaded.coursier.util.Task$WrappedException: java.lang.NoClassDefFoundError: scala/collection/compat/immutable/LazyList$
	at bloop.shaded.coursier.util.Task$.wrap(Task.scala:88)
	at bloop.shaded.coursier.util.Task$.$anonfun$flatMap$2(Task.scala:14)
	at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:307)
	at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:41)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
	at java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1402)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:175)
Caused by: java.lang.NoClassDefFoundError: scala/collection/compat/immutable/LazyList$
	at bloop.shaded.coursier.core.ResolutionProcess$.uniqueModules$1(ResolutionProcess.scala:416)
	at bloop.shaded.coursier.core.ResolutionProcess$.fetchAll(ResolutionProcess.scala:423)
	at bloop.shaded.coursier.core.ResolutionProcess.run(ResolutionProcess.scala:28)
	at bloop.shaded.coursier.Resolve$.runProcess(Resolve.scala:303)
	at bloop.shaded.coursier.Resolve.run$1(Resolve.scala:110)
	at bloop.shaded.coursier.Resolve.ioWithConflicts0(Resolve.scala:160)
[...]

PR #1468, opened 2021-02-25
Fix issues with scala-collection-compat missing in bloopgun

tgodzik:
Fixes https://github.com/scalacenter/bloop/issues/1463

scala-collection-compat is brought in by coursier and since `scala` is added to be ingored when shading, it never ends up properly on the classpath of bloopgun. This fixes the issues by adding the dependency directly. It's binary compatible, so it should not be an issue in case of conflicts.


adpi2 (approved):
LGTM

PR #1471, opened 2021-03-01
Integrations: Gradle: Fix: Ignore non bloop capable projects when getting all projects

shasts:
This is a proposed fix to address the issue in https://github.com/scalacenter/bloop/issues/1472

### Changes
1. Extract and reuse existing function `canRunBloop` to determine if a Gradle sub-project is a bloop capable project or not. 
2. Change all occurrences of  `getAllprojects`  with the newly introduced function.
3. Minor change to replace `map` and then `flatten` with `flatMap`
4. Test


tgodzik (commented):
Thanks for the contribution! The change looks good, although it would be awesome to hear some feedback from @Arthurm1 - he does a lot of the work around here.

Arthurm1:
Makes sense and LGTM - thanks @shasts

tgodzik (approved):

PR #1476, opened 2021-03-07
Gradle fixes

Arthurm1:
2 fixes...

1) javac options list was getting corrupted causing issues with annotation processing.

2) If project A depends on project B then project B's classes dir would be on project A's classpath but project B's resources dirs would **not** be included on project A's classpath when they should be.


tgodzik (approved):
LGTM! Thanks for the fixes!

Issue #1242, opened 2020-04-11
Allow hot class reloading with DCEVM


PhilAndrew:
I would like to have hot class reloading with DCEVM. http://dcevm.github.io/
Bloop is used with the Seed Scala build tool. In this issue is described the problem found with attempting to get DCEVM to work. https://github.com/tindzk/seed/issues/76

The main problem as described is boop writing class files to different directories, as it says.

The first command wrote the class files to
/tmp/build-dcevm/bloop/bloop-internal-classes/example-6P2vxtl9QjGGuCgBFSQ8wg==, whereas the second one used a different path:
/tmp/build-dcevm/bloop/bloop-bsp-clients-classes/example-bloop-21r1tSAdRkONeSDCaa_UqA==/

If these problems could be addressed then hot class reloading could work potentially.

jvican:
Thanks for opening the ticket @PhilAndrew.

Bloop has complete control over how class files are managed internally. For example, you have referenced some of those internal directories on your ticket. But Bloop still keeps the invariant that class files are written to a stable directory for the duration.

So, Bloop will write the updated class files to the external directory of your client. If you're using the CLI and there's only one CLI client, that is usually `.bloop/project-foo/bloop-bsp-clients-classes/classes-bloop-cli/` which is stable across runs. When you use seed, seed connects to bloop via BSP, so it will write the class files to a stable directory that is created when the seed client connects and deleted when it exits.
  
I don't know the specifics of how seed manages its bloop creation but integrating with DCEVM should work. The internal BSP classes dirs are private to bloop and shouldn't be used by external tools. /cc @tindzk

tindzk:
Presently, there are two limitations in Bloop which make DCEVM support slightly inconvenient:

1. The BSP protocol does not allow users to specify environment variables. This would be useful to selectively run a module with the DCEVM agent via `JAVA_TOOL_OPTIONS` (see example below).

2. Seed does not support running commands in parallel over a single BSP connection yet (as discussed [here](https://github.com/tindzk/seed/issues/72#issuecomment-567012714)). For now, it should be possible to use two terminals:

```bash
# Terminal 1
$ seed build --watch example  # uses BSP connection

# Terminal 2
$ seed run example  # uses BSP connection, but requires hard-coding the JVM options in the JSON configuration file
# or
$ JAVA_TOOL_OPTIONS="-XXaltjvm=dcevm -javaagent:/path/to/hotswap-agent.jar" bloop run example
```

[...]

Issue #627, opened 2018-08-16
Speed up test execution (hot classloading)


jvican:
I've noticed that the creation of the single-threaded pool to run tests takes around 1 second or sometimes slightly more in my laptop. This cost should go away. Bloop should cache the test loader across different test runs. The current cost also happens in tools like sbt.

tindzk:
On a similar note, I observed that small cross-compiling test suites actually run up to twice as fast with Node (#572), but for larger test suites, JVM performs better and its start-up time becomes barely noticeable.

Would it be possible to run tests without forking?

jvican:
I think having an open-source version of jrebel would be cool, but the amount of effort it would take would be staggering. I think the ideas collected here are great, but someone else would need to act on them (outside the scope of bloop, probably, and with lots of tests to make sure it works). On top of that, jrebel has some issue with regards to initialization (if a new field to a class is added, the field will not be initialized in objects that existed in the stack previous to this change), so it remains to be seen how much productivity it can be gained from jrebel after all.
  
I'll look into our test framework execution and see if we can make it faster to start up in another ticket.

PR #1480, opened 2021-03-21
Hot Swap Compilation

danilbykov:
I would like to use [dcevm](http://dcevm.github.io/) with bloop in order to reload classes without restarting application.

I run my application using command `bloop run <project>`. As I understand this command starts application with classpath `.bloop/root/bloop-bsp-clients-classes/classes-bloop-cli`. But once application started there is no way to update classes in this directory. So dcevm can not discover new class files and update application.

I added new option for hot swap compilation which starts application as well as starts watcher with recompilation. What do you think about that change? Is there some way to achieve this behavior using existing options?

Using new option I start application with following command
```
bloop run root --main main.Class --hot-swap-compilation=true -- -J-XX:HotswapAgent=core -J-javaagent:$JAVA_HOME/lib/hotswap/hotswap-agent.jar=autoHotswap=true
```
and it sees new class files without restart.

### DIFF
--- a/frontend/src/main/scala/bloop/cli/Commands.scala
+++ b/frontend/src/main/scala/bloop/cli/Commands.scala
@@ -193,6 +193,9 @@ object Commands {
       @ExtraName("w")
       @HelpMessage("If set, run the command whenever projects' source files change.")
       watch: Boolean = false,
+      @ExtraName("H")
+      @HelpMessage("Hot Swap Compilation")
+      hotSwapCompilation: Boolean = false,
       @HelpMessage("Ignore arguments starting with `-J` and forward them instead.")
       skipJargs: Boolean = false,
       @ExtraName("O")
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -596,7 +596,16 @@ object Interpreter {
         }
       }
 
-      if (cmd.watch) watch(projects, state)(runTask) else runTask(state)
+      if (cmd.hotSwapCompilation) {
+        val compilationProcess =
+          watch(projects, state)(runCompile(cmd, _, projects, false, cmd.cliOptions.noColor))
+            .runAsync(ExecutionContext.ioScheduler)
+        runTask(state).doOnFinish(_ => Task.delay(compilationProcess.cancel()))
+      } else if (cmd.watch) {
+        watch(projects, state)(runTask)
+      } else {
+        runTask(state)
+      }
     }
   }
 


tgodzik:
Thanks for the contribution! This looks pretty good, though would it be possible to maybe add a test for it?

jvican:
This won't work, see https://github.com/scalacenter/bloop/issues/627 and https://github.com/scalacenter/bloop/issues/1242

danilbykov:
What do you mean? I tested this on several projects. I just added new flag for 'bloop run' which updates class files inside directory `bloop-bsp-clients-classes/classes-bloop-cli`. DCEVM handles the rest. This feature works fine when some additional logging is needed for debugging. It may not work for more complex cases when some class is used during application initialization.

tgodzik:
Hmm... I am not sure myself. `.bloop/root/bloop-bsp-clients-classes/classes-bloop-cli.` is a stable directory so DCEVM should be able to hot swap the code when invoked via CLI. This will not work currently via BSP, but for the CLI I don't really see the exact issue. @jvican Could you elaborate a bit?

Especially if it's under a separate flag, the risk merging this seems minimal and especially if we add some proper tests, we should be good to go. At least in the CLI

jvican (commented):
Yes, the CLI use case can be made to work as I explained [here](https://github.com/scalacenter/bloop/issues/1242#issuecomment-612408265), I initially misread the PR use case.

However, the CLI already works with with DCEVM: you can just run the application with `--watch` and then just pass in the pertinent JVM options to your project. There's no need for this change.

Adding a specific option for hot swap compilation for such obvious support is confusing for users and it gives the impression that bloop handles hot swap compilation use cases in BSP as well and that it will make the right thing. It'll be extremely disappointing that a user finds this flag and uses it and just realizes it is the good old `--watch` with a different name.

tgodzik:
Thanks for explaining!

> Yes, the CLI use case can be made to work as I explained [here](https://github.com/scalacenter/bloop/issues/1242#issuecomment-612408265), I initially misread the PR use case.
> 
> However, the CLI already works with with DCEVM: you can just run the application with `--watch` and then just pass in the pertinent JVM options to your project. There's no need for this change.

Wouldn't watch always recompile and restart the JVM process? 

> Adding a specific option for hot swap compilation for such obvious support is confusing for users and it gives the impression that bloop handles hot swap compilation use cases in BSP as well and that it will make the right thing. It'll be extremely disappointing that a user finds this flag and uses it and just realizes it is the good old `--watch` with a different name.

I do agree that this is confusing and especially if we can make it work with watch, it would be best to avoid it. @danilbykov what was the issue with using watch?

danilbykov:
I know about two different watches: one is in `bloop run` and another one is in `bloop compile`.

I see following behavior for watch in `bloop run`:
- It does nothing for server-like long-living application. Application is not restarted and class files in `bloop-bsp-clients-classes/classes-bloop-cli` are not updated (I check last modification times).
- It restarts application for short-living applications. But hot swap compilation is not for short-living applications. So it is totally useless.

Maybe you was talking about watch in `bloop compile`. I did following test. I run application using `bloop run` (without watch) in one console and `bloop compile` in another console. But this doesn't update class files inside `bloop-bsp-clients-classes/classes-bloop-cli` (again checking last modification times).

So I don't know how plain `watch` can help for hot swap compilation.

tgodzik:
> I know about two different watches: one is in `bloop run` and another one is in `bloop compile`.
> 
> I see following behavior for watch in `bloop run`:
> 
> * It does nothing for server-like long-living application. Application is not restarted and class files in `bloop-bsp-clients-classes/classes-bloop-cli` are not updated (I check last modification times).
> * It restarts application for short-living applications. But hot swap compilation is not for short-living applications. So it is totally useless.
> 
> Maybe you was talking about watch in `bloop compile`. I did following test. I run application using `bloop run` (without watch) in one console and `bloop compile` in another console. But this doesn't update class files inside `bloop-bsp-clients-classes/classes-bloop-cli` (again checking last modification times).
> 
> So I don't know how plain `watch` can help for hot swap compilation.

I think plain watch is not a solution here. I think I need to dig in a bit more to see how to address this properly. I don't know enough about this part of the codebase yet to warrant @jvican reservations, so I might be missing something still. I will try to get back to it during the next month.

Issue #1487, opened 2021-04-12
Migrate back to `alexarchambault/case-app`


ckipp01:
Taking a quick peek again at #1295 , I think a necessary first step is to actually migrate away from the `ch.epfl.scala:case-app` that Bloop uses. This is a fairly large task in itself that can be done apart from migrating to 2.13, which already seems to be quite a large task.

I have no idea why this was forked and used, I don't see it under the scalacenter org at all, and I'm unsure if there is a specific issue with just using the upstream case-app, but this needs to be done before the move to 2.13.

Creating an issue for this to maybe break #1295 into smaller chunks to be tackled.

alexarchambault:
I started to have a look at this, as I'd need bloop-launcher in 2.13 in a project I'm working on.

PR #1492, opened 2021-04-16
Update to mainline case-app 2.x

alexarchambault:
Fixes https://github.com/scalacenter/bloop/issues/1487.


alexarchambault:
Just realized a few `Parser` derivations could be moved around, for things to compile faster. CI ought to be green, but I'll probably push one more commit.

alexarchambault:
I guess this is ready for review. It seems the CI failures aren't related to these changes (and the failing tests run fine locally for me).

frontend/src/main/scala/bloop/util/CommandsDocGenerator.scala
```diff
@@ -7,43 +7,49 @@ import bloop.Cli
...
+import caseapp.core.util.NameOps._
 
 import scala.util.control.NonFatal
 import bloop.io.Environment.lineSeparator
+import caseapp.core.util.Formatter
 
 object CommandsDocGenerator {
   private final val docsContentDelimiter =
     "<!-- This file was generated by bloop.util.CommandsDocGenerator (see developer docs) -->"
+  private val NL = System.lineSeparator()
```
tgodzik, line 21: Best to use `bloop.io.Environment.lineSeparator` which makes sure that even if we are on Windows and use for example ZSH, we will just get `\n`.

project/Dependencies.scala
```diff
@@ -23,7 +23,7 @@ object Dependencies {
   val scalaXmlVersion = "1.2.0"
   val lmVersion = "1.0.0"
   val configDirsVersion = "10"
-  val caseAppVersion = "1.2.0-faster-compile-time"
+  val caseAppVersion = "2.0.6"
```
tgodzik, line 26: Should we also remove the custom shapeless?  I think both forks were added because of caseapp? I checked it out locally and removed shapeless, compilation time is faster than before, so I think we no longer have use for it. Both were originally introduced here: https://github.com/scalacenter/bloop/pull/509

tgodzik (changes requested):
Great work! Thanks you for taking the time to do it! I left two very minor comments, but we should most likely do it also in this PR.

frontend/src/main/scala/bloop/util/CommandsDocGenerator.scala
alexarchambault: Good point, just fixed that!

project/Dependencies.scala
alexarchambault: I hadn't noticed the custom shapeless… I just removed it too.

alexarchambault:
About https://github.com/scalacenter/bloop/pull/509, deriving the type class instances in the companion object, as done in https://github.com/scalacenter/bloop/pull/1492/commits/b50c65b3e760bdaddbd1ffcf04dc80b6449f58a1 , usually helps a lot. coursier does that, and there's no noticeable slow down, in spite of its 100+ command-line options.

project/Dependencies.scala
tgodzik: Thanks! One less fork :+1:

tgodzik (approved):
Tests are being super flaky here, but I don't see anything consistently failing. Will need to dig into those failures.

PR #1501, opened 2021-05-12
Fix error on Apple M1

changvvb:
Fix "Could not initialize class com.zaxxer.nuprocess.osx.OsxProcess"

Relate https://github.com/scalameta/metals-feature-requests/issues/185

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ object Dependencies {
   val Scala212Version = "2.12.8"
 
   val Sbt013Version = "0.13.18"
-  val Sbt1Version = "1.4.7"
+  val Sbt1Version = "1.3.3"
 
   val nailgunVersion = "ee3c4343"
   // Used to download the python client instead of resolving
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ object Dependencies {
   val Scala212Version = "2.12.8"
 
   val Sbt013Version = "0.13.18"
-  val Sbt1Version = "1.3.3"
+  val Sbt1Version = "1.4.7"
 
   val nailgunVersion = "ee3c4343"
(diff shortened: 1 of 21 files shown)


project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ object Dependencies {
   val Scala212Version = "2.12.8"
 
   val Sbt013Version = "0.13.18"
-  val Sbt1Version = "1.3.3"
+  val Sbt1Version = "1.4.7"
```
tgodzik, line 9: We haven't updated the sbt version so that we can be sure that the plugin will work will all versions `1.3.0+` Is this needed for fix? Alternatively we could release the plugin with 1.3.0, but I am not sure if that is possible.

tgodzik (changes requested):
Thanks for the contribution! I just have a question about sbt version change.

project/Dependencies.scala
changvvb: Sbt can only work with version 1.4.5+ on Apple M1 (https://github.com/sbt/sbt/issues/6187). If I didn't change the version, bloop can not build on my machine.

changvvb: em... It seems not the version in "build.properties"

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.3
+sbt.version=1.4.7
```
tgodzik, line 1: Maybe let's just go back to `1.3.3` here and `ci.yml`? Let's see if it works then.

changvvb:
Something wrong about "worksWithScala3LatestNightly". Any idea about this?

tgodzik:
> Something wrong about "worksWithScala3LatestNightly". Any idea about this?

It's unrelated to this PR, everything else looks fine! Thanks!

tgodzik (approved):

Issue #1508, opened 2021-06-13
Scala 3 Maven integration broken (`mvn bloopInstall`) 


GavinRay97:
https://github.com/davidB/scala-maven-plugin/issues/487#issuecomment-860177157
https://github.com/davidB/scala-maven-plugin/blob/master/src/it/test_scala3/pom.xml

I have `pom.xml` like below:

```xml
<properties>
    <scala-maven-plugin.version>4.5.3</scala-maven-plugin.version>
    <scala.compat.version>3</scala.compat.version>
    <scala.version>3.0.1-RC1</scala.version>
</properties>

<dependencies>
    <dependency>
      <groupId>org.scala-lang</groupId>
      <artifactId>scala3-library_${scala.compat.version}</artifactId>
      <version>${scala.version}</version>
    </dependency>
</dependencies>

  <build>
    <sourceDirectory>src/main/scala</sourceDirectory>
    <testSourceDirectory>src/test/scala</testSourceDirectory>
    <plugins>
      <plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <version>${compiler-plugin.version}</version>
        <configuration>
          <parameters>${maven.compiler.parameters}</parameters>
        </configuration>
      </plugin>
      <plugin>
        <groupId>net.alchim31.maven</groupId>
        <artifactId>scala-maven-plugin</artifactId>
        <version>${scala-maven-plugin.version}</version>
        <executions>
          <execution>
            <id>scala-compile-first</id>
            <phase>process-resources</phase>
            <goals>
              <goal>add-source</goal>
              <goal>compile</goal>
            </goals>
          </execution>
          <execution>
            <id>scala-test-compile</id>
            <phase>process-test-resources</phase>
            <goals>
              <goal>add-source</goal>
              <goal>testCompile</goal>
            </goals>
          </execution>
        </executions>
        <configuration>
          <scalaVersion>${scala.version}</scalaVersion>
        </configuration>
      </plugin>
    </plugins>
  </build>
```

Trying to let `Metals` import it to `bloop` errors, because of inconsistency in the URL's:
- Right `Downloaded from central: https://repo.maven.apache.org/maven2/org/scala-lang/scala3-library_3/3.0.1-RC1/scala3-library_3-3.0.1-RC1.jar (1.2 MB at 2.0 MB/s)`

[...]

PR #1512, opened 2021-06-17
Fix Maven Scala 3 integration and add tests

tgodzik:
This fixes https://github.com/scalacenter/bloop/issues/1508


config/.jvm/src/test/scala/bloop/config/TestPlatform.scala
```diff
@@ -14,7 +13,7 @@ object TestPlatform {
         val isr = new InputStreamReader(stream)
         try {
           val reader = new BufferedReader(isr)
-          reader.lines().collect(Collectors.joining(lineSeparator))
+          reader.lines().collect(Collectors.joining(" "))
```
tgodzik, line 16: This is actually used in a test and it's not needed for a json to have `\n`. Also, `bloop.io.Environment.lineSeparator` is not available in config module, but that was never checked, since tests are not run on config. It came up here, because I added `test -> test` dependency to have simple base class for different config tests. I could add a separate module, but the way it's currently done would mean that I would have to add 3 new modules.

I will tidy it up further down the line.

config/src/test/scala/bloop/config/utils/BaseConfigSuite.scala
```diff
@@ -0,0 +1,182 @@
+package bloop.config.utils
+
+import bloop.config.Config
+import java.io.File
+import java.nio.file.Path
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+
+trait BaseConfigSuite {
```
tgodzik, line 9: Methods moved from Gradle suite with some minor modifications.

integrations/maven-bloop/src/test/scala/bloop/integrations/maven/MavenConfigGenerationSuite.scala
```diff
@@ -0,0 +1,148 @@
...
+
+      assertNoConfigsHaveAnyJars(List(configFile), List(s"$projectName", s"$projectName-test"))
+      assertAllConfigsMatchJarNames(List(configFile), List("scala-library", "munit"))
+    }
+  }
+
+  private def check(testProject: String, projectName: String)(
+      checking: (Config.File, String) => Unit
+  ): Unit = {
+    val tempDir = Files.createTempDirectory("mavenBloop")
```
tgodzik, line 70: In order to avoid some additional dependencies on Maven testers (which are not so simple to add in sbt) the tests are done this:
- publishM2 is run before the tests
- maven process is executed using the maven wrapper
- configuration is read and checked

project/Dependencies.scala
```diff
@@ -32,7 +32,7 @@ object Dependencies {
   val directoryWatcherVersion = "0.8.0+6-f651bd93"
   val mavenApiVersion = "3.6.1"
   val mavenAnnotationsVersion = "3.5"
-  val mavenScalaPluginVersion = "3.2.2"
+  val mavenScalaPluginVersion = "4.5.3"
```
tgodzik, line 35: This version introduced the support for Scala 3 and Bloop is just reusing the plugin to read the data and generate configuration.

integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
```diff
@@ -94,15 +94,28 @@ object MojoImplementation {
         emptyLauncher
       }
 
+    val scalaContext = mojo.findScalaContext()
```
adpi2, line 97: `findScalaContext` was introduced in `maven-scala-plugin:4.5.x`. The  `basic-scala` integration test below does work because maven must use `maven-scala-plugin:4.5.x` instead of `maven-scala-plugin:3.3.2` when running the `ch.epfl.scala:maven-bloop_2.13:$version:bloopInstall` task. So far so good... but I am crossing fingers that this will work in more elaborate builds out there.

adpi2 (approved):
Left some comments but I don't know if we can make this better. So overall it looks good to me.

integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
```diff
@@ -94,15 +94,28 @@ object MojoImplementation {
...
+      }
+      .getOrElse("org.scala-lang")
     val scalacArgs = mojo.getScalacArgs().asScala.toList.filter(_ != null)
 
     def writeConfig(
         sourceDirs0: Seq[File],
         classesDir0: File,
         classpath0: java.util.List[_],
-        resources0: java.util.List[Resource],
+        resources0: java.util.List[_],
```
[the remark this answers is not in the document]
tgodzik: No idea, it seems inferred to Object, but this doesn't matter overall and I see we had that issue in the other cases too.

integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
tgodzik: I can add a test that uses an older version to check this out. It should work, since we don't use the version defined in the pom.

integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
```diff
@@ -94,15 +94,28 @@ object MojoImplementation {
...
+    }.toList
+
+    val scalaOrganization = compilerAndDeps
+      .collectFirst {
+        case artifact
+            if artifact.getArtifactId() == "scala3-compiler_3" || artifact
+              .getArtifactId() == "scala-compiler" =>
+          artifact.getGroupId()
+      }
+      .getOrElse("org.scala-lang")
```
ckipp01, line 111: Just to make sure I'm following, is this just here to catch another distribution of the compiler or something? We can't just rely on `org.scala-lang`?

ckipp01 (approved):
Super unfamiliar with the Maven stuff, but from what I can tell this looks good 👍🏼

integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
tgodzik: We should be ok to rely on `org.scala-lang` in most cases, however they have been different compilers in the past (I think all are unused) and I don't want to change the logic here.

ckipp01: 👍🏼 figured just wanted to make sure.

build.sbt
```diff
@@ -642,15 +638,23 @@ lazy val sbtBloop013Shaded =
...
   .disablePlugins(ScriptedPlugin)
-  .dependsOn(jsonConfig212.jvm)
-  .settings(name := "maven-bloop", scalaVersion := Scala212Version)
-  .settings(BuildDefaults.mavenPluginBuildSettings)
+  .enablePlugins(BuildInfoPlugin)
+  .dependsOn(jsonConfig213.jvm % "compile->compile;test->test")
+  .settings(
+    name := "maven-bloop",
+    scalaVersion := Keys.scalaVersion.in(jsonConfig213.jvm).value,
+    publishM2 := publishM2.dependsOn(publishM2.in(jsonConfig213.jvm)).value,
```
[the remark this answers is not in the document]
tgodzik: Fixed!

integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
tgodzik: Actually, we already have a test with an older plugin version and it works without an issue, so we should be fine.

PR #1513, opened 2021-06-18
Separate scalafmt step and run using boostraped jar

tgodzik:
The separate step guarantees that other tests  will also be run, which is better for the contributors.


adpi2 (approved):
:heavy_check_mark:

PR #1516, opened 2021-06-18
Separate build integration tests and remove unused bash scripts

tgodzik:
Previously, we would run integration tests only after all normal tests succeeded. Now, we run them separately, which has two added bonuses:
- contributors can easily say if they change to integrations have failed
- integration tests are always run irrelevant if the unit tests passed

I also remove leftover drone tests that are no longer used.


tgodzik:
What do you think @Arthurm1 ? I think it should be easier to debug any issues with integrations. This came up when I was trying to add tests for Maven integration.

.github/workflows/ci.yml
```diff
@@ -10,17 +10,40 @@ on:
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macOS-latest]
+        jdk: [graalvm-ce-java11@20.1.0]
+    steps:
+      - uses: actions/checkout@v2.3.4
+        with:
+          submodules: true
```
ckipp01, line 33: I know in the past we hit on issues with not include `fetch-depth: 0` in the checkout. If I'm not mistaken we even it on it in Bloop in the past. I'd probably include that here.

ckipp01 (approved):
This totally makes sense. I'd probably add in `fetch-depth: 0`, but other than that LGTM!

.github/workflows/ci.yml
adpi2: Do you remember the kind of issue?

As long as it does not happen again, I would prefer not to include the `fetch-depth: 0` here.

adpi2 (approved):
LGTM

.github/workflows/ci.yml
tgodzik: The issue was connected to releasing, we would get a wrong dynamic version, but it's not an issue in tests themselves. We only use `fetch-depth: 0` in the release job of metals.

ckipp01: > but it's not an issue in tests themselves.

Ah yea true, good point.

PR #1517, opened 2021-06-18
Try to fix flaky `fin test classes` test

tgodzik:
Might be possible that using later scalaJs version will help out here.

### DIFF
--- a/frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
+++ b/frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
@@ -286,7 +286,7 @@ class BspProtocolSpec(
       val logger = new RecordingLogger(ansiCodesSupported = false)
       loadBspBuildFromResources("cross-test-build-scalajs-0.6", workspace, logger) { build =>
         val project = build.projectFor("test-project-test")
-        val compiledState = build.state.compile(project)
+        val compiledState = build.state.compile(project, timeout = 120)
         val expectedClasses = Set(
           "JUnitTest",
           "ScalaTestTest",
--- a/frontend/src/test/scala/bloop/bsp/BspBaseSuite.scala
+++ b/frontend/src/test/scala/bloop/bsp/BspBaseSuite.scala
@@ -174,10 +174,11 @@ abstract class BspBaseSuite extends BaseSuite with BspClientTest {
     def compile(
         project: TestProject,
         originId: Option[String] = None,
-        clearDiagnostics: Boolean = true
+        clearDiagnostics: Boolean = true,
+        timeout: Long = 30
     ): ManagedBspTestState = {
       // Use a default timeout of 30 seconds for every operation
-      TestUtil.await(FiniteDuration(30, "s")) {
+      TestUtil.await(FiniteDuration(timeout, "s")) {
         compileTask(project, originId, clearDiagnostics)
       }
     }


tgodzik:
i run the tests 3 times and the flaky test didn't fail at all, so I think this should improve the situation greatly.

frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
```diff
@@ -284,18 +284,15 @@ class BspProtocolSpec(
   test("find test classes") {
     TestUtil.withinWorkspace { workspace =>
       val logger = new RecordingLogger(ansiCodesSupported = false)
-      loadBspBuildFromResources("cross-test-build-scalajs-0.6", workspace, logger) { build =>
+      loadBspBuildFromResources("cross-test-build-scalajs-1.x", workspace, logger) { build =>
```
tgodzik, line 287: My assumption is that the older version of Scala JS project might be more flaky than the 1.x version. We can see if it continuous down the road.

adpi2:
From what I've just seen, the flaky test fails because of a 30s timeout on this line:
https://github.com/scalacenter/bloop/blob/7adb59fd1437a1c57606645e991d0f0d17f78a19/frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala#L289

When it passes it takes more than 20s:
```
+ bloop.bsp.LocalBspProtocolSpec.find test classes 26182ms  
```

Even after the changes in this PR:
```
+ bloop.bsp.LocalBspProtocolSpec.find test classes 22327ms
```

So I think the test would not be flaky if we give it an infinite amount of time. But it is really slow... Do we know what takes so much time?

tgodzik:
> So I think the test would not be flaky if we give it an infinite amount of time. But it is really slow... Do we know what takes so much time?

No idea, locally it's pretty fast, faster than some other tests that never seem to fail. I tried to investigate it locally, but couldn't reproduce the issue. We could try doing it on CI, but it's not something I have time to work on currently and I honestly just want to make it less flaky. 

I can raise the timeout additionally to make sure that the test doesn't hit the limit. What do you think?

adpi2:
I am okay to raise the timeout.

Using `cross-test-build-scalajs-1.x` instead of `cross-test-build-scalajs-0.6` does not seem to make much difference. It's faster because it tests less test frameworks.

ckipp01:
> So I think the test would not be flaky if we give it an infinite amount of time. But it is really slow... Do we know what takes so much time?

I think this may actually be on the bsp4s side of things. I noticed when working on https://github.com/build-server-protocol/build-server-protocol/pull/175 that I needed to bump the timeout for the roundtrip tests to 20 seconds and even then sometimes they were timing out. There is no way it should take that long. So it very well could be that something funky is going on with bsp4s as well.

tgodzik:
@adpi2  I updated the timeout for the test


> I think this may actually be on the bsp4s side of things. I noticed when working on [build-server-protocol/build-server-protocol#175](https://github.com/build-server-protocol/build-server-protocol/pull/175) that I needed to bump the timeout for the roundtrip tests to 20 seconds and even then sometimes they were timing out. There is no way it should take that long. So it very well could be that something funky is going on with bsp4s as well.

Weirdly enough this seems to be happening while compiling. We might need to take a look at it, but right now I would love to see the tests less flaky.

adpi2 (approved):

Issue #1505, opened 2021-05-21
Bloop generates .bloop/.json file for root project with empty project.name


stefanrybacki:
In VSCode using metals:

What can be the reason that bloop generates a json without a name with empty project.name within the json for a gradle project.

It leads to error: `WARN  no build target for:` or any scala file within the project.

Manually changing the json name and project.name fixes the error.

Arthurm1:
We do play with the project names to create the simplest version of each project while making sure project names don't clash.  I imagine it's to do with that but can't think how it would produce empty project names.

Could you create a minimised reproduction of your build file(s) and post it here please.

stefanrybacki:
Will try to create a minimalistic example. However, it is weird, using the same repository on machine, different folder, it works as expected.

Arthurm1:
Could you post the result of `./gradlew projects`

stefanrybacki:
Here it the projects output (had to change some project names for legal reasons though):

```
Root project 'mmvs'
+--- Project ':project1'
+--- Project ':project1-api'
+--- Project ':project1-mgmt'
+--- Project ':doc'
|    \--- Project ':doc:sub1'
|         +--- Project ':doc:sub1:mmvs'
|         \--- Project ':doc:sub1:subsub1'
+--- Project ':lib'
+--- Project ':util'
+--- Project ':project2'
+--- Project ':project3'
+--- Project ':project3-api'
\--- Project ':project3-mgmt'
```

I have jsons for all projects and sub projects including -test.json in my .bloop folder but for mmvs, which is the .json file with an empty project name. (maybe it has to do with :doc:sub1:mmvs?)

Arthurm1:
Ah - it looks like `rootProject#getPath` returns ":".  We only use that on name clashes.  I'll see about a fix.

I guess you have source code at the root project level?  I've not seen that in multi-project builds before.

stefanrybacki:
We don't have source at the root level, we have buildSrc though.

Arthurm1:
OK - but you must be applying the Scala or Java plugin to your root project - otherwise Bloop wouldn't try to export that project.

stefanrybacki:
We do have this:

```
plugins {
  id 'java'
}

subprojects {
  apply plugin: 'java'
}
```
I would assume `plugins` will auto apply the java plugin also for root.

Arthurm1:
AFAIK the following code will declare _and apply_ the java plugin...
```
plugins {
  id 'java'
}
```

[...]

PR #1518, opened 2021-06-18
Gradle duplicate project name fix and tests for v7.0

Arthurm1:
Handles root project having a duplicate name of a sub project. Fixes: #1505

Reduces fully qualified names of duplicate projects to minimum namespace required to make them unique e.g.
`a-b-c-Foo` and `a-b-c-d-Foo`
becomes
`c-Foo` and `d-Foo`

Updates max test version to Gradle 7.0.  Some Gradle configurations are deprecated (or made non-consumable) so those tests are excluded and another added.


Arthurm1:
`bloop.dap.DebugServerSpec` fail - spurious

tgodzik (approved):
LGTM!

PR #1521, opened 2021-06-21
Remove unused bsp submodule

tgodzik:
Previously, it was useful for quick interation of the BSP protocol, which is much more stable today.


ckipp01 (approved):
I'm all for removing these. 👍🏼 LGTM

PR #1527, opened 2021-07-01
Revert nuprocess to 2.0.1

tgodzik:
2.0.2 was released on Java 9, which breaks on JDK 8

We should most likely stop using nuprocess, it's causing us way to many issues.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -40,7 +40,7 @@ object Dependencies {
   val circeVersion = "0.9.3"
   val jsoniterVersion = "2.4.0"
   val circeVersion213 = "0.12.2"
-  val nuprocessVersion = "2.0.2"
+  val nuprocessVersion = "2.0.1"
   val shapelessVersion = "2.3.4"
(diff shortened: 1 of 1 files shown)


adpi2 (approved):
Okay with that.

But how did you discover that it breaks on Java 8? What about Apple M1 since you made the upgrade for compatibility with Apple M1?

ckipp01:
> We should most likely stop using nuprocess

Do you have an alternative in mind?

tgodzik:
> Okay with that.
> 
> But how did you discover that it breaks on Java 8? What about Apple M1 since you made the upgrade for compatibility with Apple M1 (#1507)?

https://github.com/brettwooldridge/NuProcess/issues/128 and also it breaks on Metals tests run with JDK 8

> Do you have an alternative in mind?

Probably best to use plain Java Process API, there is a related issue in Metals https://github.com/scalameta/metals-feature-requests/issues/51

I asked @dos65 to take a look at it first.

ckipp01 (approved):

PR #1531, opened 2021-07-08
Replace nuprocess by java process API

dos65:
There is an issue with `nuprocess` on apple M1 that was fixed but its latest release isn't binary compatible with jdk8.
It seems that there we aren't need any specific feature from `nuprocess` so it would be easier just replace it by plain java process api.


tgodzik (changes requested):
Great work! I just have a single question and if you could fix the formatting we should be good to merge!

bridges/scalajs-1/src/main/scala/bloop/scalajs/jsenv/NodeJsEnv.scala
```diff
@@ -262,39 +260,39 @@ object NodeJSEnv {
...
-    builder.setProcessListener(handler)
-    config.cwd.foreach(builder.setCwd)
-
-    val processEnvironment = builder.environment()
-    processEnvironment.clear()
-    processEnvironment.putAll(config.env.asJava)
-
-    val process = builder.start()
-    process.wantWrite()
+    import monix.execution.Scheduler.Implicits.global
```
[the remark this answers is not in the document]
dos65: Done

bridges/scalajs-1/src/main/scala/bloop/scalajs/jsenv/NodeJsEnv.scala
```diff
@@ -25,6 +24,11 @@ import org.scalajs.jsenv.JSUtils.escapeJS
 
 import scala.collection.JavaConverters._
 import scala.concurrent.{Future, Promise}
+import scala.concurrent.ExecutionContext.Implicits.global
```
tgodzik, line 27:
```suggestion
```

Let's no use global

bridges/scalajs-1/src/main/scala/bloop/scalajs/jsenv/NodeJsEnv.scala
```diff
@@ -96,7 +100,7 @@ object NodeJSConfig {
  *
  * Adapted from `scala-js/nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala`.
  */
-final class NodeJSEnv(logger: Logger, config: NodeJSConfig) extends JSEnv {
+final class NodeJSEnv(logger: Logger, config: NodeJSConfig, ioScheduler: Scheduler) extends JSEnv {
```
tgodzik, line 103: Sorry to comment about it again, but it seems that we are anyway using  `engine.ExecutionContext.ioScheduler`, so maybe we can use that? Maybe adding it to params was not needed?

tgodzik (changes requested):

bridges/scalajs-1/src/main/scala/bloop/scalajs/jsenv/NodeJsEnv.scala
dos65: oh. I missed that jsBridge depends on frontend. Changed.

bridges/scalajs-1/src/main/scala/bloop/scalajs/jsenv/NodeJsEnv.scala
dos65: It's used [here](https://github.com/scalacenter/bloop/pull/1531/files#diff-9e00bb9ceb3e05cac1e8ff00ead15d96710162b0d36489715128d67a0d13d80aR296). I moved that import close to its usage.

dos65:
@tgodzik 

> Looks like the test failures could be connected to the change:

Let's wait another CI run. I think this one one is a flaky test on Windows that also happens in metals tests constantly.

tgodzik (approved):
LGTM!

PR #1533, opened 2021-07-16
[SbtBloop] Fix source dependencies for meta-build

### LINKED ISSUE #1450: Sbt-bloop - no `source` artifacts in meta-builds
Bloop has `bloop.export-jar.classifiers` java option that allows Metals additionally specify `source` classifier.
It seems that this setting isn't used for meta-build and  `goto definition` doesn't work for symbols that come from Sbt plugins.

dos65:
Fixes #1450


tgodzik (approved):
Looks good! Should we add a test here? Or are we even able to properly check if it works?

dos65:
@tgodzik Now it's covered by test

tgodzik (approved):
LGTM!

PR #1535, opened 2021-07-29
Bump scala-debug-adapter

adpi2:
Show static variables in stack frame (see https://github.com/scalacenter/scala-debug-adapter/pull/46)


tgodzik (approved):
LGTM! Thanks!

PR #1536, opened 2021-07-29
Do not log debug output on info level

tgodzik:
This causes any client to log all output messages twice, one in the official output for example in the editor and the second time in the logs, which can just grow the logs without purpose.

We should still be able to print it under debug flag.

Connected to https://github.com/scalameta/metals/issues/2311

Tested it locally and it works correctly. There is no need for logging all output since everything goes through DAP anyway.


adpi2 (approved):

ckipp01 (approved):
Ooo nice, I was expected this to be more work than this 😆 . This will be great. LGTM!

PR #1537, opened 2021-07-30
WIP: Bump monix to 3.4.0

### LINKED ISSUE #1510: Update Monix to 3.x.x
Started to again take a look at https://github.com/scalacenter/bloop/issues/1295 in hopes to get launcher published for 2.13. Another large task that can be separated out is an update to Monix 3.x.x. Due to the build graph publishing launcher forces updates in other modules, which forces Monix to be updated to a version that supports Scala 2.13. This happens to only be the Monix 3.x.x series. Looking through it, this probably isn't a small task as Monix is quite intertwined throughout Bloop. However, this will need to happen before Launcher can be published for 2.13.

kpodsiad:
The aim of this PR is to close #1510
I managed to fix all errors which occurred by bumping MOnix to 3.4.0. There are plenty of warnings caused by deprecated methods but I'll fix them later. For now, I'm curious how many things will break

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/compilation/CompileGraph.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/compilation/CompileGraph.scala
@@ -316,18 +316,18 @@ object CompileGraph {
         }
 
         val compileAndDeduplicate = Task
-          .chooseFirstOf(
+          .racePair(
             obtainResultFromDeduplication,
             Task.fromFuture(deduplicateStreamSideEffectsHandle)
           )
           .executeOn(ExecutionContext.ioScheduler)
 
         val finalCompileTask = compileAndDeduplicate.flatMap {
-          case Left((result, deduplicationFuture)) =>
-            Task.fromFuture(deduplicationFuture).map(_ => result)
-          case Right((compilationFuture, deduplicationResult)) =>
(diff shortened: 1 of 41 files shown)


frontend/src/main/scala/bloop/engine/tasks/compilation/CompileGraph.scala
```diff
@@ -316,18 +316,18 @@ object CompileGraph {
...
-          case Right((compilationFuture, deduplicationResult)) =>
+          case Left((result, deduplicationFiber)) =>
+            deduplicationFiber.cancel.map(_ => result)
+          case Right((compilationFiber, deduplicationResult)) =>
             deduplicationResult match {
-              case DeduplicationResult.Ok => Task.fromFuture(compilationFuture)
+              case DeduplicationResult.Ok => compilationFiber.join
               case DeduplicationResult.DeduplicationError(t) =>
                 rawLogger.trace(t)
                 val failedDeduplicationResult = Compiler.Result.GlobalError(
```
kpodsiad, line 333: I'm not sure if I preserve the semantics of the previous version

kpodsiad:
I finally managed to find (I hope!) the main source of the problems with failing tests, it's https://github.com/monix/monix/pull/724.

PR #1539, opened 2021-08-02
Gradle: handle --release and multi-threaded dir create

Arthurm1:
Don't add `-source` or `-target` javac options if `--release` is present
Don't set `-source` or `-target` from default jvm if not set by Gradle build

Ignore `FileAlreadyExistsException` exceptions on `.bloop` dir creation.  It's a function of parallel export of multiple projects.


tgodzik (approved):
LGTM, thanks!

PR #1540, opened 2021-08-03
support sbt Optional scope in bloopInstall

Arthurm1:
I don't use SBT much so I may be misinterpreting its dependency scopes but I think `Optional` isn't currently handled properly

Consider project `A` depends on `B` which depends on `C`
Project `B` uses `Optional` scope to depend on `C`

So `C` should feature in the compile classpath of `B` but **not** in the compile classpath of 'A' (as it shouldn't be included transitively)

This was already happening but before this PR there was no **project** dependency from `B` to `C` which Bloop needs to work out that it should compile `C` before `B`.

If I'm wrong on that then don't merge this PR but I think `Optional` should behave exactly like `Provided` because the runtime classpaths aren't currently exported and that's where `Optional` and `Provided` differ.

Relevant [link](https://github.com/scalacenter/bloop/discussions/1437)

### DIFF
--- a/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
+++ b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
@@ -26,6 +26,7 @@ import sbt.{
   ThisBuild,
   ThisProject,
   KeyRanks,
+  Optional,
   Provided
 }
 import xsbti.compile.CompileOrder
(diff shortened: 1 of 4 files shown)


Arthurm1:
`bloop.dap.DebugServerSpec`  - I think that's spurious

tgodzik (commented):
LGTM!

We also have `defaultSbtConfigurationMappings` but that doesn't really seem to be needed? Or maybe it might be needed in some rare case?

Arthurm1:
Unsure - I really don't know SBT anywhere near as well as Gradle so I was reluctant to change more than needed.  This got the test passing and made sense.  I was going to look at expanding the test cases and see if runtime classpaths can be set to properly take advantage of Provided & Optional scopes.

tgodzik (approved):
Anyway, I think this is already a great improvement, so I think we can merge it unless you want to improve something here. 

Thanks for figuring this out, I had no idea about the optional scope!

PR #1541, opened 2021-08-04
Do attempt to write Scala configuration for Java only modules

tgodzik:


ckipp01 (approved):
I'm pretty unfamiliar with the Maven stuff, but from what I can tell with the tests and this, looks good. LGTM!

PR #1544, opened 2021-08-09
WIP: Monix 3.x.x

### LINKED ISSUE #1510: Update Monix to 3.x.x

kpodsiad:
I accidentally closed https://github.com/scalacenter/bloop/pull/1537 :(
The aim of this PR is to close #1510

I managed to bump the Monix version to 3.3.0, but then I encountered various problems:
- some tests were failing. After investigation, I found that this is because of https://github.com/monix/monix/pull/724. Using find & replace I added an option that restores the previous behavior. For that moment, such a quick & dirty solution was ok, but before the merge, a better one should be found and applied.
- after that, some other tests were failing, but this time it was due to the incompatibility of Monix versions. JSONRPC framework from [scalameta lsp4s](https://github.com/scalameta/lsp4s) which was used in the Bloop depended on 2.x series so I have to bump [bsp](https://github.com/build-server-protocol ) version to the newest M15. This involves further changes, mentioned earlier JSON RPC framework was replaced by [one used in Bloop](https://github.com/jvican/jsonrpc4s)
- ~~now, there is a deadlock in `bloop.bsp.TcpBspConnectionSpec` which causes `Await's` to fail, but moreover, tests also hang after this failure. I checked and this deadlock also was happening before I bumped the BSP version so it's very probable that it's connected with Monix.~~ It's gone 🥳

Unfortunately, I failed to resolve the last problem and I would appreciate any help with it. 

Test Summary (13 August):
Failed: Total 322, Failed 31, Errors 0, Passed 291
Failed tests:
- [ ] bloop.bsp.TcpBspConnectionSpec
- [ ] bloop.bsp.LocalBspProtocolSpec
- [ ] bloop.bsp.TcpBspCompileSpec
- These three above works but the assertion which is checking no diff on strings is failling :(
- [ ] bloop.dap.DebugProtocolSpec
- [ ] bloop.CompileSpec
- [ ] bloop.FileWatchingSpec
- [ ] bloop.dap.DebugServerSpec
- [ ] bloop.bsp.LocalBspConnectionSpec
- [ ] bloop.JsTestSpec
- [ ] bloop.bsp.LocalBspCompileSpec
- [ ] bloop.JvmTestSpec
- [x] bloop.io.SourceHasherSpec
- [ ] bloop.DeduplicationSpec

### DIFF
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -330,7 +330,7 @@ object Compiler {
       Task
         .gatherUnordered(tasks)
         .map(_ => ())
-        .runAsync(compileInputs.ioScheduler)
+        .runToFuture(compileInputs.ioScheduler)
     }
 
     val start = System.nanoTime()
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -328,7 +328,7 @@ object Compiler {
 
     def runAggregateTasks(tasks: List[Task[Unit]]): CancelableFuture[Unit] = {
       Task
-        .gatherUnordered(tasks)
+        .parSequenceUnordered(tasks)
         .map(_ => ())
         .runToFuture(compileInputs.ioScheduler)
     }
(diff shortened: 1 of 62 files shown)


kpodsiad:
Deadlock which I mentioned earlier is caused by https://github.com/monix/monix/pull/1161. Downgrading to `3.2.0` resolves the problem but there are still some bugs that make tests fail.

hepin1989:
@kpodsiad the issue has been fixed.

kpodsiad:
We should simplify build definition and other stuff before trying to bump Monix version

PR #1548, opened 2021-08-14
Initial Gradle Scala 3 support

Arthurm1:
@tgodzik Very nice job on the Gradle 3 support

This PR handles Scala 3 Gradle projects setup using https://github.com/gradle/gradle/pull/18001

The specific Dotty handling is gone as it's no longer needed

The tests for Scala 3 are commented out until https://github.com/gradle/gradle/pull/18001 is merged and we know what Gradle version it will be released under

Up to you if you want to merge this now and then use the snapshots.  I've tested locally and seems fine.  Shouldn't cause anyone an issue with Scala 2 unless they were using the `dottyVersion` override but then they can stick with an early version of the bloop Gradle plugin.


GavinRay97:
Thank you for this!

tgodzik (approved):
Thanks for the follow up! I think we can merge this already and have it in the next Bloop release, which I was working on slowly.

Issue #1549, opened 2021-08-21
Using "Metals -> Import Build", get `'Couldn't retrieve module org.scala-lang:compiler-bridge_2.13:3.1.0-RC1-bin-20210819-408e6d2-NIGHTLY:compile'`


GavinRay97:
For some reason this only seems to happen with the `3.1.0` series and none of the dependency versions are different between `3.1.0`, `3.0.3-RC-NIGHTLY`, and `3.0.1/2`.

It just could be: "Hey, it's a nightly, stuff is going to break, who knows." 😅 

```gradle
plugins {
    id "scala"
}

repositories {
    mavenCentral()
}

dependencies {
    implementation("org.scala-lang:scala3-compiler_3:3.1.0-RC1-bin-20210819-408e6d2-NIGHTLY")
}

group = "org.example"
version = "1.0.0-SNAPSHOT"
```

```ps1
2021.08.21 14:10:09 INFO  Attempting to connect to the build server...
2021.08.21 14:10:09 INFO  tracing is enabled: C:\Users\rayga\AppData\Local\scalameta\metals\cache\bsp.trace.json
2021.08.21 14:10:09 INFO  time: Connected to build server in 0.63s
2021.08.21 14:10:09 INFO  Connected to Build server: Bloop v1.4.8-110-3d7bbe2c
2021.08.21 14:10:13 INFO  time: indexed workspace in 3.61s
2021.08.21 14:10:13 WARN  unsupported Scala 3.1.0-RC1-bin-20210819-408e6d2-NIGHTLY
2021.08.21 14:10:13 INFO  compiling quarkus-scala-gradle (5 scala sources)
2021.08.21 14:10:13 INFO  time: compiled quarkus-scala-gradle in 0.28s
2021.08.21 14:10:13 ERROR Unexpected error when compiling quarkus-scala-gradle:
  'Couldnt retrieve module org.scala-lang:compiler-bridge_2.13:3.1.0-RC1-bin-20210819-408e6d2-NIGHTLY:compile'
2021.08.21 14:10:18 INFO  compiling quarkus-scala-gradle (5 scala sources)
2021.08.21 14:10:18 INFO  time: compiled quarkus-scala-gradle in 0.1s
```

Arthurm1:
[...]

PR #1550, opened 2021-08-23
Fix support for Scala 3.1

tgodzik:
Fixes https://github.com/scalacenter/bloop/issues/1549#issuecomment-903242035


adpi2 (approved):
Thanks!

PR #1555, opened 2021-09-02
Cache Gradle API to avoid download issues

tgodzik:
Such as https://github.com/scalacenter/bloop/runs/3487999013?check_suite_focus=true


ckipp01 (approved):
👍🏼 Look good

Issue #1132, opened 2019-12-18
Debug messages are shown


tindzk:
After upgrading to v1.4.0-RC1, Bsp debug messages started being sent to stderr even without the `--verbose` parameter.

I added some [logic](https://github.com/tindzk/seed/pull/64/commits/660051bc1784285fb72ebdb990f5387483d32fff) to detect the log level of messages. This does not always work since there are lines that do not have the `[D]` or `[E]` prefix.

See also the [CI log](http://ci.sparse.tech/tindzk/seed/480/1/3).

jvican:
Err... this is not expected. It might be a regression. Or it might be that the launcher of this version surreptitiously enables --verbose on the server. Will have a look.

jvican:
Are you by any chance running `--verbose` in the launcher? What are the specific arguments you pass to the launcher?

tindzk:
No, I am not including `--verbose`. The CLI parameters are `bsp --socket <path>` ([code](
https://github.com/tindzk/seed/blob/55124732ba48da1e9225703fceee4eccb124bf3d/src/main/scala/seed/build/Bloop.scala#L448-L450)).

jvican:
I can reproduce. Will fix

PR #1556, opened 2021-09-02
Take into account --verbose flag when creating a logger

kpodsiad:
Should fix https://github.com/scalacenter/bloop/issues/1132


tgodzik (approved):
Thanks for the contribution!

Issue #1022, opened 2019-08-29
Jar dependencies between projects not working without publishing local jars


tgodzik:
Reported by @ayoub-benali in Metals

Given this workspace:
```
root
| - A
| - B
```
with project B depending on project A. Maven can't compile project B because project's A jar isn't available. Probably the generated bloop configuration isn't correct.

Current workaround is to do a mvn install -DskipTests to force JARs to be published locally but this isn't ideal for development .

To Reproduce

1. Clone https://github.com/dcmn-com/herringbone.git and switch to upgrade-scala-2.12 branch
2. install thrift locally (apt-get install or brew)
3. import build with Metals in any text editor

See error
```
metals: [ERROR] Failed to execute goal on project herringbone-main: Could not resolve dependencies for project com.stripe:herringbone-main:jar:0.0.1: Could not find artifact com.stripe:herringbone-impala:jar:0.0.2 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
metals: [ERROR] 
metals: [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
metals: [ERROR] Re-run Maven using the -X switch to enable full debug logging.
metals: [ERROR] 
metals: [ERROR] For more information about the errors and possible solutions, please read the following articles:
metals: [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
metals: [ERROR] 
metals: [ERROR] After correcting the problems, you can resume the build with the command
metals: [ERROR]   mvn <goals> -rf :herringbone-main
metals: build tool exit: 1
metals: time: ran 'mvn bloopInstall' in 7.94s
metals: Maven command failed: /tmp/metals516909384235530052/mvnw ch.epfl.scala:maven-bloop_2.10:1.3.2:bloopInstall -DdownloadSources=true
metals: Import project partially failed, limited functionality may work in some parts of the workspace. See the logs for more details. 
metals: running installed 'bloop bsp --verbose --protocol local --socket /tmp/bsp1697843160558780888/-rbfh4dcwxpo1.socket'
metals: tracing is disabled for protocol BSP, to enable tracing of incoming and outgoing JSON messages create an empty 
```
**Expected behavior**
The build to work

**Installation:**

Operating system: Linux
Editor: Sublime
Metals version: 0.7.0+111-0039eb7c-SNAPSHOT
Bloop version: v1.3.2
Search terms
maven, multi project, bloop

[...]

PR #1557, opened 2021-09-02
Download only external dependencies in multi module Maven projects

tgodzik:
Previously, we the Bloop Maven plugin would try to download all jars including ones that come from uncompiled subprojects. Now, we will only download external dependencies.

This was done via several steps:
- `requiresDependencyResolution` was replaced by `requiresDependencyCollection` which doesn't try to resolve the jar, but resolves information about all the tranisitive dependencies using pom
- to download the jars we use aether component `RepositorySystem`
- we will resolve both file and source files, if they exist we will write the correct config for Module, but also update file reference in the Artifact, which is null without resolution (caused by the first point). Thanks to the last step we will be able to get the correct classpath of the project

Additionally, since we no longer invoke maven tasks the plugin should be considerably faster.

Fixes https://github.com/scalacenter/bloop/issues/1022

### DIFF
--- a/integrations/maven-bloop/src/main/java/bloop/integrations/maven/BloopMojo.java
+++ b/integrations/maven-bloop/src/main/java/bloop/integrations/maven/BloopMojo.java
@@ -4,23 +4,41 @@ import bloop.config.Config;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.maven.plugin.MavenPluginManager;
 import org.apache.maven.plugin.MojoExecution;
+import org.apache.maven.project.MavenProject;
 import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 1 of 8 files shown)


ckipp01 (approved):
Huh, reading through the issue I assumed this would be a lot more work than this. Nice job! From looking at everything it looks good, but I didn't actually test this out with Maven. LGTM from me.

tgodzik:
> Huh, reading through the issue I assumed this would be a lot more work than this. Nice job! From looking at everything it looks good, but I didn't actually test this out with Maven. LGTM from me.

It was actually a lot of different approaches that didn't work, but finally I got to a solution that I am happy with. It's quite faster to work and uses the native maven mechanisms, which should spare us any suprises :sweat_smile:

PR #1559, opened 2021-09-04
Don't unnecessarily parse server handle as URI

alexarchambault:
It seems `ServerHandle.uri` is only used for logging purposes, no need to parse it as a `java.net.URI`.

In particular, when using Windows named pipes (like `bloop bsp --protocol local --pipe \\.\pipe\foo`), `ServerHandle.WindowsLocal.uri` crashes because of the back-slashes when trying to parse it as a URI.


tgodzik (approved):
LGTM! Thanks!

PR #1560, opened 2021-09-07
Set correct Scala version when multiple scala libraries are on Gradle classpath

Arthurm1:
The plugin wasn't respecting the order of the scala libraries on the classpath so Scala 3 version never got taken

When Gradle 7.3 is out I'll uncomment the tests but they work locally with the snapshot distribution


tgodzik (approved):
LGTM!

Issue #1240, opened 2020-04-09
Can bloop server support socket file in addition to port number?


davidmlw:
It is better to have socket file as server entry point. Especially in a multi-user server, it makes users independent work locally.

jvican:
Are you asking for support of something like UNIX Domain Sockets? And for which clients? The CLI or BSP clients such as Metals? Bloop supports Unix Domain sockets in BSP clients and CLI clients, though in the latter case it is not straightforward to set it up.

davidmlw:
Thanks, would please share some documents or bloop repo code snippet? By Unix domain socket, I mean a file path. In this way, the users can be more isolated to each other.

The using senario includes BSP or Bloop clients or Bloop.CLI. Each user can start a server that totally private, that others in the same server cannot access through a public PORT.

Here is the requirement reason: https://github.com/sifive/api-scala-sifive/issues/32, https://github.com/sifive/api-scala-sifive/issues/14.

Many thanks!

jvican:
@davidmlw Thanks for elaborating on your use case.
  
You can use Unix Domain Sockets or named pipes (which are essentially just socket files) with Bloop via `bloop bsp --help` will tell you how to connect to it. Our launcher (used by tools to start bloop in a machine and connect to it as a client) also has support for it, see [the implementation](https://github.com/scalacenter/bloop/blob/72a856b61568233256149f39f3e7f8e08b9602a5/launcher/src/main/scala/bloop/launcher/bsp/BspBridge.scala#L92-L109). But this constrains the use of Bloop only to BSP-based clients. The CLI is not a BSP client and our nailgun integration does not currently support Unix Domain Sockets or named pipes directly.
  
You could, in theory, run the server with `coursier launch ch.epfl.scala:bloop-frontend_2.12:1.4.0-RC1-165-906110dc -m com.martiansoftware.nailgun.NGServer -- local:/path/to/comm.socket` and use the soon-to-be-deprecated python-based bloop CLI to connect to it with `bloop --nailgun-server local:/path/to/comm.socket bloop.Cli $YOUR_COMMAND` where $YOUR_COMMAND can be any valid bloop command such as `compile`, `test` or `run`.

However, I haven't used that myself and I would recommend not using it because the python bloop script will be gone with the next release.

I agree we should support these use cases better. To do that, we would need two things:

[...]

PR #1562, opened 2021-09-09
Accept connections on domain socket / named pipe too

alexarchambault:
Opening this to know if everybody's fine with the idea. Currently, this PR makes Bloop listen on both a TCP port (as Bloop does currently) and on a unix domain socket in the "data directory" (`~/.local/share/bloop/socket` on Linux, `~/Library/Application Support/Bloop/socket` on macOS) or a named pipe on Windows. So both a TCP port and a socket at the same time.

This would allow Bloop clients to more easily give a try to named sockets, and maybe progressively phase out the default opening of a TCP port.

Phasing out the TCP port would also to close a security loophole. (AFAIU, anyone that can access the Bloop port can compile code with macros, so can basically run arbitrary code with the Bloop process privileges.)

Fixes https://github.com/scalacenter/bloop/issues/1240.

(Things added in this PR need to be made configurable prior to merging.)

### DIFF
--- a/.github/nailgun-publish-local.sh
+++ b/.github/nailgun-publish-local.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+cd nailgun
+sbt nailgun-server/publishLocal
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,7 @@ jobs:
           key: ${{ hashFiles('project/GradleIntegration.scala') }}
       - name: Tests
         run: |
+          .github/nailgun-publish-local.sh &&\
           ./bin/sbt-ci.sh \
               "mavenBloop/publishM2" \
               "mavenBloop/test" \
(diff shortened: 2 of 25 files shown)


tgodzik:
Thanks @alexarchambault ! I think this is perfectly reasonable, are there other benefits aside from security here?

alexarchambault:
I dived a bit more into that, I think it requires quite some more work.

On Unix, to handle a daemon socket file properly, it's better to write the PID of the daemon process in a file alongside the socket, so that users can check if the daemon is still alive and will accept connections on the socket file. The [libdaemon](http://0pointer.de/lennart/projects/libdaemon/) libraries illustrates that in C. This needs to check if a process with a given PID is still alive (so we have to run a command like `kill -0 $PID` or use the `kill` system call via JNI…).

On Windows, I'm not sure yet what assumptions we can make with its named pipes (does the pipe vanishes when the server abruptly stops? or does it remain there like domain sockets on Unix?).

I'll get back to this later…

alexarchambault:
First [green build](https://github.com/scalacenter/bloop/runs/4353673326?check_suite_focus=true) 🎉

alexarchambault:
Closing this, this has been implemented [here](https://github.com/scala-cli/bleep).

PR #1563, opened 2021-09-10
Update JGit

tgodzik:


project/build.sbt
```diff
@@ -31,7 +31,8 @@ val `bloop-build` = project
     addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.7.2"),
     // We need to add libdeps for the maven integration plugin to work
     libraryDependencies ++= List(
-      "org.eclipse.jgit" % "org.eclipse.jgit" % "4.6.0.201612231935-r",
+      "org.eclipse.jgit" % "org.eclipse.jgit" % "5.12.0.202106070339-r",
+      "org.eclipse.jgit" % "org.eclipse.jgit.ssh.jsch" % "5.12.0.202106070339-r",
```
ckipp01, line 35: Was this separated out or something in this version?

ckipp01 (approved):
These version numbers are odd 😆 .

project/build.sbt
ckipp01: Actually one other thing I just noticed. We also have this listed in `benchmark-bridge/build.sbt`. Should we also update it there?

tgodzik: > Was this separated out or something in this version?

Yeah, you can switch implementation, but honestly I would rather invoke a process than run this. Jgit seems unnecessarily complex, but didn't want to change to much of the release infra.

> Actually one other thing I just noticed. We also have this listed in benchmark-bridge/build.sbt. Should we also update it there?

That is a separate project, but it might be nice to do it once I work on the benchmarks, haven't touched them in a while.

I wanted to introduce scalafix for the project, which is a lot of work still, but I wanted to update it so it's easier to do so in the future. And scalafix uses jgit, which clashes with the version here.

Issue #1020, opened 2019-08-27
Fails to compile maven project 


ayoub-benali:
Bloop fails to compile a maven workspace with multi project. Instead I get this error
```
metals: error while loading Object, Missing dependency 'object scala.native in compiler mirror', required by /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/rt.jar(java/lang/Object.class)
metals: time: compiled herringbone-impala in 1.57s
metals: Unexpected error when compiling herringbone-impala: 'object scala in compiler mirror not found.'
metals: Failed to compile 'herringbone-impala','herringbone-main'
```

**step to reproduce:**
1. Clone `https://github.com/dcmn-com/herringbone.git` 
2. switch to `upgrade-scala-2.12` branch
3. run `mvn install -DskipTests` see https://github.com/scalameta/metals/issues/882
4. import build with Metals in any text editor
5. Notice error in logs

Using bloop CLI also leads to the same error:
`bloop compile herringbone-impala` -> 
```
Compiling herringbone-impala (5 Scala sources)
Compiled herringbone-impala (43ms)
[E] Unexpected error when compiling herringbone-impala: 'object scala in compiler mirror not found.'
[E] Failed to compile 'herringbone-impala'
```

**Installation:** 
- Operating system: Linux
- Editor: Sublime
- Metals version: 0.7.0+111-0039eb7c-SNAPSHOT
- Bloop version: v1.3.2
- Java 8

jvican:
I'm marking this as "help wanted" because I cannot personally look into it any time soon. I would appreciate if somebody can have a look at this error.

jvican:
I've had a closer look at what's happening here. The Bloop plugin for Maven doesn't seem to be adding the scala library jar in the classpath field. However, the `scala.jars` section of the configuration file does indeed have a reference to all of the relevant scala jars (library, reflect, compiler, etc): https://gist.github.com/jvican/074499727a125d17c445d8c4fa2c9492  

Adding the Scala library jar to the classpath makes the project work.

@ayoub-benali Are you interested in taking a look at how to fix this? I'm happy to provide pointers to find the offending issue and make Bloop work with your Maven project. Realistically nobody will have time to look at this otherwise.

PR #1564, opened 2021-09-13
Automatically add Scala library if it's not added explicitly

tgodzik:
You can use the Scala Maven plugin without adding any actual Scala dependencies, which will cause the Scala library to be missing.

Previously, we would not be able to compile such project. Now it's possible both to compile and to resolve sources.

Fixes https://github.com/scalacenter/bloop/issues/1020

### DIFF
--- a/integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
+++ b/integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
@@ -107,7 +107,7 @@ object MojoImplementation {
       .getReactorProjects()
       .asScala
       .map { project =>
-        (project.getGroupId(), project.getName(), project.getVersion())
+        (project.getGroupId(), project.getArtifactId(), project.getVersion())
       }
       .toSet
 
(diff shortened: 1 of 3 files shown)


integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
```diff
@@ -172,10 +172,29 @@ object MojoImplementation {
       val analysisOut = None
       val sourceDirs = sourceDirs0.map(abs).toList
       val classesDir = abs(classesDir0)
+      val artifacts = project.getArtifacts().asScala
+      lazy val libraryAndDependencies = scalaContext.toList
+        .flatMap(_.findLibraryAndDependencies().asScala)
```
adpi2, line 177: In Scala 3, would this contain the `scala3-library`?

tgodzik: in that case it will contain both `scala-library` and `scala3-library`

adpi2 (approved):
LGTM

PR #1567, opened 2021-09-15
Add expression evaluation for Scala 2.12 and 2.13

adpi2:
Upgrade to scala 2.12.13 to be able to test expression evaluation. The tests were running with Scala 2.12.8 but expression evaluation is only available on Scala 2.12.10 or greater.

Upgrade scala-debug-adapter to 2.0.2, the debug server now needs:
- a set of all the class path entries to be able to map the source files to the class files
- the jvm runtime
- the expression compiler class loader containing the expression-compiler jar and the compiler jars

Add a simple test on expression evaluation.

Warning: After merging this PR, Bloop debug session will no longer be compatible with the current version of Metals. https://github.com/scalameta/metals/pull/2959 in Metals intends to fix this.

### DIFF
--- a/frontend/src/main/scala/bloop/dap/BloopDebuggeeRunner.scala
+++ b/frontend/src/main/scala/bloop/dap/BloopDebuggeeRunner.scala
@@ -1,21 +1,23 @@
 package bloop.dap
 
+import bloop.ScalaInstance
 import bloop.cli.ExitStatus
 import bloop.data.{JdkConfig, Platform, Project}
-import bloop.engine.State
+import bloop.engine.{Build, State}
 import bloop.engine.tasks.{RunMode, Tasks}
+import bloop.logging.Logger
 import bloop.testing.{LoggingEventHandler, TestInternals}
 import ch.epfl.scala.bsp.ScalaMainClass
-import ch.epfl.scala.debugadapter.{CancelableFuture, DebuggeeRunner, DebuggeeListener}
+import ch.epfl.scala.debugadapter._
 import monix.eval.Task
 import monix.execution.Scheduler
-import xsbti.compile.analysis.SourceInfo
 
-import java.nio.file.Path
+import java.net.URLClassLoader
+import scala.collection.mutable
+import scala.annotation.tailrec
 
 abstract class BloopDebuggeeRunner(initialState: State, ioScheduler: Scheduler)
     extends DebuggeeRunner {
-  private lazy val allAnalysis = initialState.results.allAnalysis
 
   override def run(listener: DebuggeeListener): CancelableFuture[Unit] = {
     val debugSessionLogger = new DebuggeeLogger(listener, initialState.logger)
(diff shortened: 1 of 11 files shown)


project/Dependencies.scala
```diff
@@ -3,7 +3,7 @@ package build
 object Dependencies {
   val Scala210Version = "2.10.7"
   val Scala211Version = "2.11.12"
-  val Scala212Version = "2.12.8"
+  val Scala212Version = "2.12.13"
```
tgodzik, line 6: Do we need this change? I haven't managed to bump the version due to some issues with native image, which is causing the issues in CI

frontend/src/main/scala/bloop/dap/BloopDebuggeeRunner.scala
```diff
@@ -138,10 +146,157 @@ object BloopDebuggeeRunner {
...
+  private def getEvaluationClassLoader(
+      scalaInstance: ScalaInstance,
+      logger: Logger,
+      ioScheduler: Scheduler
+  ): Option[ClassLoader] = {
+    import ch.epfl.scala.debugadapter.BuildInfo._
+    import coursier._
+    val scalaVersion = scalaInstance.version
+    val module = s"${expressionCompilerName}_$scalaVersion"
+    val expressionCompilerDep = Dependency(
```
tgodzik, line 217: Could we cache it similar to how we do it with semanticdb plugins? Take a look at `SemanticDBCache` for details. This way we will not attempt to reresolve if there are issues.

frontend/src/test/scala/bloop/dap/DebugServerSpec.scala
```diff
@@ -661,6 +651,73 @@ object DebugServerSpec extends DebugBspBaseSuite {
...
+            _ <- client.initialized
+            breakpoints <- client.setBreakpoints(breakpoints)
+            _ = assert(breakpoints.breakpoints.forall(_.verified))
+            _ <- client.configurationDone()
+            stopped <- client.stopped
+            stackTrace <- client.stackTrace(stopped.threadId)
+            topFrame <- stackTrace.stackFrames.headOption
+              .map(Task.now)
+              .getOrElse(Task.raiseError(new NoSuchElementException("no frames on the stack")))
+            evaluation <- client.evaluate(topFrame.id, "1 + 1")
```
tgodzik, line 704: Could we also try evaluating soemthing from context like a method of `foo` ?

tgodzik (changes requested):
Great work @adpi2 ! I added some minor comments, but the biggest issue is the updated to newest Scala. It would probably be best to tackle it separately ass there seem to be issues with native image.

project/Dependencies.scala
adpi2: Actually no I don't need this change since I can choose a different Scala version in the tests.

frontend/src/test/scala/bloop/dap/DebugServerSpec.scala
adpi2: Yes we can but I only want to test the integration here, not all the expression evaluation cases because those are already tested here https://github.com/scalacenter/scala-debug-adapter/blob/main/core/src/test/scala/ch/epfl/scala/debugadapter/ExpressionEvaluatorSpec.scala

frontend/src/main/scala/bloop/dap/BloopDebuggeeRunner.scala
adpi2: Done! Thanks for the pointer.

frontend/src/test/scala/bloop/dap/DebugServerSpec.scala
adpi2: I changed the test anyway

adpi2:
@tgodzik  Out of curiosity, do you know why native images are broken on 2.12.13?

tgodzik:
> @tgodzik Out of curiosity, do you know why native images are broken on 2.12.13?

It's probably a part of Scala that is not supported on Native Image, there are ways around it with https://github.com/scalameta/sbt-native-image, but I haven't had time to look properly into that.

tgodzik (approved):
LGTM! Great work and pretty exciting!

PR #1568, opened 2021-09-17
Replace all references to the Bloop gitter channel

tgodzik:
I also updated the link to never go outdated.


ckipp01 (approved):
Good call 👍🏼

PR #1570, opened 2021-09-20
Fix issues with last tag not being detected properly

tgodzik:
Previously, describe would return a wrong tag, which wasn't the lastest published tag. Now we lsit tags explicitely and get the last one.

Fixes https://github.com/scalacenter/homebrew-bloop/issues/17


ckipp01 (approved):
LGTM!

PR #1574, opened 2021-09-29
Bump scala-debug-adapter

adpi2:
Fixes expression evaluation in https://github.com/scalacenter/scala-debug-adapter/issues/103 and https://github.com/scalacenter/scala-debug-adapter/issues/106


tgodzik (approved):
:tada:

Issue #1573, opened 2021-09-26
maven-bloop doesn't include directories with generated sources


nigredo-tori:
Consider this maven project: [maven-example.zip](https://github.com/scalacenter/bloop/files/7231391/maven-example.zip). It includes a protobuf model for which Java code is generated, which is later compiled together with the unmanaged sources. Documentation instructs us to export it as follows:

```sh
$ mvn ch.epfl.scala:maven-bloop_2.13:1.4.9:bloopInstall
```
(actually, documentation says 2.12, but that artifact does not exist). However, when exported this way, we only have one source directory in the generated bloop project:
```json
        "sources": [
            "/home/nigredo/dev/tmp/foo/src/main/scala"
        ],
```
whereas the correct output would look like this:
```json
        "sources": [
            "/home/nigredo/dev/tmp/foo/src/main/scala",
            "/home/nigredo/dev/tmp/foo/target/generated-sources/protobuf/java"
        ],
```

This is because both `bloopInstall` and Protobuf code generation are bound to the same Maven phase, `generate-sources`. Thus there is no dependency between these two goals, and Protobuf code generation is not run. Since one of the things the code generation goal does is to add a compile source directory to the project description, and that does not happen, maven-bloop has no way to know about that directory.

One possible solution would be to bind `bloopInstall` to the `process-sources` phase, so that it is guaranteed to run after code generation. I'm not sure what this would break.

As a workaround, it's possible to manually micromanage the goals by exporting the project this way:

```sh
$ mvn generate-sources ch.epfl.scala:maven-bloop_2.13:1.4.9:bloopInstall
```
This gives us the expected `sources`.

tgodzik:
Thanks for reporting! We should be able to switch to `process-sources`, let me try to experiment with that.

tgodzik:
I tested it out and it seems there is not other way than adding `generate-sources`.

I updated the [docs](https://github.com/scalacenter/bloop/pull/1575) and will add the change to Metals also.

tgodzik:
I updated the docs for Bloop and changed the way Metals runs the Maven command. Otherwise, I don't see any better workaround, because when Bloop is run as an outside plugin it will not invoke other phases.

Some power users might want to add Bloop to `pom.xml` and bind it to the `process-sources` phase, but we don't want to force it on people.

Closing it for now, unless anyone comes up with a better solution. Thanks again @nigredo-tori for raising this!

PR #1575, opened 2021-09-29
Update Maven and Coursier documentation

tgodzik:
Related to https://github.com/scalacenter/bloop/issues/1573

### DIFF
--- a/docs/build-tools/maven.md
+++ b/docs/build-tools/maven.md
@@ -27,7 +27,14 @@ The Bloop plugin for Maven doesn't need to be installed, you can run
 following command.
 
 ```bash
-$ mvn ch.epfl.scala:maven-bloop_2.12:@VERSION@:bloopInstall
+$ mvn ch.epfl.scala:maven-bloop_2.13:@VERSION@:bloopInstall
+```
+
+In some cases, when you are using generated sources in you project
+you might need to run:
+
+```bash
+$ mvn generate-sources ch.epfl.scala:maven-bloop_2.13:@VERSION@:bloopInstall
 ```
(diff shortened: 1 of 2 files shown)


docs/build-tools/maven.md
```diff
@@ -27,7 +27,14 @@ The Bloop plugin for Maven doesn't need to be installed, you can run
...
 ```bash
-$ mvn ch.epfl.scala:maven-bloop_2.12:@VERSION@:bloopInstall
+$ mvn ch.epfl.scala:maven-bloop_2.13:@VERSION@:bloopInstall
+```
+
+In some cases, when you are using generated sources in you project
+you might need to run:
+
+```bash
+$ mvn generate-sources ch.epfl.scala:maven-bloop_2.13:@VERSION@:bloopInstall
```
adpi2, line 37: Would it be possible that `bloopInstall` triggers `generate-sources` automatically?

tgodzik: I tried to add it to a phase afterwards, but it still didn't run the generate-sources phase for the protobuf plugin. It might be due to the fact that we don't actually add the plugin to the build.

adpi2 (approved):

docs/build-tools/maven.md
tgodzik: Apparently, you can bind the plugin to a phase and it will be executed during that phase by default, but not invoke the phase when you run the plugin:

https://stackoverflow.com/questions/3934833/maven-mojo-plugin-how-to-define-phases-that-must-be-triggered-before-this-goal

PR #1577, opened 2021-10-02
Gradle improvements

Arthurm1:
1) Add test for Scala3 (using Gradle nightlies) and make functionality support by Gradle version clearer.
2) Refactor the way ArchiveTasks are queried for their contents.
3) Make artifact resolution more lenient

I can't reproduce either issue in #1566 but 2 and 3 may solve both of them 🤷


tgodzik (approved):
LGTM!

Issue #1584, opened 2021-10-19
Set `canDebug` in BuildTargetCapabilities


kpodsiad:
Metals since https://github.com/scalameta/metals/pull/3210 generate run lenses for sbt, bloop (there are custom checks for them), and any build server which support `canDebug`. Setting this capability in bloop would allow to simplify metals' logic.
See https://github.com/build-server-protocol/build-server-protocol/pull/161 for more info.

ckipp01:
This was completed in https://github.com/scalacenter/bloop/pull/1820.

PR #1586, opened 2021-10-19
Set canDebug capability

kpodsiad:
Closes https://github.com/scalacenter/bloop/issues/1584


kpodsiad:
There's no point in fighting with the build config related errors :(

PR #1599, opened 2021-11-02
Bump debug-adapter to 2.0.10

adpi2:
-  Fix https://github.com/scalacenter/scala-debug-adapter/issues/148
- Add support for Scala 2.13.7


tgodzik:
Och, actually it seems that the fix didn't work :/

tgodzik:
It seems ClasspathEntries only contains the dependencies not the actual compiler. We should use `project.rawClasspath` (or maybe we should calculate it for all dependencies like we do for entries) and add it to the Debugee constructors.

We should most likely also merge https://github.com/scalacenter/scala-debug-adapter/pull/154

Issue #1598, opened 2021-11-02
Bloop doesn't compile Scala 2.13.7


tgodzik:
The issue is because of the Zinc fork the compiler bridge doesn't compile anymore:
```
/tmp/sbt_f7b9f74f/xsbt/ZincSymbolLoaders.scala:17: error: value nameChars is not a member of scala.tools.nsc.util.ClassRepresentation
        val len = classRep.nameChars(nameCharBuffer0)
                           ^
```

taking a look at it but I have no estimates for when it will work.

tgodzik:
I made a fix in https://github.com/scalacenter/zinc/tree/main based on changes in https://github.com/scala/scala/commit/f51e204fb53f4155320bb5f75271991417453d76

@Duhemm would you be able to publish the fork?

Duhemm:
The updated Zinc fork is in Maven central with version `1.3.0-M4+47-d881fa2f`.

SethTisue:
Is there something we could add to the Scala 2 community build so we'd know in advance?

tgodzik:
> Is there something we could add to the Scala 2 community build so we'd know in advance?

Theoretically, the zinc fork compilation, but we want to get rid of it, so maybe it doesn't make sense. I plan to put aside all other work next week to work on it

PR #1600, opened 2021-11-02
Update zinc fork with the newest fixes

tgodzik:
Fixes https://github.com/scalacenter/bloop/issues/1598

### DIFF
--- a/backend/src/main/scala/sbt/internal/inc/BloopComponentCompiler.scala
+++ b/backend/src/main/scala/sbt/internal/inc/BloopComponentCompiler.scala
@@ -108,11 +108,6 @@ object BloopComponentCompiler {
       scalaVersion.startsWith("2.13.2")
     }
 
-    private def isTooRecentFor213Before3(bridgeSources: ModuleID): Boolean = {
-      bridgeSources.revision == "1.3.0-M4+45-d4354be3" ||
-      bridgeSources.revision == "1.3.0-M4+46-edbe573e"
-    }
-
     /**
      * Defines a richer interface for Scala users that want to pass in an explicit module id.
      *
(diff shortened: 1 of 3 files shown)


ckipp01 (approved):
If there was every another sign that we should get off our zinc fork 😅 .

If this indeed fixes the issues then LGTM. Looking at the linked Scala pr to be honest I don't fully comprehend the full ramifications of all the changes in there, so for me this is sort of a "if it's ✅ merge it".

tgodzik:
> If there was every another sign that we should get off our zinc fork sweat_smile .
> 

100% agree

ckipp01:
Great job getting this done btw 🚀 .

adpi2:
Great job @tgodzik and thanks for taking care of it.

SethTisue:
> Looking at the linked Scala pr

What PR is that? (I'm not seeing a link)

ckipp01:
> What PR is that? (I'm not seeing a link)

It was in the linked issue: https://github.com/scala/scala/commit/f51e204fb53f4155320bb5f75271991417453d76

SethTisue:
ah okay, https://github.com/scala/scala/pull/9557

SethTisue:
which is a 2.12.14 PR, I'm a bit curious why this problem didn't manifest earlier, like after the 2.12.14 or 2.12.15 releases? only curious

tgodzik:
> which is a 2.12.14 PR, I'm a bit curious why this problem didn't manifest earlier, like after the 2.12.14 or 2.12.15 releases? only curious

That class is in a 2.13.x specific directory: https://github.com/scalacenter/zinc/tree/main/internal/compiler-bridge/src/main

PR #1612, opened 2021-11-24
Use coursier-interface rather than coursier

alexarchambault:
The former is more stable, has less dependencies, and is enough for our needs here.


ckipp01 (approved):
This is great, the more thinning out with Bloop we can do the better.

Nit: Can you squash away the two `?` commits?

tgodzik:
> This is great, the more thinning out with Bloop we can do the better.
> 
> Nit: Can you squash away the two `?` commits?

We can actually do `Squash and merge` ourselves :tada:

tgodzik (approved):
LGTM! Thanks a lot for these PRs, it does simplify things :tada:

PR #1613, opened 2021-11-24
Try to make frontend tests less flaky

alexarchambault:
By writing temporary data somewhere under the current directory, rather than via Files.createTempDirectory. (In the Scala CLI repo, this got rid of many transient failures in tests involving domain sockets, https://github.com/VirtusLab/scala-cli/commit/cc5052eeeb3b729649027ce0f428097fae5d0a36).


alexarchambault:
More flakyness :|

PR #1617, opened 2021-12-02
Ignore Hydra license output in tests

tgodzik:


tgodzik:
@dotta it seems that the license is expiring, so the test will be ignored in 14 days.

frontend/src/test/scala/bloop/HydraCompileSpec.scala
```diff
@@ -13,9 +13,14 @@ import scala.tools.nsc.Properties
 
 object HydraCompileSpec extends BaseCompileSpec {
   override protected val TestProject = HydraTestProject
+
   override protected def extraCompilationMessageOutput: String =
     " [E-1] Using 1 Hydra worker to compile Scala sources."
 
+  override protected def processOutput(message: String) = {
+    message.replace(" [E-1] License will expire in 14 days.\n", "")
```
adpi2, line 21: What if we run the test tommorrow? Will the message contain `13 days` rather than `14 days`?

tgodzik: Good point :S I didn't think about it

adpi2 (approved):
LGTM

PR #1618, opened 2021-12-02
Fix regex used for HydraCompileSpec

tgodzik:
second attempt :/


tgodzik:
I think this should work better, but it might be possible that Hydra tests only run on master.

PR #1622, opened 2021-12-09
Add exports for Semanticdb plugin on Java 17

Arthurm1:
Adds module exports for Sourcegraph lsif-java plugin when the JVM is >= 17

As suggested - I've used the `JAVA_HOME/release` file to get the build target's JVM version or fallback to Bloop's JVM if the build target has no JVM specified.
`JAVA_HOME/release` seems to be available on all my installed JDKs across Windows and WSL

I've copied the version compare code from `scala.util.Properties#compareVersions` which is unfortunately not public.  I'd put this code in a util helper class but I don't think it's worth it until it's going to be used in other places as well.
`com.google.gson.internal.JavaVersion` only offers comparison to the running JVM - the rest of the code is private.

### DIFF
--- a/frontend/src/main/scala/bloop/data/Project.scala
+++ b/frontend/src/main/scala/bloop/data/Project.scala
@@ -2,17 +2,28 @@ package bloop.data
 
 import bloop.ScalaInstance
 import bloop.bsp.ProjectUris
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM! Thanks!

PR #1623, opened 2021-12-10
Update sbt to 1.4.9 and remove offloading

tgodzik:
Offloading uses internal sbt API, which means there will be issues with any update that comes up. Especially since this was never finished, I think it's better to remove it and push the current state to a separate branch.

The sbt version bump is needed to help with running the benchamarks in zinc update PR (older sbt versions keep hanging)


tgodzik:
```
[info] compiling 2 Scala sources to /home/runner/work/bloop/bloop/project/project/target/scala-2.12/sbt-1.0/classes ...
[error] bad option: '-Wconf:cat=unused-nowarn:s'
[error] (Compile / compileIncremental) Compilation failed
```
I have no idea why is this popping up.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.3
+sbt.version=1.4.9
```
ckipp01, line 1: Does bumping to 1.5.x blow things up?

ckipp01 (approved):
I didn't realize there was this much code for the offloading. LGTM!

project/build.properties
tgodzik: Yeah, and I am not sure exactly why that is. I need to report an issue upstream. 1.4.x version should already help with the benchmarks

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -1031,9 +1022,8 @@ object BloopDefaults {
           val config = {
             val c = Keys.classpathOptions.value
             val java = Config.Java(javacOptions)
-            val analysisOut = BloopKeys.bloopAnalysisOut.value.map(_.toPath)//.getOrElse(out.resolve(s"${projectName}-analysis.bin"))
             val compileSetup = Config.CompileSetup(compileOrder, c.bootLibrary, c.compiler, c.extra, c.autoBoot, c.filterLibrary)
-            val `scala` = Config.Scala(scalaOrg, scalaName, scalaVersion, scalacOptions, allScalaJars, analysisOut, Some(compileSetup))
+            val `scala` = Config.Scala(scalaOrg, scalaName, scalaVersion, scalacOptions, allScalaJars, None, Some(compileSetup))
```
adpi2, line 1026: Is this `analysisOut` parameter used anywhere else? Otherwise we should maybe remove it.

```suggestion
            val `scala` = Config.Scala(scalaOrg, scalaName, scalaVersion, scalacOptions, allScalaJars, Some(compileSetup))
```

adpi2 (approved):
:heart:

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
tgodzik: It's used in some tests currently and it might be useful if we decide to take a stab at the offloading. Don't want to change the configuration right now (this is the same as the one used in `.bloop` files)

PR #1625, opened 2021-12-13
Handle gradle project-sourceset name clashes

### LINKED ISSUE #1624: Bloop export directory name clashes
Hi, I've got a large Gradle project with 'sql' and 'sql-test' sub-projects in it.

It seems that bloop export [generates](https://scalacenter.github.io/bloop/docs/build-tools/gradle) its own 'sql-test' config based on the 'sql' project and ignores/overwrites my original 'sql-test' project in .bloop/ - I see various errors about missing classes in my IDE (vscode).

This issue would affect projects already containing both '_name_' and '_name_-test' directories.
Perhaps using a random (or configurable) suffix in bloop-exported directory names could help here?

Thanks!

Arthurm1:
I don't think there's a nice way to create names on this type of clash. I've ended up just adding a number on the end of the `project-sourceset` name.

I've also excluded tests when the testing jvm is a later version than the Gradle version supports.  See [here](https://docs.gradle.org/current/userguide/compatibility.html).
This does seem to have caused the github diff to go bananas.  There's only 1 new test `worksWithDuplicateTargetNames1` and then I've indented all the others because of adding the check `supportsCurrentJavaVersion`

Fixes #1624
That example should end up with...

| project | sourceset | bloop name |
| - | - | - |
sql|main|sql
sql|test|sql-test-2
sql-test|main|sql-test
sql-test|test|sql-test-test

Another option would have been

| project | sourceset | bloop name |
| - | - | - |
sql|main|sql
sql|test|sql-test
sql-test|main|sql-test-main
sql-test|test|sql-test-test

But that doesn't preserve the name `sql-test` for the Gradle project `sql-test` which I think is important

### DIFF
--- a/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
+++ b/integrations/gradle-bloop/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala
@@ -768,9 +768,8 @@ class BloopConverter(parameters: BloopParameters, info: String => Unit) {
     }
 
     // Need to namespace only those projects that can run bloop. Others would not cause collision.
-    val projectsWithSameName =
-      getAllBloopCapableProjects(project.getRootProject())
-        .filter(_.getName == project.getName)
+    val allProjects = getAllBloopCapableProjects(project.getRootProject())
+    val projectsWithSameName = allProjects.filter(_.getName == project.getName)
 
     val uniqueProjectName =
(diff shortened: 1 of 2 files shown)


Arthurm1:
Spurious `TcpBspMetalsClientSpec` failure

tgodzik (approved):
LGTM! 

You can always use `Hide whitespace changes` option, which reduces the diff to just a couple of dozen lines.

Issue #1479, opened 2021-03-16
bloop fish completion prints errors


ohechtl:
Hi!

When I hit <Tab> after inputting an expression like `bloop test ` in the fish terminal, I'll get this text printed before I'll get my completion:
```
bloop test test: Expected a combining operator like '-a' at index 5nTest

/usr/share/fish/vendor_completions.d/bloop.fish (line 105): 
        if test -n $parsed[1] -a -n $parsed[2] -a -n $parsed[3]
           ^
from sourcing file /usr/share/fish/vendor_completions.d/bloop.fish
in command substitution

(Type 'help test' for related documentation)
test: Expected a combining operator like '-a' at index 5

/usr/share/fish/vendor_completions.d/bloop.fish (line 107): 
        else if test -n $parsed[1] -a -z $parsed[2] -a -n $parsed[3]
                ^
from sourcing file /usr/share/fish/vendor_completions.d/bloop.fish
in command substitution

(Type 'help test' for related documentation)
test: Expected a combining operator like '-a' at index 5

/usr/share/fish/vendor_completions.d/bloop.fish (line 109): 
        else if test -n $parsed[1] -a -n $parsed[2] -a -z $parsed[3]
                ^
from sourcing file /usr/share/fish/vendor_completions.d/bloop.fish
in command substitution

(Type 'help test' for related documentation)
test: Expected a combining operator like '-a' at index 5

/usr/share/fish/vendor_completions.d/bloop.fish (line 111): 
        else if test -n $parsed[1] -a -z $parsed[2] -a -z $parsed[3]
                ^
from sourcing file /usr/share/fish/vendor_completions.d/bloop.fish
in command substitution

(Type 'help test' for related documentation)
```
                                                                     
Although I'll get my completion after this, and I won't get a new error if I try to reproduce it again, it's a minor pain point for me.

Do you have any idea what could cause this?

tgodzik:
Thanks for reporting! I am not exactly sure what might be causing these issues. I would welcome any help if anyone is more versed in fish. 

I might take a look into it at some point in the future as I also use fish, but I cannot promise anything.

a06kin:
Looks like fix is available in [v1.4.12](https://github.com/scalacenter/bloop/releases/tag/v1.4.12)
Can we close it @tgodzik?

ohechtl:
Thank you very much! :) :) :)

PR #1633, opened 2021-12-22
Fix fish completions error

a06kin:
Mainly its fix for #1479 issue.

The problem is that when we don't have all 3 parameters in parsed
`if test -n $parsed[1] -a -n $parsed[2] -a -n $parsed[3]`
after read all vars script will look like this (for example we have only $parsed[1] = fizz)
`if test -n fizz -a -n -a -n` which is not proper syntax

Another small fix (in _testsfqcn) is for redundant `Argument --project: argument missing` in autocomplete list.
Its cause sometimes we call _testsfqcn without --project arg

Screenshot 2021-12-22 at 10 03 47

### DIFF
--- a/etc/fish-completions
+++ b/etc/fish-completions
@@ -102,13 +102,13 @@ for cmd in (_commands)
     set -l flags (_flags_for_command $cmd)
     for flag in $flags
         set -l parsed (string split '#' $flag)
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM! Thanks a lot for the fix!

PR #1638, opened 2021-12-31
Fix JS compilation errors

tgodzik:
Config module is cross published for JVM and JS, which means that config-test is also compiled with both. This is causing compilation errors for `BasicConfigSuite`, although it is never actually used in tests for js, which is why it wasn't failing on CI. This PR is just to get rid of that annoying error.

### DIFF
--- a/config/.js/src/main/scala/bloop/config/PlatformFiles.scala
+++ b/config/.js/src/main/scala/bloop/config/PlatformFiles.scala
@@ -4,6 +4,7 @@ object PlatformFiles {
   type Path = String
   val emptyPath: Path = getPath("")
   def getPath(path: String): Path = path
+  def getFileName(path: Path): Path = path.split('/').last
 
   def userDir: Path = NodePath.resolve(".")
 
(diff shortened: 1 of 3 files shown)


config/.js/src/main/scala/bloop/config/PlatformFiles.scala
```diff
@@ -4,6 +4,7 @@ object PlatformFiles {
   type Path = String
   val emptyPath: Path = getPath("")
   def getPath(path: String): Path = path
+  def getFileName(path: Path): Path = path.split('/').last
```
ckipp01, line 7: Won't this break on windows?

tgodzik: Not sure, but we don't actually use it, since the method is only used in integration tools for build tools.

I can modify it to use the system separator just in case 🤔

ckipp01: > but we don't actually use it

meh, don't even worry about it then.

ckipp01 (approved):

PR #1645, opened 2022-01-03
Fix gradle211 compilation errors

kpodsiad:

PR #1646, opened 2022-01-03
Launcher 213

kpodsiad:
Contains https://github.com/scalacenter/bloop/pull/1645

What I checked:
- both `compile` and `test:compile` yield no errors.
- `+launcher/publishLocal` publishes 2.13 artifact (bloop-launcher-core). 
- `+launcherShaded/publishLocal` publishes 2.13 artifact (bloop-launcher). This artifact is used by other tools such as Metals

What I didn't check:
- does it work?

Currently, I am swamped so if anyone is willing to push this further - you'll be very welcome to do it!


ckipp01:
:) This is awesome @kpodsiad and so much simpler than all the other stuff we were trying. Today I took some time, published this, and then updated Metals to be publishable with 2.13 🚀 . From what I can tell, it was able to start up my locally published Bloop just fine:

```
❯ bloop about
bloop v1.4.11-40-39ec23e0
```

From what I can tell this seems to be working.

kpodsiad:
> Also, just for downsteam users, do you think we should keep it `bloop` instead of `bloop-core`? That will force an artifact name change for users which if possible we should avoid.

@ckipp01 `+launcherShaded/publishLocal` publishes `bloop-launcher` and this version is probably used by downstream users. I forgot to mention `launcherShaded` in the first comment.

ckipp01:
> > Also, just for downsteam users, do you think we should keep it `bloop` instead of `bloop-core`? That will force an artifact name change for users which if possible we should avoid.
> 
> @ckipp01 `+launcherShaded/publishLocal` publishes `bloop-launcher` and this version is probably used by downstream users. I forgot to mention `launcherShaded` in the first comment.

Ah yea sorry, that's what I mean. Right now this produces `bloop-launcher-core`, so we should probably keep the `bloop-launcher` name right?

kpodsiad:
> > > Also, just for downsteam users, do you think we should keep it `bloop` instead of `bloop-core`? That will force an artifact name change for users which if possible we should avoid.
> > 
> > 
> > @ckipp01 `+launcherShaded/publishLocal` publishes `bloop-launcher` and this version is probably used by downstream users. I forgot to mention `launcherShaded` in the first comment.
> 
> Ah yea sorry, that's what I mean. Right now this produces `bloop-launcher-core`, so we should probably keep the `bloop-launcher` name right?

I think there is some misunderstanding between us. I didn't change module names, they remain the same:
- `launcher` - publishes `bloop-launcher-core`
- `launcherShaded` - publishes `bloop-launcher`

Could you confirm that `++2.13.7 launcherShaded/publishLocal` publishes `bloop-launcher_2.13`?

ckipp01:
> > > > Also, just for downsteam users, do you think we should keep it `bloop` instead of `bloop-core`? That will force an artifact name change for users which if possible we should avoid.
> > > 
> > > 
> > > @ckipp01 `+launcherShaded/publishLocal` publishes `bloop-launcher` and this version is probably used by downstream users. I forgot to mention `launcherShaded` in the first comment.
> > 
> > 
> > Ah yea sorry, that's what I mean. Right now this produces `bloop-launcher-core`, so we should probably keep the `bloop-launcher` name right?
> 
> I think there is some misunderstanding between us. I didn't change module names, they remain the same:
> 
> * `launcher` - publishes `bloop-launcher-core`
> * `launcherShaded` - publishes `bloop-launcher`
> 
> Could you confirm that `++2.13.7 launcherShaded/publishLocal` publishes `bloop-launcher_2.13`?

Ahhhh yes! Sorry, I totally misunderstood you. You're correct. I see so the bloop launcher being published. for 2.13. One other thing I noticed that maybe we can easily change is that bloop launcher is bringing in `scala-xml` 1.2.0 it looks like. That's fine for 2.12 since the 2.12 compiler has also stayed on the 1.x series, but for 2.13 we should try to bump that to 2.x. If not any 2.13 project that tries to use it will hit on 

```
[error]         * org.scala-lang.modules:scala-xml_2.13:2.0.1 (early-semver) is selected over 1.2.0
[error]             +- org.scalameta:metals_2.13:0.10.10-SNAPSHOT         (depends on 2.0.1)
[error]             +- ch.epfl.scala:bloop-launcher_2.13:1.4.11-40-39ec23e0 (depends on 1.2.0)
```
If they are also using `scala-xml`.

kpodsiad:
I wonder if `xml` is really needed, it's used only by the frontend (`CompilerPluginWhitelist`) and backend (`ScalaInstance`). Since the launcher doesn't depend on them anymore maybe `xml` is unnecessary?

kpodsiad:
@tgodzik What do you think about:
- merging this PR as it is
- removing obsoleted modules
- reorganizing publishing tasks & scripts

build.sbt
```diff
@@ -418,15 +419,26 @@ lazy val bloopgunShaded = project
     fork in Test := true,
     bloopGenerate in Compile := None,
     bloopGenerate in Test := None,
-    libraryDependencies ++= List(Dependencies.scalaXml, Dependencies.scalaCollectionCompat)
```
tgodzik, line 421: I wonder what scalaXml was need for here? I can see the usage only in backend, which does not explicitly depend on it. Anyway, it looks like it's save to remove.

tgodzik (approved):
Looks good!

build.sbt
kpodsiad: It might be a leftover from past, but it's not needed now.

kpodsiad:
Publishing launcher for 2.13 required a few followup PR's - #1675, #1676, #1677, #1678

PR #1650, opened 2022-01-07
Remove obsolete modules

kpodsiad:
As discussed in https://github.com/scalacenter/bloop/discussions/1644


tgodzik (changes requested):
Could remove any references to `sbtBloop013` ?

We are getting in the CI:
```
Unspecified value parameters sbtBloop013, sbtBloop10.
        .exportCommunityBuild(
```

We could you maybe add somewhere in the docs what version of the sbt plugin needs to be used? It's the version that is used in metals for <=1.3

You should also remove the code that is being used for sbt 0.13

kpodsiad:
Sorry for the notification noise @tgodzik, PR should be created as a draft.

kpodsiad:
@Arthurm1 There are `gradleBloop211` and `gradleBloop212` modules. Does Gradle use both of them or we can safely remove `gradleBloop211`?

Arthurm1:
Sadly it depends on how people are using other Scala plugins.

Gradle plugins share the same classpath.  So if you have no other Gradle plugins _built with Scala_ in your project then it doesn't matter which you add in: `gradleBloop211` or `gradleBloop212`.  But if you do, then you need to add in the one that matches the Scala version of your other plugins.

I've no idea whether anyone is using `211`.  I'd imagine we'd get complaints about Metals if people were as I think Metals adds in `212`.  Although people can use their own init scripts to get round that - maybe some have

Maybe we should remove `211`.  There haven't been any changes to the plugin for a while.  If people have to use `211` then they can use an old version.

kpodsiad:
@tgodzik what do you think?

ckipp01:
> @Arthurm1 There are `gradleBloop211` and `gradleBloop212` modules. Does Gradle use both of them or we can safely remove `gradleBloop211`?

Is it possibly to check the sonatype downloads to see how many there are of the older?

tgodzik:
> gradleBloop211

Let me try to figure it out.

ckipp01:
This was done in https://github.com/scalacenter/bloop/pull/1852 so I'm going to go ahead and close this one.

Issue #1656, opened 2022-01-13
NumberFormatException when using zulu 1.17 jdk


rreckel:
# Description

bloop exits with exception when using the zulu 1.17 jdk

## Reproduction steps:
Install and use zulu 1.17 jdk using:
`eval $(cs java --jvm zulu:1.17 --env)`

Run bloop in project directory:
`bloop projects --verbose`

Result:
```
[E] For input string: "17-ea"
[T] java.lang.NumberFormatException: For input string: "17-ea"
[T] 	java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
[T] 	java.base/java.lang.Integer.parseInt(Integer.java:668)
[T] 	java.base/java.lang.Integer.parseInt(Integer.java:786)
[T] 	scala.collection.immutable.StringLike.toInt(StringLike.scala:304)
[T] 	scala.collection.immutable.StringLike.toInt$(StringLike.scala:304)
[T] 	scala.collection.immutable.StringOps.toInt(StringOps.scala:33)
[T] 	bloop.data.Project.versionOf$1(Project.scala:195)
[T] 	bloop.data.Project.compareVersions$2(Project.scala:205)
[T] 	bloop.data.Project.compareVersions$1(Project.scala:214)
[T] 	bloop.data.Project.javaVersionAtLeast(Project.scala:255)
[T] 	bloop.data.Project$.enableJavaSemanticdbOptions$1(Project.scala:462)
[T] 	bloop.data.Project$.enableMetalsSettings(Project.scala:513)
[T] 	bloop.engine.BuildLoader$.$anonfun$resolveSemanticDBForProjects$7(BuildLoader.scala:128)
[T] 	scala.collection.immutable.List.map(List.scala:293)
[T] 	bloop.engine.BuildLoader$.$anonfun$resolveSemanticDBForProjects$6(BuildLoader.scala:127)
[T] 	bloop.engine.BuildLoader$.$anonfun$tryEnablingSemanticDB$1(BuildLoader.scala:251)
[T] 	monix.eval.Coeval$Always.apply(Coeval.scala:666)
[T] 	monix.eval.internal.TaskRunLoop$.monix$eval$internal$TaskRunLoop$$loop$1(TaskRunLoop.scala:187)
[T] 	monix.eval.internal.TaskRunLoop$.startWithCallback(TaskRunLoop.scala:260)
[...]

PR #1657, opened 2022-01-13
Allow Java versions with a suffix

tgodzik:
Fixes https://github.com/scalacenter/bloop/issues/1656

Checked it locally and this seems to work.

### DIFF
--- a/frontend/src/main/scala/bloop/data/Project.scala
+++ b/frontend/src/main/scala/bloop/data/Project.scala
@@ -251,8 +251,15 @@ final case class Project(
           getJavaVersionFromJavaHome(f.javaHome)
       )
       .getOrElse(Properties.javaVersion)
-
-    compareVersions(compileVersion, version) >= 0
+      .split("-")
+      .head // needed for versions like 17-ea
+    try {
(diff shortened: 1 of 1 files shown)


ckipp01 (approved):
Easy fix. LGTM!

dos65:
@tgodzik To be fully safe I think we also need to wrap `toInt` call [here](https://github.com/scalacenter/bloop/blob/21be461a75f9483248d00118cf8c1967e8307e62/frontend/src/main/scala/bloop/data/Project.scala#L195) into `Try`.  Just in case if there will be more unexpected formats of writing version in release file.

tgodzik:
> @tgodzik To be fully safe I think we also need to wrap `toInt` call [here](https://github.com/scalacenter/bloop/blob/21be461a75f9483248d00118cf8c1967e8307e62/frontend/src/main/scala/bloop/data/Project.scala#L195) into `Try`. Just in case if there will be more unexpected formats of writing version in release file.

I wrapped the entire compare into try, should be safe now.

dos65 (approved):
:+1:

tgodzik:
It seems the debug tests are failing again :/ no change from our side popped up.

PR #1658, opened 2022-01-13
Fix a few broken links

vitaliis:
You can check the current state here: https://scalacenter.github.io/bloop/docs/build-tools/sbt


tgodzik (approved):
LGTM! Thanks a lot for the fix!

PR #1662, opened 2022-01-21
Update Zinc to 1.6.0 and drop the fork

tgodzik:


backend/src/main/scala/bloop/BloopClassFileManager.scala
```diff
@@ -142,12 +149,26 @@ final class BloopClassFileManager(
       }
     }
 
+    // Idea taken from the default TransactionalClassFileManager in zinc
+    // https://github.com/sbt/zinc/blob/c18637c1b30f8ab7d1f702bb98301689ec75854b/internal/zinc-core/src/main/scala/sbt/internal/inc/ClassFileManager.scala#L183
+    val toBeBackedUp = (classes ++ invalidatedExtraCompileProducts).filter(c =>
```
tgodzik, line 154: Previously, Bloop was using forked Zinc, that allowed to invalidate all symbols from invalidated files from the compiler. That API is not available for the mainline Zinc though, so we need to move them explicitely.

bloopoid:

bloopoid:

bloopoid:

bloopoid:

backend/src/main/scala/bloop/CompilerCache.scala
```diff
@@ -265,7 +281,13 @@ final class CompilerCache(
         processJavaDirArgument(cleanedOptions.indexOf("-s"))
         processJavaDirArgument(cleanedOptions.indexOf("-h"))
 
-        val newJavacOptions = cleanedOptions.toList.asJava
+        output.getSingleOutputAsPath match {
+          case p if p.isPresent => java.nio.file.Files.createDirectories(p.get)
+          case _ =>
+        }
+
+        val outputOption = CompilerArguments.outputOption(output)
```
tgodzik, line 289: It's needed to specify the output directory now, otherwise it will get compiled next to the source file. The change is based on changes to zinc/sbt

backend/src/main/scala/bloop/BloopClassFileManager.scala
```diff
@@ -142,12 +149,26 @@ final class BloopClassFileManager(
...
+      !movedFiles.contains(c) && !generatedFiles(c)
+    )
+    for (c <- toBeBackedUp)
+      if (c.exists)
+        movedFiles.put(c, move(c)).foreach(move)
+
+    classes.foreach { f =>
```
romanowski, line 161: Nit: I am not sure about mixing `forach` and `for` in one method is a good idea

backend/src/main/scala/bloop/CompilerCache.scala
```diff
@@ -309,7 +331,9 @@ final class CompilerCache(
         val invalidated = {
           zincManager match {
             case m: bloop.BloopClassFileManager => m.invalidatedClassFilesSet
-            case _ => zincManager.invalidatedClassFiles().toSet
+            // Bloop uses it's own classfile manager so this should not happen
+            case _ => new HashSet[File]()
```
romanowski, line 335: Shouldn't we at least log something here?

backend/src/main/scala/bloop/ScalaInstance.scala
```diff
@@ -60,7 +68,25 @@ final class ScalaInstance private (
     filename.startsWith(ScalacCompilerName) ||
       (isDotty && (filename.startsWith("dotty-compiler") || filename.startsWith("scala3-compiler")))
   private def hasScalaLibraryName(filename: String): Boolean =
-    filename.startsWith("scala-library")
+    filename.startsWith("scala-library") || filename.startsWith("scala3-library")
```
romanowski, line 71: I really do not like how it works... Maybe we can change bloop options at some point to pass compiler jars more directly?

backend/src/main/scala/sbt/internal/inc/BloopComponentCompiler.scala
```diff
@@ -386,8 +380,8 @@ private[inc] class BloopComponentCompiler(
           val allSourceContents =
             (hydraSourceContents ++ regularSourceContents).map(s => s -> relativize(tempDir, s).get)
 
-          zip(allSourceContents.toSeq, mergedJar)
-          Right(Vector(mergedJar))
+          zip(allSourceContents.toSeq, mergedJar, time = None)
```
romanowski, line 383: :+1: Do we have other palces in bloop where we zip some files?

backend/src/main/scala/sbt/internal/inc/bloop/BloopZincCompiler.scala
```diff
@@ -173,15 +172,18 @@ object BloopZincCompiler {
...
       manager: ClassFileManager,
       logger: ObservedLogger[_]
   ): CompileAnalysis = {
     // Copied from `Incremental` to pass in the class file manager we want
-    def prune(invalidatedSrcs: Set[File], previous0: CompileAnalysis, classfileManager: ClassFileManager): Analysis = {
+    def prune(invalidatedSrcs: Set[VirtualFile], previous0: CompileAnalysis, classfileManager: ClassFileManager): Analysis = {
       val previous = previous0 match { case a: Analysis => a }
-      classfileManager.delete(invalidatedSrcs.flatMap(previous.relations.products).toArray)
+      val toDelete = invalidatedSrcs.flatMap(previous.relations.products).toArray.collect {
+        case vf: VirtualFile => vf
```
romanowski, line 184: Can we get anything else then VirtualFile here (like a entry in the jar)? If so, shouldn't we somehow invalidate

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopIncremental.scala
```diff
@@ -80,11 +88,21 @@ object BloopIncremental {
...
+    def isJrt(path: Path) = path.getFileSystem.provider().getScheme == "jrt"
     val binaryChanges = new DependencyChanges {
-      val modifiedBinaries = initialChanges.binaryDeps.toArray
+      val modifiedLibraries = initialChanges.libraryDeps.toArray
+
+      val modifiedBinaries: Array[File] = modifiedLibraries
+        .map(converter.toPath(_))
+        .collect {
+          // jrt path is neither a jar nor a normal file
+          case path if !isJrt(path) =>
```
romanowski, line 100: What is jrt is modified? We are dropping that information here and this can lead to undercompilations

backend/src/main/scala/sbt/internal/inc/bloop/internal/ConcurrentAnalysisCallback.scala
```diff
@@ -254,15 +309,19 @@ final class ConcurrentAnalysisCallback(
     addUsedNames(addCompilation(addProductsAndDeps(Analysis.empty)))
   }
 
+  // According to docs this is used for build tools and it's not unused in Bloop
```
romanowski, line 312: It looks like a thing for pipelined compilation

romanowski (approved):
I am a bit worried about tons of allocations we introduce with all this Path / FIle >=> VirtualFile conversions. I guess we would need to profile the bloop to see how much of the effect it will have.

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopIncremental.scala
tgodzik: I asked that on discord and Eugene is not even sure if jrt is handled correctly by zinc. Anyways, this a deprecated API and I have no idea if we can get file from it (I tried)

tgodzik:
> I am a bit worried about tons of allocations we introduce with all this Path / FIle >=> VirtualFile conversions. I guess we would need to profile the bloop to see how much of the effect it will have.

I run the benchmarks and I haven't seen any impact, but I can try and minimize the conversions even further ( tried to do it as much as possible)

romanowski:
if there is no regression then it can stay at it is. Let's merge it and optimize it when needed.

lukaszwawrzyk (approved):

backend/src/main/scala/bloop/ScalaInstance.scala
tgodzik: It's not perfect, but we would need to split the Scala configuration to have two lists and we would most likely move this part to the integrations (which might also just have a single Scala classpath). Overall I feel it's better to keep it here and besides `allJars` are only Scala jars, so no user defined libraries will pop up here.

The only risk is if someone forks the compiler, but then a lot more things could break.

backend/src/main/scala/sbt/internal/inc/BloopComponentCompiler.scala
tgodzik: Not from what I can see, this is for the compiler bridge compiled sources.

backend/src/main/scala/sbt/internal/inc/bloop/BloopZincCompiler.scala
tgodzik: We should not really have anything else, classFileManager only works with VirtualFile, so I don't imagine anything else is normally handled.

backend/src/main/scala/bloop/CompilerCache.scala
tgodzik: Added logging

backend/src/main/scala/bloop/BloopClassFileManager.scala
tgodzik: Switched to for completely.

tgodzik:
The failures are unrelated, seems after windows being bumped the tests started failing.

PR #1664, opened 2022-01-26
Support Scala Native 0.4.3+

lolgab:
Scala Native 0.4.3 changed how main methods are resolved: https://scala-native.readthedocs.io/en/latest/changelog/0.4.3.html#changes-to-the-main-method-resolving

The failures in CI seem unrelated...


tgodzik:
Thanks for contributing, I had a quick look at the test failures and they seem unrelated (the debug test started failing randomly on ubuntu recently, no idea why). The windows one about JS looks new though, could it have soemthing to do with the chanegs?

lolgab:
@tgodzik I have no idea if it can be related, maybe you can re-run the workflows to see if it fails again? It seems that only the graalvm-ce one is failing on windows tho.

tgodzik (approved):
LGTM!

PR #1667, opened 2022-01-28
Force newest log4j version

tgodzik:
Turns out Zinc brings in log4j, and despite it all working locally (so no actual way to exploit it), it might cause issues if a company proxy blocks log4j older versions.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -35,6 +35,7 @@ lazy val bloopShared = (project in file("shared"))
     libraryDependencies ++= Seq(
       Dependencies.bsp4s,
       Dependencies.zinc,
+      Dependencies.log4j,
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):

lolgab:
I think the root problem is that Bloop depends on an [2 years outdated fork of Zinc](https://github.com/scalacenter/zinc) instead of depending on the current Zinc, which has the log4j dependency updated.
Also, this way Bloop misses all the improvements to Zinc.

tgodzik:
Good news is that we actually no longer depend on that outdated fork 🥳

We recently merged an update which bumped zinc to newest one and you can already test it out!

PR #1668, opened 2022-02-01
Update coursier-interface to 1.0.6

alexarchambault:
I'm [currently seeing](https://github.com/coursier/setup-action/runs/5028748027?check_suite_focus=true#step:15:40) issues with the native Bloop `1.4.2` launcher on Windows (complaining that it can't find the `sun.misc.BASE64Encoder` class, via directories, shaded in coursier-interface).

Updating coursier-interface might fix it (directories should ship with the right config since https://github.com/dirs-dev/directories-jvm/pull/33, I suspect it might be missing in coursier-interface `1.0.4`).


tgodzik (approved):
Thanks! LGTM

tgodzik:
We might want to do the 1.4.13 release just before zinc is merged so that the recent fixes are accoutned for.

PR #1670, opened 2022-02-07
chore: bump scala 2.12 and 2.13 minor versions

kpodsiad:


build.sbt
```diff
@@ -804,6 +804,7 @@ val allProjects = Seq(
   jsBridge06,
   jsBridge1,
   launcher,
+  launcherTest,
```
tgodzik, line 807: Why do we need to add it here?

tgodzik (changes requested):
It seems graalvm is not working https://github.com/scalacenter/bloop/runs/5148588464?check_suite_focus=true

We might need to bump graalvm version or https://github.com/scalameta/sbt-native-image

kpodsiad:
@tgodzik Just a moment ago I discovered that on my machine I have a newer version of graalVM, let's hope that bumping version will solve issue

kpodsiad:
@tgodzik bumping graal helped and now CI passes. One failed check is related to flaky test `bloop.dap.DebugServerSpec
`

build.sbt
kpodsiad: because typing `compile` or `test` should also be run in `launcherTest` project. I forgot to add that last time and we have undetected errors in the project which were uncaught

tgodzik (approved):
LGTM!

PR #1690, opened 2022-02-22
Fix upload artifacts version

tgodzik:
It seems that the version available on the github page is not the one we can use unfortunately.


tgodzik:
Merging it since this since it needs to be tested on main

Issue #1691, opened 2022-02-22
class and semanticdb files in wrong place and javacOptions repeated for java only + jdk 17


Arthurm1:
Using
```
Java home: /usr/lib/jvm/temurin17
Metals version: 0.11.1+114-58f06ff8-SNAPSHOT
Parse release: 17.0.1
Connected to Build server: Bloop v1.4.13-12-6d14a532
```

Download `https://github.com/Arthurm1/java-bloop-test`
Import into Metals (it's a java-only Gradle project)
Run `Clean compile workspace`
`.class` files appear in same folder as source instead of under `.bloop/XXX`
`.semanticdb` files appear under `workspace/META-INF` instead of under `.bloop/XXX/META-INF`

No `.class` or `.semanticdb` files appear under `.bloop`

Also the javacOptions for jdk 17 have been applied multiple times.
Using `Display build target`....
```

  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED
  -J--add-exports
  -Jjdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
[...]

PR #1694, opened 2022-02-23
Add proper output option for the forked Java compiler

tgodzik:
Previously, we only added the output directory for the local Java compiler, which caused compile artifacts to pop up in the source directories. Now, we properly add the output directory option.

Fixes https://github.com/scalacenter/bloop/issues/1691

I also removed 2 tests that were commented out for a long while.


backend/src/main/scala/bloop/CompilerCache.scala
```diff
@@ -211,13 +211,19 @@ final class CompilerCache(
             val newClasspathValue = newInvalidatedEntry.toFile + File.pathSeparator + classpathValue
             options(classpathValueIndex) = newClasspathValue
 
+            output.getSingleOutputAsPath match {
+              case p if p.isPresent => java.nio.file.Files.createDirectories(p.get)
+              case _ =>
+            }
```
kpodsiad, line 217:
```suggestion
            output.getSingleOutputAsPath.ifPresent{p => 
              java.nio.file.Files.createDirectories(p)
              ()
            }
```
is it better than `p.isPresent` and `p.get` or not?

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
```diff
@@ -300,7 +300,13 @@ class BspMetalsClientSpec(
 
   test("compile with semanticDB") {
     TestUtil.withinWorkspace { workspace =>
-      val `A` = TestProject(workspace, "A", dummyFooScalaAndBarJavaSources)
+
+      object JavacOptions {
+        // This will cause to use the forked javac compiler
```
kpodsiad, line 305: why it'll cause forking? Does every option cause forking?

backend/src/main/scala/bloop/CompilerCache.scala
tgodzik: Both don't look great to be honest and this follows the style in Zinc/sbt

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
tgodzik: Yes, every `-J` option causes forking.

kpodsiad: If you add ` every -J option causes forking.` as a comment it'll be awesome

backend/src/main/scala/bloop/CompilerCache.scala
kpodsiad: I personally don't like pattern matching when the side effect is executed only for one branch, it kind of blurs it, `asScala` + `foreach` would be great.

kpodsiad (approved):

backend/src/main/scala/bloop/CompilerCache.scala
tgodzik: changed to `ifPresent` here and in other place (LocalCompiler) also.

frontend/src/test/scala/bloop/bsp/BspMetalsClientSpec.scala
tgodzik: Added!

PR #1695, opened 2022-02-24
feat: adjust scala test classes bsp endpoint

kpodsiad:
I've added `[skip ci]`  because there's no point in wasting resources now.

It is an implementation of BSP proposal which is described [here](https://github.com/build-server-protocol/build-server-protocol/issues/296#issuecomment-1041525725).


kpodsiad:
CI is green 🥳

tgodzik (approved):
LGTM!

PR #1700, opened 2022-03-14
chore: add scalafix

kpodsiad:
We have to agree on scalafix settings. Currently, they are very similar to the ones that can be found in Metals.

What I've found is that `RemoveUnused` have strange behavior - sometimes it removes only `val name =` and lefts right side of the assignment :/ 
Moreover, we have to be very careful, for example here https://github.com/scalacenter/bloop/blob/1b6633ae67ce0ba508caa947cf88b3daf459ee96/frontend/src/main/scala/bloop/Cli.scala#L508-L518 
there is a side effect (`runAsync`) that is assigned to the variable and we can't remove both sides. Probably the best way would be to iterate manually over those unused warnings.

scalafix can be run on frontend by `frontend/scalafix`

cc: @tgodzik

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,0 +1,18 @@
+rules = [ 
+  DisableSyntax,
+  OrganizeImports,
+  NoValInForComprehension,
+]
+
+OrganizeImports {
+  removeUnused = true
+  expandRelative = true
+  groupedImports = Explode
+  groups = [
+  "re:javax?\\."
+  "scala."
+  "ch.epfl"
+  "bloop"
+  "*"
+]
+}
--- a/backend/src/main/scala/bloop/BloopClassFileManager.scala
+++ b/backend/src/main/scala/bloop/BloopClassFileManager.scala
@@ -1,26 +1,26 @@
 package bloop
 
+import java.io.File
+import java.io.IOException
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.nio.file.StandardCopyOption
+
+import scala.collection.mutable
+import scala.util.Failure
+import scala.util.Success
+import scala.util.Try
+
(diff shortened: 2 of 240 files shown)


.scalafix.conf
```diff
@@ -0,0 +1,22 @@
+rules = [ 
+  DisableSyntax,
+  OrganizeImports,
+  NoValInForComprehension,
```
tgodzik, line 4: What does this rule do?

tgodzik (commented):
Looks sensible to me, did you try running scalafix? I remember I had some issues with it, but I tried fixing some dependencies causing those issues, so might be ok now.

.scalafix.conf
kpodsiad: it bans 
``` scala
for {
  ... <- ...
  val x = ??? // <- this
} yield ...
```

tgodzik: makes sense!

kpodsiad:
I think there's no point in adding scalafix to the CI for now - first we should handle initial sbt load overhead.

tgodzik:
> I think there's no point in adding scalafix to the CI for now - first we should handle initial sbt load overhead.

We can just add it to a separate job, no?

project/Dependencies.scala
```diff
@@ -6,6 +6,9 @@ object Dependencies {
   val Scala212Version = "2.12.15"
   val Scala213Version = "2.13.8"
 
+  val Scala211BinaryVersion = "2.11"
```
tgodzik, line 9: Is this needed? I can't find it being used.

kpodsiad:
I'll do `rewrite generation of the test projects to not happen on sbt load` task from Bloop next steps and then I'll add a CI job.

tgodzik (approved):
LGTM! Thanks for doing this!

PR #1701, opened 2022-03-14
Fix copying tasty files on incremental compilation

tgodzik:
Previously, we removed both generatedClassFile and rebasedClassFile from invalidations, but we only did it for rebased product files such as `.tasty`. Now we do it for both

I am not 100% sure why sometimes invalidations happen for rebased and sometimes for generated files, but that can happen according to the comment above.

What happened here is that the generated files existed, but the rebased one was in the invalidation list, so later on we would skip it.

This was causing the issues in https://github.com/scalameta/metals/pull/3706

### DIFF
--- a/backend/src/main/scala/bloop/BloopClassFileManager.scala
+++ b/backend/src/main/scala/bloop/BloopClassFileManager.scala
@@ -180,14 +180,21 @@ final class BloopClassFileManager(
       // invalidations can happen in both paths, no-op if missing
       allInvalidatedClassFilesForProject.-=(generatedClassFile)
       allInvalidatedClassFilesForProject.-=(rebasedClassFile)
-      supportedCompileProducts.foreach { supportedProductSuffix =>
-        val productName = rebasedClassFile
+
+      def productFile(classFile: File, supportedProductSuffix: String) = {
+        val productName = classFile
           .getName()
           .stripSuffix(".class") + supportedProductSuffix
-        val productAssociatedToClassFile =
-          new File(rebasedClassFile.getParentFile, productName)
-        if (productAssociatedToClassFile.exists())
-          allInvalidatedExtraCompileProducts.-=(productAssociatedToClassFile)
+        new File(classFile.getParentFile, productName)
+      }
+      supportedCompileProducts.foreach { supportedProductSuffix =>
+        val generatedProductName = productFile(generatedClassFile, supportedProductSuffix)
+        val rebasedProductName = productFile(rebasedClassFile, supportedProductSuffix)
+
+        if (generatedProductName.exists() || rebasedProductName.exists()) {
+          allInvalidatedExtraCompileProducts.-=(generatedProductName)
+          allInvalidatedExtraCompileProducts.-=(rebasedProductName)
+        }
       }
     }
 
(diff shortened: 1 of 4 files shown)


kpodsiad:
Question for Zinc newbies, what are `generatedClassFile` and `rebasedClassFile`?

tgodzik:
> Question for Zinc newbies, what are `generatedClassFile` and `rebasedClassFile`? (generated I think is obvious, but the second one isn't)

I think the rebased one is the on where we put copy of files for the client.

tgodzik:
Actually, now I think of it. Rebased might be a new directory and we might always be compiling to a new directory but using the artifacts from the old one. 🤔

tgodzik:
Actually, now I think of it. Rebased might be a new directory and we might always be compiling to a new directory but using the artifacts from the old one. 🤔

tgodzik:
> @tgodzik ELI5: what is this PR about and what are invalidations. pray Incremental compilation isn't my strong side yet :D

Ach sorry, each time anything changes in the code the files get invalidated (meaning they need recompilation) and when those get recompiled we might need to recompile some additional files etc.

backend/src/main/scala/bloop/BloopClassFileManager.scala
```diff
@@ -180,14 +180,21 @@ final class BloopClassFileManager(
...
-          new File(rebasedClassFile.getParentFile, productName)
-        if (productAssociatedToClassFile.exists())
-          allInvalidatedExtraCompileProducts.-=(productAssociatedToClassFile)
+        new File(classFile.getParentFile, productName)
+      }
+      supportedCompileProducts.foreach { supportedProductSuffix =>
+        val generatedProductName = productFile(generatedClassFile, supportedProductSuffix)
+        val rebasedProductName = productFile(rebasedClassFile, supportedProductSuffix)
+
+        if (generatedProductName.exists() || rebasedProductName.exists()) {
```
adpi2, line 194: I don't think it is useful to check if they exists. It is okay to remove a element from a set even if it is not there.

frontend/src/test/scala/bloop/BaseCompileSpec.scala
```diff
@@ -140,6 +140,38 @@ abstract class BaseCompileSpec extends bloop.testing.BaseSuite {
     }
   }
 
+  test(s"compile scala 3 build incrementally") {
```
adpi2, line 143: Does the test fail without the fix?

adpi2 (approved):
While I am not sure to understand why invalidation can happen in both paths, it makes sense to remove all the products of the invalidated class files.

frontend/src/test/scala/bloop/BaseCompileSpec.scala
tgodzik: Yes, this is based on reproduction in the Metals repository.

backend/src/main/scala/bloop/BloopClassFileManager.scala
tgodzik: I agree, but it breaks another test :sweat_smile: 

There is a test that creates a fake nir files and expects them to be removed. So if you change the target from scalaJs to jvm it should remove the old files.

tgodzik:
I will go ahead and merge this since from my understanding this is the safest choice.

I think the main reason for this is that we change output directory on each compilation, which is why there might be different paths for invalidated files and new generated ones.

PR #1704, opened 2022-03-22
Implement buildTarget/inverseSources BSP endpoint

Duhemm:
Previously, the BSP endpoint `buildTarget/inverseSources` was not
implemented, even though Bloop declares itself as supporting this
endpoint.

This commit provides an implementation of this endpoint.


tgodzik (approved):
LGTM!

PR #1706, opened 2022-03-28
ci: break CI actions into smaller pieces

kpodsiad:
I removed `onload` resources generation and then I extracted two additional actions on CI. 
Finer grained CI actions should reduce main tests duration and provide a better experience when debugging CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
   pull_request:
 
 jobs:
-  scalafmt: 
+  scalafmt:
     name: Formatting
     runs-on: ubuntu-latest
     steps:
@@ -19,7 +19,7 @@ jobs:
           java-version: graalvm-ce-java11@21.1.0
       - name: Check formatting
         run: ./bin/scalafmt --test
-  integrations: 
+  integrations:
     name: Build integrations tests
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


kpodsiad:
Before
Screenshot 2022-03-28 at 15 24 12
After
Screenshot 2022-03-28 at 15 24 18

Although these are statistics from a single run they indicate that CI jobs on GraalVM are a bit faster. The most gain can be observed on the slowest windows so it's a double win.

bin/setup-test-projects.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
tgodzik, line 1: Do we need this bash script if the same can be achieved by exportProjectsInTestResources?

docs/contributing-guide.md
```diff
@@ -39,8 +39,10 @@ bsp/cli connection to it.
 1. Clone the [GitHub repository](https://github.com/scalacenter/bloop)
 1. Change directory to bloop's base directory and run `git submodule update --init`.
 1. Run the sbt shell in the base directory.
+1. If it's the first time you should run `exportProjectsInTestResources` command to generate test projects which are needed for tests.
```
tgodzik, line 42: Do we show a sensible exception if the project is not present?

tgodzik (commented):
The approach looks sensible, thanks for looking into that.

dos65:
@kpodsiad I did an alternative fix for this in #1708

tgodzik:
@kpodsiad looks like we can close this one?

kpodsiad:
Actually, after https://github.com/scalacenter/bloop/pull/1708 was merged I pivoted into breaking CI actions into smaller pieces. I think it will be valuable to have more fine-grained actions which are responsible only for testing something more narrow than currently.

kpodsiad:
@tgodzik wdyt?

tgodzik (approved):
LGTM!

PR #1707, opened 2022-03-28
Remove the possibility to read proeprties from.jvmopts

tgodzik:
This is no longer a supported method of specifying the server properties. Users should instead us `$HOME/.bloop/bloop.json`


kpodsiad (approved):

ckipp01 (approved):
😆 I should have never mentioned I used this.

LGTM!

PR #1708, opened 2022-04-01
fix: move resource generation out from onLoad

dos65:


kpodsiad:
@dos65 so this will run generate files lazily?
These resources are also used in bridges' tests, will your fix work too for bridges?

dos65:
> so this will run generate files lazily?

Kind of. Probably not perfect, in this case they will be generated at the moment of computing test resources that is part of import-build. I can move their generation into `test` and `testOnly` but in this case running tests from Metals won't work.

> These resources are also used in bridges' tests, will your fix work too for bridges?
Is it about `jsBridge` and `becnhmarkBridge` ?
Could you specify commands that requires these resources too?

I'm not familiar fully with all these aspects. My plan was to check that `frontend/test` works and CI pass

kpodsiad:
> > These resources are also used in bridges' tests, will your fix work too for bridges?
> > Is it about `jsBridge` and `becnhmarkBridge` ?
> > Could you specify commands that requires these resources too?
> 
> I'm not familiar fully with all these aspects. My plan was to check that `frontend/test` works and CI pass

So, the best part, `frontend/test` resources aren't used only in `frontend/test` module :) They are used also in `ScalaJsToolchainSpec`, which is part of the `jsBridge06` and `jsBridge1`. (`nativeBridge04` tests aren't tested on CI.

I wonder if adding a similar trick to these modules will be good enough?

dos65:
> So, the best part, frontend/test resources aren't used only in frontend/test module :) They are used also in ScalaJsToolchainSpec, which is part of the jsBridge06 and jsBridge1. (nativeBridge04 tests aren't tested on CI.

These project that you mentioned depends on `frontend` including its test axis.
No additional changes is needed.

project/BuildPlugin.scala
```diff
@@ -426,82 +427,72 @@ object BuildImplementation {
...
-      Keys.onLoad.value.andThen { state =>
-        if (sys.env.isDefinedAt("SKIP_TEST_RESOURCES_GENERATION")) state
-        else exportProjectsInTestResources(state, enableCache = true)
-      }
-    }
-
-    def exportProjectsInTestResources(state: State, enableCache: Boolean): State = {
+    def exportProjectsInTestResources(
+        baseDir: File,
+        log: Logger,
```
kpodsiad, line 432: format

kpodsiad (approved):
> These project that you mentioned depends on frontend including its test axis.
No additional changes is needed.

Ok, you're right. Awesome!

Few nitpicks, but it looks good!

kpodsiad (approved):
Works like a charm!

Issue #1713, opened 2022-04-06
Couldn't run bloopGenerate for project


ioleo:
Hi,

I'm using VSCode with Metals + Bloop. Until today everything worked perfectly. now I'm getting `[error] Couldn't run bloopGenerate for ...`:

```
2022.04.06 11:25:04 INFO  [success] Generated .bloop/root-test.json
2022.04.06 11:25:04 INFO  [success] Generated .bloop/externalProtos-test.json
2022.04.06 11:25:04 INFO  [success] Generated .bloop/esApi-test.json
2022.04.06 11:25:04 INFO  [success] Generated .bloop/root.json
2022.04.06 11:25:04 INFO  [success] Generated .bloop/externalProtos.json
2022.04.06 11:25:04 INFO  [success] Generated .bloop/esApi.json
2022.04.06 11:25:04 INFO  [error] Couldn't run bloopGenerate for app-test
2022.04.06 11:25:04 INFO  [error] Couldn't run bloopGenerate for app
2022.04.06 11:25:04 INFO  [success] Total time: 1 s, completed 6 Apr 2022, 11:25:05
```

Which results reduced IDE functionality.

I have tried cloning a fresh repo (just to make sure all the bloop stuff is generated from scratch), as well as manually removing `.metals`, `.bloop`, `.bsp`, `target`, `project/metals.sbt`. The result stays the same.

The only thing that changes since yesterday is I run software updates for my Fedora 35 system. SBT is installed globally (from repository) so it might have been updated. However my understanding is, it does use a local sbt version based on `project/build.properties` so it shouldn't matter.

I have no idea how to progress from this point. Please help me resolve this issue.

ioleo:
* `sbt.version=1.6.2`
* Bloop version `1.4.13`
* Metals extension version `1.13.0`

ioleo:
For some reason switching to `sbt` as build server did not help aswell. Then I uninstalled the `java-latest-openjdk` and replaced it with `java-11-openjdk` (although in both cases `java -version` reported the same), and switched to `sbt` as build server and now got my VSCode hints back.

Although still not clear whats going on with bloop, and I would, if possible, love to get it working.

ioleo:
@dos65 Thanks for takeing interest in this issue. How can I use the latest version with this commit?

tgodzik:
[...]

PR #1714, opened 2022-04-07
fix: sbt-bloop - add cause to error log if bloopGenerate fails

dos65:
This should help with investigations in #1713


tgodzik (approved):
LGTM! Great idea!

PR #1715, opened 2022-04-08
fix: stabilize test project resources generation

dos65:
Fixes the issue that was found by @kpodsiad in https://github.com/scalacenter/bloop/pull/1706#issuecomment-1091992342


kpodsiad (approved):
Few more words about why this PR is needed.
Previously, if one of the `bloop-config` directories was removed it wasn't regenerated.
Now, it works like a charm! I tested it locally so LGTM!

PR #1720, opened 2022-04-12
fix: prevent ignorance of initial changes

dos65:


backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
```diff
@@ -84,7 +84,10 @@ private final class BloopNameHashing(
           initialChangedSources,
           allSources.map(v => v: VirtualFileRef),
           previous
-        ).collect { case f: VirtualFile => f }
```
dos65, line 87: `initialChagedSources` are received as `VirtualFileRef` and `mapInvalidationToSources` doesn't lift them into `VirtualFile`.
Using `collect` here is incorrect as it just drops correctly invalidated sources.

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
```diff
@@ -84,7 +84,10 @@ private final class BloopNameHashing(
...
+        ).map {
+          case f: VirtualFile => f
+          case ref: VirtualFileRef => PlainVirtualFileConverter.converter.toVirtualFile(ref)
+        }
```
dos65, line 90: @tgodzik Is `PlainVirtualFileConverter.converter` usage fine here?
I've seen that in some methods `converter` is passed as parameter.

tgodzik: I think so, we don't use any other converters currently, so this should be fine.

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
tgodzik: och! That looks like the issue i've been having! Awesom!

tgodzik (approved):
LGTM!

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
tgodzik: what is the actual type returned? VirtualFileRef is only an interface, no?

PR #1721, opened 2022-04-13
Fix duplicate compiler classpath and bootclasspath options

tgodzik:
Previously we created compiler arguments on the Bloop side and it turns out that it's already being done in Zinc, so this way we duplicate some of the options. Now, we just use the user scalacOptions, while Zinc takes cares of the other options.

Fixes https://github.com/scalameta/metals/issues/3766


ckipp01 (approved):
LGTM!

kpodsiad (approved):

Issue #1669, opened 2022-02-04
Maven import: support attached tests


nigredo-tori:
Maven allows one submodule to reference compiled tests of another submodule (similar to `.dependsOn(foo % Test)` in SBT). This is achieved using the special `test-jar` dependency type, as described [here](https://maven.apache.org/guides/mini/guide-attached-tests.html). At the moment, Bloop import does not propagate these dependencies to `project.dependencies` in the bloop project file, but it does include the related outputs in `project.classpath`. This means that the dependee project is not recompiled when the dependent is recompiled, leading to various issues (missing classes/methods etc.).

Here's an example:
1. Extract [example.zip](https://github.com/scalacenter/bloop/files/8001048/example.zip) into a temporary directory.
2. Run `mvn test-compile -pl bar -am`.
3. Observe that test files in both `foo` and `bar` submodules were compiled.
4. Run `mvn ch.epfl.scala:maven-bloop_2.13:1.4.12:bloopInstall`
5. Observe that, in `.bloop/example-bar-test.json`, `project.classpath` contains `foo/target/test-classes`, but `project.dependencies` does not contain `example-foo-test`.
6. Run `mvn clean`.
7. Run `bloop compile example-bar-test`. This fails because of a missing type.

PR #1727, opened 2022-04-25
Check for test-jar dependencies

mnd999:
When adding dependencies, check the project artifacts to see if there
are any test-jar dependencies and if so, add the test project as a
dependency as well.

Fixes: #1669


integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
```diff
@@ -122,7 +120,14 @@ object MojoImplementation {
...
     val dependencies =
       session.getProjectDependencyGraph.getUpstreamProjects(project, true).asScala.toList
-    val dependencyNames = dependencies.map(_.getArtifactId()).toList
+    val dependencyNames = dependencies.flatMap(dep => {
+      val matchingArtifacts = project.getArtifacts.asScala.filter(a =>
+        ArtifactUtils.versionlessKey(dep.getArtifact) == ArtifactUtils.versionlessKey(a)
+      )
+      matchingArtifacts.collect {
+        case artifact if artifact.getType == "test-jar" => dep.getArtifactId + "-test"
+      }.toList :+ dep.getArtifactId
```
kpodsiad, line 129: Could we use method instead of ot alias? I find colon operators very confusing.
```suggestion
      }.toList.appended(dep.getArtifactId)
```

mnd999:
I've added the test project from the linked issue.

tgodzik (approved):
Thanks! :tada:  I run the formatting and after the CI passes it should be good to go.

PR #1729, opened 2022-04-27
Fix scoop install script

tgodzik:
Previously it would try to find `bloop.json` in $dir, which did not exists as we were downloading coursier-bloop.json.

Now, we just install specific Bloop version using the default channels, which should work correctly.


kpodsiad (approved):
I trust you that it works ;)

PR #1738, opened 2022-06-18
Use Maven for downloading Gradle test kit

Arthurm1:
Previously, downloading the Gradle test kit for building/testing the Gradle plugin required custom code.

Now the jars are all published [here](https://search.maven.org/search?q=g:dev.gradleplugins) so we can use that instead.  It also means the source jars are available which is a boon for debugging.

Caveat - I'm not that familiar with SBT so hopefully these changes don't have any side effects on the rest of the build.


Arthurm1:
Spurious `bloop.dap.DebugProtocolSpec` test failure.

tgodzik (approved):
LGTM! Looks great to get rid of the additional code.

PR #1740, opened 2022-06-22
Remove class directories in `clean`

### LINKED ISSUE #1388: BSP clean request does not clean IntelliJ classes directory
Followup of https://youtrack.jetbrains.com/issue/SCL-18354

Steps to reproduce the issue:

* Import sbt project via BSP (Bloop)
* Build the project, observe that classfiles are generated in the out/ directory for the IntelliJ BSP client
* Run "Rebuild project" action, the IntelliJ `out/` directory is not cleaned even if the IntelliJ BSP client sends a clean request to Bloop.

I expected the IntelliJ `out/` directory to be cleaned  because the IntelliJ BSP client sent a BSP clean request.

Duhemm:
Previously, `bloop clean` would only clear the results cache, but the
class files would be left untouched on the disk. This commit changes
this, so that the `clean` command doesn't exit until the classfiles of
the last result, and the classes directory of the current client are
deleted.

Fixes #1388

### DIFF
--- a/frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
+++ b/frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
@@ -73,11 +73,29 @@ final case class ResultsCache private (
 
   def allResults: Iterator[(Project, Compiler.Result)] = all.iterator
   def allSuccessful: Iterator[(Project, LastSuccessfulResult)] = successful.iterator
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
This is great! I've been meaning to look into this since clean compiler has not been very helpful in a lot of cases :rocket:

PR #1741, opened 2022-06-23
Bump scala-debug-adapter with fixes for Scala 3

tgodzik:
The most recent milestone works much better thanks to @adpi2 :tada: !

I plan to release 1.5.1 next, because the clean functionality got fixed.

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -631,7 +631,10 @@ final class BloopBspServices(
             }
           )
         case bsp.DebugSessionParamsDataKind.ScalaAttachRemote =>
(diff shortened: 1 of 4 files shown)


adpi2:
I added the `scalaVersion` field in the `DebuggeeRunner`, so it needs to be added in the `BloopDebuggeeRunner` implementation.

adpi2 (approved):
That looks good for now. But maybe I should declare `scalaVersion` as an `Option` so that we can use the scala-debug-adapter to debug pure Java projects.

tgodzik:
> That looks good for now. But maybe I should declare `scalaVersion` as an `Option` so that we can use the scala-debug-adapter to debug pure Java projects.

That might be a good idea. I also noticed that we might needed to add `scalacOptions` debugger will not work currently with `-Xsource:3` etc.

PR #1745, opened 2022-06-29
bugfix: Allow to run using DAP in Java only projects

tgodzik:
This popped up in https://github.com/scalameta/metals/pull/4078

I will follow up with a quick 1.5.2 release.


adpi2 (approved):
LGTM

PR #1747, opened 2022-06-30
Fix resolution of test-jar artifacts

keynmol:
Update: this is indeed a separate issue with the fix provided in https://github.com/scalacenter/bloop/pull/1747/commits/3a1d935f9c37bc8ad26ffdeb79ad367f6302971a

---

_context:_

The problem I saw is related to #1727, but in the case of a test-jar external dependency - it wasn't being added to test scope classpath at all.

~~While tinkering at it, I also noticed that manual translation between Aether and Maven's `Artifact` types is not required, as there seems to be a method for that, used in Maven's own sources.~~ Turns out it's not part of public API so it's better not to use it.

And finally, I added a test to specifically verify that `<...>-tests.jar` is resolved.

I think this "fix" is somewhat different to #1727, and is worth a separate review.

### DIFF
--- a/integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
+++ b/integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
@@ -10,6 +10,7 @@ import scala.util.Success
 import scala.util.Try
 import bloop.config.Config
 import bloop.config.Tag
+import org.apache.maven.RepositoryUtils
 import org.apache.maven.artifact.Artifact
 import org.apache.maven.artifact.ArtifactUtils
 import org.apache.maven.execution.MavenSession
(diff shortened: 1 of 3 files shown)


integrations/maven-bloop/src/main/scala/bloop/integrations/maven/MojoImplementation.scala
```diff
@@ -84,15 +85,7 @@ object MojoImplementation {
...
-        request.setArtifact(
-          new DefaultArtifact(
-            artifact.getGroupId(),
-            artifact.getArtifactId(),
-            classifier,
-            artifact.getType(),
-            artifact.getVersion()
-          )
-        )
+        request.setArtifact(RepositoryUtils.toArtifact(artifact))
```
keynmol, line 88: Reading https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/RepositoryUtils.java#L56-L60 I can see it's an internal tool and shouldn't be used, hence I will revert it.

Will need to duplicate some of its logic.

keynmol: Removed in 8a606df

tgodzik (approved):
LGTM! Thanks :tada:

PR #1750, opened 2022-07-02
feat: capture diagnostic code from Problem

ckipp01:
Needs https://github.com/lampepfl/dotty/pull/15565 to actually do anything. Although it's not working currently and I have no idea why. Just putting this here to test and tie all together.

Shoot, I just realized I opened this up on a branch and not in my fork 😬 sorry about that.

### DIFF
--- a/frontend/src/main/scala/bloop/logging/BspServerLogger.scala
+++ b/frontend/src/main/scala/bloop/logging/BspServerLogger.scala
@@ -102,7 +102,7 @@ final class BspServerLogger private (
     val problemPos = event.problem.position
     val problemSeverity = event.problem.severity
     val sourceFile = toOption(problemPos.sourceFile())
-    // TODO: Add code content
+    val code = toOption(event.problem.diagnosticCode()).map(_.code())
 
     (problemPos, sourceFile) match {
       case (ZincInternals.ZincExistsStartPos(startLine, startColumn), Some(file)) =>
@@ -120,7 +120,7 @@ final class BspServerLogger private (
         val source = Some("bloop")
         val uri = bsp.Uri(file.toPath.toUri)
         val severity = bspSeverity(problemSeverity)
-        val diagnostic = bsp.Diagnostic(pos, Some(severity), None, source, message, None)
+        val diagnostic = bsp.Diagnostic(pos, Some(severity), code, source, message, None)
         val textDocument = bsp.TextDocumentIdentifier(uri)
         val buildTargetId = bsp.BuildTargetIdentifier(event.projectUri)
         client.notify(
--- a/shared/src/main/scala/bloop/reporter/Problem.scala
+++ b/shared/src/main/scala/bloop/reporter/Problem.scala
@@ -1,5 +1,7 @@
 package bloop.reporter
 
+import java.util.Optional
+
 import xsbti.Severity
 
 /** Describes a problem (error, warning, message, etc.) given to the reporter. */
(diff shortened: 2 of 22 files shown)


frontend/src/main/scala/bloop/logging/BspServerLogger.scala
```diff
@@ -91,7 +91,7 @@ final class BspServerLogger private (
     val problemPos = event.problem.position
     val problemSeverity = event.problem.severity
     val sourceFile = toOption(problemPos.sourceFile())
-    // TODO: Add code content
+    val code = toOption(event.problem.diagnosticCode()).map(_.code())
```
ckipp01, line 94: @adpi2 I might need your insight in here like I did in sbt. I'm hitting on the same thing here as I am in Mill and this seems to always be empty when I'm testing with `3.2.1-RC1-bin-20220805-e560c2d-NIGHTLY`. It works find with sbt and is here, but when I'm trying with Mill or Bloop here I can't figure out why it's always empty.

adpi2: FYI I managed to reproduce it, and I am trying to understand why this is happening.

adpi2: So I tried to do inception: debugging Bloop by starting a debugger in Bloop and it just froze after a few steps... :laughing:

adpi2: Here is the :bug::
https://github.com/scalacenter/bloop/blob/a0ca82de1c0239e395045f2a73a1799b89f10ead/backend/src/main/scala/bloop/reporter/Reporter.scala#L140 

The diagnostic code must be copied here:
https://github.com/scalacenter/bloop/blob/a0ca82de1c0239e395045f2a73a1799b89f10ead/backend/src/main/scala/bloop/reporter/Reporter.scala#L97

ckipp01: Bingo
```
[Trace - 08:11:45 pm] Received notification 'build/publishDiagnostics'
Params: {
  "textDocument": {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/minimal/minimal3/src/Main.scala"
  },
  "buildTarget": {
    "uri": "file:/Users/ckipp/Documents/scala-workspace/minimal/minimal3/?id\u003dminimal3"
  },
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 4,
          "character": 18
        },
        "end": {
          "line": 4,
          "character": 19
        }
      },
      "severity": 1,
      "code": "7",
      "source": "bloop",
      "message": "Found:    (\u001b[31m3\u001b[0m : Int)\nRequired: String"
    }
  ],
  "reset": true
}
```
Amazing. So happy to see this. Thanks again!

shared/src/main/scala/bloop/reporter/Problem.scala
```diff
@@ -13,12 +15,21 @@ final case class Problem private (
     /** Position in the source code where the message was triggered */
     position: xsbti.Position,
     /** The category of this problem. */
-    category: String
+    category: String,
+    /** Unique code attatched to the diagnostic being reported */
+    override val diagnosticCode: Optional[xsbti.DiagnosticCode]
```
ckipp01, line 20: The `override val` here is odd but I believe it's needed because of the signature in Javaland. Is there a better way to do this?

ckipp01:
🤔 CI is blowing up, and I don't really get why.

shared/src/main/scala/bloop/reporter/Problem.scala
adpi2: Looks good to me

tgodzik:
Looks like an issue was an old sbt version that didn't use the latest interfaces, but I can't find exactly why this would happen. Maybe some dependencies are added from the current sbt version?

tgodzik (approved):
Looks to be passing! Thanks @ckipp01

ckipp01:
> Looks to be passing! Thanks @ckipp01

Awesome, thanks @tgodzik !

PR #1751, opened 2022-07-04
sbt-bloop and buildpress don't depend on launcher

alexarchambault:
It seems these are spurious dependencies (maybe these are there for historical reasons?). sbt-bloop only needs bloop-config, and buildpress is fine with just bloopgun and bloop-shared.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -687,7 +687,7 @@ def defineShadedSbtPlugin(
 }
 
 lazy val sbtBloop10: Project = project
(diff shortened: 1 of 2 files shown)


alexarchambault:
(It seems I can't re-run the failing CI jobs myself, but the errors look transient, and definitely unrelated to the changes here.)

tgodzik (approved):
LGTM! So of that might been made irrelevant by the changes related to Zinc

PR #1756, opened 2022-07-11
Correctly pick up test-jar dependency sources

keynmol:
### fixes

test-jar dependencies not only have a different suffix for regular jars `-tests.jar`, but they also have a different one for test sources :D `-test-sources.jar` notice that it's `test` not `tests`.

They also need to have a different classifier (`test-sources`).

### refactoring

`resolveArtifact` now takes `sources: Boolean` parameter instead of a wider `classifier: Option[String]` - we only ever use it to download sources, so we can tighten down the method contract.

I would also argue that `resolveArtifact` needs to return `case class ResolvedArtifact(af: Artifact, sources: Boolean, file: File)` so that we don't need to do this string manipulation with jar names, but it would involve larger refactoring than what I'm ready to commit to in this PR.


tgodzik (approved):
LGTM! Thanks for digging into this!

PR #1761, opened 2022-07-15
Cache ScalaJS linkers for incremental linking

lolgab:


bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
```diff
@@ -51,6 +53,46 @@ object JsBridge {
...
+      val scalaJSModuleKind = input.moduleKind match {
+        case ModuleKindJS.NoModule => ScalaJSModuleKind.NoModule
+        case ModuleKindJS.CommonJSModule => ScalaJSModuleKind.CommonJSModule
+        case ModuleKindJS.ESModule => ScalaJSModuleKind.ESModule
+      }
+
+      val useClosure = input.isFullLinkJS && input.moduleKind != ModuleKindJS.ESModule
+
+      val config = StandardConfig()
+        .withOptimizer(true)
```
lolgab, line 88: `.withOptimizer(true)` should be used also with `fastLinkJS`.

bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
```diff
@@ -51,6 +53,41 @@ object JsBridge {
       }
     override def trace(t: => Throwable): Unit = logger.trace(t)
   }
+  private object ScalaJSLinker {
+    private val cache = TrieMap.empty[Path, WeakReference[(JsConfig, Linker)]]
```
sjrd, line 57: `WeakReference`s are not the best choice for memory-sensitive caches. Use a [`SoftReference`](https://docs.oracle.com/javase/8/docs/api/java/lang/ref/SoftReference.html) instead.

bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
sjrd: `withOptimizer(true)` is the default anyway. ;)

bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
```diff
@@ -51,6 +53,41 @@ object JsBridge {
...
+        case ModuleKindJS.NoModule => ScalaJSModuleKind.NoModule
+        case ModuleKindJS.CommonJSModule => ScalaJSModuleKind.CommonJSModule
+        case ModuleKindJS.ESModule => ScalaJSModuleKind.ESModule
+      }
+
+      val useClosure = isFullLinkJS && config.kind != ModuleKindJS.ESModule
+
+      val linkerConfig = StandardConfig()
+        .withOptimizer(true)
+        .withClosureCompilerIfAvailable(useClosure)
```
sjrd, line 83:
```suggestion
        .withClosureCompiler(useClosure)
```
This is JVM-only code, so this stronger method exists.

PR #1762, opened 2022-07-19
Don't give a local compiler when disallowed

Duhemm:
Previously, Bloop's `CompilerCache` could return instances of
`JavaCompiler` relying on using the local JDK Java compiler API
("unforked" compilation) in cases where the javac options included
options expected to be passed to the runtime system (options starting
with `-J`, eg. `-J-Dfoo=bar`).

This happened, because the compiler instances are cached and the path to
the `javac` binary was used as cache key. This means that if 2 projects
A and B share the same path to the `javac` binary, but only project B
has javac options that should be passed to the runtime system, then
Bloop could use a local compiler to compile project B, because the cache
would have been populated with the local compiler used to compile
project A. See the test cases included in this commit.

This commit fixes this issue by encoding in the cache key whether a
local compiler can be used.


tgodzik (approved):
LGTM!

PR #1764, opened 2022-07-21
fix: limit the amount of logs that launcher keeps in memory

dos65:
I'm noticed that this outputStream might produce an unhandled OOM error when it used from Metals (~80MB in memory).
Can be reproduced on long running session with a lot of diagnostics.

### DIFF
--- a/launcher-core/src/main/scala/bloop/launcher/bsp/BspBridge.scala
+++ b/launcher-core/src/main/scala/bloop/launcher/bsp/BspBridge.scala
@@ -41,11 +41,8 @@ final class BspBridge(
     bspServerStatus = None
   }
 
-  case class RunningBspConnection(bsp: BspConnection, out: ByteArrayOutputStream) {
-    def logs: List[String] = {
-      val contents = new String(out.toByteArray(), StandardCharsets.UTF_8)
-      contents.linesIterator.toList
-    }
+  case class RunningBspConnection(bsp: BspConnection, out: BspBridge.LogsRecordingStream) {
+    def logs: Seq[String] = out.logs
   }
 
   /**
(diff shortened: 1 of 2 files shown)


launcher-core/src/main/scala/bloop/launcher/bsp/BspBridge.scala
```diff
@@ -41,11 +41,8 @@ final class BspBridge(
     bspServerStatus = None
   }
 
-  case class RunningBspConnection(bsp: BspConnection, out: ByteArrayOutputStream) {
-    def logs: List[String] = {
```
tgodzik, line 45: Actually, where are these logs used? Don't we send them out immediately?

dos65: in `BspBridge.waitForOpenBsp`

There is some logic over log message - https://github.com/scalacenter/bloop/blob/c1fd321b3f62da4fab79b19794616990c7b656c8/launcher-core/src/main/scala/bloop/launcher/bsp/BspBridge.scala#L123

tgodzik (approved):
LGTM!

tgodzik:
Actually, just realised that the test fail due to the new test: https://github.com/scalacenter/bloop/runs/7469473959?check_suite_focus=true

tgodzik:
@dos65 I think this one only needs scalafmt and we should be good to merge!

dos65:
@tgodzik it's workflow failure. Could you rerun it?
```
 Installing graalvm-ce-java11@21.1.0
  Downloading graalvm-ce-java11@21.1.0 (https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-21.1.0/graalvm-ce-java11-linux-amd64-21.1.0.tar.gz) 
  Extracting /tmp/jabba-d-667052340 to /home/runner/.jabba/jdk/graalvm-ce-java11@21.1.0 
  gzip: invalid header 
  0/212
  Error: exec: Downloading graalvm-ce-java11@21.1.0 (https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-21.1.0/graalvm-ce-java11-linux-amd64-21.1.0.tar.gz) 
  Extracting /tmp/jabba-d-667052340 to /home/runner/.jabba/jdk/graalvm-ce-java11@21.1.0 
```

PR #1766, opened 2022-07-26
Add sbt-vspp for publishing the BLOOP SBT plug-in in a Maven-consistent format

ScalaWilliam:
This is to enable usage of Bloop/Metals in Enterprise environments where only the valid-POM format is accepted for JAR downloads (meaning most SBT plug-ins don't work -- currently making Scala unusable from VS Code at organizations with stricter security policies), for thousands of developers.

--

How this is tested locally:

`publishM2`, then in the target project that uses sbt-bloop, in its `project/metals.sbt` add, e.g.:

```
libraryDependencies += "ch.epfl.scala" % "sbt-bloop_2.12_1.0" % "1.5.2-13-15ded987-20220726-1436"
// only needed if published to local .m2 via publishM2
resolvers += Resolver.mavenLocal
```

More background here - https://github.com/esbeetee/sbt-vspp - more than happy to hop on a video call to discuss - as I appreciate that this is a non-issue for those that have a direct connection to Maven Central.

### DIFF
--- a/docs/build-tools/sbt.md
+++ b/docs/build-tools/sbt.md
@@ -21,6 +21,11 @@ Install bloop by adding the following line to your `project/plugins.sbt`:
 addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "@VERSION@")
 ```
 
+If this does not work due to using an internal Artifactory, use instead:
+```sbt
+libraryDependencies += "ch.epfl.scala" % "sbt-bloop_2.12_1.0" % "@VERSION@"
+```
+
 After that, start up sbt or reload your sbt shell to load the plugin.
 
 ## Export your build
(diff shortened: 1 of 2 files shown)


dos65:
@ScalaWilliam Thanks for the contribution!

I went trough all attached info and original issue in sbt - didn't know about this issue :sweat_smile: 
I'm wondering what sbt mantainers think about this approach. If it's valid one maybe it would be better just to include right it into sbt and all plugin will start supporting this scheme after sbt upgrade.

tgodzik:
> @ScalaWilliam Thanks for the contribution!
> 
> I went trough all attached info and original issue in sbt - didn't know about this issue sweat_smile I'm wondering what sbt mantainers think about this approach. If it's valid one maybe it would be better just to include right it into sbt and all plugin will start supporting this scheme after sbt upgrade.

Yeah, I am kind of curious what it's not being done by default :thinking: I am totally ok with merging, just I am surprised it's not being handled properly in the first place.

dos65:
@adpi2 Could you take a look at this? Is there any chance that it will be ported to sbt?

tgodzik (approved):

tgodzik:
Let's merge it an see if it works correctly in Metals.

PR #1768, opened 2022-07-30
Support scala android plugin on Gradle

Arthurm1:
Currently any Gradle Android projects only get a `java` section not a `scala` section in the Bloop json exports because there's no real link between the Android sourcesets and java/scala sourcesets so the `ScalaCompile` info is just floating about.

This PR assumes that if the project has `ScalaCompile` applied to it then that info should be exported.

So https://github.com/dsvdsv/scala-android-plugin is now supported.  It looks dead as it doesn't support later versions of Gradle or Android, but why not.
I've left the test for it commented out as the `GradleRunner` seems to apply plugins in a different order than real-life Gradle so the test fails on configuration.  Testing this on a local `build.gradle` file appears to work.

Hopefully this should solve https://github.com/scalameta/metals/issues/4036


Arthurm1:
Spurious test failure `bloop.launcher.LatestStableLauncherSpec`

tgodzik (approved):
LGTM!

Issue #960, opened 2019-07-12
Respond with adequate test reports to BSP test


samvel1024:
[Here](https://gist.github.com/samvel1024/c3a434a0ef6f3a555730d14ef9392166) is the structure of my project. Basically it's the hello world project from SBT examples. I have successfully exported it to bloop.

Now I am trying to run it with BSP buildTaregt/test. [Here](https://gist.github.com/samvel1024/9b582103df7d2003253efc97ee4e8af4) is the log of all JSON RPC messages. 

1. I successfully receive all the logs generated during testing. Pay attention to line 191, it says `1 tests, 1 failed`
2. Line 207 says `Missing configured test frameworks in root` which is probably why, on line 214 I get statusCode 1, which probably maps to ch.epfl.scala.bsp4j.StatusCode.OK (ordinal = 1). Seems like bloop is not able to understand the results generated by the test framework. 

However, when running from CLI I get information about failing tests (at least exit code is not 0), which means that behaviour of BSP server in Bloop is inconsistent with the CLI (if I don't miss some important configuration).

<img width="432" alt="Screen Shot 2019-07-12 at 15 17 16" src="https://user-images.githubusercontent.com/19408688/61124523-28fb3400-a4b8-11e9-98c5-05a8229a2246.png">

jvican:
Bloop uses the same logic for both, the only difference here is in how the logs are reported to the client and, most importantly, the exit code, which is OK via BSP (a bug) but failed in the CLI. The `buildTarget/test` endpoint has not yet been completely finished and these are well-known issues. I'll have a look at fixing them soon.

PR #1769, opened 2022-07-30
Improve `test` BSP service

oyvindberg:
- add `TestRun` and `TestRuns` structures to aggregate all produced `TestSuiteEvent.Results`.
- propagate all test results back up to service
- use test results to return `bsp.StatusCode.Error` on failed tests (Fixing #960)
- ensure that the compilation before test finished with `StatusCode.Ok`

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -20,8 +20,12 @@ import ch.epfl.scala.bsp
 import ch.epfl.scala.bsp.BuildTargetIdentifier
 import ch.epfl.scala.bsp.MessageType
 import ch.epfl.scala.bsp.ShowMessageParams
+import ch.epfl.scala.bsp.CompileResult
+import ch.epfl.scala.bsp.StatusCode
 import ch.epfl.scala.bsp.Uri
 import ch.epfl.scala.bsp.endpoints
+import ch.epfl.scala.bsp.SbtBuildTarget.encodeSbtBuildTarget
+import ch.epfl.scala.bsp.ScalaBuildTarget.encodeScalaBuildTarget
 import ch.epfl.scala.debugadapter.DebugServer
 import ch.epfl.scala.debugadapter.DebuggeeRunner
 
@@ -689,21 +693,11 @@ final class BloopBspServices(
   }
 
   def test(params: bsp.TestParams): BspEndpointResponse[bsp.TestResult] = {
-    def test(
-        project: Project,
-        state: State
-    ): Task[State] = {
+    def test(project: Project, state: State): Task[Tasks.TestRuns] = {
       val testFilter = TestInternals.parseFilters(Nil) // Don't support test only for now
-      val handler = new LoggingEventHandler(state.logger)
-      Tasks.test(
-        state,
-        List(project),
-        Nil,
-        testFilter,
-        ScalaTestSuites.empty,
-        handler,
-        mode = RunMode.Normal
-      )
+      val handler = new BspLoggingEventHandler(state.logger, client)
+
+      Tasks.test(state, List(project), Nil, testFilter, ScalaTestSuites.empty, handler, mode = RunMode.Normal)
     }
 
     val originId = params.originId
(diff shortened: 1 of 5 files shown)


frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
```diff
@@ -112,53 +128,43 @@ object Tasks {
...
-    import state.logger
-    implicit val logContext: DebugFilter = DebugFilter.Test
+  ): Task[TestRuns] = {
 
-    var failure = false
     val testTasks = projectsToTest.map { project =>
-      /* Intercept test failures to set the correct error code */
-      val failureHandler = new LoggingEventHandler(state.logger) {
+      val resultsBuilder = List.newBuilder[TestSuiteEvent.Results]
+      val handleAndStore = new LoggingEventHandler(state.logger) {
```
kpodsiad, line 135: I'd add explicit warning/information that `resultsBuilder` will be changed by event handler (`handleAndStore`) during test run but it's safe to use that because `result` is called after test run is finished - in TestTask.runTestSuites(...).map(...)

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
```diff
@@ -112,53 +128,43 @@ object Tasks {
...
 
-    runAll(testTasks).map { exitCodes =>
+    runAll(testTasks).map { testRuns =>
       // When the test execution is over report no matter what the result is
       testEventHandler.report()
-      logger.debug(s"Test suites failed: $failure")
-      val isOk = !failure && exitCodes.forall(_ == 0)
-      if (isOk) state.mergeStatus(ExitStatus.Ok)
-      else state.copy(status = ExitStatus.TestExecutionError)
+      TestRuns(testRuns)
```
kpodsiad, line 167: I wonder if `testEventHandler` is suitable for handling parallel execution of task, I see mutable collections there and plain Ints :/ but ofc this isn't concern of this PR

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -708,25 +702,33 @@ final class BloopBspServices(
...
-
-                  sequentialTestExecution.materialize.map(_.toEither).map {
-                    case Right(newState) =>
+            case (newState, Right(CompileResult(_, StatusCode.Ok, _, _))) =>
+              val sequentialTestExecution: Task[Seq[Tasks.TestRuns]] =
+                Task.sequence(mappings.map { case (_, p) => test(p, state) })
+
+              sequentialTestExecution.materialize.map {
+                case Success(testRunss) =>
+                  testRunss.reduceOption(_ ++ _) match {
```
kpodsiad, line 711:
```suggestion
                case Success(testRunsSeq) =>
                  testRunsSeq.reduceOption(_ ++ _) match {
```

kpodsiad (commented):
Solid job @oyvindberg. I think we should add a testcase for fixed scenario (something near e.g. BspMetalsClientSpec?)

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
```diff
@@ -93,6 +93,22 @@ object Tasks {
...
   }
 
+  case class TestRun(project: Project, exitCode: Int, results: List[TestSuiteEvent.Results]) {
+    def testsFailed: Boolean =
+      results.exists(_.events.exists(e => TestFailedStatus.contains(e.status())))
+    def processFailed: Boolean =
+      exitCode != 0
+    def failed: Option[ExitStatus] =
+      if (processFailed || testsFailed) Some(ExitStatus.TestExecutionError) else None
+  }
```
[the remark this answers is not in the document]
oyvindberg: The idea was to hold on to this state a bit longer before discarding it. by propagating it back out to the BSP test service, we gain the opportunity to return a structure for instance in `TestResult#data` which summarizes all the test runs. Or write test assertions based on it. Or print a summary to screen. or something else.

That said, it's clearly not needed, I can remove it if you want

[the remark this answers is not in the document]
kpodsiad: Extending `TestResult#data` with structured summary would be nice indeed. Can you create an issue and add todo for that?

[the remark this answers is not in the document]
oyvindberg: #1771

frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
oyvindberg: the test code seems a bit half baked, so I wouldn't be surprised if some fixes are necessary. A related issue I came across is that clients are not (AFAIU) able to request parallel execution at all for now. Perhaps that could be a bloop specific field inside the request data strucure again.

oyvindberg:
I agree there should be tests. I'm thinking to duplicate the structure for the compile tests and adapt the structure for test end point. or something like that. It'll be a day or two before I get to it

oyvindberg:
added a test now

oyvindberg:
Just an update.

I've had an hour here and there over the last week, and numerous times I've jumped in to fix this test. I'll probably get there, but it's incredibly slow and difficult work.

tgodzik:
> Just an update.
> 
> I've had an hour here and there over the last week, and numerous times I've jumped in to fix this test. I'll probably get there, but it's incredibly slow and difficult work.

Do you need help with that? What is slowing you down most?

oyvindberg:
It would help a lot. I'm working towards a MVP of [bleep](https://github.com/oyvindberg/bleep) and really need this feature, but I have much less time than I would have wanted.

It's a deep call stack, and some state handling I haven't had time to understand. 

- I'm currently stuck debugging why my tests within the test harness are not run because `compileAnalysis` is empty (see TestTask.scala:83)
- I commented out that check, and then it's very puzzling that I don't see all the test events this PR needs when running the tests. I hope that is not an junit vs scalatest issue. if it is, the main code in this PR needs to adapt to whatever cues sbt uses to determine failed tests

oyvindberg:
Re slow, it's a combination of many things. I'm unable to run sbt in arm64 mode (is there even an issue for that?), so everything with the build is just incredibly slow. roundtrip for compiling `frontend` plus tests and then `testOnly` the test I want to run is like half a minute. I needed to change the default timeout value for the tests from 5 to 10 seconds for things to work at all. I know much of this responsibility lies outside of bloop, and I know the build itself is beeing looked at, just wanted to air some frustration :)

tgodzik:
I will try and take a look, but that might be next week unfortunately.

oyvindberg:
That would be perfect. Gives me some time to try to figure it out myself first :)

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -1094,7 +1102,7 @@ final class BloopBspServices(
     def matchesGlobs(document: Path, project: Project): Boolean =
       project.sourcesGlobs.exists(glob => glob.matches(document))
     val document = AbsolutePath(request.textDocument.uri.toPath).underlying
-    ifInitialized(None) { (state: State, _: BspServerLogger) => 
+    ifInitialized(None) { (state: State, _: BspServerLogger) =>
```
dos65, line 1105: might be reverted

dos65 (commented):
@oyvindberg Thanks for contribution!
Overall looks good. The only issue is that `BspTestSpec.`bsp test fails`` test case doesn't pass. No idea why. Could you look at it?

tgodzik (approved):

oyvindberg:
thanks @tgodzik !

Issue #980, opened 2019-07-25
Watch changes while compilation is happening


pkolaczk:
1. Run
 
```
bloop compile `bloop projects` --watch
```

2. Open a multi-module bloop-enabled project in an IDE.
3. Go to Refactoring and Rename a class that's used in multiple modules.
4. See plenty compilation failures (missing class: (new name))
5. Bloop does *not* retry the failed compilation after the renaming is finished.
6. Rerun compilation manually - compiles fine.

Proposed solutions:
A. Watch for changes happening *during* compilation (or any watched action). Detect and record any change since the time the compilation *started*. If such change happens, force another round, because the current compilation session might have missed it. You can be even quite smart with that - e.g. don't restart the whole process, but just add changed source files and source files that failed to compile in the current stage to the invalidated set and start from there.

B. Wait a 0.1 second after detecting a file change before launching compilation. If another change is detected during that time, wait another 0.1s, etc. This is simple and I guess it would work in most cases, avoiding needless recompilation.

jvican:
> Watch for changes happening during compilation (or any watched action)
  
I've tried it but it's quite difficult to get it working reliably. The file watching logic is far from trivial and all changes there heavily affect the cancellation of the processes in negative ways, so this is not a trivial change.
  
> . You can be even quite smart with that - e.g. don't restart the whole process, but just add changed source files and source files that failed to compile in the current stage to the invalidated set and start from there.
  
Thought about it, we have the infrastructure about it but don't want to implement it because it's too complex for little gain.
  
> Wait a 0.1 second after detecting a file change before launching compilation.
  
We're already doing this, too, but not for such a long time. I want file watching to be as snappy as possible so increasing timeout is not an option, assuming the timeout is an issue.
  
It'd be helpful if we had a tiny reproduction or even a test case in `FileWatchingSpec`.

pkolaczk:
> I've tried it but it's quite difficult to get it working reliably. 

I don't understand what's the difficulty here. Can you elaborate a bit more?

There is a file-change detection logic already implemented. The project recompiles fine if I trigger compilation manually after refactoring, even though background compilation failed.

If a change happens *during* compilation, just store the change event into a queue instead of ignoring. When the compilation finishes, check the pending file changes queue. If not empty, force recompile.

> I want file watching to be as snappy as possible so increasing timeout is not an option.
Fair enough.

pkolaczk:
Possibly related: 
https://github.com/sbt/zinc/issues/683

PR #1772, opened 2022-08-02
fix: check if products were removed in inc compilation

dos65:
Do it in the same way how it's implemeted in zinc - https://github.com/sbt/zinc/blob/189d33bcf6678fcc86e7e3e438c551d6cc51be8a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala#L396-L405

Technically it shouldn't be an issue, if user don't touch clasees then
there is no need to check `removedProducts`.

However, sometimes when I do a lot of switches between branches in
combination with Metals by some reason I'm getting an invalid state of project where some products were
removed and their sources aren't presented in changed files.

### DIFF
--- a/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
+++ b/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
@@ -1,5 +1,7 @@
 package sbt.internal.inc.bloop.internal
 
+import scala.collection.parallel.immutable.ParVector
+
 import _root_.bloop.UniqueCompileInputs
 import _root_.bloop.tracing.BraveTracer
 import monix.eval.Task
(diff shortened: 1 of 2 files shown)


tgodzik (commented):
What kind of issues do you get? Bloop compiles things to a separate readonly directory and then it copies over the results (at least I think this is the intention). So no unused files should show up. Maybe it's an issue that we don't recompile when removing files?

dos65:
@tgodzik 
Sometimes after switching git branches I'm getting errors like: `object $name is no member of package` | `not found value $name` and others about not having a class and there is no way to get rid of them. This happens once in 2-3 weeks on a different projects.

The situation is exactly as it described in a test that I added with an only difference that class files just don't exists while the current analysis doesn't know about that. Unfortunately, I haven't found a reliable way to reproduce it.

I guess maybe Metals might trigger compilation in a half way of checkout (git performs a file rewrite as delete + write a fresh one) or it somehow related to https://github.com/sbt/zinc/issues/683 or https://github.com/scalacenter/bloop/issues/980

tgodzik (approved):
Ok, let's merge this in that case. If we can reduce false positives that's only for the better.

PR #1773, opened 2022-08-03
fix: avoid resource file updates in test

dos65:
Previously, if you run `bloop.CompileSpec` locally you might notice that it rewrites `App.scala` in `bloop/frontend/src/test/resources/`.


tgodzik (approved):
LGTM!

Issue #1718, opened 2022-04-11
RuntimeException: Could not find any viable Base64 encoder


michelou:
Here are 2 scenarios (Windows 11 Pro) illustrating the same issue : 

### Scenario 1 : Java 8

```
> where python java
c:\opt\Python-3.10.4\python.exe
C:\opt\jdk-temurin-1.8.0u322-b06\bin\java.exe

>echo %JAVA_HOME%& echo %PYTHON_HOME%
C:\opt\jdk-temurin-1.8.0u322-b06
C:\opt\Python-3.10.4
```

**Test 1a** with standalone executable [`bloop-x86_64-pc-win32.exe`](https://github.com/scalacenter/bloop/releases/tag/v1.4.13) : 

```
> c:\opt\bloop-1.4.13\bloop.exe server
Attempting a connection to the server...
Resolving ch.epfl.scala:bloop-frontend_2.12:1.4.13...
Exception in thread "main" java.lang.ExceptionInInitializerError
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.initialize(ClassInitializationInfo.java:315)
        at coursierapi.shaded.coursier.cache.shaded.dirs.ProjectDirectories.from(ProjectDirectories.java:298)
        [...]
        at coursierapi.Fetch.create(Fetch.java:99)
        at bloop.bloopgun.core.DependencyResolution$.resolveWithErrors(DependencyResolution.scala:62)
        [...]
        at bloop.bloopgun.Bloopgun.main(Bloopgun.scala)
Caused by: java.lang.RuntimeException: Could not find any viable Base64 encoder! (java.util.Base64 failed with: java.util.Base64$Encoder.encodeToString([B))
        at coursierapi.shaded.coursier.cache.shaded.dirs.Util.initBase64Encoding(Util.java:188)
        at coursierapi.shaded.coursier.cache.shaded.dirs.Util.toUTF16LEBase64(Util.java:170)
        at coursierapi.shaded.coursier.cache.shaded.dirs.Util.<clinit>(Util.java:49)
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.invokeClassInitializer(ClassInitializationInfo.java:375)
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.initialize(ClassInitializationInfo.java:295)
        ... 20 more
[...]

PR #1775, opened 2022-08-05
bugfix: Try and fix issue with Base64 reflection

tgodzik:
Based on https://github.com/dirs-dev/directories-jvm/blob/aa87fc1705fec69bc7f486dae6910306a50c221a/src/main/resources/META-INF/native-image/dev.dirs/reflect-config.json

Fixes https://github.com/scalacenter/bloop/issues/1718

### DIFF
--- a/bloopgun/src/main/graal/reflection.json
+++ b/bloopgun/src/main/graal/reflection.json
@@ -7,5 +7,23 @@
         "allPublicMethods": true,
         "allDeclaredClasses": true,
         "allPublicClasses": true
+    },
+    {
+        "name": "java.util.Base64",
+        "allDeclaredConstructors": true,
+        "allPublicConstructors": true,
+        "allDeclaredMethods": true,
+        "allPublicMethods": true,
+        "allDeclaredClasses": true,
+        "allPublicClasses": true
+    },
+    {
+        "name": "java.util.Base64$Encoder",
+        "allDeclaredConstructors": true,
+        "allPublicConstructors": true,
+        "allDeclaredMethods": true,
+        "allPublicMethods": true,
+        "allDeclaredClasses": true,
+        "allPublicClasses": true
     }
 ]


tgodzik:
Anyone willing to test it out on windows? Otherwise I can do that probably next week and run a new release.

tgodzik:
I tried releasing it locally on Windows, but it seems a bottomless rabbit hole :/

Try is a simple instruction:
`gu.cmd install native-image` which should install `native-image.cmd` in the same folder, but it doesn't! 

Spent an hour trying to google/fix that, but to no avail. I found `native-image.exe`, is that the same thing? It's in a totally different folder, but maybe it does work?

Then it started failing with not finding cl.exe, which needed Visual Studio being installed (!!!) and it still didn't help.

I think we will check this when the release runs :/

ckipp01:
I honestly know nothing about what this is doing. Do you have any links or reference to what these are?

tgodzik:
There is a bit more here: https://www.graalvm.org/22.1/reference-manual/native-image/Reflection/

But I ma kind of guessing here and copying from dev-dirs, which seems to have that setting

adpi2 (approved):
Ok, so this dirs-dev library tries to load `java.util.Base64` 
by reflection [here](https://github.com/dirs-dev/directories-jvm/blob/aa87fc1705fec69bc7f486dae6910306a50c221a/src/main/java/dev/dirs/Util.java#L185-L197) and to call `getEncoder` which probably returns a `java.util.Base64$Encoder`.

I wonder if we should also declare `sun.misc.BASE64Encoder` here. Probably not if we compile with Java 11 or greater.

I cannot test this change now but it seems perfectly legitimate to me.

tgodzik:
> Ok, so this dirs-dev library tries to load `java.util.Base64` by reflection [here](https://github.com/dirs-dev/directories-jvm/blob/aa87fc1705fec69bc7f486dae6910306a50c221a/src/main/java/dev/dirs/Util.java#L185-L197) and to call `getEncoder` which probably returns a `java.util.Base64$Encoder`.
> 
> I wonder if we should also declare `sun.misc.BASE64Encoder` here. Probably not if we compile with Java 11 or greater.
> 
> I cannot test this change now but it seems perfectly legitimate to me.

I will try and build it on CI, might be possible to download it and test on windows.

tgodzik:
> I will try and build it on CI, might be possible to download it and test on windows.

That actually worked! :tada:

ckipp01 (approved):
Alrighty, finally read through the link. Thanks! From what I can tell, yea this makes sense. LGTM

PR #1749, opened 2022-07-01
WIP: use setup-java

keynmol:

PR #1776, opened 2022-08-16
Truncate file timestamp in `AttributedPath` to milliseconds

oyvindberg:
- this makes it consistent between old and new JVMs, and makes a bunch of tests happy which among other things diff file timestamps

### DIFF
--- a/shared/src/main/scala/bloop/io/Paths.scala
+++ b/shared/src/main/scala/bloop/io/Paths.scala
@@ -19,6 +19,8 @@ import scala.collection.mutable
 
 import dev.dirs.ProjectDirectories
 
+import java.time.temporal.ChronoUnit
+
 object Paths {
   private val projectDirectories = ProjectDirectories.from("", "", "bloop")
   private def createDirFor(filepath: String): AbsolutePath =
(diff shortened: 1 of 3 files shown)


keynmol:
Haha, I thought I was going crazy :D https://github.com/scalacenter/bloop/pull/1749/files#r912112085

oyvindberg:
> Haha, I thought I was going crazy :D https://github.com/scalacenter/bloop/pull/1749/files#r912112085

I guess everybody meets this error for the first time, it's certainly very confusing!

shared/src/main/scala/bloop/io/Paths.scala
```diff
@@ -80,7 +82,24 @@ object Paths {
     out.toList
   }
 
-  case class AttributedPath(path: AbsolutePath, lastModifiedTime: FileTime, size: Long)
+  // sealed abstract is an abomination for scala 2.12 to get private which actually works
+  sealed abstract case class AttributedPath(
```
tgodzik, line 86: Why change it to an abstract class?

oyvindberg: What the comment says really. To force all usage to go through the `of` constructor method. 

Older scala versions and private case class constructors is a surprisingly complex topic. For instance this article touches on it https://tuleism.github.io/blog/2020/scala-smart-constructors/ 

That said, now that the compiler showed me all usage i can change it back to something more normal if this is too weird. I just like having the guarantee

tgodzik: Ach, sorry! I saw that previously when checking and then forgot :sweat:

tgodzik (approved):

PR #1778, opened 2022-08-16
show stack traces of `Error`s thrown in asynchronous code

oyvindberg:
Bloop will only report `Boxed error`, see https://stackoverflow.com/questions/17265022/what-is-a-boxed-error-in-scala

I guess there might be a better place to put this. I wound up with some `ClassDefNotFoundError` and the likes when debugging

### DIFF
--- a/frontend/src/main/scala/bloop/Cli.scala
+++ b/frontend/src/main/scala/bloop/Cli.scala
@@ -96,7 +96,7 @@ object Cli {
       case x: java.util.concurrent.ExecutionException =>
         // print stack trace of fatal errors thrown in asynchronous code, see https://stackoverflow.com/questions/17265022/what-is-a-boxed-error-in-scala
         // the stack trace is somehow propagated all the way to the client when printing this
-        x.getCause.printStackTrace()
+        x.getCause.printStackTrace(ngContext.out)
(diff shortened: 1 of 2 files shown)


frontend/src/main/scala/bloop/Cli.scala
```diff
@@ -89,8 +89,16 @@ object Cli {
...
-    val exitStatus = run(cmd, NailgunPool(ngContext))
-    ngContext.exit(exitStatus.code)
+    try {
+      val exitStatus = run(cmd, NailgunPool(ngContext))
+      ngContext.exit(exitStatus.code)
+    } catch {
+      case x: java.util.concurrent.ExecutionException =>
+        // print stack trace of fatal errors thrown in asynchronous code, see https://stackoverflow.com/questions/17265022/what-is-a-boxed-error-in-scala
+        // the stack trace is somehow propagated all the way to the client when printing this
+        x.getCause.printStackTrace()
```
tgodzik, line 99: Maybe?
```suggestion
        x.getCause.printStackTrace(ngContext.out)
```
This should probably also show up correctly via BSP

oyvindberg: Yeah, makes sense. Let's try

tgodzik (approved):

PR #1781, opened 2022-08-17
fix: initialize response - specify debugProvider

dos65:


ckipp01:
This was contained inside of https://github.com/scalacenter/bloop/pull/1820 so closing here.

PR #1783, opened 2022-08-20
Update gradle docs

Arthurm1:
Just Gradle documentation updates...

Update supported Gradle version info

Add check for plugin already being applied so build files can have bloop plugin setup and that won't clash with the Metals import as it also applies the bloop plugin

### DIFF
--- a/docs/build-tools/gradle.md
+++ b/docs/build-tools/gradle.md
@@ -16,7 +16,10 @@ Learn how to get set up by following the instructions below.
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


Arthurm1:
Is there a way to submit a PR without triggering the tests - as this is just a documentation update?

tgodzik:
> Is there a way to submit a PR without triggering the tests - as this is just a documentation update?

I never remember it, but this should work https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

tgodzik (approved):
LGTM!

PR #1819, opened 2022-09-15
Run source generators in topological order

Duhemm:
Previously, the `buildTarget/sources` BSP request could fail in cases where the build included source generators that depend on the output of other source generators. This issue would happen because, in this case, the ordering in which the source generators would run wasn't clearly defined.

During a compilation task, this problem would not appear because the source generators would always be run after their dependencies have run.

To fix this problem in the BSP services, Bloop will now run the source generators in the topological order.


tgodzik (approved):
LGTM!

PR #1821, opened 2022-09-28
bugfix: Try to download semanticdb even if it might not be supported

tgodzik:
Previously, we would not download semanticdb even if it was backpublished. Now, we will try to download, but not print any warning if it's unsupported.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/BuildLoader.scala
+++ b/frontend/src/main/scala/bloop/engine/BuildLoader.scala
@@ -211,11 +211,9 @@ object BuildLoader {
   ): Option[AbsolutePath] = {
     val (scalaVersion, scalaSemanticdbSettings) = scalaSemanticdbVersionAndSettings
     // Recognize 2.12.8-abdcddd as supported if 2.12.8 exists in supported versions
-    val isUnsupportedVersion =
-      !scalaSemanticdbSettings.supportedScalaVersions.exists(scalaVersion.startsWith(_))
-    if (isUnsupportedVersion) {
-      if (!scalaVersion.startsWith("3."))
-        logger.debug(Feedback.skippedUnsupportedScalaMetals(scalaVersion))(DebugFilter.All)
+    val isSupportedVersion =
+      scalaSemanticdbSettings.supportedScalaVersions.exists(scalaVersion.startsWith(_))
+    if (scalaVersion.startsWith("3.")) {
       None
     } else {
       SemanticDBCache.fetchScalaPlugin(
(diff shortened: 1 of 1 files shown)


ckipp01 (approved):

frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -211,11 +211,9 @@ object BuildLoader {
   ): Option[AbsolutePath] = {
     val (scalaVersion, scalaSemanticdbSettings) = scalaSemanticdbVersionAndSettings
     // Recognize 2.12.8-abdcddd as supported if 2.12.8 exists in supported versions
-    val isUnsupportedVersion =
-      !scalaSemanticdbSettings.supportedScalaVersions.exists(scalaVersion.startsWith(_))
-    if (isUnsupportedVersion) {
-      if (!scalaVersion.startsWith("3."))
-        logger.debug(Feedback.skippedUnsupportedScalaMetals(scalaVersion))(DebugFilter.All)
+    val isSupportedVersion =
+      scalaSemanticdbSettings.supportedScalaVersions.exists(scalaVersion.startsWith(_))
```
kpodsiad, line 215: How bloop knows which scala version supports semanticDB and which doesn't?

frontend/src/main/scala/bloop/engine/BuildLoader.scala
```diff
@@ -226,9 +224,13 @@ object BuildLoader {
         case Right(path) =>
           logger.debug(Feedback.configuredMetalsScalaProjects(projects))(DebugFilter.All)
           Some(path)
-        case Left(cause) =>
+        case Left(cause) if isSupportedVersion =>
           logger.displayWarningToUser(Feedback.failedMetalsScalaConfiguration(scalaVersion, cause))
           None
+
+        // We try to download anyway, but don't print the exception
+        case _ =>
```
kpodsiad, line 232: Can you explain what will happen if downloading semanticDB fails? Is this possible that user won't have code navigation and no indication why?

Also, I think comment might have better description, as this action (downloading) already took place and we handling failed case. For instance:
```scala
        // We tried to download semanticDB but scala version is unsupported - don't warn the user about it.
        case Left(cause) =>
          logger.debug(cause) (or sth along these lines?)
          None
```

kpodsiad (commented):
Couple of late questions, I'm trying to understand what's going on.

frontend/src/main/scala/bloop/engine/BuildLoader.scala
tgodzik: Metals sends that information, but since we backpublish Metals no longer has that information.

frontend/src/main/scala/bloop/engine/BuildLoader.scala
tgodzik: > Can you explain what will happen if downloading semanticDB fails? Is this possible that user won't have code navigation and no indication why?

No, they will always get the information that semanticdb options are missing. I decided not to show it to the user as a notification if the version is not expected to be supported. I could change that, but this should ideally not happen.

tgodzik: I will update it when doing another change.

PR #1829, opened 2022-10-03
bugfix: Revert to bsp 2.0.0, since later versions don't run on JDK 8

tgodzik:


kpodsiad (approved):
Have you managed to find what's causing problems?

tgodzik:
Not yet, I will need to dig into it a bit more, but it's something between 2.0.0 and 2.1.0-M2

PR #1832, opened 2022-10-05
fix classpath hashing

kpodsiad:
* remove previous mechanism which tried to share work between concurrent requests - it was causing too much mental overhead and was error prone. I'd say that gains from it were pretty negligibly small - if the same jar was being hashed _at the same time_ workload was shared - I'd say that it's very rare situation and it doesn't cost that much. After computation results are cached so subsequent requests will use cache. 
* uncomment already existing tests & fix them
* remove unused code

### DIFF
--- a/backend/src/main/scala/bloop/io/ClasspathHasher.scala
+++ b/backend/src/main/scala/bloop/io/ClasspathHasher.scala
@@ -192,8 +192,6 @@ class ClasspathHasher {
 
     hashes
       .map { result =>
-        pprint.log(this.cacheMetadataJar.size())
-        pprint.log(this.hashingPromises.size())
         if (isCancelled.get || cancelCompilation.isCompleted) {
           cancelCompilation.trySuccess(())
           Left(())
--- a/backend/src/main/scala/bloop/io/ClasspathHasher.scala
+++ b/backend/src/main/scala/bloop/io/ClasspathHasher.scala
@@ -11,6 +11,7 @@ import scala.concurrent.Promise
 import scala.concurrent.duration._
 import scala.util.control.NonFatal
 
+import bloop.logging.DebugFilter
 import bloop.logging.Logger
 import bloop.task.Task
 import bloop.tracing.BraveTracer
@@ -19,31 +20,34 @@ import monix.execution.atomic.AtomicBoolean
 import sbt.internal.inc.bloop.internal.BloopStamps
 import sbt.io.IO
 import xsbti.compile.FileHash
-import bloop.logging.DebugFilter
 
-private final case class JarMetadata(
-    modifiedTime: FileTime,
-    size: Long
-)
+class ClasspathHasher {
+  import ClasspathHasher._
 
-object ClasspathHasher {
-
+  private[this] val hashingPromises = new ConcurrentHashMap[Path, Promise[FileHash]]()
   private[this] val cacheMetadataJar = new ConcurrentHashMap[Path, (JarMetadata, FileHash)]()
 
   /**
    * Hash the classpath in parallel
    *
-   * NOTE: When the task returned by this method is cancelled, the promise
-   * `cancelCompilation` will be completed and the returned value will be
-   * empty. The call-site needs to handle the case where cancellation happens.
+   * NOTE: When the task returned by this method is cancelled, the promise `cancelCompilation` will
+   * be completed and the returned value will be empty. The call-site needs to handle the case where
+   * cancellation happens.
    *
-   * @param classpath The list of files to be hashed (if they exist).
-   * @param parallelUnits The amount of classpath entries we can hash at once.
-   * @param cancelCompilation A promise that will be completed if task is cancelled.
-   * @param scheduler The scheduler that should be used for internal Monix usage.
-   * @param logger The logger where every action will be logged.
-   * @param tracer A tracer to keep track of timings in Zipkin.
-   * @return A task returning an error if the task was cancelled or a complete list of hashes.
+   * @param classpath
+   *   The list of files to be hashed (if they exist).
+   * @param parallelUnits
+   *   The amount of classpath entries we can hash at once.
+   * @param cancelCompilation
+   *   A promise that will be completed if task is cancelled.
+   * @param scheduler
+   *   The scheduler that should be used for internal Monix usage.
+   * @param logger
+   *   The logger where every action will be logged.
+   * @param tracer
+   *   A tracer to keep track of timings in Zipkin.
+   * @return
+   *   A task returning an error if the task was cancelled or a complete list of hashes.
    */
   def hash(
       classpath: Array[AbsolutePath],
@@ -53,13 +57,12 @@ object ClasspathHasher {
       tracer: BraveTracer,
       serverOut: PrintStream
   ): Task[Either[Unit, Vector[FileHash]]] = {
-    val timeoutSeconds: Long = 2L
+    val timeoutSeconds: Long = 10L
 
     implicit val debugFilter: DebugFilter = DebugFilter.Compilation
 
     val isCancelled = AtomicBoolean(false)
-    def makeHashingTask(path: Path): Task[FileHash] = {
-      val isHashingComplete = AtomicBoolean(false)
+    def makeHashingTask(path: Path, promise: Promise[FileHash]): Task[FileHash] = {
 
       def hashFile(): FileHash = {
         val attrs = Files.readAttributes(path, classOf[BasicFileAttributes])
@@ -77,8 +80,10 @@ object ClasspathHasher {
                 logger.debug(
                   s"Cached entry for $file has different metadata, hash will be recomputed"
                 )
+              } else {
+                logger.debug(s"Cache miss for $file, hash will be computed")
               }
-              tracer.traceVerbose(s"computing hash $file") { _ =>
+              tracer.traceVerbose(s"computing hash for $file") { _ =>
                 val newHash =
                   FileHash.of(path, ByteHasher.hashFileContents(path.toFile))
                 cacheMetadataJar.put(path, (currentMetadata, newHash))
@@ -89,21 +94,21 @@ object ClasspathHasher {
       }
 
       def hash(path: Path): Task[FileHash] = Task {
-        try {
-          if (cancelCompilation.isCompleted) {
-            BloopStamps.cancelledHash(path)
-          } else if (isCancelled.get) {
-            cancelCompilation.trySuccess(())
-            BloopStamps.cancelledHash(path)
-          } else {
-            val hash = hashFile()
-            isHashingComplete.set(true)
-            hash
+        val hash =
+          try {
+            if (isCancelled.get) {
+              cancelCompilation.trySuccess(())
+              BloopStamps.cancelledHash(path)
+            } else {
+              hashFile()
+            }
+          } catch {
+            // Can happen when a file doesn't exist, for example
+            case NonFatal(_) => BloopStamps.emptyHash(path)
           }
-        } catch {
-          // Can happen when a file doesn't exist, for example
-          case NonFatal(_) => BloopStamps.emptyHash(path)
-        }
+        hashingPromises.remove(path, promise)
+        promise.trySuccess(hash)
+        hash
       }
 
       /*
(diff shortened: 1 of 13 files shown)


backend/src/main/scala/bloop/io/ClasspathHasher.scala
```diff
@@ -66,219 +49,100 @@ object ClasspathHasher {
...
-  def containsMacroDefinition(classpath: Seq[File]): Task[Seq[(File, Boolean)]] = {
-    import org.zeroturnaround.zip.commons.IOUtils
-    import org.zeroturnaround.zip.{ZipEntryCallback, ZipUtil}
-    def readJar(jar: File): Task[(File, Boolean)] = Task {
-      if (!jar.exists()) sys.error(s"File ${jar} doesn't exist")
-      else {
-        def detectMacro(jar: File): Boolean = {
-          var found: Boolean = false
-          ZipUtil.iterate(
-            jar,
```
kpodsiad, line 246: Testcase for this was commented, method is unused so I decided to delete it.

frontend/src/test/scala/bloop/testing/BaseSuite.scala
```diff
@@ -564,14 +574,22 @@ abstract class BaseSuite extends TestSuite with BloopHelpers {
...
   }
 
+  def testAsyncT(name: String, maxDuration: Duration)(
+      run: => Task[Unit]
+  ): Unit = {
+    test(name) {
+      Await.result(run.runAsync(ExecutionContext.scheduler), maxDuration)
+    }
+  }
+
```
kpodsiad, line 592: new utility method which accepts task - no more Thread.sleep and Await.Result in tests. Logic can be composed via Task's map/flatMap and we can have Await.Result only at the end.

frontend/src/test/scala/bloop/util/TestUtil.scala
```diff
@@ -469,6 +469,13 @@ object TestUtil {
     finally delete(AbsolutePath(temp))
   }
 
+  /** Creates an empty workspace where operations  described as Tasks can happen. */
+  def withinWorkspaceT[T](fun: AbsolutePath => Task[T]): Task[T] = for {
+    temp <- Task(Files.createTempDirectory("bloop-test-workspace").toRealPath())
+    tempDirectory = AbsolutePath(temp)
+    result <- fun(tempDirectory).doOnFinish(_ => Task(delete(tempDirectory)))
+  } yield result
```
kpodsiad, line 477: similar to the `withinWorkspaceT` but creating and deleting directory is bounded to Task's lifetime

project/Dependencies.scala
```diff
@@ -82,7 +82,9 @@ object Dependencies {
   val slf4jNop = "org.slf4j" % "slf4j-nop" % "1.7.2"
 
   val utest = "com.lihaoyi" %% "utest" % "0.6.9"
-  val pprint = "com.lihaoyi" %% "pprint" % "0.5.5"
+  val pprint = "com.lihaoyi" %% "pprint" % "0.8.0"
+  val oslib = "com.lihaoyi" %% "os-lib" % "0.8.1"
```
kpodsiad, line 86: `pprint` ids must have when developing, `oslib` reduce boilerplate in written testcases

backend/src/main/scala/bloop/io/ClasspathHasher.scala
```diff
@@ -66,219 +49,100 @@ object ClasspathHasher {
...
 
-        Option(definedMacrosJarCache.get(jar)) match {
-          case Some((metadata, hit)) if metadata == currentMetadata => jar -> hit
-          case _ =>
-            val detected = detectMacro(jar)
-            definedMacrosJarCache.put(jar, (currentMetadata, detected))
-            jar -> detected
+    val onCancel = Task(isCancelled.compareAndSet(false, true)).void
+
+    val tasks = classpath.toVector.map(path => makeHashingTask(path.underlying))
```
kpodsiad, line 135: core logic is pretty much the same, but expressed much simpler. Instead of creating array with nulls and mutable listBuffers, current logic just creates list of tasks and then sequence them.

backend/src/main/scala/bloop/io/ClasspathHasher.scala
```diff
@@ -66,219 +49,103 @@ object ClasspathHasher {
...
+      /*
+       * As a protective measure, set up a timeout for hashing which complete hashing with empty value to unlock
+       * downstream clients on the assumption that the hashing of this
+       * entry is too slow because of something that happened to this
+       * process. The completion of the downstream promise will also log a
+       * warning to the downstream users so that they know that a hashing
+       * process is unusually slow.
+       *
+       * There is no need for specyfying `doOnCancel` callback, this task won't start before timeout (see timeoutTo implementation)
+       */
```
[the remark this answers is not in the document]
kpodsiad: > Is there a case for this measure?

Good question, my answer to this is.. I have no idea :)

> btw, what will happen if we return emptyHash instead of correct one?

Another good question without answer :)

I can only say that I didn't want to change too much, I just wanted to fix NPE, uncomment tests and simplify this mechanism. As for me, we can remove this timeout fallback and see what will happen. In cases like this, correctness shouldn't be put before throughput.

kpodsiad:
> Performance test finished successfully.
> 
> Benchmarks is based on merging with master

[Gimme Gimme Gimme](https://www.youtube.com/watch?v=XEjLoHdbVeE)  results from perf tests after midnight.

backend/src/main/scala/bloop/io/ClasspathHasher.scala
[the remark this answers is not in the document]
dos65: Ouch. I missed that this timeout already was there. 
I incorrectly decided that you added that.

Ok, let's keep it as is.

tgodzik:
> > > Performance test finished successfully.
> > > Benchmarks is based on merging with master
> > 
> > 
> > [Gimme Gimme Gimme](https://www.youtube.com/watch?v=XEjLoHdbVeE) results from perf tests after midnight.
> 
> It seems to have run correctly but I don't see any results yet, I think there is some delay there :/

Ok, I don't see any regressions, so we should be good.

backend/src/main/scala/bloop/io/ClasspathHasher.scala
```diff
@@ -66,219 +49,103 @@ object ClasspathHasher {
...
+      val timeoutFallback: Task[FileHash] = Task {
+        val msg =
+          s"Hashing ${path} is taking more than ${timeoutSeconds}s"
+        try {
+          pprint.log(msg)
+          logger.warn(msg)
+          serverOut.println(msg)
+        } catch { case NonFatal(_) => () }
```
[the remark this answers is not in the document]
kpodsiad: `pprint.log(msg)` is a leftover, but the rest was already present and I'd say it's better to keep it.

dos65 (approved):
LGTM!

@tgodzik 
> @Duhemm do you have any idea what kind of gains this might offer?

From my side it looks more like refactoring + fix for non-correct usage of array across threads. Actually, I wanted to do the same while I was doing monix upgrade.

backend/src/main/scala/bloop/io/ClasspathHasher.scala
```diff
@@ -1,284 +1,231 @@
...
         }
       }
-    }
+      .doOnCancel(onCancel)
+  }
+}
 
-    Task.gatherUnordered(classpath.map(readJar(_)))
+object ClasspathHasher {
+  final val global: ClasspathHasher = new ClasspathHasher
```
kpodsiad, line 210: It was impossible to test `ClasspathHasher` because tests were sharing global cache. I turned it into class and create `global` instance because:
- I didn't want to touch already existing code which was referring to `ClasspathHasher` as an object. Now it's using `ClasspathHasher.global` which is almost the same
- in tests, each testcase is creating a fresh instance of `ClasspathHasher` - no pollution

backend/src/main/scala/bloop/io/ClasspathHasher.scala
```diff
@@ -1,284 +1,231 @@
...
+   *     session of hashing is computing it)
+   *
+   * It's important to distinguish between them, because `Computed` case should be treated will
+   * higher priority as it needs CPU and `FromPromise` is just waiting on the result.
+   */
+  private sealed trait FileHashingResult {
+    def task: Task[FileHash]
   }
+  private final case class Computed(task: Task[FileHash]) extends FileHashingResult
+  private final case class FromPromise(task: Task[FileHash]) extends FileHashingResult
```
kpodsiad, line 230: Is scaladoc good enough to explain why this is needed?

kpodsiad (commented):
Ok, I brought back workload sharing. Implementation is almost pure (updating concurrent hashmap isn't) and should be more straightforward to understand.

kpodsiad:
I must have messed up cancelation somehow :/

tgodzik:
Looks like the tests are not finishing now :thinking:

tgodzik:
`ClasspathHasher.scala:196 this.hashingPromises.size(): 0` <- looks like no promises are ever added?

kpodsiad:
@tgodzik @dos65 you guys probably want to unsubscribe from his PR for while, I can push a lot of trash commits here to debug what's going on.

> `ClasspathHasher.scala:196 this.hashingPromises.size(): 0` <- looks like no promises are ever added?

This is actually ok I think, every hash request for file is adding/removing or just waiting for promise to complete. Logging is placed at the very end of task, when all task were resolved so it means that all promises should be completed & removed too. 

I'm a bit lost here currently as locally there is no such deadlock like there is on CI.

Duhemm:
Hi! Thanks a lot for looking into this @kpodsiad ! I did some testing on a small subset of our build, and it looks like there's a small performance regression for us:

|                   | v1.5.4-4-00074600 | This PR |
|-------------------|-------------------|---------|
| First full build  | 5m53s             | 6m15s   |
| Second full build | 4m25s             | 5m8s    |

Some info about the Bloop projects that are being built:
- 709 projects
- 3915 sources
- Average classpath length is 2335
- 7299 unique classpath entries

frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
```diff
@@ -159,7 +160,10 @@ object ResultsCache {
       logger: Logger
   ): ResultsCache = {
     val handle = loadAsync(build, cwd, cleanOrphanedInternalDirs, logger)
-    Await.result(handle.runAsync(ExecutionContext.ioScheduler), Duration.Inf)
+    Await.result(
```
tgodzik, line 163: Looks like this is causing timeouts for DAP tests, any reason for the change?

frontend/src/test/scala/bloop/testing/BaseSuite.scala
```diff
@@ -571,6 +581,33 @@ abstract class BaseSuite extends TestSuite with BloopHelpers {
...
+  def testAsyncT(name: String, maxDuration: Duration)(
+      run: => Task[Unit]
+  ): Unit = {
+    test(name) {
+      Await.result(run.runAsync(ExecutionContext.scheduler), maxDuration)
+    }
+  }
+
+  /*
```
tgodzik, line 600: Remove the commented out method?

frontend/src/test/scala/bloop/testing/BaseSuite.scala
```diff
@@ -533,6 +533,16 @@ abstract class BaseSuite extends TestSuite with BloopHelpers {
     )
   }
 
+  @nowarn("msg=parameter value run|maxDuration in method ignore is never used")
```
tgodzik, line 536: Is this not removed to avoid too many changes?

backend/src/main/scala/bloop/io/ClasspathHasher.scala
tgodzik: In the description you wrote:
```
remove previous mechanism which tried to share work between concurrent requests - it was causing too much mental overhead and was error prone. 
```
should we not have only computed since we don't want to depend on another task?

tgodzik (changes requested):
Thanks for coming back to this! It looks good, though I have some questions and comments.

tgodzik:
Let's get back to it when someone has time to work on it.

PR #1838, opened 2022-10-16
Support traversal mode in `Dag.dfs`

Duhemm:
Previously, `Dag.dfs` would always do pre-order traversal. This commit adds a `mode` parameter which allows to select pre-order or post-order traversal. Also, this commit replaces calls to `Dag.topologicalSort` with `Dag.dfs` with `mode = PostOrder`.


tgodzik (approved):
LGTM!

PR #1841, opened 2022-10-21
Use scala-debug-adapter 3.0

adpi2:
This new major version of the scala-debug-adapter brings a number of new features including:
- Step filtering for Scala 2 and Scala 3
- Better expression evaluation for Scala 3
- Configuration of the ScalacOptions of the expression compiler
- Show returned value of method call
- Log-points
- Conditional breakpoints
- Restart top frame of the call stack
- Filtering of class loading

To learn more about those new features you can refer to the [release notes](https://github.com/scalacenter/scala-debug-adapter/releases/tag/v3.0.1)


adpi2:
The tests do not pass because the new step filter does not like when a class is defined in the `<empty>` package. I need to do another release of the scala-debug-adapter to fix that.

adpi2:
Ready to be reviewed. There is one test that failed but it seems totally unrelated.

frontend/src/main/scala/bloop/dap/BloopDebugToolsResolver.scala
```diff
@@ -0,0 +1,70 @@
...
+        .map(jars => toClassLoader(jars, false))
+        .toTry
+    }
+  }
+
+  override def resolveStepFilter(scalaVersion: ScalaVersion): Try[ClassLoader] = {
+    getOrTryUpdate(stepFilterCache, scalaVersion) {
+      val stepFilterModule = s"${BuildInfo.scala3StepFilterName}_${scalaVersion.binaryVersion}"
+      val stepFilter = Artifact(BuildInfo.organization, stepFilterModule, BuildInfo.version)
+      val tastyCore = Artifact("org.scala-lang", "tasty-core_3", scalaVersion.value)
```
tgodzik, line 41: Why do we need scala 3 artifacts on the classpath. Do we need it even with Scala 2?

tgodzik (changes requested):
Wow! That's a lot of amazing features! That's super exciting. One comment from me.

frontend/src/main/scala/bloop/dap/BloopDebugToolsResolver.scala
adpi2: We only need to load the step filter for Scala 3.

The Scala 2 step filter is a direct dependency of the `scala-debug-adapter` so it does not need to be class loaded in its own class loader.

tgodzik (approved):
LGTM!

PR #1845, opened 2022-11-03
Update bsp to 2.1.0-M3

lwronski:


lwronski:
> Could you make sure you are able to start it with JDK 8? Last time I tried the updated BSP would on work on older JDK. I haven't been able to find why.

Bsp uses `jsoniter-scala` which has recently dropped support for JDK 8 - https://github.com/plokhotnyuk/jsoniter-scala/pull/860/files. That's probably why you couldn't run bloop with JDK 8

tgodzik:
Ach thanks! We should probably change jsoniter to an older version in BSP then. Unless we want to start dropping JDK 8 everywhere :thinking:

tgodzik:
> Ach thanks! We should probably change jsoniter to an older version in BSP then. Unless we want to start dropping JDK 8 everywhere thinking

Or we should change BSP to using JDK 11 at the least. Can you raise an issue in BSP so that they are aware of that? I think at this point it might be fine to start dropping JDK 8, especially since Bloop isn't used as a library anywhere.

ckipp01:
Regarding jdk 8, did sbt not have any issues with this? https://github.com/sbt/sbt/pull/6985

tgodzik:
@lwronski Let's revert jsoniter update in BSP for now to the last one working with JDK8, we should also be able to pin the version in scala steward for now.

build.sbt
```diff
@@ -68,6 +68,8 @@ lazy val bloopShared = (project in file("shared"))
     name := "bloop-shared",
     libraryDependencies ++= Seq(
       Dependencies.bsp4s,
+      Dependencies.jsoniterCore,
```
tgodzik, line 71: Could you also bump jsoniter to the latest version that supports JDK 8 ?

tgodzik (changes requested):
Looks good! Could you only update some of the expect tests? They don't expect the additional None. Alternatively, we could  change the string representation in tests so that they don't fail with new fields.

tgodzik (approved):
LGTM!

Issue #1847, opened 2022-11-08
Bloop CLI hangs after several runs


brunofarache:
We noticed Bloop's CLI starts hanging after several project compilations at this line: `Compiled ${project.name} (${durationMs}ms)`.

The only to fix it is to restart the server.

Took some thread dumps: 
[threaddump-1667905915033.tgz](https://github.com/scalacenter/bloop/files/9960758/threaddump-1667905915033.tgz)

Apparently there are many threads hanging on `cleanUpNonStableCliDirectories`: https://github.com/scalacenter/bloop/blob/586cfbff1de4dde023e7d04517e1e6ed01fb1b64/frontend/src/main/scala/bloop/Cli.scala#L440-L464.

```
"scala-execution-context-global-1657" #1657 daemon prio=5 os_prio=31 cpu=5.70ms elapsed=3929.90s tid=0x00007fc1d8013800 nid=0x8b17 waiting for monitor entry  [0x0000700014968000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at java.io.PrintStream.println(java.base@11.0.11/PrintStream.java:881)
	- waiting to lock <0x000000060001b920> (a java.io.PrintStream)
	at bloop.Cli$.$anonfun$cleanUpNonStableCliDirectories$1(Cli.scala:441)
	at bloop.Cli$.$anonfun$cleanUpNonStableCliDirectories$1$adapted(Cli.scala:438)
	at bloop.Cli$$$Lambda$5185/0x0000000801358840.apply(Unknown Source)
	at scala.collection.immutable.List.map(List.scala:293)
	at bloop.Cli$.cleanUpNonStableCliDirectories(Cli.scala:438)
	at bloop.Cli$.$anonfun$runWithState$3(Cli.scala:374)
	at bloop.Cli$$$Lambda$271/0x0000000800282040.apply(Unknown Source)
	at monix.eval.internal.TaskRunLoop$.monix$eval$internal$TaskRunLoop$$loop$1(TaskRunLoop.scala:220)
	at monix.eval.internal.TaskRunLoop$RestartCallback$1.onSuccess(TaskRunLoop.scala:119)
	at monix.eval.Callback$Extensions$.monix$eval$Callback$Extensions$$$anonfun$asyncOnSuccess$1(Callback.scala:100)
	at monix.eval.Callback$Extensions$$anonfun$asyncOnSuccess$extension$1.run(Callback.scala:100)
[...]

PR #1850, opened 2022-11-14
bugfix: Remove potentially problematic println

tgodzik:
It seems that https://github.com/scalacenter/bloop/issues/1847 is hanging on trying to lock the out object and since the println here is not really useful I think it's fine to remove and check if that helps.

The stack trace mostly shows a lot of thread stuck on that println actually.


kpodsiad (approved):
Alternatively, `cleanUpNonStableCliDirectories` can accept logger and log that message at info/debug level. I have no idea how useful it'll though.

tgodzik:
> Alternatively, `cleanUpNonStableCliDirectories` can accept logger and log that message at info/debug level. I have no idea how useful it'll though.

I was thinking about it, but I don't think I have ever seen this being useful.

PR #1851, opened 2022-11-14
chore: Try to setup Scala Steward

tgodzik:
We might want to migrate to a new bot later on since I can't find access to this one, but let's test it out first.


ckipp01 (approved):
👍🏼 godspeed

PR #1852, opened 2022-11-15
chore: Remove support for sbtBloop for 0.13 version

tgodzik:
I think it's finally the time to stop releasing for 0.13 and users can anyway use an older version of Bloop if they don't want to migrate.


build.sbt
```diff
@@ -152,29 +152,6 @@ val testResourceSettings = {
   })
 }
 
-lazy val jsonConfig210 = crossProject(JVMPlatform)
```
ckipp01, line 155: If we're removing this I believe we can also then remove `config/src/main/scala-2.10/*` right? Since we'll no longer be publishing `bloop-config` in 2.10.

tgodzik: Yay! More removed code :D

build.sbt
```diff
@@ -152,29 +152,6 @@ val testResourceSettings = {
...
-      compilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full),
-    testResourceSettings
-  )
-  .jvmSettings(
-    testSettings,
-    target := (file("config") / "target" / "json-config-2.10" / "jvm").getAbsoluteFile,
-    libraryDependencies ++= Seq(
-      Dependencies.circeParser,
-      Dependencies.circeCore,
-      Dependencies.circeGeneric
```
ckipp01, line 174: Also it looks like these three Deps are only used here. I think you can remove them from `Dependencies.scala`.

tgodzik: Done! I will go through some more deps later on.

ckipp01 (approved):
Getting rid of code. Love it.

LGTM!

PR #1864, opened 2022-11-21
chore: Remove java-debug and ignore updates for nailgun, release-early

tgodzik:
javaDebug is brought in by scala-debug-adapter, so no need to define it here.

Nailgun version should fixed, since it's the one Bloop works with (dropping the fork would be awesome.)

Release-early should be replaced by sbt-ci-release so don't want to change the version and risk regression. Especially since it's not actively maintained.

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,5 +1,7 @@
 updates.ignore = [
   { groupId = "com.github.plokhotnyuk.jsoniter-scala" },
+  { groupId = "ch.epfl.scala", artifactId = "nailgun-server" },
+  { groupId = "ch.epfl.scala", artifactId = "sbt-release-early" }
 ]
 commits.message = "build(deps): Update ${artifactName} from ${currentVersion} to ${nextVersion}"
 updates.limit = 5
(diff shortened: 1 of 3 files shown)


.scala-steward.conf
```diff
@@ -1,5 +1,7 @@
 updates.ignore = [
   { groupId = "com.github.plokhotnyuk.jsoniter-scala" },
+  { groupId = "ch.epfl.scala", artifactId = "nailgun-server" },
```
ckipp01, line 3: For some reason I always thought Bloop was using https://github.com/jvican/snailgun not yet another fork of nailgun. I'm curious how much changed in that fork and if the migration off it would be easy.

.scala-steward.conf
```diff
@@ -1,5 +1,7 @@
...
+  { groupId = "ch.epfl.scala", artifactId = "sbt-release-early" }
```
ckipp01, line 4: Haha no I see where the `release-early` came from in the Metals commit.

ckipp01 (approved):
LGTM

.scala-steward.conf
tgodzik: I need to investigate what's the best solution here, especially that the Bloop fork is also using snailgun I think.

tgodzik: Actually it does use snailgun :thinking:

PR #1866, opened 2022-11-22
bugfix: Don't fail test if socket was already closed

tgodzik:
Seems that the socket might already be closed at that point and an exception is caused causing frequent flaky tests.

### DIFF
--- a/frontend/src/test/scala/bloop/bsp/BspBaseSuite.scala
+++ b/frontend/src/test/scala/bloop/bsp/BspBaseSuite.scala
@@ -750,7 +750,12 @@ abstract class BspBaseSuite extends BaseSuite with BspClientTest {
       val closeTask = {
         lsClient.request(endpoints.Build.shutdown, bsp.Shutdown()).flatMap { _ =>
           Task.fromFuture(lsClient.notify(endpoints.Build.exit, bsp.Exit())).map { _ =>
-            socket.close()
+            try {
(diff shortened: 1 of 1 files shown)


frontend/src/test/scala/bloop/bsp/BspBaseSuite.scala
```diff
@@ -750,7 +750,12 @@ abstract class BspBaseSuite extends BaseSuite with BspClientTest {
       val closeTask = {
         lsClient.request(endpoints.Build.shutdown, bsp.Shutdown()).flatMap { _ =>
           Task.fromFuture(lsClient.notify(endpoints.Build.exit, bsp.Exit())).map { _ =>
-            socket.close()
+            try {
+              socket.close()
```
ckipp01, line 754: Maybe I'm totally missing it, but how would it be closed already if there was still a connection until `closeTask` was called?

tgodzik: hmm.... not sure exactly, but this seems only a clean up, so it shouldn't fail anything

ckipp01 (approved):
👍🏼 if it helps, LGTM

tgodzik:
> 👍🏼 if it helps, LGTM

I will be trying more bugfixes to stabilize the CI, so I might need to rework that even further, but let's see if it helps.

Issue #1680, opened 2022-02-16
[EPIC] Bloop Next Steps


kpodsiad:
Based https://contributors.scala-lang.org/t/bloop-next-steps/5554. These changes should simplify working with Bloop and ease the maintenance burden on any current and future contributors.

TODO:
- [x] publish bloop launcher artifacts for Scala 2.13 - #1646
- [ ] bump used Scala and sbt versions to the newest ones #1670, #1679 (blocked on #1662)
- [ ] use upstream Zinc instead of forked one - #1662
- [ ] remove modules which are no longer needed - #1650
- [ ] rewrite generation of the test projects to not happen on sbt load
- [ ] remove custom shading plugin and use [sbt-shading](https://github.com/coursier/sbt-shading) if possible
- [ ] migrate from sbt-release-early to sbt-ci-release
- [ ] add Scala steward
- [ ] move integrations to separate repositories, where they can be released independently.
- [ ] create a community suite (most likely some modification of buildpress) that would test out some of the more popular repositories.
- [ ] make benchmarks easier to run and use.
- [ ] investigate tests and try to make them less flaky

tgodzik:
Thanks for putting it into an actual issue! I should have done it previously after the post on contributors.

tgodzik:
Most of the things are done. but the last two ones we don't plan on working on msot likely/

PR #1868, opened 2022-11-22
refactor: start using bloop-config as a dependency

ckipp01:
For context, this was based off a conversation we had on Discord where there was a desire to break config out of Bloop. This would make it easier to setup MiMa, simplify this build a bit, and allow it to be versioned separately. You can find bloop-config [here](https://github.com/scalacenter/bloop-config).


ckipp01:
🤔 looking at how the `BaseConfigSuite` is used all over I wonder if I should publish that as like a `bloop-config-testkit` or something and then rely on that for testing.

build.sbt
```diff
@@ -585,43 +472,90 @@ def isJdiJar(file: File): Boolean = {
...
+lazy val integrationUtils213 = project
+  .in(integrations / "utils")
+  .settings(scalafixSettings)
+  .settings(
+    scalaVersion := Dependencies.Scala213Version,
+    libraryDependencies += Dependencies.bloopConfig,
+    target := (file(
+      "integrations"
+    ) / "utils" / "target" / "utils-2.13").getAbsoluteFile
+  )
```
ckipp01, line 513: So it's a little annoying that I had to define 3 like this, but that's it's sort of due to how the build is defined and not using something like sbt-projectmatrix. Either way this seems to do the trick... All for one file that we need cross built 😆 .

ckipp01 (commented):
Alrighty this is finally ready to take a look. I wanted to run scalafix on my files and then realized that it had a larger change then expected. However it did clean up so other unused imports, so I hope that's ok. If you're alright with the pr my plan is the following:

- [ ] Release a new patch in https://github.com/scalacenter/bloop-config
- [ ] Bump this to use that new patch
- [ ] Merge this
Then moving forward after this pr config will be completely standalone and not versioned along with Bloop.

build.sbt
adpi2: Should we introduce `sbt-projectmatrix` then? Would it simplify the build?

build.sbt
```diff
@@ -650,17 +584,19 @@ lazy val gradleBloop212 = project
...
   )
   .settings(
-    libraryDependencies += Dependencies.classgraph % Test,
-    publishLocal := publishLocal.dependsOn((jsonConfig212.jvm / publishLocal)).value
+    libraryDependencies ++= Seq(
+      Dependencies.classgraph % Test,
+      Dependencies.bloopConfig
+    )
   )
+  .dependsOn(integrationUtils212 % "test->compile")
```
adpi2, line 592: It's kind of unrelated to this PR but why not put all the settings together in a single `.settings(...)` and all the internal dependencies in a single `.dependsOn(...)`.

I find it misleading that there are two `.dependsOn` on the same project, and spread over many `.setttings`. It is easy to miss one.

adpi2 (changes requested):

build.sbt
ckipp01: We should yes, but I'd rather not do it in this pr. Because if we do it here we also have to all the other modules that are defined this way as well. I know in https://github.com/scalacenter/bloop/issues/1680 switching to `sbt-ci-release` is mentioned. I think maybe doing that migration and switching to `sbt-projectmatrix` at the same time would be a good idea. If we all agree, after I get this merged in I can try to tackle that.

julienrf: Why is `crossTarget` not working here? It seems that `target/utils-x.yz` is reproducing what `crossTarget` does?

build.sbt
ckipp01: Ah I actually didn't realize there was another `dependsOn` up above. I'll move this.

In general, I totally agree with you. I have a couple small things on my list that I hit on while working on this and cleaning up some of this stuff is one of them. However I want to do it in a separate pr.

adpi2: :+1:

build.sbt
adpi2: > Why is crossTarget not working here? It seems that target/utils-x.yz is reproducing what crossTarget does?

I believe `crossTarget` is only for Scala-version dependent targets: like class files, zinc cache...

Here we want the whole target to be independent. From sbt's point of view, those projects are completely separate and they can't share the same target directory.

adpi2 (approved):
LGTM

build.sbt
```diff
@@ -585,100 +472,141 @@ def isJdiJar(file: File): Boolean = {
...
   else file.getAbsolutePath.contains(SbtJdiTools.JavaTools.getAbsolutePath.toString)
 }
+val publishJsonModuleSettings = List(
+  publishM2Configuration := publishM2Configuration.value.withOverwrite(true),
+  publishLocalConfiguration := publishLocalConfiguration.value.withOverwrite(true),
+  // We compile in both so that the maven integration can be tested locally
+  publishLocal := publishLocal.dependsOn(publishM2).value
+)
+
+lazy val integrationUtils211 = project
```
tgodzik, line 482: We could probably remove 2.11 and just note that for Gradle you need to use the last version which is 1.5.4. We can do it later though.

tgodzik (approved):
LGTM! Now this make me thinking whether we should separate the plugins, or at least the gradle and maven ones, into separate repositories.

project/build.sbt
```diff
@@ -5,7 +5,13 @@ val mvnPluginToolsVersion = "3.6.0"
 val `bloop-shaded-plugin` = project
   .settings(
     sbtPlugin := true,
-    libraryDependencies += "ch.epfl.scala" %% "sbt-bloop-build-naked" % "1.0.0-SNAPSHOT"
+    libraryDependencies ++= Seq(
+      "ch.epfl.scala" %% "sbt-bloop-build-naked" % "1.0.0-SNAPSHOT",
+      // TODO idn if this is needed.. no idea why we are getting a
+      // java.lang.NoClassDefFoundError: bloop/config/Config$LinkerMode
+      // for the shaded plugin in CI, but no there.
+      "ch.epfl.scala" %% "bloop-config" % "1.5.4-6-7dc3eb-SNAPSHOT"
```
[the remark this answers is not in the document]
tgodzik: Old comment, nvm me.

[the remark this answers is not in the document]
ckipp01: Yup, I removed it before the last commit 👍🏼

build.sbt
ckipp01: Sounds good. Supporting 11 isn't causing any real issues at the moment, but the moment it does yea, we should rip it out.

ckipp01:
> Now this make me thinking whether we should separate the plugins, or at least the gradle and maven ones, into separate repositories.

I go back and forth. It's certainly simplify things, but I also see the benefit of being able to quickly test new thing with the integration tests in here.

tgodzik:
> > Now this make me thinking whether we should separate the plugins, or at least the gradle and maven ones, into separate repositories.
> 
> I go back and forth. It's certainly simplify things, but I also see the benefit of being able to quickly test new thing with the integration tests in here.

The main issue is config, that's what we test, so this would not depend on actual Bloop server at all

ckipp01:
Alrighty, we're all 🟢 so let's go!

PR #1869, opened 2022-11-23
Discover test classes only in test projects

Duhemm:
Previously, Bloop would inspect all projects to try and find test classes. This is not necessary, because only classes that live in test projects should be run.

This commit filters out test projects from the analysis, so that it runs faster.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
@@ -42,6 +42,20 @@ object TestTask {
   implicit private val logContext: DebugFilter = DebugFilter.Test
 
   /**
+   * Does the given project represent a test project? Test project are projects tagged with either
+   * `test` or `integration-test`.
+   *
+   * See [[Tag.Test]], [[Tag.IntegrationTest]]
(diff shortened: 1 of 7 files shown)


frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
```diff
@@ -41,6 +41,20 @@ final case class TestFrameworkWithClasses(
...
+   * Does the given project represent a test project? Test project are projects tagged with either
+   * `test` or `integration-test`.
+   *
+   * See [[Tag.Test]], [[Tag.IntegrationTest]]
+   *
+   * @param project The project to test
+   * @return true if `project` is a test project, false otherwise.
+   */
+  def isTestProject(project: Project): Boolean = {
+    project.tags.contains(Tag.Test) ||
```
tgodzik, line 54: The tests seem to be failing, maybe they don't have the tags set properly?

Duhemm: That was exactly the problem!

tgodzik (approved):
LGTM!

PR #1870, opened 2022-11-23
refactor: Use the actual timeouts defined for tests in DebugServerSpec

tgodzik:
Previously, we would always timeout after 35 seconds, because that was the timeout set in startDebugServer. Now, we forward the task up to the test and wait the amount specified in each test.


adpi2 (approved):
LGTM

PR #1871, opened 2022-11-24
improvement: Change the stamps to timewrapped ones

tgodzik:
Previously we would use last modfied stamps for classfiles (and other products) and hashes for jars. Now, we use hashes for both, but only if the modified time changed.

Should help with https://github.com/scalameta/metals/issues/4623

### DIFF
--- a/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopStamps.scala
+++ b/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopStamps.scala
@@ -15,14 +15,14 @@ import xsbti.compile.analysis.Stamp
 
 object BloopStamps {
   private val converter = PlainVirtualFileConverter.converter
-  def initial: ReadStamps = {
-    Stamps.initial(
-      Stamper.forLastModifiedInRootPaths(converter),
-      // The hash is for the sources
-      BloopStamps.forHash,
-      Stamper.forHashInRootPaths(converter)
-    )
-  }
+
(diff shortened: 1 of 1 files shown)


tgodzik:
Locally run benchmark seems to produce the same timings and I am running the server one to confirm.

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopStamps.scala
```diff
@@ -34,8 +34,11 @@ object BloopStamps {
   def directoryHash(path: Path): FileHash = FileHash.of(path, directoryHash)
   def isDirectoryHash(fh: FileHash): Boolean = fh.hash == directoryHash
 
-  def forHash(file: VirtualFileRef): Hash = {
-    fromBloopHashToZincHash(ByteHasher.hashFileContents(converter.toPath(file).toFile()))
+  def forHash(fileRef: VirtualFileRef): Hash = {
+    val file = converter.toPath(fileRef).toFile()
+    if (file.exists())
```
Duhemm, line 39: Why do we need to check if the file exists now but not before?

Duhemm (approved):
Looks good to me, thank you!

I'll try that out and let you know how that may affect the performance on larger builds. Don't wait for me though 😄

bloopoid:

bloopoid:

bloopoid:

PR #1882, opened 2022-11-28
chore: Remove sbt asembly since it's no longer used

tgodzik:


adpi2 (commented):
Does anyone remember when and why we used sbt-assembly?

tgodzik:
> Does anyone remember when and why we used sbt-assembly?

Looks like it was used for building a fat jar for benchmarks, but this is no longer the case.

adpi2 (approved):
Looks good, then!

PR #1886, opened 2022-11-29
chore(ci): remove local steward

ckipp01:
We no have a scalacenter-wide one that we can use in `scalacenter/steward`

refs: https://github.com/scalacenter/steward/pull/1


tgodzik:
Is the local steward yet set up though? Should we not first merge it and only then remove?

ckipp01:
> Is the local steward yet set up though? Should we not first merge it and only then remove?

Yea the timing is a bit funky since I don't know which will all happen first. I'd say just approve now if you're ok with it, then I'll leave it open until I can verify the org one is working. Then I'll merge.

tgodzik (approved):
Thanks!

PR #1888, opened 2022-11-29
refactor: remove unused cloneKafka stuff.

ckipp01:
This isn't used at all and in the future if we really want to use this
  we can revisit it and pull it back out. For now, all the cleanup we can
  get makes thing easier to understand


adpi2 (approved):
LGTM

tgodzik (approved):
LGTM! `Need it in the future` code is not really needed with git.

PR #1924, opened 2022-12-02
docs: Add release notes for 1.5.5

tgodzik:


notes/v1.5.5.md
```diff
@@ -0,0 +1,146 @@
+# bloop `v1.5.5`
+
+Bloop v1.5.5 is a bugfix release.
```
ckipp01, line 3: I'd also add something like the following

> One notable change this release for anyone using the `bloop-config` artifact is that it's not longer contained inside of this repo or versioned along with it. You can find its new location in [`scalacenter/bloop-config`](https://github.com/scalacenter/bloop-config). The latest version of `bloop-config` is still 1.5.5 and you'll continue to use it as your normally have.

adpi2, line 3: Bloop v1.5.5 is a bugfix release, but it also contains a major version bump of the `scala-debug-adapter` with [many new features](https://github.com/scalacenter/scala-debug-adapter/releases/tag/v3.0.1).

So I would appreciate some advertisement of that.

> Bloop v1.5.5 ships a new major version of the Scala Debug Adapter, featuring better debugging steps (step filter), conditional breakpoint, logpoints and more. See a full lists of features in the releases notes of the Scala Debug Adapter [3.0.1](https://github.com/scalacenter/scala-debug-adapter/releases/tag/v3.0.1) and [3.0.3](https://github.com/scalacenter/scala-debug-adapter/releases/tag/v3.0.3).

julienrf: If the classpath of projects depending on bloop artifacts now contains the new version of the debugger, it would be better to also bump the major version of bloop (since the new classpath is not binary compatible with the old classpath).

adpi2: Bloop frontend is a tool not a library, so I don't think we need to follow the semantic versioning rules. Scala Debug Adapter uses `VersionScheme.Strict` which means that we can break the binary compatibility between any two versions, even a patch version. As a consequence, on Bloop, it can also break the binary compatibility between any two patch versions.

tgodzik: Added! Good idea!

tgodzik: Let's do that in a follow up then.

>So I would appreciate some advertisement of that. Something like the following:

Added! Though I think it will be better visible within Metals, which I also plan to add.

julienrf (approved):

notes/v1.5.5.md
```diff
@@ -0,0 +1,174 @@
...
+One of the notable changes in this release for anyone using the bloop-config
+artifact is that it's no longer contained inside of this repo or versioned along
+with it. You can find its new location in
+[scalacenter/bloop-config](https://github.com/scalacenter/bloop-config). The
+latest version of bloop-config is still 1.5.5 and you'll continue to use it as
+your normally have.
+
+Bloop v1.5.5 also ships a new major version of the Scala Debug Adapter,
+featuring better debugging steps (step filter), conditional breakpoint,
+logpoints and more. See a full lists of features in the release notes of the
```
adpi2, line 12:
```suggestion
featuring better debugging steps (step filter), conditional breakpoints,
logpoints and more. See a full list of features in the release notes of the
```

adpi2 (approved):
Apart from my own typos :worried:, LGTM !

notes/v1.5.5.md
```diff
@@ -0,0 +1,174 @@
+# bloop `v1.5.5`
+
+One of the notable changes in this release for anyone using the bloop-config
+artifact is that it's no longer contained inside of this repo or versioned along
+with it. You can find its new location in
+[scalacenter/bloop-config](https://github.com/scalacenter/bloop-config). The
+latest version of bloop-config is still 1.5.5 and you'll continue to use it as
+your normally have.
```
ckipp01, line 8:
```suggestion
you normally have.
```

notes/v1.5.5.md
```diff
@@ -0,0 +1,174 @@
+# bloop `v1.5.5`
+
+One of the notable changes in this release for anyone using the bloop-config
```
ckipp01, line 3:
```suggestion
One of the notable changes in this release for anyone using the `bloop-config`
```

notes/v1.5.5.md
```diff
@@ -0,0 +1,174 @@
...
+artifact is that it's no longer contained inside of this repo or versioned along
+with it. You can find its new location in
+[scalacenter/bloop-config](https://github.com/scalacenter/bloop-config). The
+latest version of bloop-config is still 1.5.5 and you'll continue to use it as
```
ckipp01, line 7:
```suggestion
latest version of `bloop-config` is still 1.5.5 and you'll continue to use it as
```

ckipp01 (approved):
Couple really small nits on y comment that I missed as well, but other than that LGTM!

PR #1943, opened 2022-12-06
chore: downgrade release ubuntu version

tgodzik:
It seems that if we build with the newer version and try to run on ubuntu 20 it will fail with missing `GLIBC_2.34'

I updated my local ubuntu version and it seems to work, but it might be problematic for some users with older version.


ckipp01 (approved):
I'm not super familiar with this and don't use ubuntu, but this seems to make sense. LGTM

PR #1947, opened 2022-12-09
bugfix: Use the project working directory for running

lwronski:


tgodzik (approved):
LGTM! 

Side note, this doesn't really change anything for other tools aside from ScalaCLI, which without this change would run mains in .scala-build directory

PR #1948, opened 2022-12-09
refactor: move the Maven plugin out of Bloop

ckipp01:
This pr removes the Maven integration from this repo since it's been moved to its new home in https://github.com/scalacenter/bloop-maven-plugin.

There are a few different motivations for this:

- Gets rid of the custom sbt code to publish a Maven plugin since now it's just published with Maven.
- Simplifies this build.
- Allows for the Maven integration to be versioned standalone.

Refs #1680


ckipp01:
Alright, this should be ready to rock and roll now. I realeased `bloop-maven-plugin-RC4` in https://github.com/scalacenter/bloop-maven-plugin and also had a couple people that use Maven at work test it out to ensure it's actually working. I also sent in a PR to Metals to update this in https://github.com/scalameta/metals/pull/4728. I think once we get this ripped out, the Metals PR merged, and then ensure everything is working I'll release the 2.0.0 stable.

docs/build-tools/maven.md
```diff
@@ -27,32 +31,24 @@ The Bloop plugin for Maven doesn't need to be installed, you can run
...
 
+**Note**: if you've used this command before you'll have noticed the artifact
+name has changed. In the past you would have used `maven-bloop_2.13`, however as
+of the 2.x series the artifact name as changed to `bloop-maven-plugin`.
 
 To build with a single Scala project `foo` generates two configuration files:
 
 ```bash
-$ mvn ch.epfl.scala:maven-bloop_2.10:@VERSION@:bloopInstall
+$ mvn ch.epfl.scala:bloop-maven-plugin:@BLOOP_MAVEN_VERSION@:bloopInstall
```
adpi2, line 51: Just out of curiosity, what is the Scala version of the Bloop Maven plugin? Can it conflict with other Maven plugins compiled with Scala?

docs/build-tools/maven.md
```diff
@@ -6,13 +6,17 @@ sidebar_label: Maven
 
 ## Getting Started
 
-Configuring bloop with Maven projects speeds up developer workflows and gives you access to a modern Scala toolchain, despite Maven traditionally lacking good Scala suppport.
+Configuring bloop with Maven projects speeds up developer workflows and gives
```
adpi2, line 9: Not really related to this PR, but I noticed that Bloop is often written `bloop` with no capital B.

```suggestion
Configuring Bloop with Maven projects speeds up developer workflows and gives
```

adpi2 (approved):
Some minor comments and a big :tada:

docs/build-tools/maven.md
ckipp01: So the Scala version used for the plugin is 2.13. And honestly I'm not 100% sure how that works with Maven. I'm not sure if the same classpath for the build is shared with the plugin goals or not, so I'm not sure if there will be a conflict. I guess I didn't even think about that when I was testing on various repos.

It's currently only being published for 2.13 in this repo as well, so I didn't change that and instead just kept the behavior the same.

adpi2: It seems that the Scala Maven Plugin also [depends on Scala 2.13](https://github.com/davidB/scala-maven-plugin/blob/b64ca9c868466fb741dd05fc16e024c34d021d02/pom.xml#L184-L186) so that's probably why.

PR #1954, opened 2022-12-15
refactor: get rid of leftover integration utils

ckipp01:
Now that the Gradle and Maven integrations are no longer in this repo we're not actually using the integration utils for anything. This PR removes them and also removes any reference to sbt 0.x.

Refs: https://github.com/scalacenter/bloop/issues/1680


adpi2 (approved):
:+1:

PR #1955, opened 2022-12-15
refactor(build): a bit of build cleanup

ckipp01:
This does a few different things:

- removes no longer needed deps
- removes a couple defs/Keys that are no longer used
- deduplicates settings entries

And just general cleanup


tgodzik (approved):
LGTM :rocket:

PR #1961, opened 2022-12-22
refactor: stop shading launcher and bloopgun

ckipp01:
Looking back I can't find a reason as to why this was done, so my
assumption is that this was done as a precaution. Looking through the
deps they are quite minimal. Testing around with Metals using the
unshaded version of launcher doesn't cause any issues. I also see that
multiple apps are using bloop-rifle instead, and that's also not shaded,
arguably has more deps, and isn't causing issues. If this is ok, we can
get rid of the entire custom shading logic we have that really
simplifies things and also allows us to no longer publish both shaded
and not shaded versions of launcher and bloopgun. If we detect problems,
we can easily then bring in sbt-shading.

There is a discussion related to this with more details in https://github.com/scalacenter/bloop/discussions/1958.

Also note that if we want to actually slim down the deps in these, I think it's doable. For example we bring in jsoniter in bloopgun _only_ to read the global config since it's json, and that config has two fields.


adpi2 (approved):
Sorry, for some reason I did not see this PR earlier.

It's crazy good to see how much stuff you can remove from the build without breaking anything.

tgodzik (approved):
LGTM! Let's drop it for the time being and if needed we can add it back, but using an existing plugin.

ckipp01:
lol I feel like I've just been taking an 🪓 to this thing. Love it.

PR #1962, opened 2022-12-22
refactor(ci): migrate to sbt-ci-release

ckipp01:
This pr changes the way we do releases and migrates us away from the
forked sbt-release-early that's no longer maintained and is also
bringing in some old versions of scala-xml which will be problematic
when we try to update our sbt stuff.

Plus this simplifies the release a little bit. The one difference is
that we'll now be publishing actual `-SNAPSHOT` into sonatype snapshots.

Refs: https://github.com/scalacenter/bloop/issues/1680

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,10 +248,9 @@ jobs:
     env:
       CI: true
       SKIP_TEST_RESOURCES_GENERATION: true
-      SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
+      SONATYPE_USERNAME: ${{ secrets.SONATYPE_USER }}
       SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
-      PGP_PASSWORD: ${{ secrets.PGP_PASSWORD }}
-      PGP_PASSPHRASE: ${{ secrets.PGP_PASSWORD }}
+      PGP_SECRET: ${{ secrets.PGP_SECRET }}
       PGP_PUBLIC_KEY: ${{ secrets.PGP_PUBLIC_KEY }}
       PGP_PRIVATE_KEY: ${{ secrets.PGP_PRIVATE_KEY }}
       GITHUB_DEPLOY_KEY: ${{ secrets.GITHUB_DEPLOY_KEY }}
@@ -283,18 +282,13 @@ jobs:
         run: |
           yarn install
           touch ~/.profile
-          mkdir "$HOME/.ssh"
-          mkdir "$HOME/.sbt"
-          mkdir "$HOME/.sbt/gpg"
           git config --global user.name "Bloopoid"
           git config --global user.email "bloop@vican.me"
           git config --global push.default simple
-          echo "$PGP_PUBLIC_KEY" > "$HOME/.sbt/gpg/pubring.asc"
-          echo "$PGP_PRIVATE_KEY" > "$HOME/.sbt/gpg/secring.asc"
           ls -l bloop-artifacts
           mkdir -p frontend/target
           mv bloop-artifacts frontend/target/graalvm-binaries
-          ./bin/sbt-ci-publish.sh "releaseBloop" "docs/docusaurusPublishGhpages"
+          ./bin/sbt-ci-publish.sh "ci-release" "docs/docusaurusPublishGhpages"
         shell: bash
       - name: Cut GitHub release and update installers on ${{ matrix.os }}
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,11 +248,10 @@ jobs:
     env:
       CI: true
       SKIP_TEST_RESOURCES_GENERATION: true
-      SONATYPE_USERNAME: ${{ secrets.SONATYPE_USER }}
+      SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
       SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+      PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
       PGP_SECRET: ${{ secrets.PGP_SECRET }}
-      PGP_PUBLIC_KEY: ${{ secrets.PGP_PUBLIC_KEY }}
-      PGP_PRIVATE_KEY: ${{ secrets.PGP_PRIVATE_KEY }}
       GITHUB_DEPLOY_KEY: ${{ secrets.GITHUB_DEPLOY_KEY }}
       BLOOPOID_GITHUB_TOKEN: ${{ secrets.BLOOPOID_GITHUB_TOKEN }}
       BLOOPOID_SSH_PUBLIC_KEY: ${{ secrets.BLOOPOID_SSH_PUBLIC_KEY }}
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -248,10 +248,9 @@ jobs:
     env:
       CI: true
       SKIP_TEST_RESOURCES_GENERATION: true
-      SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
+      SONATYPE_USERNAME: ${{ secrets.SONATYPE_USER }}
       SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
-      PGP_PASSWORD: ${{ secrets.PGP_PASSWORD }}
-      PGP_PASSPHRASE: ${{ secrets.PGP_PASSWORD }}
```
ckipp01, line 254: From what I can tell these aren't used anywhere.

.github/workflows/ci.yml
```diff
@@ -283,18 +282,13 @@ jobs:
         run: |
           yarn install
           touch ~/.profile
-          mkdir "$HOME/.ssh"
-          mkdir "$HOME/.sbt"
-          mkdir "$HOME/.sbt/gpg"
```
ckipp01, line 288: It's seems that this and also the couple things that I removed down below are all covered by sbt-ci-release, so there shouldn't be any reason to manually do this anymore. We won't really be able to tell until we just try.

ckipp01 (commented):
One other thing I noticed is that we actually publish `bloop-shared`. However the other modules are using it via `dependsOn` and I don't think anything else actually uses this. Do we need to keep publishing it separately?

.github/workflows/ci.yml
```diff
@@ -248,10 +248,9 @@ jobs:
     env:
       CI: true
       SKIP_TEST_RESOURCES_GENERATION: true
-      SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
+      SONATYPE_USERNAME: ${{ secrets.SONATYPE_USER }}
```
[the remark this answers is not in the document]
ckipp01: Fixed

.github/workflows/ci.yml
ckipp01: Ahh you're correct. Good catch! Fixed

.github/workflows/ci.yml
```diff
@@ -248,10 +248,9 @@ jobs:
...
       SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
-      PGP_PASSWORD: ${{ secrets.PGP_PASSWORD }}
-      PGP_PASSPHRASE: ${{ secrets.PGP_PASSWORD }}
+      PGP_SECRET: ${{ secrets.PGP_SECRET }}
       PGP_PUBLIC_KEY: ${{ secrets.PGP_PUBLIC_KEY }}
       PGP_PRIVATE_KEY: ${{ secrets.PGP_PRIVATE_KEY }}
```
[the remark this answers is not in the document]
ckipp01: They're actually referenced in `set-up-key.sh`, although that doesn't seem to be used anywhere either. I'll remove both.

build.sbt
```diff
@@ -581,53 +577,13 @@ val allProjects = Seq(
...
-  bloopgunShaded,
-  bloopgun213,
-  bloopgunShaded213,
-  launcher,
-  launcherShaded,
-  launcher213,
-  launcherShaded213,
-  buildpressConfig,
-  buildpress
-)
```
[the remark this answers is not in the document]
ckipp01: Yes, they were before as well actually. I did double-check though.

ckipp01:
> Could you please check that versionScheme is defined?

It's actually not. I'm assuming we want to `early-semver` for this, or did you have something else in mind?

julienrf:
> > Could you please check that versionScheme is defined?
> 
> It's actually not. I'm assuming we want to `early-semver` for this, or did you have something else in mind?

`early-semver` would be perfect, but now that I start thinking about it I wonder if things are more complex than that. Especially the fact that we publish artifacts that contain shaded dependencies and others that don’t shade (probably those types of artifacts should have independent version numbers…).

In any case, if we claim that this project follows early-semver, then we should also set up checks to make sure we effectively conform to that claim. That means setting up MiMa or using sbt-version-policy.

I think that should be addressed separately.

ckipp01:
> I think that should be addressed separately.

Yes, I agree. Regarding even the versioning of the shaded stuff, I don't even think we need to shade. I was talking to the other offline about this, created [this discussion](https://github.com/scalacenter/bloop/discussions/1958) and also this [pr](https://github.com/scalacenter/bloop/pull/1961) to address / talk about it because the shading also brings a ton of complexity that is all to avoid hypothetical issues as far as I can tell.

build.sbt
```diff
@@ -581,53 +577,13 @@ val allProjects = Seq(
...
 val bloop = project
   .in(file("."))
   .disablePlugins(ScriptedPlugin)
   .aggregate(allProjectReferences: _*)
   .settings(
-    releaseEarly := { () },
     (publish / skip) := true,
-    publishLocalAllModules := {
-      BuildDefaults
-        .publishLocalAllModules(allProjectsToRelease)
```
adpi2, line 615: It seems that `allProjectsToRelease` contained more project than `allProjectReferences`. In the diff there are all the shaded and the buildpress projects. Will they be released by `ci-release` or, is it okay not to release them?

ckipp01: Ah good catch, yes I am missing a couple that I didn't realize weren't in there. I knew the shaded stuff wasn't in there, but I was hoping to get some thoughts on https://github.com/scalacenter/bloop/pull/1961 before finishing up this pr 😆. Apart from that it looks like I'm also missing the build press stuff. Just to compare here's what a `publishLocal` will give you:

```
 bloop-backend_2.12
 bloop-frontend_2.12
 bloop-js-bridge-0-6_2.12
 bloop-js-bridge-1_2.12
 bloop-launcher-core_2.12
 bloop-launcher-core_2.13
 bloop-launcher-test_2.12
 bloop-native-bridge-0-4_2.12
 bloop-shared_2.12
 bloopgun-core_2.12
 bloopgun-core_2.13
 sbt-bloop
 sockets
```
vs what was in the publish list before:

```
  bloopShared, // TODO do we actually need to publish this???
  backend,
  frontend,
  sbtBloop,
  nativeBridge04,
  jsBridge06,
  jsBridge1,
  sockets,
  bloopgun,
  bloopgunShaded,
  bloopgun213,
  bloopgunShaded213,
  launcher,
  launcherShaded,
  launcher213,
  launcherShaded213,
  buildpressConfig,
  buildpress
```
I'll make sure the others get added, but it'll be nice to get some thoughts on the shaded stuff first, because ideally we wouldn't have to include those here as well.

tgodzik (approved):
This looks awesome! Thanks @ckipp01 for looking into all those issues! :heart_eyes:

build.sbt
ckipp01: Alright now that the shaded stuff is gone, the new publish list will look like this:

```
bloop-backend_2.12
bloop-frontend_2.12
bloop-js-bridge-0-6_2.12
bloop-js-bridge-1_2.12
bloop-launcher-core_2.12
bloop-launcher-core_2.13
bloop-launcher-test_2.12
bloop-native-bridge-0-4_2.12
bloop-shared_2.12
bloopgun-core_2.12
bloopgun-core_2.13
buildpress_2.12
buildpressconfig_2.12
sbt-bloop
sockets
```

PR #1963, opened 2022-12-22
chore(deps): bump scalaz to 7.3.7

ckipp01:
So this isn't used very much at all, only for the Show typeclass, and
that's what was breaking here due to the introduction of `Cord`. I think
my little hack is fine, it' just creates the cord out of the result of
the `shows` method. However, I'm not 100% sure, so let's run the tests
and see. I also removed scalaz concurrent because from what I can tell,
it wasn't being used and with the new version, it's not there anymore.

Supersedes #1938


tgodzik (approved):
LGTM!

PR #1968, opened 2022-12-28
refactor(build): get rid of build warnings and unused things

ckipp01:
This pr is just some general cleanup.

- Moves to slash syntax in sbt
- Removes some unused sbt things
- Moves some sbt setting inline if they are only used once


tgodzik (approved):
LGTM! I think we could probably add the scalafix config and job now :thinking:

ckipp01:
> I think we could probably add the scalafix config and job now

Yea sure! Ha, there already is a `.scalafix.conf`, it's just not enforced.

PR #1973, opened 2022-12-29
Throw error from bloopGenerate task

G1ng3r:
Sbt task bloopGenerate should throw occurred errors to indicate failures during build import


integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -1139,6 +1135,12 @@ object BloopDefaults {
     BloopKeys.bloopGenerate
       .all(filter)
       .map(_.toSet)
+      .map(_.map {
+        case Inc(th) =>
+          logger.error(s"Couldn't run bloopGenerate. Cause:\n$th")
```
tgodzik, line 1140: So this should still allow the behaviour introduced in https://github.com/scalacenter/bloop/pull/1663 ? Meaning, fail, but still run bloop generate for all the projects it could.

tgodzik (changes requested):
Thanks for the great contribution! I have one question and it seems the jobs are failing due to explicit usage of `bloopGenerate := None` in some places. You should be able to change those occurrences to the new format.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
G1ng3r: Hmm, nope, that behaviour wont work after. Should we introduce some flag to bloopInstall to allow #1663 behaviour, cause it's look like failing import is preferred action?

G1ng3r: maybe we should introduce some new task which one should be run manually for #1663 behaviour

G1ng3r: Guess we can do some changes to keep both behaviours, I'll try

tgodzik: Thanks!

G1ng3r: I'm not sure about how tasks evaluate, I've read some time ago that there's no order guarantees. If at this point of evaluation all bloopGenerate tasks finished then #1663 is satisfied

tgodzik: Could we maybe just flatten the result first? Or do a fold on it? And once we get the final result we can throw the exception.

ckipp01: Just jumping in here to saw that I agree with @tgodzik. I think it's importing to keep the behavior to complete as much as possible and to continue to the end. Once we have the collection of results of all the `blooGenerate`s we should then be able to still determine if there was a single `Inc` and if so report on it.

G1ng3r: I made some changes, need your approve on that

tgodzik: Done! Sorry for the delay!

G1ng3r:
@tgodzik some checks failed with timeouts, is it some problem with my changes? or restart until it goes away?

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -1139,9 +1136,25 @@ object BloopDefaults {
     BloopKeys.bloopGenerate
       .all(filter)
       .map(_.toSet)
+      .map(_.foldLeft(Set.empty[Option[File]] -> Set.empty[Incomplete]) {
+        case ((succ, fail), res) =>
+          res match {
+            case Inc(th) => succ -> (fail + th)
```
ckipp01, line 1142:
```suggestion
            case Inc(incomplete) => succ -> (fail + incomplete)
```
or failure or something, I'm not 100% sure. But `th` isn't very clear to me.

ckipp01 (commented):
Just a nit on naming. As for the CI, don't worry about it the tests that failed have been very flaky recently.

Thanks a lot for working on this!

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
G1ng3r: done

ckipp01 (approved):
Great, thanks a lot for this @G1ng3r! 

@tgodzik do you have any thoughts or are we ok to merge this?

tgodzik (approved):
LGTM! Thanks for the great work!

PR #1979, opened 2023-01-06
chore: remove twitterIntegrationProjects

ckipp01:
This is failing in the dependency job and testing locally I
can't actually get these to resolve anymore either. Not
sure if Twitter is just nuking their maven instance or what
but I don't even think these are used, so we just get rid of
them.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -438,17 +438,6 @@ lazy val nativeBridge04 = project
     (Test / fork) := true
   )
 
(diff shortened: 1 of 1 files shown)


adpi2 (approved):
`https://maven.twttr.com` returns

> upstream connect error or disconnect/reset before headers. reset reason: connection failure

ckipp01:
> `https://maven.twttr.com` returns
> 
> > upstream connect error or disconnect/reset before headers. reset reason: connection failure

Yea exactly.

julienrf (approved):

PR #1980, opened 2023-01-06
Small test improvements

ckipp01:
This just cheery picks a couple of the commits from https://github.com/scalacenter/bloop/pull/1749 that I wanted in here. Then I'll close that seeing we've updated to setup-java already.


tgodzik (approved):
LGTM!

PR #1981, opened 2023-01-06
refactor(test): bump maxDuration in js tests

ckipp01:
This is consistently failing with messages like:

```
[error] Test bloop.scalajs.ScalaJsToolchainSpec.canLinkScalaJsProject failed: java.util.concurrent.TimeoutException: Futures timed out after [45 seconds], took 46.973 sec
```

My naive hope is that just bumping this might make this test less flaky.

### DIFF
--- a/bridges/scalajs-1/src/test/scala/bloop/scalajs/ScalaJsToolchainSpec.scala
+++ b/bridges/scalajs-1/src/test/scala/bloop/scalajs/ScalaJsToolchainSpec.scala
@@ -69,7 +69,7 @@ class ScalaJsToolchainSpec {
     logger.getMessages.assertContain("Hello, world!", atLevel = "info")
   }
 
-  private final val maxDuration = Duration.apply(45, TimeUnit.SECONDS)
+  private final val maxDuration = Duration.apply(60, TimeUnit.SECONDS)
   private implicit class RichLogs(logs: List[(String, String)]) {
     def assertContain(needle: String, atLevel: String): Unit = {
       def failMessage: String =


kpodsiad:
My 2c, totally without looking what does actually does: 45s is awfully a lot (and 60 even more), what are those tests doing? 😮

ckipp01:
> My 2c, totally without looking what does actually does: 45s is awfully a lot (and 60 even more), what are those tests doing? 😮

You're calling out my laziness. Let me spend a bit more time on them and get a better understanding before I answer that.

tgodzik (approved):
Fine with me! We might need to fix these tests properly at some points. but I haven't looked at them yet.

tgodzik:
I think we can 

> My 2c, totally without looking what does actually does: 45s is awfully a lot (and 60 even more), what are those tests doing? open_mouth

Btw. looking at this tests again the thing that is happening is:
```
Non-compiled module 'compiler-bridge_2.11' for Scala 2.11.12. Compiling...
  Compilation completed in 39.365s.
```
which leaves 6 seconds for the test to finish, which itself invokes other processes, so I think this is safe to bump. What do you think @ckipp01 ? Should we rebase and see if the tests become more stable? I think this is the main cause of flakiness nowadays.,

ckipp01:
> I think we can
> 
> > My 2c, totally without looking what does actually does: 45s is awfully a lot (and 60 even more), what are those tests doing? open_mouth
> 
> Btw. looking at this tests again the thing that is happening is:
> 
> ```
> Non-compiled module 'compiler-bridge_2.11' for Scala 2.11.12. Compiling...
>   Compilation completed in 39.365s.
> ```
> 
> which leaves 6 seconds for the test to finish, which itself invokes other processes, so I think this is safe to bump. What do you think @ckipp01 ? Should we rebase and see if the tests become more stable? I think this is the main cause of flakiness nowadays.,

Ah good context to have. Yea, honestly I'm fine with this. Let me rebase it and see if everything is 🟢 . I do think it should be pretty safe to bump.

tgodzik:
Looks all green :tada:

PR #1986, opened 2023-01-17
improvement: Use diagnostic related information from the compiler

tgodzik:
Looks like previously, this was totally ignored, so this should help, but not sure how to write tests until it's available in the compiler release.

I also changed 2 deprecated usages of the zinc api.


ckipp01 (approved):
Yes, I was excited to see https://github.com/lampepfl/dotty/pull/16677 get merged! Agree, it's hard to test this without anything actually emitting this. I think when the compiler starts actually adding some of these we can add some tests.

LGTM!

PR #1995, opened 2023-01-26
refactor: take care of a bunch of warnings

ckipp01:
These are mainly low-hanging fruit related to deprecations.

### DIFF
--- a/backend/src/main/scala/bloop/task/Task.scala
+++ b/backend/src/main/scala/bloop/task/Task.scala
@@ -132,7 +132,7 @@ sealed trait Task[+A] { self =>
 
   final def transform[R](fa: A => R, fe: Throwable => R): Task[R] =
     Task.Transform(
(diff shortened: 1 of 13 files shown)


kpodsiad (approved):

tgodzik (approved):
Looks good, but it seems a single test fails in at least two jobs:
`bloop.dap.DebugProtocolSpec.starts a debug session `

Might be unrelated, they have been flaky or maybe there is something going with that test even before.

ckipp01:
> Might be unrelated, they have been flaky or maybe there is something going with that test even before.

Let me re-run them and see.

ckipp01:
Alright, so it didn't fail on all of them, so it must be crazy flaky. In https://github.com/scalacenter/bloop/pull/1997 it failed for all of them as well. I'll merge this in, and then try to dig in a bit as to what's going on with that test.

PR #1996, opened 2023-01-26
test: re-enable hydra tests

ckipp01:
We have a fresh Hydra license that I've added to the secrets. Let's see
how re-enabling the tests work.


tgodzik:
I think at this point it's no longer relevant since we would have request the license again and I am not sure if Hydra is being very actively maintained.

So the tests will stay ignored until they want or need to send us a license.

PR #2016, opened 2023-03-09
fix: do not expose mutable env properties in State

kpodsiad:
fixes https://github.com/scalameta/metals/issues/5025

- change mutable PrettyProperties to wrapper over immutable Map
- adjust forker's implementations logic so they are not even aware of env parameters. They're handled one step before

I was thinking about some tests but immutable data structure should be just enough to prevent such cases in the future


kpodsiad:
Debug tests seems to be flaky as they passed on few checks and failed on 2

ckipp01 (approved):
👍🏼 LGTM

And yes, the failing tests have been super flaky lately.

PR #2029, opened 2023-04-04
Fix timeout in DebugProtocolTests

adpi2:
For some mysterious reason, traversing the `src.zip` of JDK Temurin 11.0.18 installed by `actions/setup-java` is very very slow. It takes up to 40 second, the first time, and then it takes about 500 ms. That's why the `DebugProtocolSpec` times out.

But if I install the exact same JDK with `coursier/setup-action`, the issue disappears. Similarly, if I manually download and extract the JDK, it fixes the issue. So it seems there is something wrong with `actions/setup-java`, about how it installs this JDK, but I don't know exactly what.

Switching to `coursier/setup-action` solves the issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,12 +112,12 @@ jobs:
           submodules: true
           fetch-depth: 0
 
-      - uses: actions/setup-java@v3
+      - uses: coursier/setup-action@v1
         with:
-          distribution: 'temurin'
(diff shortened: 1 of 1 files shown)


ckipp01:
> Switching to coursier/setup-action solves the issue.

It's been so long since I've seen a fully green CI, so if this truly fixes, I say let's go.

tgodzik (approved):

PR #2032, opened 2023-04-06
bugfix: Properly clean up orphan directories for internal classes

tgodzik:
It seems that the genericClassesDir is not being used at all by the compiler, so cleaning orphan directories in it was not achieving anything.

We might need to figure out how to use `genericClassesDir` and whether it needs to be used at all, but for now this fixes the issue of large .bloop directory.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
+++ b/frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
@@ -170,6 +170,16 @@ object ResultsCache {
   ): Task[ResultsCache] = {
     import bloop.util.JavaCompat.EnrichOptional
 
+    /**
+     * Orphaned internal directories are directories that are not used by any
+     * client and that are currently not the internal classes directory of any project.
+     *
+     * These are usually deleted by Bloop right away, but in some cases there might be left,
+     * which means it's safe to delete them.
+     *
+     * An example path cleaned up by this method is:
+     * .bloop/mtest/bloop-internal-classes/classes-Metals-12MpgYmRSGOOK5fxouxTqg==-iWRxh-luRTOxk3bOx9UAfg==
+     */
     def cleanUpOrphanedInternalDirs(
         project: Project,
         analysisClassesDir: AbsolutePath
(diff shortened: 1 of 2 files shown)


frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
```diff
@@ -177,7 +177,7 @@ object ResultsCache {
       if (!cleanOrphanedInternalDirs) Task.unit
       else {
         val internalClassesDir =
-          CompileOutPaths.createInternalClassesRootDir(project.genericClassesDir)
+          CompileOutPaths.createInternalClassesRootDir(project.out)
```
kpodsiad, line 180: Could you provide concrete values for `project.out` and `project.genericClassesDir`? Is one of them `bloop-internal-classes`?

tgodzik: `project.out` is the currently used one for example `/.bloop/input3/` while `genericClassesDir` is `.bloop/input3/scala-3/classes`, which is actually defined in the json file.

ckipp01: I'm actually a bit surprised by this, I thought it _would_ use the one specific to the scala version and therefore the nested on in `.bloop/input3/scala-3/classes`.

tgodzik: That is kind of redundant since you can't have multiple Scala versions for one project, so I don't think it's really useful. Also, it's a smaller change to just ignore that directory :S

tgodzik: also, it seems that we always use `bloop-bsp-clients-classes` always instead of `classesDir`, so this looks legit as if we had only a single output we could have issues with mutiple clients. (whether we could have mutiple BSP client for the same workspace is another thing altogether.)

kpodsiad: Can you add scaladoc which will explain what orphaned directories are? Giving some concrete examples you did int his thread will awesome

tgodzik: Ok, more digging it seems the is an option for the tool to use managed classes directory, but there is even more logic on top of it to dermine it.

ckipp01: > That is kind of redundant since you can't have multiple Scala versions for one project, so I don't think it's really useful. Also, it's a smaller change to just ignore that directory :S

ah ok cool, thanks for the explanations 😄

ckipp01 (approved):

frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
tgodzik: I added scaladoc to this method. I will probably dig in a bit more later on.

PR #2049, opened 2023-05-12
Support Scala 3's Best Effort compilation

jchyb:
## Related PRs
* dotty: https://github.com/lampepfl/dotty/pull/17582
* metals: https://github.com/scalameta/metals/pull/5219

## Explanation

This commit allows bloop to support the hopefully-upcoming Scala 3 best effort compilation for Metals, it being enabled with the other Metals options.

Best Effort is meant to be a set of Scala 3 compiler options that allow the errored, but typed programs, to be able to be serialized into a TASTy aligned format (Best Effort TASTy), and later to reuse those typed program trees in both the dependent projects, and in the presentation compiler.

~~Those best effort tasty files are always serialized to a seperate directory, which in this PR is managed by bloop.~~ It is worth noting that the best effort compilation may fail, similarly to the regular compilation.

~~The best effort directory is kept outside of the regular build directory, in `.bloop/project-id/scala-3/best-effort`. There, two directories are managed, `build` and `dep`. Whenever metals compiles a project the `build` directory is populated with best effort artifacts. Then, if the best effort compilation is successful, they are copied to the 'dep' directory, which is used for dependencies, metals presentation compiler etc. This way if a compilation fails, the artifacts from the previous compilation can still be kept. This is not completely dissimilar to how the regular compilation works here, with the difference being the more straightforward implementation, without any counters that monitor how many tasks are being executed with the `dep` directory concurrently.~~

There may be some unnecessary redundancies in this draft PR,~~ where sometimes information about best effort directories of a project is unnecessarily passed around~~. Also, I probably need to add some tests.

### DIFF
--- a/backend/src/main/scala/bloop/BloopClassFileManager.scala
+++ b/backend/src/main/scala/bloop/BloopClassFileManager.scala
@@ -216,19 +216,36 @@ final class BloopClassFileManager(
         ) => {
           clientTracer.traceTaskVerbose("copy new products to external classes dir") { _ =>
             val config = ParallelOps.CopyConfiguration(5, CopyMode.ReplaceExisting, Set.empty)
-            ParallelOps
-              .copyDirectories(config)(
-                newClassesDir,
-                clientExternalClassesDir.underlying,
-                inputs.ioScheduler,
-                enableCancellation = false,
-                inputs.logger
-              )
-              .map { walked =>
-                readOnlyCopyDenylist.++=(walked.target)
+            val clientExternalBestEffortDir =
+              clientExternalClassesDir.underlying.resolve("META-INF/best-effort")
+
+            // Deletes all previous best-effort artifacts to get rid of all of the outdated ones.
+            // Since best effort compilation is not affected by incremental compilation,
+            // all relevant files are always produced by the compiler. Because of this,
+            // we can always delete all previous files and copy newly created ones
+            // without losing anything in the process.
+            val deleteClientExternalBestEffortDir =
+              Task {
+                if (Files.exists(clientExternalBestEffortDir)) {
+                  BloopPaths.delete(AbsolutePath(clientExternalBestEffortDir))
+                }
                 ()
-              }
-              .flatMap(_ => deleteAfterCompilation)
+              }.memoize
+
+            deleteClientExternalBestEffortDir *>
+              ParallelOps
+                .copyDirectories(config)(
+                  newClassesDir,
+                  clientExternalClassesDir.underlying,
+                  inputs.ioScheduler,
+                  enableCancellation = false,
+                  inputs.logger
+                )
+                .map { walked =>
+                  readOnlyCopyDenylist.++=(walked.target)
+                  ()
+                }
+                .flatMap(_ => deleteAfterCompilation)
           }
         }
       )
--- a/backend/src/main/scala/bloop/BloopClassFileManager.scala
+++ b/backend/src/main/scala/bloop/BloopClassFileManager.scala
@@ -40,8 +40,6 @@ final class BloopClassFileManager(
   private[this] val weakClassFileInvalidations = new mutable.HashSet[Path]()
   private[this] val generatedFiles = new mutable.HashSet[File]
 
-  // Supported compile products by the class file manager
-  private[this] val supportedCompileProducts = List(".sjsir", ".nir", ".tasty", ".betasty")
   // Files backed up during compilation
   private[this] val movedFiles = new mutable.HashMap[File, File]
 
@@ -140,7 +138,7 @@ final class BloopClassFileManager(
 
     val invalidatedExtraCompileProducts = classes.flatMap { classFile =>
       val prefixClassName = classFile.getName().stripSuffix(".class")
-      supportedCompileProducts.flatMap { supportedProductSuffix =>
+      BloopClassFileManager.supportedCompileProducts.flatMap { supportedProductSuffix =>
         val productName = prefixClassName + supportedProductSuffix
         val productAssociatedToClassFile = new File(classFile.getParentFile, productName)
         if (!productAssociatedToClassFile.exists()) Nil
@@ -186,7 +184,7 @@ final class BloopClassFileManager(
           .stripSuffix(".class") + supportedProductSuffix
         new File(classFile.getParentFile, productName)
       }
-      supportedCompileProducts.foreach { supportedProductSuffix =>
+      BloopClassFileManager.supportedCompileProducts.foreach { supportedProductSuffix =>
         val generatedProductName = productFile(generatedClassFile, supportedProductSuffix)
         val rebasedProductName = productFile(rebasedClassFile, supportedProductSuffix)
 
@@ -318,6 +316,9 @@ final class BloopClassFileManager(
 }
 
 object BloopClassFileManager {
+  // Supported compile products by the class file manager
+  val supportedCompileProducts = List(".sjsir", ".nir", ".tasty", ".betasty")
+
   def link(link: Path, target: Path): Try[Unit] = {
     Try {
       // Make sure parent directory for link exists
--- a/backend/src/main/scala/bloop/BloopClassFileManager.scala
+++ b/backend/src/main/scala/bloop/BloopClassFileManager.scala
@@ -41,7 +41,7 @@ final class BloopClassFileManager(
   private[this] val generatedFiles = new mutable.HashSet[File]
 
   // Supported compile products by the class file manager
-  private[this] val supportedCompileProducts = List(".sjsir", ".nir", ".tasty")
+  private[this] val supportedCompileProducts = List(".sjsir", ".nir", ".tasty", ".betasty")
   // Files backed up during compilation
   private[this] val movedFiles = new mutable.HashMap[File, File]
 
@@ -215,7 +215,8 @@ final class BloopClassFileManager(
             clientTracer: BraveTracer
         ) => {
           clientTracer.traceTaskVerbose("copy new products to external classes dir") { _ =>
-            val config = ParallelOps.CopyConfiguration(5, CopyMode.ReplaceExisting, Set.empty)
+            val config =
+              ParallelOps.CopyConfiguration(5, CopyMode.ReplaceExisting, Set.empty, Set.empty)
             val clientExternalBestEffortDir =
               clientExternalClassesDir.underlying.resolve("META-INF/best-effort")
 
@@ -291,7 +292,8 @@ final class BloopClassFileManager(
             else
               clientTracer.traceTask("populate empty classes dir") { _ =>
                 // Prepopulate external classes dir even though compilation failed
-                val config = ParallelOps.CopyConfiguration(1, CopyMode.NoReplace, Set.empty)
+                val config =
+                  ParallelOps.CopyConfiguration(1, CopyMode.NoReplace, Set.empty, Set.empty)
                 ParallelOps
                   .copyDirectories(config)(
                     Paths.get(readOnlyClassesDirPath),
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -402,15 +402,19 @@ object Compiler {
               val clientClassesDir = clientClassesObserver.classesDir
               clientLogger.debug(s"Triggering background tasks for $clientClassesDir")
               val firstTask = Task { BloopPaths.delete(AbsolutePath(newClassesDir)) }
-              val secondTask = updateExternalClassesDirWithReadOnly(
-                clientClassesDir,
-                clientTracer,
-                clientLogger,
-                compileInputs,
+              val config =
+                ParallelOps.CopyConfiguration(
+                  parallelUnits = 5,
+                  CopyMode.ReplaceIfMetadataMismatch,
+                  denylist = Set.empty,
+                  denyDirs = Set.empty
+                )
+              val secondTask = ParallelOps.copyDirectories(config)(
                 compileProducts.newClassesDir,
-                readOnlyCopyDenylist = mutable.HashSet.empty,
-                allInvalidatedClassFilesForProject,
-                allInvalidatedExtraCompileProducts
+                clientClassesDir.underlying,
+                compileInputs.ioScheduler,
+                enableCancellation = false,
+                compileInputs.logger
               )
               Task
                 .gatherUnordered(List(firstTask, secondTask))
@@ -716,8 +720,11 @@ object Compiler {
           allInvalidatedExtraCompileProducts.iterator.map(_.toPath).toSet
         val invalidatedInThisProject = invalidatedClassFiles ++ invalidatedExtraProducts
         val denyList = invalidatedInThisProject ++ readOnlyCopyDenylist.iterator
+        // Let's not copy outdated betasty from readOnly, since we do not have a mechanism
+        // for tracking that otherwise
+        val denyDir = Set(readOnlyClassesDir.resolve("META-INF/best-effort"))
         val config =
-          ParallelOps.CopyConfiguration(5, CopyMode.ReplaceIfMetadataMismatch, denyList)
+          ParallelOps.CopyConfiguration(5, CopyMode.ReplaceIfMetadataMismatch, denyList, denyDir)
         val lastCopy = ParallelOps.copyDirectories(config)(
           readOnlyClassesDir,
           clientClassesDir.underlying,
@@ -934,6 +941,8 @@ object Compiler {
                   syntax.replace(readOnlyClassesDirPath, clientClassesDirPath)
                 )
                 if (rebasedFile.exists) {
+                  // In practice this deletes previous successful compilation artifacts
+                  // (like '.tasty' and ".class"), which we do not need in clientDir
                   Files.delete(rebasedFile.underlying)
                 }
               }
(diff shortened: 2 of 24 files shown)


backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -70,6 +71,7 @@ case class CompileOutPaths(
     analysisOut: AbsolutePath,
     genericClassesDir: AbsolutePath,
     externalClassesDir: AbsolutePath,
+    bestEffortDirs: Option[BestEffortDirs],
```
tgodzik, line 74: Could we instead put it into META-INF directory similar to semanticdb? Or would that be picked up as normal classfiles?

Alternatively, if META-INF is not an option I would just put into a totally separate directory and ideally not handle it specifically in Bloop. The reason for that is that later we would need to reimplement this in sbt, mill, bazel, which will be a lot of problems.

jchyb: There should not be any conflict between the best effort and regular tasty, so that is not an issue. The problem is that I needed to add a special handling of these "best effort tasty" files into bloop, regardless of where they will be put (but for it to work I needed 2 separate directories for those files). I'll try to explain:

There is a subset of compilation errors (minority), for which the compiler will not be able to produce those betasty files for now (and it does not sound easily fixable), instead just crashing (probably not ok, will need to have a separate error for best effort compilation fail). In those cases, metals would work as before. That creates an issue, that when a successful best effort compilation is followed by an unsuccessful BE compilation we are left with nothing (unlike following successful comp with a failing one, where we still have info from the successful one), as we delete the previous be-tasty files before recompiling to avoid stale signatures, etc. So I came up with having a separate directories for compilation (always starts empty, is filled with new betasty files), and dependencies (where we copy from the previous directory after confirming the best-effort compilation was successful, and actually use it in metals presentation compiler and downstream projects). Bloop actually already does something similar for regular compilation, but in a much more complex manner, with semaphore-like counters, etc. Now, we don't do that with the semanticdb files - I do not remember why this never became an issue there, will need to reinvestigate, maybe I am missing something.

I will admit I knew I would need to reimplement this for sbt, but did not know there were even more build servers to support (I incorrectly assumed that when using other build tools only bloop build server was available in metals). I guess we could keep the bloop behavior here, and just use a single directory (like META-INF) in other build servers (without the nice fail with successful best effort to fail with unsuccessful best effort fallback)

tgodzik: > Now, we don't do that with the semanticdb files - I do not remember why this never became an issue there, will need to reinvestigate, maybe I am missing something.

What we actually do currently is that when files are invalidated, we back them up and move them back if the compilation failed:

https://github.com/scalacenter/bloop/blob/02774ea2a473be856e4c9415908c58b9392ed11b/backend/src/main/scala/bloop/BloopClassFileManager.scala#L235

We also do that for supported products:
`val supportedCompileProducts = List(".sjsir", ".nir", ".tasty")`

I think semanticdb files are not invalidated (that might be bug though :thinking: )

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -606,26 +672,158 @@ object Compiler {
...
+              Result.Failed(Nil, Some(t), elapsed, backgroundTasks, None)
           }
       }
   }
 
+  /**
+   * Handles successful Best Effort compilation.
+   * Does not persist incrementalCompilation analysis.
+   */
+  def handleBestEffortSuccess(
```
tgodzik, line 699: So this is needed so that we don't persist analysis? What happens if we do? Could explain a bit more in the comment?

jchyb: Oh sorry, this was a forget-me-not I ironically forgot about. I'm still going to have to run through this method, clean it up and make things less redundant - but basically now everything should at the very least work. About the comment, for now I reused a some methods from the way successful compilation is handled, however I had to remove handling of analysis.bin files from those for two reasons:
* to work it needs xsbti.compile.MiniSetup obtained from "successful" zinc compilation which we do not always have
* I guess we could reimplement it but I saw no purpose to it for those "best-effort" cases - it does not seem to affect anything there

jchyb:
I fixed the issue mentioned by me above. I also removed the custom path handling for best effort compilation as asked. Now instead of specifying custom directory, we get best-effort-tasty files in META-INF/best-effort, similarly to semanticdb. 

Now, I could not just reuse the aforementioned behavior in `BloopClassFileManager.scala` for handling outdated files, since it predominantly relies on recognizing which classifies were generated/deleted on making decisions based on that (not a luxury we have). Still, I could very easily remove the weird custom two-directory structure and custom handlings that came with it. I also could remove changes to the `bsp` and `bloop-config` projects which is a big plus, as previously we would have to use them to pass those directories around. 

The PR is still very much WIP, the things I need to add:
[] most important - I believe it now does full recompilation of every upstream "failed" best-effort project, so I am going to need to find what causes this
[] add (any) tests
[] proper handling of failed best-effort compilations (now we rely on seeing if the compiler crashes, this of course will be changed into a separate error with non crash stack-trace, like is done for failing macros)
[] cleanup of the handleBestEffortSuccess, I kind of hacked it together quickly for now and it has some redundant elements (sorry)

backend/src/main/scala/bloop/Compiler.scala
jchyb: I looked into the semanticdb thing very quickly while working on this now and it turns out there indeed seems to be a bug, but just adding `.semanticdb` will not fix it. The thing is the above file types are all tied to classfiles, which is a fact used during the invalidation. Semanticdb is tied to a source file, which means that it's harder to notice the invalid signatures, but they can be generated by simultaneous renamings of the file and changes to it.

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -357,13 +366,74 @@ object Compiler {
...
+            }
+          }
+          Some(Failed(problems, t, elapsed, backgroundTasks, bestEffortProducts))
+        } else None
+      case _ => None
+    }
+
+    if (noopBestEffortResult.isDefined) {
+      logger.debug("Skipping redundant best-effort compilation")
+      return Task { noopBestEffortResult.get }
```
jchyb, line 427: This is probably a bit ugly as nonLocal returns are not used anywhere else in bloop, but any other solution would destroy the diff in Compiler.scala. Maybe I should squash the commits into one and then add a change from this to an else/pattern match, so that the review can be done on the basis of the first commit, but the actual merged code will be cleaner?

tgodzik: Might be good to squash as it's already a lot of commits with contradicting changes

frontend/src/main/scala/bloop/engine/tasks/compilation/CompileGraph.scala
```diff
@@ -197,7 +183,7 @@ object CompileGraph {
                 case a: ReporterAction.ProcessEndCompilation =>
                   a.code match {
                     case BspStatusCode.Cancelled | BspStatusCode.Error =>
-                      reporter.processEndCompilation(previousSuccessfulProblems, a.code, None, None)
+                      reporter.processEndCompilation(previousProblems, a.code, None, None)
                       reporter.reportEndCompilation()
```
jchyb, line 187: Changes here (and similar below) are not strictly related to best effort compilation, but I found out that clicking go-to-reference from metals on a dependency project pointing to a dependent project, I would get the "Deduplicating compilation of (...)" log and I would lose all of the diagnostics from metals. This (and the same below) is the culprit, and it seems to me that returning previousProblems (not necessarily from last successful run) is more correct here

tgodzik: och, good catch! Any chance to get it as a separate fix maybe?

jchyb: Sure!

backend/src/main/scala/bloop/BloopClassFileManager.scala
```diff
@@ -216,19 +216,36 @@ final class BloopClassFileManager(
...
-                inputs.ioScheduler,
-                enableCancellation = false,
-                inputs.logger
-              )
-              .map { walked =>
-                readOnlyCopyDenylist.++=(walked.target)
+            val clientExternalBestEffortDir =
+              clientExternalClassesDir.underlying.resolve("META-INF/best-effort")
+
+            // Deletes all previous best-effort artifacts to get rid of all of the outdated ones.
```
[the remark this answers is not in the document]
jchyb: Not at the moment unfortunately. Maybe I should add an option in metals to be able disable best-effort? It should be able to start the compilation with the files from last successful compilation, but I do not remember if that's the case to be honest (it should be)

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -357,13 +366,74 @@ object Compiler {
 
     val uniqueInputs = compileInputs.uniqueInputs
     reporter.reportStartCompilation(previousProblems)
+    val fileManager = newFileManager
+
+    // Manually skip redundant best-effort compilations. This is necessary because compiler
+    // phases supplying the data needed to skip compilations in zinc remain unimplemented for now.
```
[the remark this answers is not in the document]
jchyb: Yes, I tried to enable those phases, but it would have taken way too much time to do it properly (without excessive crashing). I planned to go back and start implementing that after the initial merges (and I would still love to do that), but to be completely transparent I do not know if that will be possible (time-wise).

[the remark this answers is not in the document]
tgodzik: ach, right. Could we make sure there is an issue about it in the compiler?

frontend/src/main/scala/bloop/data/Project.scala
```diff
@@ -429,6 +433,26 @@ object Project {
       version != "3.0.0-M2"
     }
 
+    def canEnableBestEffortFlag(version: String): Boolean = {
+      val split = version.split('.')
+      val major = split(0).toInt
```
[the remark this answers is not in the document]
jchyb: Looks like string interpolation unapply and toIntOption were not in scala 2.12 yet, so I did the next best things (wrapped with a try)

backend/src/main/scala/bloop/BloopClassFileManager.scala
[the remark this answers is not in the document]
tgodzik: Do you mean that a new betasty directory is always spawned? It's fine if we remove them later on.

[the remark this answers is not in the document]
jchyb: It is always spawned, but I was talking about the new internal bloop directory being created (and the previous one not being deleted) when following failing compilation with a successful one. That's not a big issue (outside of spamming a little bit with files in .bloop), but I also saw that we delete the previous successful artifacts (.class and .tasty) altogether in client directory when ending up with betasty, which I think is what I wanted to write before, since you were asking about deleting old artifacts - sorry about that.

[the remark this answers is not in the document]
jchyb: Basically:
- we have a successful compilation, we end up with META-INF/betasty, semanticdb and regular classifies and tasty in the client directory (which metals uses)
- we follow this with an unsuccessful compilation with some kind of error, we end up with just META-INF/betasty, semanticdb (other files get deleted, I already found where that happens)

backend/src/main/scala/bloop/Compiler.scala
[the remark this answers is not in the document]
jchyb: Done: https://github.com/scala/scala3/issues/20704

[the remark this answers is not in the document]
tgodzik: Thanks!

backend/src/main/scala/bloop/BloopClassFileManager.scala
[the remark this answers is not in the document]
tgodzik: The flow is a bit complex with the different directories, I was thinking of simplifying it, but didn't get around to it yet. You mean to fix it in the current PR? Let me know if you need any help.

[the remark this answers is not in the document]
jchyb: Yes, fixing it right now

jchyb:
@tgodzik Here is what I actually changed since the last review:
* I made it so that betasty coming from successful compilation is not copied into a clientDir (`classes-Metals…`).
* I discovered that before, it was easy to produce an outdated betasty file that would not be cleaned up by bloop, by alternating between successful compilations and unsuccessful ones. I added a test for that too.

About copying last successful compilation artifacts into a clientDir to reuse in incremental compilations. I got a little confused before and it looks like there are two bloop-internal-classes directories used in a compilation: `readOnlyClassesDir` and `newClassesDir`. In best effort compilation, `readOnlyClassesDir` from what I understand will always represent the last successful compilation (with .class and .tasty), and newClassesDir will include newly compiled artifacts (`.betasty`  when compiling erroring file). So perhaps we do not need to copy successful artifacts into a clientDir as I mentioned before (since it is still saved and reused on the bloop side).

tgodzik:
>  So perhaps we do not need to copy successful artifacts into a clientDir as I mentioned before (since it is still saved and reused on the bloop side).

Ach yes, pretty sure incremental compilation has nothing to do with the client dir. From what I remember there are 3 different directories:
- readonly for current compilation later to be copied over
- internal  bloop directory
- client directories for each client (pretty sure we can get rid of it.)

For incremental compilation only the second is important for sure.

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -934,6 +941,8 @@ object Compiler {
                   syntax.replace(readOnlyClassesDirPath, clientClassesDirPath)
                 )
                 if (rebasedFile.exists) {
+                  // In practice this deletes previous successful compilation artifacts
+                  // (like '.tasty' and ".class"), which we do not need in clientDir
```
tgodzik, line 945: Why wouldn't we need class files in client directory? That would be where metals reads them from no?

jchyb: Yes, but after a best effort compilation with compilation errors (producing only betasty), metals would only need to read betasty there - other files (coming from readOnly) would be outdated at that point and would contain outdated symbols. This is the part of the code I was looking for after the last review, after incorrectly assuming clientDir participated in incremental compilation, so I chose it fit to add this comment here (since I was confused where that was happening)

backend/src/main/scala/bloop/BloopClassFileManager.scala
```diff
@@ -41,7 +41,7 @@ final class BloopClassFileManager(
   private[this] val generatedFiles = new mutable.HashSet[File]
 
   // Supported compile products by the class file manager
-  private[this] val supportedCompileProducts = List(".sjsir", ".nir", ".tasty")
+  val supportedCompileProducts = List(".sjsir", ".nir", ".tasty", ".betasty")
```
tgodzik, line 44: Could we just put it into an object? We wouldn't need to pass manager in the params

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -357,13 +366,78 @@ object Compiler {
...
+
+    // Manually skip redundant best-effort compilations. This is necessary because compiler
+    // phases supplying the data needed to skip compilations in zinc remain unimplemented for now.
+    val noopBestEffortResult = compileInputs.previousCompilerResult match {
+      case Failed(
+            problems,
+            t,
+            elapsed,
+            _,
+            bestEffortProducts @ Some(BestEffortProducts(compileProducts, previousHash))
```
tgodzik, line 379:
```suggestion
            bestEffortProducts @ Some(BestEffortProducts(previousCompilationResults, previousHash))
```
minor, but this seems misleading down since you delete newClassesDir and then copy from compileProducts.newClassesDir

And those are two different directories, right?  You delete the current directory and copy over old classes dir?

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -357,13 +366,78 @@ object Compiler {
...
+          val backgroundTasks = new CompileBackgroundTasks {
+            def trigger(
+                clientClassesObserver: ClientClassesObserver,
+                clientReporter: Reporter,
+                clientTracer: BraveTracer,
+                clientLogger: Logger
+            ): Task[Unit] = Task.defer {
+              val clientClassesDir = clientClassesObserver.classesDir
+              clientLogger.debug(s"Triggering background tasks for $clientClassesDir")
+              val firstTask = Task { BloopPaths.delete(AbsolutePath(newClassesDir)) }
```
tgodzik, line 404: Could you explain a bit more in a comment why is this remove and later we copy from `compileProducts.newClassesDir`?

backend/src/main/scala/bloop/Compiler.scala
jchyb: Yes, `newClassesDir` Is a newly created directory for new scalac compilation artifacts (which we will not have in this case so we delete it). The we copy to the previous (complete) results to clientDir (useful when we e.g restart metals and it's assigned a new empty directory)

backend/src/main/scala/bloop/Compiler.scala
jchyb: Sure! Just added some comments there

tgodzik (approved):
Thanks! LGTM

tgodzik:
Btw. you test out `1.5.18-38-79f9e6e0-SNAPSHOT` version on the Metals PR

PR #2055, opened 2023-05-24
Report java error column

Arthurm1:
Currently Bloop reports java errors as always being at the start of the line.  Now it includes the column as well.

I think this worked at one point but maybe got lost in the move away from the Zinc fork?

I don't know why Zinc reports the column of the error in the `offset` field instead of the `pointer` field as the code assumed - maybe it's an issue with Zinc?  Anyway - I don't see a problem in checking both fields.

I've tested this locally with Java code and Metals now underlines the correct error range.


Arthurm1:
`bloop.dap.DebugServerSpec` error - doesn't look relevant

tgodzik (approved):
Thanks! This must have been missed when moving from the fork, though I thought it was added to the main fork?

Issue #2053, opened 2023-05-18
Bloop fails to compile project with weird error


ktonga:
It happens for only one of the modules (maybe some other after passing that one too) in a multi-module project, there are many modules before that one is attempted which finish compiling without issue. Also if I compile using sbt it doesn't happen.

I wasn't able to isolate the issue in a smaller self-contained project. I can help with any debugging if I'm guided, TBH I don't know where I could start from. To see that error I had to add the `--verbose` flag.

```
[E] Unexpected error when compiling merchant-model: 'null'
[T] java.nio.file.ClosedFileSystemException
[T] 	jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.ensureOpen(ZipFileSystem.java:1628)
[T] 	jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.getFileAttributes(ZipFileSystem.java:536)
[T] 	jdk.zipfs/jdk.nio.zipfs.ZipPath.readAttributes(ZipPath.java:767)
[T] 	jdk.zipfs/jdk.nio.zipfs.ZipPath.readAttributes(ZipPath.java:777)
[T] 	jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.readAttributes(ZipFileSystemProvider.java:276)
[T] 	java.base/java.nio.file.Files.readAttributes(Files.java:1851)
[T] 	java.base/java.nio.file.Files.size(Files.java:2468)
[T] 	sbt.internal.inc.HashUtil$.farmHash(HashUtil.scala:31)
[T] 	sbt.internal.inc.FarmHash$.ofPath(Stamp.scala:89)
[T] 	sbt.internal.inc.Stamper$.$anonfun$forHashInRootPaths$3(Stamp.scala:233)
[T] 	sbt.internal.inc.Stamper$.tryStamp(Stamp.scala:188)
[T] 	sbt.internal.inc.Stamper$.$anonfun$forHashInRootPaths$1(Stamp.scala:233)
[T] 	sbt.internal.inc.UncachedStamps.library(Stamp.scala:468)
[T] 	sbt.internal.inc.InitialStamps.$anonfun$library$1(Stamp.scala:422)
[T] 	scala.collection.mutable.HashMap.getOrElseUpdate(HashMap.scala:86)
[T] 	sbt.internal.inc.InitialStamps.library(Stamp.scala:422)
[...]

PR #2058, opened 2023-05-25
improvement: Don't fail on hashing, but use a random hash

tgodzik:
Previously, when hashing would fail the whole compilation would also fail and while this seems incredibly rare let's try to guard ourselves here.

Connected to https://github.com/scalacenter/bloop/issues/2053

It's the safest option, we could also try another stamp in this case, but I think there is something wrong with that jar


adpi2 (approved):

ckipp01 (approved):
👍🏼

PR #2060, opened 2023-05-30
bugfix: Don't accept files other than .class or .jar as binary deps

tgodzik:
Should help with https://github.com/scalacenter/bloop/issues/2053

The binary is described as:
```
A binary entry represents either the jar or the
concrete class file from which the Scala compiler
knows that <code>onBinaryClassName</code> comes from.
```

I am kind of lost here, since no idea how would this even pop up


tgodzik:
Ok seems, this was not as easy :/

adpi2 (approved):

adpi2:
Clicked the :heavy_check_mark: button a bit fast: some tests are failing

PR #2081, opened 2023-06-21
improvement: Automatically include semanticdbjar for tests

tgodzik:
Previously, we would use a commited semanticdb jar for tests in frontend, however this caused two issues:

- semanticdb would be always out of date
- scala version updates would require also bumping semanticdb manually and fetching the jar

Now, we use the latest semanticdb plugin always, which should work properly.


ckipp01 (approved):
👍🏼

PR #2086, opened 2023-06-25
chore: Remove unused python check

tgodzik:
The tests work on my machine despite not having python on path, so most likely this is no longer valid as the launcher uses snailgun.


ckipp01 (approved):
LGTM!

PR #2087, opened 2023-06-25
chore: Remove nailgun module and NailgunSpec, which is no longer rele…

tgodzik:
…vant


ckipp01 (approved):
Yay for getting rid of stuff that is no longer relevant. LGTM!

PR #2089, opened 2023-06-26
refactor: Remove cancelPromise, which was used mostly for logging

tgodzik:
I am trying to remove some cancelation logic, which doesn't seem 100% neccessary or try to handle it differently. This might help in migrating off monix.

Here, I think we mostly used the cancelation for logging purposes, which is not really something users care a lot.

### DIFF
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -57,7 +57,6 @@ case class CompileInputs(
     reporter: ZincReporter,
     logger: ObservedLogger[Logger],
     dependentResults: Map[File, PreviousResult],
-    cancelPromise: Promise[Unit],
     tracer: BraveTracer,
     ioScheduler: Scheduler,
     ioExecutor: Executor,
@@ -165,10 +164,7 @@ object CompileOutPaths {
 object Compiler {
   private implicit val filter: DebugFilter.Compilation.type = bloop.logging.DebugFilter.Compilation
   private val converter = PlainVirtualFileConverter.converter
-  private final class BloopProgress(
-      reporter: ZincReporter,
-      cancelPromise: Promise[Unit]
-  ) extends CompileProgress {
+  private final class BloopProgress(reporter: ZincReporter) extends CompileProgress {
     override def startUnit(phase: String, unitPath: String): Unit = {
       reporter.reportNextPhase(phase, new java.io.File(unitPath))
     }
@@ -179,12 +175,8 @@ object Compiler {
         prevPhase: String,
         nextPhase: String
     ): Boolean = {
-      val isNotCancelled = !cancelPromise.isCompleted
-      if (isNotCancelled) {
-        reporter.reportCompilationProgress(current.toLong, total.toLong)
-      }
-
-      isNotCancelled
+      reporter.reportCompilationProgress(current.toLong, total.toLong)
+      true
     }
   }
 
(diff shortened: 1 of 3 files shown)


backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -386,14 +368,11 @@ object Compiler {
...
         uniqueInputs,
         newFileManager,
-        cancelPromise,
         tracer,
         classpathOptions
       )
       .materialize
-      .doOnCancel(Task(cancel()))
       .map {
-        case Success(_) if cancelPromise.isCompleted => handleCancellation
```
tgodzik, line 396: @dos65 Could you double check my logic?

If cancel promise is completed in the doOnCancel above, which is the only place it's completed, then we would never get to this map unless we already went through it.

In other words, would this branch would ever be achievable?

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopHighLevelCompiler.scala
```diff
@@ -140,12 +139,7 @@ final class BloopHighLevelCompiler(
             case NonFatal(t) =>
               // If scala compilation happens, complete the java promise so that it doesn't block
               JavaCompleted.tryFailure(t)
-
-              t match {
-                case _: NullPointerException if cancelPromise.isCompleted =>
-                  throw new InterfaceCompileCancelled(Array(), "Caught NPE when compilation was cancelled!")
```
tgodzik, line 146: This seems to be useful to just wrap the exception, which doesn't seem neccessary.

backend/src/main/scala/bloop/Compiler.scala
tgodzik: The reason why I am thinking this is that if cancelation cancel a particular task we don't expect it to go through the next chains, right? We wanted all the cancels to happen, but not really anything else.

tgodzik: I might need to dig in here a bit more. I haven't noticed other usage.

dos65: > If cancel promise is completed in the doOnCancel above, which is the only place it's completed, then we would never get to this map unless we already went through it

@tgodzik No, we always goes trough all task computation. The `cancel` in monix2 only performs lamdas passed in `doOnCancel`. In this case this removed `handleCancellation` is actually reachable and valid

tgodzik: Ach, right so that is why this is not really cancelation :thinking: 

I remember you had some small experiments with Monix, do you have it still somewhere?

dos65: 😄 what about cancellation, it was just small samples - I think I placed them in tests: `cancel1`, `cancel2` https://github.com/scalacenter/bloop/blob/fa0d88309eb88aea86376f6613d5923aaaced2dc/backend/src/test/scala/bloop/task/TaskSpec.scala#L24

tgodzik: If we do cancelation this way I wonder if we shouldn't just use Future :thinking: and bsp4j library, where we could handle things via completable futures + promises. Is there an actual benefit of using current monix cancelation in comparison to just Future?

I wonder if rewriting to cats will just make us use another library not really suited for out purposes.

dos65: From the perspective of getting rid of monix, I don't think that there will be major difference in choice between cats-io vs delayed version of future `() => Future`. The amount of work will be similar.

Right now, we are using an embedded cancelation mechanic in monix2, while it would be better to separate that thing.
I mean that instead of cancelling a whole bsp-request computation, what we need is to interrupt a compilation + maybe test/run tasks which would fit to any async data-type.
I mean:
```scala
// 
trait BloopTask[A] {
  def result: F[A]
  def cancel: F[Unit]
}

object Compiler {
  //  Return a task that return a specific bloop task (for example compile)
  //  then at CLI/BSP side if we want to cancel, we cancel only long-running operations that are designed to be interrupted 
  def compile(...): Task[BloopTask[Result]] = {
     ...
  }
}
```
However,  I have no idea how much time it will take.

PR #2097, opened 2023-06-30
improvement: Try and find latest supported semanticdb version

tgodzik:
Previously, we wouldn't produce semanticdb files for older version of Scala since we wouldn't find the correct artifact if we stopped publishing for a particular version.

Now, we try and find the latest version that still supported that Scala version.

This is similar to the change we did in Metals in regards to the presentation compiler interface, where we would fetch old mtags for unsupported scala version and this way always support a particular version.

Will do a similar thing for our Metals sbt plugin if it makes sense.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/caches/SemanticDBCache.scala
+++ b/frontend/src/main/scala/bloop/engine/caches/SemanticDBCache.scala
@@ -2,21 +2,30 @@ package bloop.engine.caches
 
 import java.nio.file.Path
 
+import scala.collection.JavaConverters._
+import scala.concurrent.Await
+import scala.concurrent.duration.FiniteDuration
 import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
 
 import bloop.DependencyResolution
 import bloop.SemanticDBCacheLock
+import bloop.engine.ExecutionContext
 import bloop.io.AbsolutePath
 import bloop.io.Paths
 import bloop.logging.Logger
+import bloop.task.Task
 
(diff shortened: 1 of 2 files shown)


frontend/src/main/scala/bloop/engine/caches/SemanticDBCache.scala
```diff
@@ -57,6 +65,44 @@ object SemanticDBCache {
...
+        coursierapi.Complete
+          .create()
+          .withScalaVersion(scalaVersion)
+          .withScalaBinaryVersion(scalaVersion.split('.').take(2).mkString("."))
+          .withInput(s"org.scalameta:semanticdb-scalac_$scalaVersion:")
+          .complete()
+          .getCompletions()
+          .asScala
+          .lastOption
+      }
```
kpodsiad, line 85: nitpick - `versionFuture` returns Task despite its name. Moreover, you're instantly converting it to future via `versionFuture.runAsync(ExecutionContext.ioScheduler)`, maybe you could go to the Future directly? There is no point in using Task in this case.

kpodsiad (changes requested):

frontend/src/main/scala/bloop/engine/caches/SemanticDBCache.scala
tgodzik: Sure, I just wanted to be consistent, but maybe it doesn't really make sense. Changed!

kpodsiad (approved):

PR #2099, opened 2023-07-04
chore: Update coursier.json and run release on Java 8 for the time being

tgodzik:


.github/workflows/ci.yml
```diff
@@ -241,7 +241,7 @@ jobs:
           fetch-depth: 0
       - uses: coursier/setup-action@v1
         with:
-          jvm: 'temurin:11.0.18'
+          jvm: 'temurin:8'
```
ckipp01, line 244: So is this the main culprit here? Because we released on 11?

ckipp01 (approved):

.github/workflows/ci.yml
tgodzik: I think so, I will test it on a snapshot version

PR #2101, opened 2023-07-05
chore: Revert JGit to a version compatible with JDK 8

tgodzik:
Alternative would be adding everywhere the -release option, but since we plan to drop the support later on, I think it's easier to do this.

The release job failed because of it, I will be updating the integrations or Bloop itself later on.


project/build.sbt
```diff
@@ -17,8 +17,8 @@ addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.11.0")
 
 updateOptions := updateOptions.value.withLatestSnapshots(false)
 libraryDependencies ++= List(
-  "org.eclipse.jgit" % "org.eclipse.jgit" % "6.6.0.202305301015-r",
-  "org.eclipse.jgit" % "org.eclipse.jgit.ssh.jsch" % "6.6.0.202305301015-r",
+  "org.eclipse.jgit" % "org.eclipse.jgit" % "5.13.2.202306221912-r",
+  "org.eclipse.jgit" % "org.eclipse.jgit.ssh.jsch" % "5.13.2.202306221912-r",
```
ckipp01, line 21: Can we add a comment here as to why we are leaving at this version? I know we'll try to fix all this later, but just in case it might be good to make sure we have this noted so we don't accidentally let steward update it again.

ckipp01 (changes requested):

project/build.sbt
tgodzik: The good thing here is that they currently have concurrent 5 and 6 versions, so this is not too old. I will add that!

ckipp01 (approved):
LGMT!

PR #2104, opened 2023-07-06
bugfix: Don't deduplicate all options when using Metals

tgodzik:
PReviously, we would use `.distinct` on all project scalac options which would remove values for different options if the happened to be the same. Now, we only deduplicate semanticdb options.

Fixes https://github.com/scalameta/metals/issues/5400

### DIFF
--- a/frontend/src/main/scala/bloop/data/Project.scala
+++ b/frontend/src/main/scala/bloop/data/Project.scala
@@ -454,10 +454,10 @@ object Project {
       // engine so that semanticdb files are replicated in those directories
       val hasSemanticDB = hasScalaSemanticDBEnabledInCompilerOptions(options)
       val pluginOption = if (hasSemanticDB) Nil else List(s"-Xplugin:$pluginPath")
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
Makes sense and this seems to be safer, although it makes me wish there was an easier way to actually detect duplicates. However I know it's hard since some will be `-key:value` and others `key, value` so detecting which shape they are isn't easy. Will bloop emit a bunch of warnings in situations where options are duplicated now though?

Either way, LGTM

tgodzik:
> Makes sense and this seems to be safer, although it makes me wish there was an easier way to actually detect duplicates. However I know it's hard since some will be `-key:value` and others `key, value` so detecting which shape they are isn't easy. Will bloop emit a bunch of warnings in situations where options are duplicated now though?
> 
> Either way, LGTM

We should not get any duplicates really, the only ones we could have are the ones from adding semanticdb, which are handled here.

tgodzik:
Anyway, one of the changes test was testing that, but the logic was reworked so the exact order changed.

PR #2105, opened 2023-07-06
improvement: Send debug logs to BSP client

tgodzik:
Previously, we would only print the logs when using Bloop from CLI. We still will need to add `-verbose` to compile params to get this behaviour


tgodzik:
Damn, I need to update the tests.

adpi2 (approved):
Don't you want to forward the traces too?

tgodzik:
> Don't you want to forward the traces too?

Aren't they only used for zipkin? I think they are not show on console when running with `-verbose`, or is it always forwarded?

tgodzik:
> > Don't you want to forward the traces too?
> 
> Aren't they only used for zipkin? I think they are not show on console when running with `-verbose`, or is it always forwarded?

Ok, nevermind, looks like we can include the, done!

PR #2107, opened 2023-07-07
bugfix: Don't try to discover test frameworks if none exist

tgodzik:
Previously, even if node_modules wouldn't exist we would run test discovery, which requires the node_modules and the scripts it contains. Now, we show an error in such a case. This seems to have hanged the node process causing no tests to be discovered.

I am not sure how to properly test it yet, but I need to dig into the JS part at some point since from earlier discussions with Sebastien the bridges might be problematic.

### DIFF
--- a/bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
+++ b/bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
@@ -147,25 +147,32 @@ object JsBridge {
       env: Map[String, String]
   ): (List[sbt.testing.Framework], () => Unit) = {
     implicit val debugFilter: DebugFilter = DebugFilter.Test
-    val nodeModules = baseDirectory.resolve("node_modules").toString
-    logger.debug("Node.js module path: " + nodeModules)
-    val fullEnv = Map("NODE_PATH" -> nodeModules) ++ env
-    val config =
-      NodeJSConfig().withExecutable(nodePath).withCwd(Some(baseDirectory)).withEnv(fullEnv)
-    val nodeEnv =
-      if (!jsConfig.jsdom.contains(true)) new NodeJSEnv(logger, config)
-      else new JsDomNodeJsEnv(logger, config)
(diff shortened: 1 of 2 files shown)


tgodzik:
Related to https://github.com/scalameta/metals/issues/4782

bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
```diff
@@ -147,25 +147,32 @@ object JsBridge {
...
-    val adapter = new TestAdapter(nodeEnv, Seq(input), testConfig)
-    val result = adapter.loadFrameworks(frameworkNames).flatMap(_.toList)
-    (result, () => adapter.close())
+      val testConfig = TestAdapter.Config().withLogger(new Logger(logger))
+      val adapter = new TestAdapter(nodeEnv, Seq(input), testConfig)
+      val result = adapter.loadFrameworks(frameworkNames).flatMap(_.toList)
+      (result, () => adapter.close())
+    } else {
+      logger.error(
+        s"Cannot discover test frameworks, missing node_modules in test project, expected them at $nodeModules"
```
ckipp01, line 173: Just trying to understand the full picture here. In the initial metals report about the tests lenses not showing up, that's because there was no `node_modules`? If that is the case, will this error float up and actually be shown to the user on the client side?

ckipp01: To be honest, I don't know enough about how the node stuff works to really answer that. Before the process never finished so we got nothing back. In this situation would it actually fix the original issue and still find the necessary frameworks?

tgodzik: No, since there is no node_modules, but since Metals doesn't run Scala JS tests currently, we wouldn't show them anyway. So this will just show the error. I don't fully understand when the `node_modules` should show up in that build, but we shouldn't break Metals on that for sure

ckipp01 (approved):
Thanks for the explanations. From me this is a LGTM

PR #2121, opened 2023-07-17
chore: Remove verbose option otherwise it's always verbose

tgodzik:
It seems because we add an option to log verbose option to BSP, now it's always doing it, since verbose is specified at the start. I changed that in order to be able to specify verbose on demand for compile.

It doesn't seem like it's needed currently, since the initial logs are printed to output stream set in the launcher.


kpodsiad (approved):

ckipp01 (approved):

adpi2 (approved):

launcher-core/src/main/scala/bloop/launcher/bsp/BspBridge.scala
```diff
@@ -78,7 +78,7 @@ final class BspBridge(
         bspServerStatus = Some {
           bspCmd.mkString(" ") -> {
             // Make it verbose so that bsp preparation logs are emitted
```
filipwiech, line 80: This comment seems no longer valid? :slightly_smiling_face:

tgodzik: It's not, thanks! I removed it

PR #2122, opened 2023-07-21
bugfix: Handle compiler position that is not a range

tgodzik:
I was experimenting with @lrytz with code actions for Scala 2 and manged it working this way. There will need to be some more changes, but this ones we can already merge.

bridge-scala-test-1689949673344


ckipp01 (approved):
Looks good, LGTM!

PR #2127, opened 2023-07-28
bugfix: Fix infinite recursion

tgodzik:
:scream_cat:


tgodzik:
It seems not have popped up in the normal tests, not sure why. Looks like maybe the method is invoked sometimes still.

PR #2128, opened 2023-07-28
specify scip javac plugin targetroot exactly

Arthurm1:
Currently Bloop uses `-targetroot:javac-classes-directory` to specify the `targetroot` dir for the `scip-java` plugin, which means  Bloop delegates to the plugin to work out the correct dir to place the semanticdb files in.

`scip-java` has to use the internal JDK API to perform the substitution which requires `--add-exports` and I see no way round that  - hence in this PR Bloop specifies the exact classes directory


tgodzik (approved):
Would love to specify that right when we add the option, but we don't have it unfortunately.

PR #2132, opened 2023-08-07
Add support for absolute paths in Bloopgun error handling

andrzejressel:
On Windows with scala-cli I got something like:
```
java.io.IOException: Cannot run program "C:\Program Files\Microsoft\jdk-17.0.7.7-hotspot\bin\java" (in directory "d:\MojeProgramy\bloop-core"): CreateProcess error=2, Nie mo?na odnale?? okre?lonego pliku
```

### DIFF
--- a/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
+++ b/bloopgun/src/main/scala/bloop/bloopgun/Bloopgun.scala
@@ -503,10 +503,15 @@ class BloopgunCli(
         (exitServerStatus, usedExtraJvmOpts)
       } catch {
         case e: IOException =>
+          val javaBinaryName = if (Environment.isWindows) {
+            "java.exe"
+          } else {
+            "java"
+          }
           val javaExec = sys.env
             .get("JAVA_HOME")
             .orElse(sys.props.get("java.home"))
-            .map(Paths.get(_).resolve("bin/java"))
+            .map(Paths.get(_).resolve(s"bin/${javaBinaryName}"))
             .filter(_.toFile().isFile())
             .map(_.toString())
           val javaErrorMessage = "Cannot run program \".*java\"".r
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

tgodzik:
Ach, I know remember why I didn't merge it. Could you run formatting? I totally forgot about it, sorry!

tgodzik:
Interestingly enough one of the tests now fails (though not on windows) and I don't think there is a way to make if pass if there is any java available :thinking: 

However, I think that is fine, though we would need to adjust the tests and see if the error on windows is different and maybe regex needs adjusting? @andrzejressel do you have time to look into that?

andrzejressel:
I would argue that test is wrong (and Windows setup). The exception thrown is:
```
java.io.IOException: Cannot run program "does-not-exist/bin/java" (in directory "/tmp/cwd-test1220557323768451443"): error=2, No such file or directorycherTest / Test / executeTests 11s
        at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
        at bloop.bloopgun.BloopgunCli.sysproc$1(Bloopgun.scala:487)
        at bloop.bloopgun.BloopgunCli.sysprocWithJava$1(Bloopgun.scala:502)
        at bloop.bloopgun.BloopgunCli.startServer(Bloopgun.scala:549)
        at bloop.bloopgun.BloopgunCli$FireInBackground$.$anonfun$fire$1(Bloopgun.scala:314)
        at bloop.bloopgun.core.Shell$$anon$2.run(Shell.scala:154)
Caused by: java.io.IOException: error=2, No such file or directory
        at java.lang.UNIXProcess.forkAndExec(Native Method)
        at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
        at java.lang.ProcessImpl.start(ProcessImpl.java:134)
        at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
        ... 5 more
```

That means that bloop is supposed to run - after all we do all the JAVA_HOME finding stuff. It was working fine incorrectly earlier because of matching.

The issue with Windows runner is that it's not setting JAVA_HOME:
image

andrzejressel:
Also that `.map(Paths.get(_).resolve("bin/java"))` - unix only stuff

andrzejressel:
Should be fine now

tgodzik (approved):
LGTM!

PR #2153, opened 2023-09-13
chore: Change frameworks not supported message to warn

tgodzik:
Previously, we would show an error whenever a native project was loaded, since we can't currently run tests for native. However, this will pop up currently in Metals as a server error, while it's more of a warning. There is nothing the user can do about it. Now, we report this as a warning.


ckipp01 (approved):
LGTM!

PR #2171, opened 2023-10-06
improvement: trim spaces when filtering the only tests that should be run

shasts:
Often, when copying the failed test path from Jenkins CI or Gradle Enterprise webpage, the path copied might contain leading or trailing space(s). Pasting the failed test path into the `-o` switch requires carefully removing the space from both ends.  

Example:
` bloop test project-it -o ' a.b.c.Test' --config-dir=../.bloop`
This command would not run the test `a.b.c.Test`

This PR fixes the annoyance by trimming the path before matching the pattern.


tgodzik (approved):
Thanks for the contribution! LGTM

PR #2172, opened 2023-10-10
feature: Automatically add -release option when needed

tgodzik:
Previously, if we compiled with a Bloop running an newer JDK we would not see any errors even when using methods belonging to a newer JDK. Now, we automatically add -release flag, which makes sure that we will fail any compilation that should not pass with an older JDK (if specified in configuration)

This is needed for incorporating the forks.

### DIFF
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -280,72 +280,12 @@ object Compiler {
       )
     }
 
-    var isFatalWarningsEnabled: Boolean = false
+    val isFatalWarningsEnabled: Boolean =
+      compileInputs.scalacOptions.exists(_ == "-Xfatal-warnings")
     def getInputs(compilers: Compilers): Inputs = {
-      val options = getCompilationOptions(compileInputs)
+      val options = getCompilationOptions(compileInputs, logger, newClassesDir)
       val setup = getSetup(compileInputs)
       Inputs.of(compilers, options, setup, compileInputs.previousResult)
-    }
-
-    def getCompilationOptions(inputs: CompileInputs): CompileOptions = {
-      // Sources are all files
-      val sources = inputs.sources.map(path => converter.toVirtualFile(path.underlying))
-      val classpath = inputs.classpath.map(path => converter.toVirtualFile(path.underlying))
-      def existsReleaseSetting = inputs.scalacOptions.exists(opt =>
-        opt.startsWith("-release") ||
-          opt.startsWith("--release") ||
-          opt.startsWith("-java-output-version")
-      )
-      def sameHome = inputs.javacBin match {
-        case Some(bin) => bin.getParent.getParent == JavaRuntime.home
-        case None => false
-      }
-
-      val scalacOptions = inputs.javacBin.flatMap(binary =>
-        // <JAVA_HOME>/bin/java
-        JavaRuntime.getJavaVersionFromJavaHome(binary.getParent.getParent)
-      ) match {
-        case None => inputs.scalacOptions
-        case Some(_) if existsReleaseSetting || sameHome => inputs.scalacOptions
-        case Some(version) =>
-          try {
-            val numVer = if (version.startsWith("1.8")) 8 else version.takeWhile(_.isDigit).toInt
-            val bloopNumVer = JavaRuntime.version.takeWhile(_.isDigit).toInt
-            if (bloopNumVer > numVer) {
-              inputs.scalacOptions ++ List("-release", numVer.toString())
-            } else {
-              logger.warn(
-                s"Bloop is run with ${JavaRuntime.version} but your code requires $version to compile, " +
-                  "this might cause some compilation issues when using JDK API unsupported by the Bloop's current JVM version"
-              )
-              inputs.scalacOptions
-            }
-          } catch {
-            case NonFatal(_) =>
-              inputs.scalacOptions
-          }
-      }
-
-      val optionsWithoutFatalWarnings = scalacOptions.flatMap { option =>
-        if (option != "-Xfatal-warnings") List(option)
-        else {
-          if (!isFatalWarningsEnabled) isFatalWarningsEnabled = true
-          Nil
-        }
-      }
-
-      // Enable fatal warnings in the reporter if they are enabled in the build
-      if (isFatalWarningsEnabled)
-        inputs.reporter.enableFatalWarnings()
-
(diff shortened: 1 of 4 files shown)


backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -289,7 +291,42 @@ object Compiler {
...
+        case Some(_) if existsReleaseSetting || sameHome => inputs.scalacOptions
+        case Some(version) =>
+          try {
+            val numVer = if (version.startsWith("1.8")) 8 else version.takeWhile(_.isDigit).toInt
+            val bloopNumVer = JavaRuntime.version.takeWhile(_.isDigit).toInt
+            if (bloopNumVer > numVer) {
+              inputs.scalacOptions ++ List("-release", numVer.toString())
+            } else {
+              logger.warn(
+                s"Bloop is run with ${JavaRuntime.version} but your code requires $version to compile, " +
```
ckipp01, line 318:
```suggestion
                s"Bloop is running with ${JavaRuntime.version} but your code requires $version to compile, " +
```
Should we maybe change it to `running` to hint at the fact that it needs to stop and be restarted with that version if necessary?

ckipp01 (approved):

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -289,7 +291,42 @@ object Compiler {
       // Sources are all files
       val sources = inputs.sources.map(path => converter.toVirtualFile(path.underlying))
       val classpath = inputs.classpath.map(path => converter.toVirtualFile(path.underlying))
-      val optionsWithoutFatalWarnings = inputs.scalacOptions.flatMap { option =>
+      def existsReleaseSetting = inputs.scalacOptions.exists(opt =>
+        opt.startsWith("-release") ||
+          opt.startsWith("--release") ||
+          opt.startsWith("-java-output-version")
+      )
```
kpodsiad, line 298: nit: I'd extract `getCompilationOptions` to standalone function, not nested one. Compile is already too long and complicated without your changes

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -289,7 +291,42 @@ object Compiler {
...
+      def sameHome = inputs.javacBin match {
+        case Some(bin) => bin.getParent.getParent == JavaRuntime.home
+        case None => false
+      }
+
+      val scalacOptions = inputs.javacBin.flatMap(binary =>
+        // <JAVA_HOME>/bin/java
+        JavaRuntime.getJavaVersionFromJavaHome(binary.getParent.getParent)
```
kpodsiad, line 306: why is checking java home relevant? I thought that what's important is checking current java version of the bloop process?

tgodzik: This is the java version of Java we are compiling with, not the Java version Bloop is using. The latter is below in line 313  `  val bloopNumVer = JavaRuntime.version.takeWhile(_.isDigit).toInt`

backend/src/main/scala/bloop/Compiler.scala
tgodzik: Done!

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -627,6 +604,66 @@ object Compiler {
...
+            inputs.scalacOptions
+          }
+        } catch {
+          case NonFatal(_) =>
+            inputs.scalacOptions
+        }
+    }
+
+    val optionsWithoutFatalWarnings = scalacOptions.filter(_ != "-Xfatal-warnings")
+    val isFatalWarningsEnabled = scalacOptions.length != optionsWithoutFatalWarnings.length
```
kpodsiad, line 651: nit: should it be `areFatalWarningsEnabled`?

backend/src/main/scala/bloop/Compiler.scala
kpodsiad: Extracting that method helped me realize one thing, the main complexity of `getCompilationOptions` is connected with adjusting scalac options with release flag if needed. That is why I think it's worth to extract another function, which I hope makes getCompilationOptions even easier and allows to explain why there is need for adding release flag in the first place. I put some scaladoc based on your replies, but it definitely needs to be improved. I'm leaving this as a suggestion, it's up to you to decide whether extract another function or not.

<details><summary>adjustScalacOptions</summary>

```scala
  /**
   * Bloop runs Scala compilation in the same process as the main server,
   * so the compilation process will use the same JDK that Bloop is using.
   * That's why we must ensure that produce class files will be compliant with expected JDK version.
   */
  private def adjustScalacOptions(
      scalacOptions: Array[String],
      javacBin: Option[AbsolutePath],
      logger: Logger
  ): Array[String] = {
    def existsReleaseSetting = scalacOptions.exists(opt =>
      opt.startsWith("-release") ||
        opt.startsWith("--release") ||
        opt.startsWith("-java-output-version")
    )
    def sameHome = javacBin match {
      case Some(bin) => bin.getParent.getParent == JavaRuntime.home
      case None => false
    }

    javacBin.flatMap(binary =>
      // <JAVA_HOME>/bin/java
      JavaRuntime.getJavaVersionFromJavaHome(binary.getParent.getParent)
    ) match {
      case None => scalacOptions
      case Some(_) if existsReleaseSetting || sameHome => scalacOptions
      case Some(version) =>
        try {
          val numVer = if (version.startsWith("1.8")) 8 else version.takeWhile(_.isDigit).toInt
          val bloopNumVer = JavaRuntime.version.takeWhile(_.isDigit).toInt
          if (bloopNumVer > numVer) {
            scalacOptions ++ List("-release", numVer.toString())
          } else {
            logger.warn(
              s"Bloop is runing with ${JavaRuntime.version} but your code requires $version to compile, " +
                "this might cause some compilation issues when using JDK API unsupported by the Bloop's current JVM version"
            )
            scalacOptions
          }
        } catch {
          case NonFatal(_) =>
            scalacOptions
        }
    }
  }
  ```

</details>

backend/src/main/scala/bloop/Compiler.scala
tgodzik: Changed.

backend/src/main/scala/bloop/Compiler.scala
tgodzik: Sure, done!

kpodsiad (approved):
Looks good!

PR #2185, opened 2023-11-07
Update sbt `bloopGenerate` docs to use `Value(None)`

kjh618:
After #1973, you should use `Value(None)` instead of `None` but the docs haven't been updated.


tgodzik (approved):
Thanks!

PR #2192, opened 2023-11-16
Log debuggee command

### LINKED ISSUE #1456: Remove stripPrefix("-J") workaround for Metals
In order not to leak internal Bloop implementation we decided to no longer user `-J` when running via DAP. This is no longer needed since we have a separate field for jvm options. For a smooth transition however, we decided to add a workaround that will strip `-J` prefix when it's passed from Metals.

We should be good to remove it once Metals is release with 1.4.7 version of Bloop.

PR that introduces this: https://github.com/scalacenter/bloop/pull/1455/files

adpi2:
- Stop stripping `-J` prefix to close #1456
- Log all forked commands with info level to address https://github.com/scalameta/metals-feature-requests/issues/352


frontend/src/main/scala/bloop/exec/Forker.scala
```diff
@@ -200,6 +200,11 @@ object Forker {
       val envMap = builder.environment()
       envMap.putAll(env.asJava)
       builder.redirectErrorStream(false)
+      logger.info("Forking process:")
```
tgodzik, line 203: Would we want it under debug rather?

adpi2: How can we get debug logs with Bloop?

tgodzik: we would need to add a flag to starting params. Maybe it's fine to do info.

tgodzik (approved):

Issue #1606, opened 2021-11-12
Dependencies from Runtime configuration are not visible


eikek:
Hello! I have an issue with `bloop run …` in one of my projects where it seems that dependencies that are declrared in sbts `Runtime` config are not visible. I prepared a small example project to reproduce it. Thank you very much!

Using this source:
```scala
package bloop.demo

object Main {
  def main(args: Array[String]): Unit = {
    println(">>> hello world! <<<")
    val cn = Class.forName("ch.qos.logback.classic.Logger")
    println(s"$cn")
  }
}
```

Where the logback dependency is declared for the `Runtime` configuration.

**With SBT**:

```
$ sbt run
[info] welcome to sbt 1.5.5 (Oracle Corporation Java 11.0.12)
[info] loading settings for project bloop-demo-build from plugins.sbt ...
[info] loading project definition from /home/eike/workspace/bloop-demo/project
[info] loading settings for project bloop-demo from build.sbt ...
[info] set current project to bloop-demo (in build file:/home/eike/workspace/bloop-demo/)
[info] compiling 1 Scala source to /home/eike/workspace/bloop-demo/target/scala-2.12/classes ...
[info] running bloop.demo.Main
>>> hello world! <<<
class ch.qos.logback.classic.Logger
[success] Total time: 3 s, completed 12 Nov 2021, 17:44:30
```

**With bloop**:

After running `sbt bloopInstall`

```
$ bloop run bloop-demo
Compiling bloop-demo (1 Scala source)
Compiled bloop-demo (292ms)
>>> hello world! <<<
[E] Exception in thread "main" java.lang.ClassNotFoundException: ch.qos.logback.classic.Logger
[E]     at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
[E]     at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
[E]     at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
[E]     at java.base/java.lang.Class.forName0(Native Method)
[E]     at java.base/java.lang.Class.forName(Class.java:315)
[E]     at bloop.demo.Main$.main(Main.scala:6)
[E]     at bloop.demo.Main.main(Main.scala)
```

[bloop-demo.tar.gz](https://github.com/scalacenter/bloop/files/7529480/bloop-demo.tar.gz)

tgodzik:
[...]

PR #2195, opened 2023-11-16
bugfix: Export runtime jars correctly from sbt

tgodzik:
Previously, we would not save runtime jars which could lead to users not being able to run their programs.

Now, we export runtime classpath properly from sbt.

Related to https://github.com/scalacenter/bloop/issues/1606 though I want to check out if it will also work via DAP

### DIFF
--- a/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
+++ b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
@@ -35,6 +35,7 @@ import sbt.Optional
 import sbt.ProjectRef
 import sbt.Provided
 import sbt.ResolvedProject
+import sbt.Runtime
 import sbt.TaskKey
 import sbt.Test
 import sbt.ThisBuild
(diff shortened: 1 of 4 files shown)


tgodzik:
Looks like tests are discovered from runtime classpath and the changes break things. Will take a look tomorrow.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -806,8 +788,19 @@ object BloopDefaults {
         val isForkedExecution = (Keys.fork in configuration in forkScopedTask).value
         val workingDir = if (isForkedExecution) Keys.baseDirectory.value else rootBaseDirectory
         val extraJavaOptions = List(s"-Duser.dir=${workingDir.getAbsolutePath}")
+        lazy val runtimeClasspath = Keys.fullClasspath.in(Runtime).value.map(_.data.toPath()).toList
+        lazy val javaRuntimeHome = (Keys.javaHome.in(Runtime) ).value.map(_.toPath())
+        lazy val javaRuntimeOptions = (Keys.javaOptions.in(Runtime)).value
```
adpi2, line 793: The `in` syntax is deprecated since sbt `1.5.0`. Can we use the slash syntax instead.

```suggestion
        lazy val runtimeClasspath = (Runtime /  Keys.fullClasspath).value.map(_.data.toPath()).toList
        lazy val javaRuntimeHome = (Runtime / Keys.javaHome).value.map(_.toPath())
        lazy val javaRuntimeOptions = (Runtime / Keys.javaOptions).value
```

integrations/sbt-bloop/src/sbt-test/sbt-bloop/runtime-dependency/build.sbt
```diff
@@ -0,0 +1,60 @@
+import bloop.integrations.sbt.BloopDefaults
+
+name := "runtime-dependency"
+
+libraryDependencies +=
+  "ch.qos.logback" % "logback-classic" % "1.2.7" % Runtime
+
+val bloopConfigFile = settingKey[File]("Config file to test")
+bloopConfigFile in ThisBuild := {
```
adpi2, line 9: Same in this file
```suggestion
ThisBuild / bloopConfigFile := {
```

adpi2 (approved):
LGTM. Apart from the deprecated `in` syntax.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
tgodzik: Thanks! Fixed here and in any code I added

PR #2197, opened 2023-11-19
Implement BSP buildTargetDependencyModules

Arthurm1:
DependencyModules is an improvement on DependencySources


tgodzik (approved):
LGTM! I think we had some issues in the past that could have been solved by this, but I don't really remember (bsp part might have been an effort to help us)

PR #2199, opened 2023-11-22
bugfix: Don't use fullclasspath in sbt plugin

tgodzik:
Previously, I used fullclasspath for getting the runtime classpath, but I forgot this causes sbt to always compile on bloopInstaall.

Now, I changed to use the same mechanism we use for compile classpath and added a test to make sure this doesn't show up again.

### DIFF
--- a/integrations/sbt-bloop/src/sbt-test/sbt-bloop/non-compiling/test
+++ b/integrations/sbt-bloop/src/sbt-test/sbt-bloop/non-compiling/test
@@ -1,0 +1,3 @@
+> bloopGenerate
+> test:bloopGenerate
+> checkBloopFiles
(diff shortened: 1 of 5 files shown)


integrations/sbt-bloop/src/sbt-test/sbt-bloop/non-compiling/test
```diff
@@ -0,0 +1,3 @@
+> bloopGenerate
+> test:bloopGenerate
+> checkBloopFiles
```
adpi2, line 3: How does this test check that the compilation is not triggered?

tgodzik: What we really want to check is that the files are properly generated despite the error. Previously, when we used fullclasspath the `bloopInstall` task itself would fail and nothing would get generated

adpi2: Oh I see there is an error in the `Obj.scala` file

adpi2 (approved):

PR #2200, opened 2023-11-23
chore: Print better errors when bloopInstall failed

tgodzik:
Previously, we would get soemthing like:

```
Cause: Incomplete(node=Some(Task(_)), tpe=Error, msg=None, causes=List(Incomplete(node=Some(Task(_)), tpe=Error, msg=None, causes=List(Incomplete(node=Some(Task(_)), tpe=Error, msg=None, causes=List(Incomplete(node=Some(Task((taskDefinitionKey: ScopedKey(Scope(Select(ProjectRef(file:/tmp/sbt_14ffb38d/no-compile-incremental/,no-compile-incremental)), Select(ConfigKey(runtime)), Zero, Zero),fullClasspath)))), tpe=Error, msg=None, causes=List(Incomplete(node=Some(Task(_)), tpe=Error, msg=None, causes=List(Incomplete(node=Some(Task((taskDefinitionKey: ScopedKey(Scope(Select(ProjectRef(file:/tmp/sbt_14ffb38d/no-compile-incremental/,no-compile-incremental)), Select(ConfigKey(runtime)), Zero, Zero),dependencyClasspath)))), tpe=Error, msg=None, causes=List(Incomplete(node=Some(Task(_)), tpe=Error
```

which doesn't contain any useful information.

Now, I change to print errors if we have any direct causes of the failure, which would come from bloop task itself.

### DIFF
--- a/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
+++ b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
@@ -1147,7 +1147,22 @@ object BloopDefaults {
       )
       .map(fail =>
         if (fail.nonEmpty) {
-          logger.error(s"Couldn't run bloopGenerate. Cause: ${fail.mkString("\n")}")
+          val messages = fail.flatMap { incomplete =>
+            incomplete.directCause.zip(incomplete.message)
+          }
+
+          if (messages.nonEmpty) {
+            logger.error(s"Couldn't run bloopGenerate. The issue might be caused by:")
+            messages.foreach {
+              case (t, message) =>
+                logger.error(message)
+                logger.trace(t)
+            }
+          } else {
+            logger.error(
+              s"Couldn't run bloopGenerate. The issue could be caused by failing to compile the project."
+            )
+          }
           throw fail.head
         }
       )


integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -1147,7 +1147,22 @@ object BloopDefaults {
...
+          if (messages.nonEmpty) {
+            logger.error(s"Couldn't run bloopGenerate. The issue might be caused by:")
+            messages.foreach {
+              case (t, message) =>
+                logger.error(message)
+                logger.trace(t)
+            }
+          } else {
+            logger.error(
+              s"Couldn't run bloopGenerate. The issue could be caused by failing to compile the project."
```
adpi2, line 1163: Is it not supposed to not even try to compile the project?

tgodzik: it should compile if there is source generation, so this might happen.

adpi2 (approved):

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
adpi2: I would argue that if there is source generation it should only generate the sources, not compile.

But in general we don't know what's in the build and a compilation could happen.

PR #2204, opened 2023-12-06
chore: Update scala-debug-adapter to newest

tgodzik:


crt-31:
@tgodzic, or anyone: I'm curious, why wasn't the new scala-debug-adapter detected and updated by the scala-steward? Previous versions have been but this one wasn't.

tgodzik:
I was faster than the steward :D

crt-31:
Sure, but isn't the stewed supposed to run every day?

tgodzik:
> Sure, but isn't the stewed supposed to run every day?

Actually, you are right. https://github.com/scalacenter/steward seems to not be running anymore

PR #2207, opened 2023-12-07
chore: Use the default download location for download action

tgodzik:
Looks like maybe target is being cleaned which is why we can publish properly later on. Tested it locally and seems to work ok.


tgodzik:
We'll see if it works on the next release.

PR #2221, opened 2023-12-27
refactor: Remove forked parts of scala js envs

tgodzik:
As far as I can see this should work properly and it hits us even more to have duplicated part of code that is not maintained.


bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
```diff
@@ -151,11 +144,12 @@ object JsBridge {
     if (nodeModules.toFile().exists()) {
       logger.debug("Node.js module path: " + nodeModules.toString())
       val fullEnv = Map("NODE_PATH" -> nodeModules.toString()) ++ env
-      val config =
-        NodeJSConfig().withExecutable(nodePath).withCwd(Some(baseDirectory)).withEnv(fullEnv)
       val nodeEnv =
-        if (!jsConfig.jsdom.contains(true)) new NodeJSEnv(logger, config)
-        else new JsDomNodeJsEnv(logger, config)
+        if (!jsConfig.jsdom.contains(true))
```
ckipp01, line 148: I know you didn't change this logic, but is this actually needed seeing that it will always be `false`?

ckipp01 (commented):
I'm not really familiar with a lot of what this is doing, but giving this a look over looks alright. I just had one question on a piece of logic that is in there and doesn't really seem useful.

bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
tgodzik: Is it? If it's empty it will be false and also if it contains false the flag will be true. Or am I not seeing it?
```scala
!Some(false).contains(true) == true
!Some(true).contains(true) == false
!None.contains(true) == true
```

ckipp01: Ahhh, no you're right, sorry I read this too quickly and didn't understand it.... ignore me.

ckipp01 (approved):
LGTM!

PR #2224, opened 2023-12-28
refactor: Backport changes from bloop-core

tgodzik:


build.sbt
```diff
@@ -308,22 +297,8 @@ lazy val launcher213 = project
     target := (file("launcher-core") / "target" / "launcher-2.13").getAbsoluteFile
   )
 
-lazy val bloop4j = project
```
ckipp01, line 311: Ah, I'm sure this will all work out fine, but we should make sure we remove this and the actual files here at the same time. So maybe include this in the other PR where you remove the code and then just rebase this.

ckipp01 (approved):

build.sbt
tgodzik: I removed it in the other PR

PR #2231, opened 2024-01-02
refactor: Remove bloopgun and luancher 2.12 modules

tgodzik:
I don't think this is needed for any client. Metals uses 2.13 and Intellij is also on 2.13.x


tgodzik:
Ach, nvm. We still need this for tests.

PR #2232, opened 2024-01-02
bugfix: Print copying error together with stacktrace

tgodzik:
It seems we don't get that properly forwarded to Bloop otherwise.

### DIFF
--- a/backend/src/main/scala/bloop/io/ParallelOps.scala
+++ b/backend/src/main/scala/bloop/io/ParallelOps.scala
@@ -173,7 +173,7 @@ object ParallelOps {
           ()
         } catch {
           case NonFatal(t) =>
-            logger.report(
+            logger.error(
               s"Unexpected error when copying $originFile to $targetFile, you might need to restart the build server.",
               t
             )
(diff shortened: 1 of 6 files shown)


backend/src/main/scala/bloop/io/ParallelOps.scala
```diff
@@ -173,9 +174,9 @@ object ParallelOps {
           ()
         } catch {
           case NonFatal(t) =>
-            logger.report(
-              s"Unexpected error when copying $originFile to $targetFile, you might need to restart the build server.",
-              t
+            logger.error(
```
adpi2, line 177: What's the difference between `logger.report` and `logger.error`?

Would it be better to add a `def error(msg: String, throwable: Throwable)` method in `Logger` instead?

tgodzik: Good idea! I removed reporrt and just changed it to error, this also prompted a bit of refactoring.

adpi2 (approved):
LGTM

PR #2262, opened 2024-02-01
bugfix: revert logback to one still supporting JDK 8

tgodzik:


ckipp01 (approved):

ckipp01:
Do we have an idea of when we'll drop Java 8 support?

PR #2269, opened 2024-02-12
Fix classpath options

### LINKED ISSUE #2268: Bloop can't compile Dotty
It's a known issue: using Bloop as the build server with Metals when working on the Dotty codebase does not work. For this reason, the official "Getting Started" instructions advise to use "SBT" as the build server instead: https://dotty.epfl.ch/docs/contributing/setting-up-your-ide.html#metals-1.

However, using SBT as the build server is suboptimal because 1. it's an additional step to explain to newcomers, and 2. commands issues by Metals are mixed and matched with user-issued commands, making it troublesome to work with.

Reproduce with:

```bash
$ git clone git@github.com:lampepfl/dotty.git dotty-fresh
$ cd dotty-fresh
$ code .
```

After `bloopInstall`, `Indexing…` and re-compilation terminate, Bloop reports a compilation error for `scala3-compiler` and autocompletion doesn't work (try for example with `compiler/src/dotty/tools/dotc/typer/Typer.scala`).

Note some other features such as "Go to definition" do work as expected.

Metals Doctor shows `scala3-compiler` compilation failed:

image

The compilation error reported by Bloop is in `QuotesImpl.scala`:

image

Same error from the CLI:

```bash
$ bloop compile scala3-compiler
Compiling scala3-compiler (567 Scala sources and 8 Java sources)
[E] [E1] compiler/src/scala/quoted/runtime/impl/QuotesImpl.scala:285:31
[E]      Not found: type ValOrDefDefMethods
[E]      L285:     given ValOrDefDefMethods: ValOrDefDefMethods with
[E]                                          ^^^^^^^^^^^^^^^^^^
[E] compiler/src/scala/quoted/runtime/impl/QuotesImpl.scala: L285 [E1]
Compiled scala3-compiler (26779ms)
[E] Failed to compile 'scala3-compiler'
```

adpi2:
Fix #2268 

We should cache the `ScalaInstance` to reuse classloaders as much as possible but we should take into account the `ClasspathOptions` of a project when we create the `ScalaCompiler`.

Tested on Dotty: compilation :heavy_check_mark: auto-completion :heavy_check_mark:

### DIFF
--- a/backend/src/main/scala/bloop/CompilerCache.scala
+++ b/backend/src/main/scala/bloop/CompilerCache.scala
@@ -38,9 +38,9 @@ import xsbti.compile.ClassFileManager
 import xsbti.compile.Compilers
 import xsbti.compile.JavaCompiler
 import xsbti.compile.Output
-import xsbti.compile.ScalaCompiler
 import xsbti.{Logger => XLogger}
 import xsbti.{Reporter => XReporter}
+import xsbti.compile.ClasspathOptions
 
 object CompilerCache {
   final case class JavacKey(javacBin: Option[AbsolutePath], allowLocal: Boolean)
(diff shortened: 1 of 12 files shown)


backend/src/main/scala/bloop/CompilerCache.scala
```diff
@@ -50,19 +50,21 @@ final class CompilerCache(
     logger: Logger
 ) {
 
-  private val scalaCompilerCache = new ConcurrentHashMap[ScalaInstance, ScalaCompiler]()
-
+  private val scalaInstanceCache = new ConcurrentHashMap[ScalaInstance, ScalaInstance]()
```
adpi2, line 53: Cache the `ScalaInstance` instead of the `ScalaCompiler` because two projects can have the same `ScalaInstance` but different `ClasspathOptions`.

backend/src/main/scala/bloop/CompilerCache.scala
```diff
@@ -50,19 +50,21 @@ final class CompilerCache(
...
   private val javaCompilerCache = new ConcurrentHashMap[JavacKey, JavaCompiler]()
 
   def get(
       scalaInstance: ScalaInstance,
+      classpathOptions: ClasspathOptions,
       javacBin: Option[AbsolutePath],
       javacOptions: List[String]
   ): Compilers = {
-    val scalaCompiler = scalaCompilerCache.computeIfAbsent(
+    val scalaInstanceFromCache = scalaInstanceCache.computeIfAbsent(
```
rochala, line 62: We don't need this cache any more, we can pass Scala instance directly

rochala (approved):

backend/src/main/scala/bloop/CompilerCache.scala
adpi2: No we really need it. It duplicates `ScalaInstance` that are equal but not reference equal. The goal is to avoid using the class loaders from new `ScalaInstance` that are equal to old ones.

PR #2274, opened 2024-02-16
refactor: don't send task start/end for no-op compilations

ckipp01:
This might be a bit opinionated and I'm not 100% sure on the reason this was done in the first place, but this changes the current behavior around reporting start and end compilations for no-ops. This still retains the compile report if it was a no-op and also still checks the diagnostics and reports them, but skips the task start/end.

Originally this was added in a commit without any context... but there was a comment that said:

> // When no-op, we keep reporting the start and the end of compilation for consistency

However, I'm not really sure that consistency matters here. In reality this ends up creating a bunch of noise on the client side, especially when these tasks turn into LSP progress notifications that aren't useful for the user to see. You can see more context about this change in the issue reported [here](https://github.com/scalameta/metals/issues/6099) and also the discussion found
[here](https://github.com/build-server-protocol/build-server-protocol/discussions/654). It also seems that in some projects like scala-cli these are even [ignored](https://github.com/VirtusLab/scala-cli/blob/6b7a10007e4eefde717079255e0df38c027f788b/modules/build/src/main/scala/scala/build/ConsoleBloopBuildClient.scala#L109).

### DIFF
--- a/frontend/src/test/scala/bloop/bsp/BspCompileSpec.scala
+++ b/frontend/src/test/scala/bloop/bsp/BspCompileSpec.scala
@@ -83,13 +83,7 @@ class BspCompileSpec(
         assertSameExternalClassesDirs(compiledState, secondCompiledState, projects)
         assertNoDiff(
           secondCompiledState.lastDiagnostics(`A`),
-          """#2: task start 2
-            |  -> Msg: Start no-op compilation for a
-            |  -> Data kind: compile-task
-            |#2: task finish 2
-            |  -> errors 0, warnings 0
-            |  -> Msg: Compiled 'a'
-            |  -> Data kind: compile-report""".stripMargin
+          "" // no-op
         )
       }
     }
@@ -576,17 +570,10 @@ class BspCompileSpec(
         assertSameExternalClassesDirs(thirdCompiledState, compiledState, projects)
         assertNoDiff(
           thirdCompiledState.lastDiagnostics(`A`),
-          """#3: task start 3
-            |  -> Msg: Start no-op compilation for a
-            |  -> Data kind: compile-task
-            |#3: a/src/A.scala
+          """#3: a/src/A.scala
             |  -> List()
             |  -> reset = true
-            |#3: task finish 3
-            |  -> errors 0, warnings 0
-            |  -> Msg: Compiled 'a'
-            |  -> Data kind: compile-report
-            """.stripMargin
+            """.stripMargin // no-op so it only gives the compile-report
         )
 
         writeFile(`A`.srcFor("/A.scala"), Sources.`A3.scala`)
@@ -603,7 +590,7 @@ class BspCompileSpec(
         assertSameExternalClassesDirs(fourthCompiledState, compiledState, projects)
         assertNoDiff(
           fourthCompiledState.lastDiagnostics(`A`),
-          """#4: task start 4
+          """#4: task start 3
             |  -> Msg: Compiling a (1 Scala source)
             |  -> Data kind: compile-task
             |#4: a/src/A.scala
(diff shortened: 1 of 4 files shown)


frontend/src/test/scala/bloop/bsp/BspCompileSpec.scala
```diff
@@ -576,17 +570,10 @@ class BspCompileSpec(
...
-            |#3: a/src/A.scala
+          """#3: a/src/A.scala
             |  -> List()
             |  -> reset = true
-            |#3: task finish 3
-            |  -> errors 0, warnings 0
-            |  -> Msg: Compiled 'a'
-            |  -> Data kind: compile-report
-            """.stripMargin
+            """.stripMargin // no-op so it only gives the compile-report
```
tgodzik, line 576: Actually, compile report is in the task finish, so we only get the compile response, no? But that should be fine as nothing really changes

frontend/src/test/scala/bloop/bsp/BspCompileSpec.scala
```diff
@@ -603,7 +590,7 @@ class BspCompileSpec(
         assertSameExternalClassesDirs(fourthCompiledState, compiledState, projects)
         assertNoDiff(
           fourthCompiledState.lastDiagnostics(`A`),
-          """#4: task start 4
+          """#4: task start 3
```
tgodzik, line 593: Any idea why these changed if there was no noop compilation beforehand?

tgodzik (approved):
Looks good, just two questions, but nothing blocking. Thanks for looking into this!

frontend/src/test/scala/bloop/bsp/BspCompileSpec.scala
ckipp01: Ahh, yea sorry that's my mistake. I was switching up the names of the compile response and compile report. You're correct, it's only the compile _response_ that is now returned.

frontend/src/test/scala/bloop/bsp/BspCompileSpec.scala
ckipp01: Yea exactly, since there was a no-op in there, there is no one less task which lowers the count.

adpi2:
> Ahh, yea sorry that's my mistake. I was switching up the names of the compile response and compile report. You're correct, it's only the compile response that is now returned.

So the compile response is sent but not the compile report. Yet there is no `noOp` field in the compile response, so how does the client know it is a no-op compilation ? Does it even need to know ? All of this should be clearly documented in BSP if we want some sort of consistency across all build servers.

Also it seems it would be harder to implement the same behavior in sbt, because we don't know beforehand if a compilation is going to be a no-op.

ckipp01:
> So the compile response is sent but not the compile report. Yet there is no noOp field in the compile response, so how does the client know it is a no-op compilation? Does it even need to know? 

That's a good question. I guess it'd be ideal to still know in the actual response in case clients handle them in specific ways, but then again it's the task notifications that end up being noisy, not the compile response, so maybe it doesn't matter that much.

> All of this should be clearly documented in BSP if we want some sort of consistency across all build servers.

Agreed.

adpi2 (approved):
I am okay to merge this but I also recommend to add the `noOp` field in the [CompileResult](https://github.com/build-server-protocol/build-server-protocol/blob/eac6038263510dad9f299e1bce3a02b6a272763b/spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy#L1197-L1208)

ckipp01:
> I am okay to merge this but I also recommend to add the `noOp` field in the [CompileResult](https://github.com/build-server-protocol/build-server-protocol/blob/eac6038263510dad9f299e1bce3a02b6a272763b/spec/src/main/resources/META-INF/smithy/bsp/bsp.smithy#L1197-L1208)

Sounds good. When I get a bit of time I'll outline all of this in the discussion I started in the BSP repo and re-phrase my original question and change the discussion to just aligning on _how_ no-ops are meant to be handled.

PR #2291, opened 2024-03-07
fix: correctly parse java version

kasiaMarek:
Changing how java version is parsed to get rid of an unwanted warning:
` WARN  Bloop is running with 1.8.0_382 but your code requires 1.8.0_382 to compile, this might cause some compilation issues when using JDK API unsupported by the Bloop's current JVM version`


tgodzik (approved):
LGTM! Thanks!

PR #2295, opened 2024-03-11
Revert "build(deps): Update sbt, test-agent from 1.9.8 to 1.9.9"

tgodzik:
This reverts commit 0032c7fa743db98adb14e9430592eb7143948250.

Looks like native image executable is not being found since that version


tgodzik:
Ok, it seems like this was only circumstancial

PR #2313, opened 2024-04-10
fix: correctly decode project name from build target uri

kasiaMarek:
fix for: https://github.com/scalacenter/bloop/pull/2311#discussion_r1558063599
I think that `split("&")` is unnecessary here, since the build target `uri` is an identifier and shouldn't really be ever modified to encode more information.


tgodzik (approved):
LGTM!

PR #2323, opened 2024-04-15
Implement BSP buildTarget/outputPaths

jturcotte:
This fixes build-server-protocol/build-server-protocol#205 when importing from bloop by reporting the `target/` folder of each project as an outputPath.

In my case I am using it with bloop-maven-plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,5 +30,5 @@ support many different workflows and allow bespoke integrations.
 | ❓Questions? Unsure if bloop is useful for your use case? Ask right away in our [Discord channel](https://discord.gg/KWF9zMhJWS)! |
 
 [discord]: https://discord.gg/KWF9zMhJWS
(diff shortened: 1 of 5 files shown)


jturcotte:
My understanding of [this comment](https://github.com/build-server-protocol/build-server-protocol/issues/205#issuecomment-1198486634) is that support was added to IntelliJ's Scala plugin a while ago to generate the module settings, but that initially only sbt's BSP server was providing it.

tgodzik:
> My understanding of [this comment](https://github.com/build-server-protocol/build-server-protocol/issues/205#issuecomment-1198486634) is that support was added to IntelliJ's Scala plugin a while ago to generate the module settings, but that initially only sbt's BSP server was providing it.

Makes sense, I vaguely remember some discussions about it.

PR #2339, opened 2024-05-24
Fix reporting after deduplicating compilations

jchyb:
Previously in some cases when deduplicating bloop would report the `previousSuccessfulProblems` in which case bsp clients like metals could have their diagnostics overwritten with outdated ones (so here we use `previousProblems` instead).


tgodzik (approved):
LGTM!

Issue #2348, opened 2024-06-05
Scrollback cleared when in `--watch` mode


s-nel:
In #2267 a change was made to clear scrollback when in `--watch` mode. This has degraded my development experience significantly. Before when I was in watch mode, I would to scroll back to look at compiler errors or test failures while I'm simultaneously making changes. Now I have to stop everything and wait until the next compile or test run finishes which is much slower.

As an example, say that I have 100 compile errors to fix. In my IDE I fix the errors at the bottom of the Bloop output, then when I switch back to Bloop, the IDE automatically saves which kicks off a new Bloop compile and clears the scrollback. I can't keep fixing other compiler errors because the scrollback has been cleared and I have to wait.

Another example: I have a test that takes 5 minutes to execute. It just finished executing and I have some failed tests. I accidentally change a comment in the code and save. The test starts running again, and I can't look at the previous failures until the test finishes again. Much time wasted.

Another example: hours ago I got some exception and made some changes to address it. Now I want to paste the stack trace in an issue. Previously I would find it in the scrollback, but now it's gone and I would have to make changes to reproduce it.

cc @kpodsiad

PR #2349, opened 2024-06-05
Revert "fix: erase scrollback when clearing the screen"

s-nel:
Reverts scalacenter/bloop#2267 which clears backscroll of the console when in `--watch` mode.

Fixes: #2348


tgodzik:
@kpodsiad what do you think?

kpodsiad:
Fixing one's workflow breaks another one - https://xkcd.com/1172/ 

Scroll was being cleared previously, but it was cleared improperly. So I think by design and by default watch mode should clear the scroll. 

However, I understand @s-nel preference to not clear it. I wonder, maybe this can be an additional flag? What do you guys think about that?

tgodzik:
> However, I understand @s-nel preference to not clear it. I wonder, maybe this can be an additional flag? What do you guys think about that?

I think flag is the only option to have both cases satisfied

s-nel:
> Scroll was being cleared previously, but it was cleared improperly

It wasn't clearing scroll for me previously. When I start a watch command the console adds a bunch of space so that new lines appear at the top of the console, but everything that ran before is still present in the scrollback. E.g. below I have one `--watch` running, I cancel it, then I start a new `--watch`. The new watch starts at the top of the console but it hasn't cleared anything in the scroll.

```
Total duration: 14s
All 1 test suites passed.
Watching 50 directories... (press Ctrl-C to interrupt)
^C
$ bloop test ... --watch






























^C
$
```

In any case, I would be happy with a flag

PR #2357, opened 2024-06-22
bugfix: Fix wrong comparison as reproted by the compiler

tgodzik:
I recently swa this and realized this might actually cause some issues, however not idea what test would be good to verify it.


adpi2 (approved):
It's kind of weird that this did not cause any issue before, even though I don't know the use case of this `definesClass` method.

I guess if it does not break any test we can merge this.

tgodzik:
> It's kind of weird that this did not cause any issue before, even though I don't know the use case of this `definesClass` method.
> 
> I guess if it does not break any test we can merge this.

This could have caused some issues with Bloop that were hard to identify is my guess.

PR #2359, opened 2024-06-25
Updated js-bridge with newest configuration

KristianAN:
Adds support for modulesplitstyle when linking scalajs projects with esmodules

see: https://github.com/scala-cli/bloop-core/pull/167


tgodzik (approved):
LGTM!

PR #2360, opened 2024-06-26
Add support for BSP TestParamsData kinds

KristianAN:
Adds support for TestParamsData kinds as defined in the scala extension of BSP. This allows clients to specify which test classes / tests to run

Relevant bsp documentation:  https://build-server-protocol.github.io/docs/extensions/scala#testparamsdata-kinds

Fixed a bug? in the TestTask.discoverTestSuites where it did not filter on which test should should
run when explicitly passing list of test classes (fully qualified)

Included tests for running specific test suites, and tests within suites


KristianAN:
I could not find any tests for test bap service, but maybe I was looking in the wrong place?

tgodzik:
> I could not find any tests for test bsp service, but maybe I was looking in the wrong place?

Did you take a look at `BspTestSpec.scala` ?

tgodzik (approved):
Looks good! Thanks! Do you want to do anything with this PR or is it good to emrgE?

KristianAN:
It's good to merge as is I think

PR #2361, opened 2024-06-28
bugfix: Forward standard output to logger

tgodzik:
Causes https://github.com/VirtusLab/scala-cli/issues/1023


tgodzik:
The biggest issue with this approach is that we might be getting wrong compiler messages (from concurrent compilation) if compiling a couple of things at the same time. On the other hand, it would really be a rare scenario where we would get printlns from compiler at the same time for different clients as normal compilation does not do it. This should not cause real issues especially that nowadays Bloop really just works with one.

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopHighLevelCompiler.scala
```diff
@@ -132,6 +147,13 @@ final class BloopHighLevelCompiler(
               config.progress.toOptional,
               logger
             )
+
+            withTee { tee =>
+              val result = baos.toString()
+              if (result.nonEmpty)
+                logger.info(baos.toString)
+              tee.removeListener(baos)
+            }
```
adpi2, line 156: Does it mean we only get the println at the end of the compilation? So what if the compiler crashes? Should we add a `try catch` around it?

adpi2 (commented):
I don't really like it but I guess we don't really have a choice. It would be nice if the compiler uses the logger to print those trees.

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopHighLevelCompiler.scala
tgodzik: We should, otherwise might hard debugging compiler crashes

tgodzik: Moved it to a finally block.

tgodzik:
> I don't really like it but I guess we don't have a choice. It would be nice if the compiler uses the logger to print those trees.

Good news is that with the recent changes this is only an issue if the same user compiles with additional print options two projects at the same time. Which should be super rare.

adpi2 (approved):

tgodzik:
I will merge this after 2.0.0 just to be on the safe side

PR #2363, opened 2024-07-04
Fix recent regressions not allowing Metals to pass its tests 

jchyb:
Previously, if `--best-effort` option was set, every project would be compiled in the best effort mode, including those using scala 2. This first commit fixes the `CodeLensLspSuite` test. `DebugProtocolSuite` might have been broken by 7154823d8acdd3ceb05bd5000d62d869515851a7 (before that merge it works), <s>but I am still investigating that (or maybe the broken metals test should change?) </s>

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
@@ -337,7 +337,7 @@ object TestTask {
         taskDefs.map {
           case TaskDefWithFramework(taskDef, _) =>
             selectedTests.get(taskDef.fullyQualifiedName()) match {
-              case None =>
+              case None | Some(Nil) =>
                 taskDef
               case Some(value) =>
                 new TaskDef(
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -436,7 +436,7 @@ final class BloopBspServices(
     }
 
     val isPipeline = compileArgs.exists(_ == "--pipeline")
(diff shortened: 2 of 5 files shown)


tgodzik:
I will look into the DAP changes and if need to adjust anything. Looks the CI here is failing?

frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
```diff
@@ -337,7 +337,7 @@ object TestTask {
         taskDefs.map {
           case TaskDefWithFramework(taskDef, _) =>
             selectedTests.get(taskDef.fullyQualifiedName()) match {
-              case None =>
+              case None | Some(Nil) =>
```
jchyb, line 340: This fixes the DebugProtocolSuite tests (previously it was
```scala
 selectedTests.get(taskDef.fullyQualifiedName()).getOrElse(Nil) match {
              case Nil =>
              case selectedTests =>
```
and I think the Some(Nil) got lost in the refactor

jchyb:
I removed some leftover logs, hopefully that should help with the CI tests here (no idea why they broke)

frontend/src/main/scala/bloop/engine/tasks/TestTask.scala
tgodzik: Ach! Good catch!

jchyb:
Ready for review! This should fix all of the tests in metals from https://github.com/scalameta/metals/actions/runs/9760409386/job/26939422405?pr=5219 outside of MillServerCodeLensSuite in `sbt integration` (which was broken by my metals PR)

tgodzik (approved):
Awesome! LGTM

PR #2370, opened 2024-07-11
refactor: Keep daemon socket used by Scala Cli

tgodzik:
This way at least Scala CLI will be able to restart the old server without an issue and not start a totally new one. For Bloop core unfortunately people might need to stop it manually


tgodzik:
I will go ahead and merge this since it unblocks Scala CLI

PR #2382, opened 2024-07-16
improvement: Rename server class so that it's easier to recognize if …

tgodzik:
…the old Bloop is running on the system

We can prompt the user to shut it down in Metals


adpi2 (approved):
Such a great improvement! :smile:

tgodzik:
> Such a great improvement! 😄

I know :P

PR #2388, opened 2024-07-29
improvement: Generate static fallback map if can't use complete deps …

tgodzik:
…from coursier

This might happen with proxies or custom repositories. coursier complere really just works with Maven central


adpi2 (approved):
LGTM

PR #2395, opened 2024-08-12
fix: report compile progress when no-op after unsuccessful compilation

kasiaMarek:
resolves: https://github.com/scalameta/metals/issues/6656


tgodzik (approved):
LGTM!

Issue #2403, opened 2024-08-15
java.nio.file.NoSuchFileException: /usr/sbin


ornicar:
bloop: 2.0.0 release
platform: linux x86_64, kernel 6.10.2
java: openjdk version "21.0.3"
OS: NixOs

bloop fails on start with the following exception:
```
Aug 15 21:52:50 fw bloop[2596718]: Exception in thread "main" java.nio.file.NoSuchFileException: /usr/sbin
Aug 15 21:52:50 fw bloop[2596718]:         at java.base@17.0.5/sun.nio.fs.UnixFileSystemProvider.newDirectoryStream(UnixFileSystemProvider.java:440)
Aug 15 21:52:50 fw bloop[2596718]:         at java.base@17.0.5/java.nio.file.Files.newDirectoryStream(Files.java:482)
Aug 15 21:52:50 fw bloop[2596718]:         at os.list$stream$$anon$1.generate(ListOps.scala:33)
Aug 15 21:52:50 fw bloop[2596718]:         at geny.Generator.foreach(Generator.scala:49)
Aug 15 21:52:50 fw bloop[2596718]:         at geny.Generator.foreach$(Generator.scala:49)
Aug 15 21:52:50 fw bloop[2596718]:         at os.list$stream$$anon$1.foreach(ListOps.scala:31)
Aug 15 21:52:50 fw bloop[2596718]:         at geny.Generator.toBuffer(Generator.scala:132)
Aug 15 21:52:50 fw bloop[2596718]:         at geny.Generator.toBuffer$(Generator.scala:130)
Aug 15 21:52:50 fw bloop[2596718]:         at os.list$stream$$anon$1.toBuffer(ListOps.scala:31)
Aug 15 21:52:50 fw bloop[2596718]:         at geny.Generator.toArray(Generator.scala:136)
Aug 15 21:52:50 fw bloop[2596718]:         at geny.Generator.toArray$(Generator.scala:136)
Aug 15 21:52:50 fw bloop[2596718]:         at os.list$stream$$anon$1.toArray(ListOps.scala:31)
Aug 15 21:52:50 fw bloop[2596718]:         at os.list$.apply(ListOps.scala:19)
Aug 15 21:52:50 fw bloop[2596718]:         at os.list$.apply(ListOps.scala:23)
Aug 15 21:52:50 fw bloop[2596718]:         at bloop.cli.util.OsLibc$.isMusl$lzycompute(OsLibc.scala:48)
[...]

PR #2405, opened 2024-08-16
bugfix: Check is lib and /usr/sbin paths exists before listing

tgodzik:
Fixes https://github.com/scalacenter/bloop/issues/2403

### DIFF
--- a/cli/src/main/scala/bloop/cli/util/OsLibc.scala
+++ b/cli/src/main/scala/bloop/cli/util/OsLibc.scala
@@ -40,12 +40,15 @@ object OsLibc {
       if (foundMusl)
         Some(true)
       else {
-        val inLib = os.list(os.Path("/lib")).map(_.last)
+        val libPath = os.Path("/lib")
+        val inLib = if (libPath.toIO.exists()) os.list(libPath).map(_.last) else Nil
         if (inLib.exists(_.contains("-linux-gnu"))) Some(false)
         else if (inLib.exists(name => name.contains("libc.musl-") || name.contains("ld-musl-")))
           Some(true)
         else {
-          val inUsrSbin = os.list(os.Path("/usr/sbin")).map(_.last)
+          val sbinPath = os.Path("/usr/sbin")
+          val inUsrSbin =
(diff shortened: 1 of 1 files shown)


cli/src/main/scala/bloop/cli/util/OsLibc.scala
```diff
@@ -40,12 +40,15 @@ object OsLibc {
       if (foundMusl)
         Some(true)
       else {
-        val inLib = os.list(os.Path("/lib")).map(_.last)
+        val libPath = os.Path("/lib")
+        val inLib = if (libPath.toIO.exists()) os.list(libPath).map(_.last) else Nil
```
adpi2, line 44: A small suggestion:
```suggestion
        val libPath = Some(os.root / "lib").filter(os.exists)
        val inLib = libPath.flatMap(os.list).map(_.last)
```

cli/src/main/scala/bloop/cli/util/OsLibc.scala
```diff
@@ -40,12 +40,15 @@ object OsLibc {
...
         if (inLib.exists(_.contains("-linux-gnu"))) Some(false)
         else if (inLib.exists(name => name.contains("libc.musl-") || name.contains("ld-musl-")))
           Some(true)
         else {
-          val inUsrSbin = os.list(os.Path("/usr/sbin")).map(_.last)
+          val sbinPath = os.Path("/usr/sbin")
+          val inUsrSbin =
+            if (sbinPath.toIO.exists()) os.list(sbinPath).map(_.last) else Nil
```
adpi2, line 51:
```suggestion
          val sbinPath = Some(os.root / "usr" / "sbin").filter(os.exists)
          val inUsrSbin = sbinPath.flatMap(os.list).map(_.last)
```

adpi2 (approved):
Some suggestions. But overall it looks good to me.

cli/src/main/scala/bloop/cli/util/OsLibc.scala
tgodzik: I modified it to Seq instead of Some because I would have to do `toIterable` otherwise, same below

PR #2410, opened 2024-08-26
Remove betasty directory on successful compilation backgroundTasks

jchyb:
Fix to https://github.com/scalameta/metals/issues/6628
Usually it was fine to always put betasty files at the end of the classpath, but in the reproduction for some reason despite the code being correct, those files were still being used, which resulted in an incoherent state in bloop (we obtained only betasty files on what was being considered non-best effort compilation). This lead to trying to apply incremental compilation there, producing singular betasty files, and losing information about symbols from other files, producing errors.

The fix is very simple, but so far I have been unable to properly add any tests for it, as the `backgroundTasks` do not seem to be run in the tests suites.

### DIFF
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -5,7 +5,6 @@ import java.io.PrintWriter
 import java.io.StringWriter
 import java.nio.file.Files
 import java.nio.file.Path
-import java.nio.file.Paths
 import java.util.Optional
 import java.util.concurrent.Executor
 
@@ -48,7 +47,6 @@ import sbt.util.InterfaceUtil
 import xsbti.T2
 import xsbti.VirtualFileRef
 import xsbti.compile._
-
 
 case class CompileInputs(
     scalaInstance: ScalaInstance,
@@ -295,11 +293,6 @@ object Compiler {
       )
     }
 
-    val previousWasBestEffort = compileInputs.previousCompilerResult match {
-      case Failed(_, _, _, _, Some(BestEffortProducts(_, _, _))) => true
-      case _ => !firstCompilation
-    }
-
     val isFatalWarningsEnabled: Boolean =
       compileInputs.scalacOptions.exists(_ == "-Xfatal-warnings")
     def getInputs(compilers: Compilers): Inputs = {
@@ -384,6 +377,9 @@ object Compiler {
 
     reporter.reportStartCompilation(previousProblems, wasPreviousSuccessful)
     val fileManager = newFileManager
+
+    val shouldAttemptRestartingCompilationForBestEffort =
+      firstCompilation && !isBestEffortDep && previousAnalysis.isDefined
 
     // Manually skip redundant best-effort compilations. This is necessary because compiler
     // phases supplying the data needed to skip compilations in zinc remain unimplemented for now.
@@ -470,7 +466,8 @@ object Compiler {
         fileManager,
         cancelPromise,
         tracer,
-        classpathOptions
+        classpathOptions,
+        !(isBestEffortMode && isBestEffortDep)
       )
       .materialize
       .doOnCancel(Task(cancel()))
@@ -483,11 +480,9 @@ object Compiler {
             () => elapsed,
             reporter,
             backgroundTasksWhenNewSuccessfulAnalysis,
-            allInvalidatedClassFilesForProject,
-            allInvalidatedExtraCompileProducts,
             previousSuccessfulProblems,
             errorCause = None,
-            previousWasBestEffort
+            shouldAttemptRestartingCompilationForBestEffort
           )
         case Success(result) =>
           // Report end of compilation only after we have reported all warnings from previous runs
@@ -553,16 +548,25 @@ object Compiler {
                 val clientClassesDir = clientClassesObserver.classesDir
                 clientLogger.debug(s"Triggering background tasks for $clientClassesDir")
                 val updateClientState =
-                  updateExternalClassesDirWithReadOnly(
-                    clientClassesDir,
-                    clientTracer,
-                    clientLogger,
-                    compileInputs,
-                    readOnlyClassesDir,
-                    readOnlyCopyDenylist,
-                    allInvalidatedClassFilesForProject,
-                    allInvalidatedExtraCompileProducts
-                  )
+                  Task
+                    .gatherUnordered(
+                      List(
+                        deleteClientExternalBestEffortDirTask(clientClassesDir),
+                        deleteBestEffortDir()
+                      )
+                    )
+                    .flatMap { _ =>
+                      updateExternalClassesDirWithReadOnly(
+                        clientClassesDir,
+                        clientTracer,
+                        clientLogger,
+                        compileInputs,
+                        readOnlyClassesDir,
+                        readOnlyCopyDenylist,
+                        allInvalidatedClassFilesForProject,
+                        allInvalidatedExtraCompileProducts
+                      )
+                    }
 
                 val writeAnalysisIfMissing = {
                   if (compileOut.analysisOut.exists) Task.unit
@@ -585,11 +589,9 @@ object Compiler {
                   )
                 }
                   .flatMap(clientClassesObserver.nextAnalysis)
-
-                deleteBestEffortDir() *> Task
+                Task
                   .gatherUnordered(
                     List(
-                      deleteBestEffortDir,
                       deleteNewClassesDir,
                       updateClientState,
                       writeAnalysisIfMissing,
@@ -598,7 +600,8 @@ object Compiler {
                   )
                   .flatMap(_ => publishClientAnalysis)
                   .onErrorHandleWith(err => {
-                    clientLogger.debug("Caught error in background tasks"); clientLogger.trace(err);
+                    clientLogger.debug("Caught error in background tasks");
+                    clientLogger.trace(err);
                     Task.raiseError(err)
                   })
                   .doOnFinish(_ => Task(clientReporter.reportEndCompilation()))
@@ -642,10 +645,19 @@ object Compiler {
               ): Task[Unit] = {
                 val clientClassesDir = clientClassesObserver.classesDir
                 val successBackgroundTasks =
-                  deleteBestEffortDir() *> Task.gatherUnordered(
-                    backgroundTasksWhenNewSuccessfulAnalysis
-                      .map(f => f(clientClassesDir, clientReporter, clientTracer))
-                  )
+                  Task
+                    .gatherUnordered(
+                      List(
+                        deleteBestEffortDir(),
+                        deleteClientExternalBestEffortDirTask(clientClassesDir)
+                      )
+                    )
+                    .flatMap { _ =>
+                      Task.gatherUnordered(
+                        backgroundTasksWhenNewSuccessfulAnalysis
+                          .map(f => f(clientClassesDir, clientReporter, clientTracer))
+                      )
+                    }
                 val persistTask =
                   persistAnalysis(analysisForFutureCompilationRuns, compileOut.analysisOut)
                 val initialTasks = List(persistTask, successBackgroundTasks)
@@ -724,11 +736,9 @@ object Compiler {
             () => elapsed,
             reporter,
             backgroundTasksWhenNewSuccessfulAnalysis,
-            allInvalidatedClassFilesForProject,
-            allInvalidatedExtraCompileProducts,
             previousSuccessfulProblems,
             errorCause = Some(cause),
-            previousWasBestEffort
+            shouldAttemptRestartingCompilationForBestEffort
           )
 
         case Failure(_: xsbti.CompileCancelled) => handleCancellation
@@ -747,7 +757,7 @@ object Compiler {
               t.printStackTrace()
               val sw = new StringWriter()
               t.printStackTrace(new PrintWriter(sw))
-              logger.info(sw.toString())
+              logger.error(sw.toString())
               val backgroundTasks =
                 toBackgroundTasks(backgroundTasksForFailedCompilation.toList)
               val failedProblems = findFailedProblems(reporter, None)
@@ -954,14 +964,11 @@ object Compiler {
       elapsed: () => Long,
       reporter: ZincReporter,
       backgroundTasksWhenNewSuccessfulAnalysis: mutable.ListBuffer[CompileBackgroundTasks.Sig],
-      allInvalidatedClassFilesForProject: mutable.HashSet[File],
-      allInvalidatedExtraCompileProducts: mutable.HashSet[File],
       previousSuccessfulProblems: List[ProblemPerPhase],
       errorCause: Option[xsbti.CompileFailed],
-      previousWasBestEffort: Boolean
+      shouldAttemptRestartingCompilation: Boolean
   ): Result = {
     val uniqueInputs = compileInputs.uniqueInputs
-    val readOnlyClassesDir = compileOut.internalReadOnlyClassesDir.underlying
     val newClassesDir = compileOut.internalNewClassesDir.underlying
 
     reporter.processEndCompilation(
@@ -978,7 +985,7 @@ object Compiler {
       )
 
     val products = CompileProducts(
-      readOnlyClassesDir,
+      newClassesDir, // let's not use readonly dir
       newClassesDir,
       noOpPreviousResult,
       noOpPreviousResult,
@@ -995,9 +1002,13 @@ object Compiler {
       ): Task[Unit] = {
         val clientClassesDir = clientClassesObserver.classesDir
         val successBackgroundTasks =
-          backgroundTasksWhenNewSuccessfulAnalysis
-            .map(f => f(clientClassesDir, clientReporter, clientTracer))
-        val allClientSyncTasks = Task.gatherUnordered(successBackgroundTasks.toList).flatMap { _ =>
+          deleteClientExternalBestEffortDirTask(clientClassesDir).flatMap { _ =>
+            Task.gatherUnordered(
+              backgroundTasksWhenNewSuccessfulAnalysis
+                .map(f => f(clientClassesDir, clientReporter, clientTracer))
+            )
+          }
+        val allClientSyncTasks = successBackgroundTasks.flatMap { _ =>
           // Only start this task after the previous IO tasks in the external dir are done
           Task {
             // Delete everything outside of betasty and semanticdb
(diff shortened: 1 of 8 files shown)


tgodzik:
Looks like this indeed does help in case of the repository I sent your way, but the Scala compiler is failing with cryptic stuff like:
```
Could not verify that the method argument is transitively initialized (Hot). It was found to be the original object of type (class SymDenotation) where initialization checking started. Only transitively initialized arguments may be passed to methods (except constructors).
Non initialized field(s): variable myFlags, variable myPrivateWithin, variable myAnnotations, variable myParamss, variable myTargetName. Calling trace:
├── class SymDenotation private[SymDenotations] (	[ SymDenotations.scala:36 ]
│   ^
├── if (Config.checkNoSkolemsInInfo) assertNoSkolems(initInfo)	[ SymDenotations.scala:53 ]
│                                    ^^^^^^^^^^^^^^^^^^^^^^^^^
├── def assertNoSkolems(tp: Type): Unit =	[ SymDenotations.scala:1622 ]
│   ^
├── assert(!hasSkolems(tp), s"assigning type $tp containing skolems to $this")	[ SymDenotations.scala:1624 ]
│   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
├── 	[ Predef.scala:10 ]
│                                                                   ^^^^^^^
└── assert(!hasSkolems(tp), s"assigning type $tp containing skolems to $this")	[ SymDenotations.scala:1624 ]
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
any idea why?

jchyb:
This is still pretty unfinished, however I am almost done with the working version. After just this change the incremental compiler interactions can be incorrect, so I also implemented a thing where we properly restart the compilation after the first unsuccessful one (this was also incorrect before, but it was not really noticable, due to us still having betasty we could use from the previous compilation - I'll try to explain this specific issue and the fix later, after I push the code here). Also one race condition I noticed caused a crash from one of the reports we got this week, so that also should be fixed.

With that said none of these fixes fix the behavior in the compiler unfortunately. The main issue there seems to be with the java source files, which I am pretty sure we do not handle yet in best effort compilation :(. When I tested the best effort compilation in the compiler I ended up running into a compiler crash and errors related to missing java symbols - now I can't be 100% sure whether the crash was also caused by the java files, but the compiler is not always able to recover gracefully from the missing symbols, so that might be it... I am sure there is some roundabout way to work with those even without updating the compiler, however it might take more time (I'll work on it next week).

tgodzik:
> When I tested the best effort compilation in the compiler I ended up running into a compiler crash and errors related to missing java symbols - now I can't be 100% sure

Wouldn't that be caused by the problems with compilation? So if the Scala source do not compile and produce classfiles, the Java sources which use them will not be able to compile either. Especially that the compilation mode in the compiler is "mixed". I would say the Java compilation issues are not relevant to the actual problem.

jchyb:
Also, obviously it should always recover after fixing the issues, no idea why it does not do that in the compiler codebase, so I'll probably start investigating with this.

jchyb:
(I pushed it but I still want to recheck it/ clean it up, so I am leaving this as a draft for now)

jchyb:
In the meantime let me quickly explain why we do those recompilations here now:
Let's say we have a `Utils` module, consisting of:
  * UtilsA.scala
  * UtilsB.scala
 
Lets say the first Utils compilation would give us:
UtilsA.class/tasty... and UtilsB.class/tasty... <- we copy everything to client dir and to readonlyDir
We add an error to UtilsB.scala. Now we have from incremental compilation (**without any data on what else to recompile**) UtilsB.betasty.
Previously we would first copy everything (including betasty!) from readonlyDir, and copy UtilsB.betasty, giving an effect of complete best effort recompilation, but actually compiling only UtilsB.betasty. This was an issue on my part, that I did not notice. Now suddenly, after these fixes, we no longer have those outdated betasty in readonly, so we need to recompile the whole Utils, which we do in this PR. It's worth noting again that the previous behavior was also incorrect, but it was harder to notice.

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -710,9 +738,13 @@ object Compiler {
...
               t.printStackTrace()
+              val sw = new StringWriter()
+              t.printStackTrace(new PrintWriter(sw))
+              logger.info(sw.toString())
               val backgroundTasks =
                 toBackgroundTasks(backgroundTasksForFailedCompilation.toList)
-              Result.Failed(Nil, Some(t), elapsed, backgroundTasks, None)
+              val failedProblems = findFailedProblems(reporter, None)
+              Result.Failed(failedProblems, Some(t), elapsed, backgroundTasks, None)
           }
```
jchyb, line 748: Changes here are not strictly related to best effort, but not returning failedProblems in the case of compiler crash caused those annoying leftover errors in the problem tab in the compiler codebase (I can't see why it wouldn't behave the same way even for crashes unrelated to best effort). I also figured that it might be useful to also log the stacktraces of those crashes, although maybe a different log level would be better (debug?).

jchyb:
About the compiler crash itself (as seen while using best effort in the compiler codebase), the fix for that will have to be in the compiler (I already have a preliminary fix working there)

frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
```diff
@@ -39,6 +39,11 @@ import bloop.util.BestEffortUtils.BestEffortProducts
 import monix.execution.CancelableFuture
 import monix.reactive.MulticastStrategy
 import monix.reactive.Observable
+import sbt.internal.inc.BloopComponentCompiler
```
tgodzik, line 42: Any chance to run organize imports?

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -555,9 +578,11 @@ object Compiler {
                   )
                 }
                   .flatMap(clientClassesObserver.nextAnalysis)
-                Task
+
+                deleteBestEffortDir() *> Task
```
tgodzik, line 582: Isn't this already added below? Or do you need to run it first? In that case we can probably remove it from below. Ideally though, we would remove it together for it to be more efficient.

Also, is there a non symbolic way of defining it? Would that not be just a flatMap?

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -710,9 +738,13 @@ object Compiler {
               Result.Failed(failedProblems, None, elapsed, backgroundTasks, None)
             case t: Throwable =>
               t.printStackTrace()
+              val sw = new StringWriter()
+              t.printStackTrace(new PrintWriter(sw))
+              logger.info(sw.toString())
```
tgodzik, line 743:
```suggestion
              logger.error(sw.toString())
```
?

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -938,49 +971,52 @@ object Compiler {
...
-          val secondTask = Task {
+          // Only start this task after the previous IO tasks in the external dir are done
+          Task {
             // Delete everything outside of betasty and semanticdb
             val deletedCompileProducts =
               BloopClassFileManager.supportedCompileProducts.filter(_ != ".betasty") :+ ".class"
             Files
               .walk(clientClassesDir.underlying)
-              .filter(path => Files.isRegularFile(path))
+              .filter(path => if (Files.exists(path)) Files.isRegularFile(path) else false)
```
tgodzik, line 981:
```suggestion
              .filter(path => Files.exists(path) && Files.isRegularFile(path))
```

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -938,49 +971,52 @@ object Compiler {
...
               .filter(path => deletedCompileProducts.exists(path.toString.endsWith(_)))
-              .forEach(Files.delete(_))
-          }
-          Task
-            .gatherUnordered(List(firstTask, secondTask))
-            .map(_ => ())
+              .forEach(path => if (Files.exists(path)) Files.delete(path))
```
tgodzik, line 983:
```suggestion
              .forEach(path => Files.delete(path)
```
Isn't this checked above? We could probably join the 2 filters and a foreach now

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -938,49 +971,52 @@ object Compiler {
           backgroundTasksWhenNewSuccessfulAnalysis
             .map(f => f(clientClassesDir, clientReporter, clientTracer))
         val allClientSyncTasks = Task.gatherUnordered(successBackgroundTasks.toList).flatMap { _ =>
-          // Only start these tasks after the previous IO tasks in the external dir are done
-          val firstTask = updateExternalClassesDirWithReadOnly(
```
tgodzik, line 942: Should this be removed? Will be sync the needed files to the client directory?

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -938,49 +971,52 @@ object Compiler {
...
-      compileInputs.sources,
-      compileInputs.classpath
-    )
+    val recompile =
+      if (
+        !previousWasBestEffort && !(compileOut.internalReadOnlyClassesDir.exists && BloopPaths
+          .list(compileOut.internalReadOnlyClassesDir)
+          .length == 0)
+      ) {
+        if (compileOut.analysisOut.exists) BloopPaths.delete(compileOut.analysisOut)
```
tgodzik, line 997: deleting analysis should probably be enough as that will result in a full recompilation. I think we do that for clean

Also what is the exact case we want to remove everything?

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -938,49 +971,52 @@ object Compiler {
...
     }
 
-    val newHash = BestEffortUtils.hashResult(
-      products.newClassesDir,
-      compileInputs.sources,
-      compileInputs.classpath
-    )
+    val recompile =
+      if (
+        !previousWasBestEffort && !(compileOut.internalReadOnlyClassesDir.exists && BloopPaths
```
tgodzik, line 993: Maybe you could extract the part after `&&` and name it to explain what is actually being checked or why?

frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
```diff
@@ -171,9 +176,72 @@ object CompileTask {
...
+                case result @ Compiler.Result.Failed(
+                      _,
+                      _,
+                      _,
+                      _,
+                      Some(BestEffortProducts(_, _, recompile))
+                    ) if recompile =>
+                  // we restart the compilation, starting from scratch (without any previous artifacts)
+                  inputs.reporter.reset()
+                  val foundSrcs = getAllSourceInputs(project)
```
tgodzik, line 233: This should not be needed here. We already calculate sources in other places. Maybe if we just remove analysis it will not be needed?

tgodzik (changes requested):
Thanks for working on this!

backend/src/main/scala/bloop/Compiler.scala
jchyb: Yes, I need to tun this before `updateClientState` (I missed the repeat below, thank you for the catch!), but I'll try rearranging the tasks so that the other ones can be run concurrently

backend/src/main/scala/bloop/Compiler.scala
jchyb: Yes, but in that case do not want to use the readonlyDir (previous compilation), and instead use just newDir (current compilation)

frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
jchyb: Bit embarassed about this but thanks for catching. I kept looking for how to properly recompile everything, came up with this (since I saw that not everything was passed into CompileInputs.sources), and only later added the key `inputs.reporter.reset()` - which it looks like it's actually all I needed, so I was able to remove that getAllSourceInputs, thankfully.

jchyb:
Since the last review I figured out an elegant way to keep the previous successful artifacts along with CompileAnalysis and reuse those, while also being able to have the best effort information simultaneously. To do this, if we already have an analysis and successful artifacts, we always first try compiling everything changed/tasked with recompilation since last successful, and then if something fails along the way recompiling the whole project in best effort mode.

Compiling twice in case of every failure might not seem performant, but it's actually the opposite. The first compilation is going to be short, since (likely) very few files are going to be recompiled. Second will be a bit longer, but still manageable, as we end the compilation at an earlier phase. This is better then the alternative, when in case of failure we replace everything with `betasty`, and when we fix the errors, everything has to be recompiled, including long transform and backend phases.

To put this into perspective, in the reproduction repository I was given, when we add a typer error to a compiled file, the first erroring compilation on my machine takes around `1s`, then the best effort recompilation takes around `5s`, and if we fix the error we end up with a single `1s` recompilation. However if we were to clean the successful output (like I tried before), then we end up having to withstand full `20s` recompilations when we fix the errors, and switching back end forth between failing and successful programs becomes annoying.

I did make sure to test whether the standard incremental compilation behaviors still work - zinc schedules recompilations and updates the analyses only for successful compilations, so it all still works (e.g. I add an indirect error -> the file is successfully compiled -> another few files are scheduled for recompilation -> this recompilation still fails -> every "second" compilation will try to recompile those files and also newly changed ones). I made it a bit confusing here, but the main thing is that it I did not have to add any extra logic for that to work.

Since that compilation order aspect started to get a bit complex, I added a chapter in the contributors guide about it (which might be a bit weird, since there is not much else like it there, so having implementation details there might be a bit confusing, but it's better than nothing in my opinion).

backend/src/main/scala/bloop/Compiler.scala
jchyb: I ended up completely changing this part since the last review to match the new design, but it also should be way less confusing now

frontend/src/main/scala/bloop/engine/tasks/compilation/CompileGraph.scala
```diff
@@ -453,7 +453,7 @@ object CompileGraph {
                           .+=(newProducts.readOnlyClassesDir.toFile -> newResult)
                       case (p, ResultBundle(f: Compiler.Result.Failed, _, _, _)) =>
                         f.bestEffortProducts.foreach {
-                          case BestEffortProducts(products, _) =>
+                          case BestEffortProducts(products, _, _) =>
```
tgodzik, line 456: Btw. I just noticed `isBestEffortDep` is always false?

frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
```diff
@@ -172,8 +177,29 @@ object CompileTask {
...
+                  // we restart the compilation, starting from scratch (without any previous artifacts)
+                  inputs.reporter.reset()
+                  val emptyResult =
+                    PreviousResult.of(Optional.empty[CompileAnalysis], Optional.empty[MiniSetup])
+                  val newInputs = inputs.copy(
+                    sources = inputs.sources,
+                    previousCompilerResult = result,
+                    previousResult = emptyResult
+                  )
+                  Compiler.compile(newInputs, isBestEffort, isBestEffortDep, false)
```
tgodzik, line 199:
```suggestion
                  Compiler.compile(newInputs, isBestEffort, isBestEffortDep, firstCompilation = false)
```

backend/src/main/scala/bloop/Compiler.scala
tgodzik: Looks like this is unchanged? Would you mind including the change?

tgodzik (changes requested):

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -500,6 +508,17 @@ object Compiler {
             Task(persist(out, analysis, result.setup, tracer, logger)).memoize
           }
 
+          // .betasty files are always produced with -Ybest-effort, even when
+          // the compilation is successful.
+          // We might want to change this in the compiler itself...
```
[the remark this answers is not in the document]
jchyb: Just made one: https://github.com/scala/scala3/issues/21833

frontend/src/main/scala/bloop/engine/tasks/compilation/CompileGraph.scala
jchyb: Thank you for catching it, it must have become redundant in one of the iterations of the first PR and I didn't notice that before. Removed now

backend/src/main/scala/bloop/Compiler.scala
jchyb: Fixed now, sorry about that

backend/src/main/scala/bloop/Compiler.scala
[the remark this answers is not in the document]
tgodzik: Thanks!

frontend/src/main/scala/bloop/engine/tasks/compilation/CompileGraph.scala
```diff
@@ -389,14 +389,15 @@ object CompileGraph {
...
-    def loop(dag: Dag[Project], isBestEffortDep: Boolean): CompileTraversal = {
+    def loop(dag: Dag[Project]): CompileTraversal = {
       tasks.get(dag) match {
         case Some(task) => task
         case None =>
           val task: Task[Dag[PartialCompileResult]] = dag match {
             case Leaf(project) =>
               val bundleInputs = BundleInputs(project, dag, Map.empty)
               setupAndDeduplicate(client, bundleInputs, computeBundle) { bundle =>
+                val isBestEffortDep = false
```
tgodzik, line 400: Why do we need it here? This is the only place we invoke `compile` so isBestEffortDep is always false. Shouldn't we remove it everywhere?

jchyb: It used in line 466/467 (I don't think I can link there as part of the review), where we set it if the dependency was unsuccessful and had betasty artifacts. I left this redundant `val isBestEffortDep = false` as I thought having a false with no context in the lambda would be annoying (but I can change that of course)

tgodzik: Ach, alright I haven't noticed that. Thanks for explaning!

tgodzik (approved):
LGTM!

PR #2427, opened 2024-09-10
bugfix: Don't republish old errors on successful compilation

tgodzik:
We need to report all problems if a BSP client just connected to the server and never compiled. `reportAllPreviousProblems` will be set to true in that case.

However, if previously we had an error, which was reverted and got back to the state before the error, we will get a successfull noop compilation, which will not produce any problems.

In this case we don't want to republish everything again, since diagnostics will contain the error which was reverted.

Connected to https://github.com/VirtusLab/scala-cli/issues/2226 where CLI always connected anew.

### DIFF
--- a/frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
+++ b/frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
@@ -301,14 +301,31 @@ final class BspProjectReporter(
   ): Unit = {
     val problemsInPreviousAnalysisPerFile = Reporter.groupProblemsByFile(previousSuccessfulProblems)
 
+    /**
+     * We need to report all problems if a BSP client just connected to the server and never compiled.
+     * `reportAllPreviousProblems` will be set to true in that case.
+     *
+     * However, if previously we had an error, which was reverted and got back to the state before the
+     * error, we will get a successfull noop compilation, which will not produce any problems.
+     *
+     * In this case we don't want to republish everything again, since diagnostics will contain the
+     * error which was reverted.
+     *
+     * Connected to https://github.com/VirtusLab/scala-cli/issues/2226 where CLI always connected anew.
+     */
+    def shouldPublishAllProblems =
+      reportAllPreviousProblems && !wasPreviousSuccessful && code != bsp.StatusCode.Ok
     def mockNoOpCompileEventsAndEnd: Option[CompilationEvent.EndCompilation] = {
       recentlyReportProblemsPerFile.foreach {
-        case (sourceFile, problemsPerFile) if reportAllPreviousProblems =>
+        case (sourceFile, problemsPerFile) if shouldPublishAllProblems =>
           reportAllProblems(sourceFile, problemsPerFile)
         case (sourceFile, problemsPerFile) =>
           problemsInPreviousAnalysisPerFile.get(sourceFile) match {
             case Some(problemsInPreviousAnalysis) =>
-              if (problemsInPreviousAnalysis.map(_.problem) == problemsPerFile.map(_.problem)) {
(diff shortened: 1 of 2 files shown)


frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
```diff
@@ -301,14 +301,31 @@ final class BspProjectReporter(
...
+     * However, if previously we had an error, which was reverted and got back to the state before the
+     * error, we will get a successfull noop compilation, which will not produce any problems.
+     *
+     * In this case we don't want to republish everything again, since diagnostics will contain the
+     * error which was reverted.
+     *
+     * Connected to https://github.com/VirtusLab/scala-cli/issues/2226 where CLI always connected anew.
+     */
+    def shouldPublishAllProblems =
+      reportAllPreviousProblems && !wasPreviousSuccessful && code != bsp.StatusCode.Ok
```
adpi2, line 317: But if it was successful and it is still, Bloop should report the warnings, right?

frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
```diff
@@ -301,14 +301,31 @@ final class BspProjectReporter(
...
 
+    /**
+     * We need to report all problems if a BSP client just connected to the server and never compiled.
+     * `reportAllPreviousProblems` will be set to true in that case.
+     *
+     * However, if previously we had an error, which was reverted and got back to the state before the
+     * error, we will get a successfull noop compilation, which will not produce any problems.
+     *
+     * In this case we don't want to republish everything again, since diagnostics will contain the
+     * error which was reverted.
```
adpi2, line 312: I am not sure to understand this. If the previous was a noop compilation, why would diagnostics contain any error? In other words, I think a noop compilation should clear the diagnostics that were stored anywhere.

adpi2 (commented):
I am not sure to understand everything. But maybe a noop compilation should clear the errors that were stored in the reporter.

frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
tgodzik: Yes, but the warnings will be contained in the previous analysis. So they will get republished properly

tgodzik:
> I am not sure to understand everything. But maybe a noop compilation should clear the errors that were stored in the reporter.

That's the whole issue actually. If a new client connects (which Scala CLI does every time it's run) and noop compilation is run then if we clear the errors nothing will be reported the new client.

frontend/src/main/scala/bloop/reporter/BspProjectReporter.scala
adpi2: Okay but if it is a success, we should first clear `problemsPerFiles` so that it does not contain any error anymore.

tgodzik: That would make sense, but then when we have reportAllPreviousProblems set (this is set when first compilation for a target and client is run) then we would not report old warnings, since we would clean it. We can try to clean just the errors out, but we don't know if some warnings weren't generate along those errors.

So in essence, we don't know what exact problems were generated in the compilation before and those are the ones we should clear. 

It's much easier to just special case this situation and just report everything in the previous successful compilation. And while I agree with you that something is wrong here, I can't figure out a more sensible solution, all alternative paths seem more complicated and carry their own drawbacks. This might be due to the fact that noop compilation is a bit of a trick, which is not really something that a lot of other tools using zinc do.

adpi2 (approved):

PR #2446, opened 2024-09-18
improvement: Respect bspEnabled when generating bloop config files

tgodzik:
I noticed we have to work around it in a number of places and might be useful in the scala3 repo for example.

We didn't do it previously to keep the sbt version as low as possible, but it should be fine for users of older sbt to use older plugins (will be done automatically in Metals)


adpi2 (approved):
That's a good idea

tgodzik:
Btw. I had to fix some tests mostly caused by bumping sbt version. It seems the scripted tests are now being run in a random temporary directory, while previously it would be named as the directory for the tests. Adjusted the tests accordingly.

PR #2450, opened 2024-09-25
bugfix: Use proper rt.jar if it exists

tgodzik:
Previously, we would automatically generate our own rt.jar from the Bloop's JDK, however this would put all the current JDK classes into that jar, which means some things might compile while they should not. This limitation is present both in Mill and sbt. We would only fail on added methods.

Now, we try to locate rt.jar if javaHome is set, so we can use it explicitely.

### DIFF
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -874,8 +874,17 @@ object Compiler {
       case _ => false
     }
     val updatedClasspath =
-      if (needsRtJar) inputs.classpath ++ RtJarCache.create(JavaRuntime.version, logger)
-      else inputs.classpath
+      if (needsRtJar) {
+        val possibleRtJar = inputs.javacBin
+          .map { binary =>
+            binary.getParent.resolve("../jre/lib/rt.jar")
+          }
+          .filter(_.exists)
+        possibleRtJar match {
+          case Some(rtJar) => inputs.classpath :+ rtJar
+          case None => inputs.classpath ++ RtJarCache.create(JavaRuntime.version, logger)
+        }
+      } else inputs.classpath
     val classpathVirtual = updatedClasspath.map(path => converter.toVirtualFile(path.underlying))
     CompileOptions
       .create()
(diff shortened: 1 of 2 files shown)


backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -874,8 +874,17 @@ object Compiler {
       case _ => false
     }
     val updatedClasspath =
-      if (needsRtJar) inputs.classpath ++ RtJarCache.create(JavaRuntime.version, logger)
-      else inputs.classpath
+      if (needsRtJar) {
+        val possibleRtJar = inputs.javacBin
+          .map { binary =>
+            binary.getParent.resolve("../jre/lib/rt.jar")
```
adpi2, line 880: It looks suspicious to do `getParent` and then resolve `..`. Would it be better to do `getParent.getParent`, or `../..`?

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -874,8 +874,17 @@ object Compiler {
...
+          }
+          .filter(_.exists)
+        possibleRtJar match {
+          case Some(rtJar) => inputs.classpath :+ rtJar
+          case None => inputs.classpath ++ RtJarCache.create(JavaRuntime.version, logger)
```
adpi2, line 885:
```suggestion
        inputs.classpath ++ possibleRtJar.orElse(RtJarCache.create(JavaRuntime.version, logger))
```

adpi2: I would even rewrite the entire thing to:
```scala
val possibleRtJar = 
  if (needsRtJar) 
    inputs.javacBin
      .map(_.resolve("../../jre/lib/rt.jar"))
      .filter(_.exists)
      .orElse(RtJarCache.create(JavaRuntime.version, logger))
  else None
val updatedClasspath = inputs.classpath ++ possibleRtJar
```

adpi2 (approved):
Some nitpicking

backend/src/main/scala/bloop/Compiler.scala
tgodzik: Done!

backend/src/main/scala/bloop/Compiler.scala
tgodzik: Switched to getParent and wrapped in Try (if someone has java bin in `/java`, which sounds ridiculous, but I've seen things.)

Issue #2479, opened 2024-10-20
Fork or vendor jsonrpc4s to address bugs in it


alexarchambault:
https://github.com/scalacenter/bloop/issues/2478 originates from a bug in [jsonrpc4s](https://github.com/jvican/jsonrpc4s).

Could you create a fork of it, in order to publish a version fixing this bug and use it in Bloop?

For easier maintenance, the jsonrpc build could be switched to scala-cli, like done [here](https://github.com/alexarchambault/jsonrpc4s/commit/0ec4a357f8542d00c5e9fc8d6a6cab076731b877) for example.

tgodzik:
Sure, I was actually hoping to switch to  just bsp4j, but that hasn't happened yet. I can also just use your fork now? Only fork it once needed

alexarchambault:
Sure, you can use my fork, I can give you write access

alexarchambault:
Just sent the invite

Issue #2478, opened 2024-10-20
Malformed jsonrpc method not found responses


alexarchambault:
When the Bloop BSP server receives requests that it doesn't support, it often responds with answers like
```json
{
    "id": "6",
    "error": {
        "code": -32601,
        "message": "buildTarget/wrappedSources"
    }
},{
    "error": {
        "code": -32601,
        "message": "buildTarget/wrappedSources"
    },
    "id": "6",
    "jsonrpc": "2.0"
}
```

That is, the response is malformed, somehow concatenating two JSON objects, separated by a comma. BSP clients fail to parse such a response, and thus do not properly report the method as being unsupported by the BSP server, usually failing with a more severe error (JSON parsing error).

alexarchambault:
The bug in more detail originates from jsonrpc4s relying on [jsoniter-scala](https://github.com/plokhotnyuk/jsoniter-scala) to (de)serialize JSON. Since jsoniter-scala [`2.10.0`](https://github.com/plokhotnyuk/jsoniter-scala/releases/tag/v2.10.0), its main methods to write JSON aren't re-entrant anymore (a codec called by a write method shouldn't itself call a write method, for example, because the write methods are stateful and can't be called twice at the same time).

jsonrpc4s needs to use the re-entrant write methods of jsoniter-scala instead, like done [here](https://github.com/alexarchambault/jsonrpc4s/commit/05248042a3ba7a63577395eb433b1a02119dbed3#diff-8bacf79d6cc2979674c6160546b0ccebf0cb424f6deeafad769b330a2de14ca0R14).

alexarchambault:
Note that the [original Scala CLI fork of Bloop](https://github.com/scala-cli/bloop-core) doesn't suffer this bug, thanks to [12a823a1fccd9e469a408e4af4a380f5945c02b3](https://github.com/scala-cli/bloop-core/commit/12a823a1fccd9e469a408e4af4a380f5945c02b3) and its use of [this jsonrpc fork](https://github.com/alexarchambault/jsonrpc4s)

PR #2483, opened 2024-10-22
chore: Replace original jsonrpc4s

tgodzik:
Fixes https://github.com/scalacenter/bloop/issues/2479

Fixes https://github.com/scalacenter/bloop/issues/2478

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -58,9 +58,15 @@ lazy val bloopShared = project
     libraryDependencies ++= Seq(
       Dependencies.jsoniterCore,
       Dependencies.jsoniterMacros,
-      Dependencies.bsp4s excludeAll ExclusionRule(
-        organization = "com.github.plokhotnyuk.jsoniter-scala"
+      Dependencies.bsp4s.excludeAll(
+        ExclusionRule(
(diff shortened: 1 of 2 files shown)


tgodzik:
CC @alexarchambault

alexarchambault (approved):

build.sbt
```diff
@@ -58,9 +58,15 @@ lazy val bloopShared = project
...
-        organization = "com.github.plokhotnyuk.jsoniter-scala"
+      Dependencies.bsp4s.excludeAll(
+        ExclusionRule(
+          organization = "com.github.plokhotnyuk.jsoniter-scala"
+        ),
+        ExclusionRule(
+          organization = "me.vican.jorge"
+        )
       ),
+      Dependencies.jsonrpc4s,
```
adpi2, line 69: So I guess we should also switch the jsonrpc4s version in [build-server-protocol](https://github.com/build-server-protocol/build-server-protocol)?

tgodzik: Hmm... that might be an option too, but it's only really used by Bloop currently as far as I know

adpi2 (approved):

PR #2487, opened 2024-10-28
WIP/experiment: try to fix ci

jchyb:
No idea what is going on but I'll try some things and see how those pan out


tgodzik:
Looks like I can't reproduce it on my linux either. I will try on another machine, but this is pretty weird.

jchyb:
The partial revert of the related part of my commit didn't work so now I am trying to revert the whole thing and see if the issue really is there

tgodzik:
Looks like it might be a coincidence that this happen around the same time as merging your PR then :thinking: 

Maybe it was ubuntu update?

tgodzik:
Thanks @jchyb for checking this out! I managed to fix it separately

PR #2489, opened 2024-10-31
 bugfix: Use last modified time instead of creation time for detecting orphaned directories

tgodzik:
Somehow this was an issue on linux and JDK 17, but not sure why not others.
    
I also fixed getting stdout to get it redirected to the logger, my previous change did not fully fix it.

### DIFF
--- a/frontend/src/main/scala/bloop/data/ClientInfo.scala
+++ b/frontend/src/main/scala/bloop/data/ClientInfo.scala
@@ -292,7 +292,7 @@ object ClientInfo {
                   val dirName = clientDir.underlying.getFileName().toString
                   val attrs =
                     Files.readAttributes(clientDir.underlying, classOf[BasicFileAttributes])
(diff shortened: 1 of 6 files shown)


frontend/src/main/scala/bloop/data/ClientInfo.scala
```diff
@@ -292,7 +292,7 @@ object ClientInfo {
                   val dirName = clientDir.underlying.getFileName().toString
                   val attrs =
                     Files.readAttributes(clientDir.underlying, classOf[BasicFileAttributes])
-                  val isOldDir = attrs.creationTime.toInstant.isBefore(deletionThresholdInstant)
+                  val isOldDir = attrs.lastModifiedTime.toInstant.isBefore(deletionThresholdInstant)
```
tgodzik, line 295: I think it's better to use last modified and it's possible to check in the tests actually.

adpi2 (approved):

PR #2490, opened 2024-10-31
bugfix: Compile only existing classes

tgodzik:
Previously, we would ask zinc to compile even a file that was removed and I think this was causing https://github.com/scalameta/metals/issues/6478

Now, we invalidated all changed sources, but only compile new ones.

### DIFF
--- a/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
+++ b/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
@@ -90,59 +90,66 @@ private final class BloopNameHashing(
           case ref: VirtualFileRef => PlainVirtualFileConverter.converter.toVirtualFile(ref)
         }
 
-      recompileClasses(invalidatedSources, binaryChanges, previous, compileTask, manager).flatMap {
-        current =>
-          // Return immediate analysis as all sources have been recompiled
-          if (invalidatedSources == allSources) Task.now(current)
-          else {
-            val recompiledClasses: Set[String] = {
-              // Represents classes detected as changed externally and internally (by a previous cycle)
-              classesToRecompile ++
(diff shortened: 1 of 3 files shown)


backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
```diff
@@ -231,21 +238,22 @@ private final class BloopNameHashing(
   }
 
   def recompileClasses(
-      sources: Set[VirtualFile],
+      currentInvalidatedSources: Set[VirtualFile],
```
tgodzik, line 241: This is the main change, we would previously send all invalidated files even if they were no longer contained in the sources. This make sense for class files to remove them, but no some much to compile new files.

With this change I was no longer able to reproduce the issue in Metals where we would get all compilation artifacts somehow.

frontend/src/test/scala/bloop/bsp/BspCompileSpec.scala
```diff
@@ -349,6 +349,77 @@ class BspCompileSpec(
...
+          compiledState.client.getUniqueClassesDirFor(buildProject, forceGeneration = true)
+
+        val classFilesAfterSuccess = takeDirectorySnapshot(externalClassesDirA)
+        assertEquals(
+          classFilesAfterSuccess.map(_.path.toString()),
+          List(s"/scala/meta/internal/metals/Foo$suffix.class")
+        )
+      }
+
+      loadBspState(workspace, projects, logger) { state =>
```
tgodzik, line 385: I haven't been able to reproduce the issue in the test unfortunately, this seems somehow connected to multiple thing being done at the same time via metals and VS Code, but the change in BloopNameHashing is sound anyway.

And having this test won't hurt

adpi2 (approved):
LGTM

PR #2494, opened 2024-11-05
chore: Fix variable refering to older macOS

tgodzik:
I missed it since it was capitalized differently and uploading artifacts failed


tgodzik:
Let's see if that helps

PR #2516, opened 2024-11-26
bugfix: Catch fatal errors when persisting zinc analysis

tgodzik:
Related to https://github.com/scala/scala3/issues/22026


adpi2 (approved):
I know there are several analysis formats in Zinc, including one that is now the default in sbt 2.x. So maybe you should try it to see if it solves your `StackOverflowError`.

tgodzik:
> I know there are several analysis formats in Zinc, including one that is now the default in sbt 2.x. So maybe you should try it to see if it solves your `StackOverflowError`.

Hmm... might be worth just trying that snippet in sbt 2

Issue #1304, opened 2020-06-05
Bloop link does not work for Scala.js projects without a main class


MaximeKjaer:
# Bug description

`bloop link` is not able to link Scala.js projects without a main class. It reports the following error:

```console
> bloop link root
[E] No main classes defined in project 'root'
```

However, Scala.js projects do not always need to have a main class: for instance, a valid use-case is to have no main class, but have methods with `@JSExport` annotations in order to export functionality to JavaScript code.

# Reproduction

I created a simple project with a small `build.sbt` config:

```scala
lazy val root = project
  .in(file("."))
  .enablePlugins(ScalaJSPlugin)
  .settings(
    scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) },
  )
```

The project only contains a single Scala file, which `JSExport`s a single hello world function:

```scala
import scalajs.js.annotation._

@JSExportTopLevel("HelloWorld")
object HelloWorld {
  @JSExport
  def sayHello(): Unit = {
    println("Hello world!")
  }
}
```

If I run `sbt fastOptJS`, sbt produces a bundle from which JS code can import things:

```js
> require("./target/scala-2.12/root-fastopt.js").HelloWorld.sayHello()
Hello world!
```

Running `bloop compile root` works as expected, but `bloop link root` produces the following output:

```console
> bloop link root
[E] No main classes defined in project 'root'
```

Here is the generated Bloop configuration:

<details><summary><code>.bloop/root.json</code></summary>

```json
{
    "version": "1.4.0",
    "project": {
        "name": "root",
        "directory": "/home/maxime/code/tmp",
        "workspaceDir": "/home/maxime/code/tmp",
        "sources": [
            "/home/maxime/code/tmp/src/main/scala-2.12",
            "/home/maxime/code/tmp/src/main/scala",
            "/home/maxime/code/tmp/src/main/java",
            "/home/maxime/code/tmp/target/scala-2.12/src_managed/main"
        ],
        "dependencies": [
            
        ],
        "classpath": [
            "/home/maxime/.sbt/boot/scala-2.12.10/lib/scala-library.jar",
            "/home/maxime/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-js/scalajs-library_2.12/1.1.0/scalajs-library_2.12-1.1.0.jar"
        ],
        "out": "/home/maxime/code/tmp/.bloop/root",
[...]

PR #2535, opened 2024-12-12
Add --link flag to compile to allow linking through BSP

KristianAN:
Before this commit linking could only be done through the run command with BSP. Because the BSP protocol does describe a linking command (as it is often a part of the compile step) the flag --link has been added to the compile command. This is useful for those kinds of projects that need the linker to be run to produce source code for further toolchains, e.g. a ScalaJS project might often use @JSExport and then run the code from some other JS entrypoint.


KristianAN:
@tgodzik  I'm not quite sure where/how I can write a test for this. I looked at BspCompileSpec, but I'm not quite sure how I can pass in the compileArgs there.
 
 I also want to further investigate some bugs with the JS linker config as it does not seem to respect everything that is set from bloop-config, but maybe that should be it's own PR?

KristianAN:
this should also fix: https://github.com/scalacenter/bloop/issues/1304

KristianAN:
The JS-bridge defines this:

```scala
    val moduleInitializers = mainClass match {
      case Some(mainClass) if runMain =>
        logger.debug(s"Setting up main module initializers for $project")
        List(ModuleInitializer.mainMethodWithArgs(mainClass, "main"))
      case _ =>
        if (runMain) {
          logger.debug(s"Setting up no module initializers, commonjs module detected $project")
          Nil // If run is disabled, it's a commonjs module and we link with exports
        } else {
          // There is no main class, install the test module initializers
          logger.debug(s"Setting up test module initializers for $project")
          List(
            ModuleInitializer.mainMethod(
              TestAdapterInitializer.ModuleClassName,
              TestAdapterInitializer.MainMethodName
            )
          )
        }
    }
```
I disagree here. Not running main is indicative of JSExport and the likes I think, not that we want to link with tests (I don't think this is very common). I propose here that we instead add a test flag for the test case and add that same flag to the Commands.Test (it can also be an additional flag in the compileArgs).

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -421,6 +424,48 @@ final class BloopBspServices(
...
+        linkJS(cmd, userProjects._2, state, None, targetDir, platform).map { state =>
+          val result = if (state.status == ExitStatus.LinkingError) {
+            Right(bsp.CompileResult(originId, bsp.StatusCode.Error, None, None))
+          } else {
+            Right(bsp.CompileResult(originId, bsp.StatusCode.Ok, None, None))
+          }
+          (state, result)
+        }
+      case Jvm(_, _, _, _, _, _) =>
+        compileProjects(userProjects :: Nil, state, compileArgs, originId, logger)
```
KristianAN, line 448: Alternatively here we can do the same as in the interpreter when it hits a JVM project on link. I'm not sure if there is much difference.

tgodzik:
> @tgodzik I'm not quite sure where/how I can write a test for this. I looked at BspCompileSpec, but I'm not quite sure how I can pass in the compileArgs there.

This should be similar to https://github.com/scalacenter/bloop/commit/32dbfb14dbaec488eccb515db67a2b0df8339f54#diff-9458dec34378d1535538ca59de86e441bcef70b5739a0cc484281ac8ed1ef1aaR695

> I also want to further investigate some bugs with the JS linker config as it does not seem to respect everything that is set from bloop-config, but maybe that should be it's own PR?

Would probably be easier to review, but I can also ask for help reviewing

tgodzik:
> I disagree here. Not running main is indicative of JSExport and the likes I think, not that we want to link with tests (I don't think this is very common). I propose here that we instead add a test flag for the test case and add that same flag to the Commands.Test (it can also be an additional flag in the compileArgs).

I haven't worked in that part too much, but I think that might have been a simplification since Bloop wasn't used for publishing anything

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -421,6 +424,84 @@ final class BloopBspServices(
...
+    compileProjects(userProjects, state, compileArgs, originId, logger).flatMap {
+      case (newState, Right(CompileResult(_, StatusCode.Ok, _, _))) =>
+        val linkExecution: Task[Seq[(State, Either[Response.Error, CompileResult])]] =
+          Task.sequence(userProjects.map { case (_, p) => doLink(p, newState) })
+
+        linkExecution.materialize.map {
+          case Success(linkRunsSeq) =>
+            linkRunsSeq
+              .map(_._2)
+              .foldLeft(Option.empty[CompileResult])((_, res) => // TODO propogate stuff better here
```
KristianAN, line 483: This logic is somewhat lifted from the test function, but it doesn't really work that well here. I'm not entirely sure what the best strategy is for combining the compileresults (actually linker results)

KristianAN:
So far I have tested that the changes in this PR allows for linking using JSExport and with a main method without having to call run.

This is also tested through BSP in relation to this PR: https://github.com/oyvindberg/bleep/pull/445.

I have also added support for the new minification in ScalaJS which is now set to true for release esmodule builds. We can add an option for this in bloop-config so that users can override this setting if they want.

Further TODOs for this PR:

- Fix the TODO I have commented
- Add some tests
- Look at the ScalaNative linker

@tgodzik anything else you think should be done?

tgodzik:
> anything else you think should be done?

Nothing comes to mind, but maybe @sjrd will be able to take a quick look here?

bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
```diff
@@ -85,7 +85,7 @@ object JsBridge {
       (config.kind, scalaJSModuleKindSplitStyle) match {
         case (ModuleKindJS.ESModule, Some(value)) =>
           StandardImpl.clearableLinker(
-            linkerConfig.withModuleSplitStyle(value)
+            linkerConfig.withModuleSplitStyle(value).withMinify(isFullLinkJS)
```
sjrd, line 88: In order to be consistent with sbt-scalajs, you should enable `minify` for *all* `fullLink` builds, regardless of `ModuleKind`.

Even when used in conjunction with the Closure compiler, `minify` still brings some code size improvements: the optimizer works extra hard to reduce the need for temporary local variables.

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -421,6 +424,76 @@ final class BloopBspServices(
...
+          val targetDir = ScalaJsToolchain.linkTargetFrom(project, config)
+          linkJS(cmd, project, newState, false, None, targetDir, platform).map { linkState =>
+            val result = if (linkState.status == ExitStatus.LinkingError) {
+              Right(bsp.CompileResult(originId, bsp.StatusCode.Error, None, None))
+            } else {
+              Right(bsp.CompileResult(originId, bsp.StatusCode.Ok, None, None))
+            }
+            (linkState, result)
+          }
+        case Jvm(_, _, _, _, _, _) => ???
```
KristianAN, line 452: @tgodzik Should we error here or just do a NoOp? if we do a NoOp a client can request linking all projects and only the ones that can link will be linked and trying to link JVM doesn't stop the linking. If we error  the client has to be explicit and only try to link targets that can be linked.

tgodzik:
> @tgodzik this is now blocked by [scalacenter/bloop-config#102](https://github.com/scalacenter/bloop-config/pull/102)

Running the release here https://github.com/scalacenter/bloop-config/actions/runs/12394117349

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
tgodzik: We should probably just compile normally in this case and ignore link option. A warning could be printed

bridges/scalajs-1-test/src/test/scala/bloop/scalajs/ScalaJsToolchainSpec.scala
KristianAN, line 1: Moving the these tests to a different module lets us remove the frontend dependency from the bridges. This allows us to create working tests for bsp in the bridges. However it breaks the ToolchainSpecs because the loading the bridge toolchain with the classloader no longer works. 

I am no guru on these classloader issues.  This is the part of the test that no longer works when separating the modules.

```scala
  val state0: State = {
    def setUpScalajs(p: Project): Project = {
      val platform = p.platform match {
        case jsPlatform: Platform.Js =>
          jsPlatform.copy(toolchain = Some(ScalaJsToolchain.apply(this.getClass.getClassLoader)))
        case _ => p.platform
      }
      p.copy(platform = platform)
    }

    val configDir = TestUtil.getBloopConfigDir("cross-test-build-scalajs-1.x")
    val logger = bloop.logging.BloopLogger.default(configDir.toString())
    TestUtil.loadTestProject(configDir, logger, true, _.map(setUpScalajs))
  }
```
@tgodzik do you have any neat trick to make this work?

KristianAN: Actually I think the solution here is just to move these tests into frontend

KristianAN:
Chages introduced in PR:
  - new --link flag for bsp compile, which tries to link the given projects, compiles jvm projects like normal if a jvm project is supplied
  - Removed dependency on frontend in all bridges
  - Moved CrossPlatform.scala to shared
  - Moved all bridge tests to frontend
  - frontend local publishes all bridges before running tests
  - New tests for linking projects through bsp
  - New version of bloop-config that allows clients to specify which type of native project is being linked (application, dynamiclibrary, staticlibrary)
  - Changes made to linking logic that now allows for linking without specifying a main class. This is mainly useful for ScalaJS projects with JSExport and for scala-native projects using dynamicLibrary or staticLibrary application type.
  
Overall This should bring bloop much closer to sbt/mill and the likes with regards to linking capabilities. There is probably still more work we can do to better the client capabilities when it comes to passing configuration, but I think we can add that later. Hopefully this PR will let us see more linking through bloop so we can discover possible bugs etc.

KristianAN:
One thing I forgot to add is that I have not really done much with the scalajs bridge for 0.6. I have not added tests, or decoupled it from frontend like the others. I highly doubt there are many projects using bloop for such an old version. It should work as it did before however.

build.sbt
```diff
@@ -366,7 +369,7 @@ lazy val jsBridge06 = project
 
 val jsBridge1Name = "bloop-js-bridge-1"
 lazy val jsBridge1 = project
-  .dependsOn(frontend % Provided, frontend % "test->test")
+  .dependsOn(bloopShared % Provided /* frontend % "test->test" */ )
```
tgodzik, line 372: if no longer needed we can remove the test dependency

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
```diff
@@ -421,6 +424,90 @@ final class BloopBspServices(
...
+      originId: Option[String],
+      logger: BspServerLogger
+  ): BspResult[bsp.CompileResult] = {
+    import bloop.engine.tasks.LinkTask.{linkJS, linkNative}
+
+    val isRelease = compileArgs.exists(_ == "--release")
+    val isDebug = compileArgs.exists(_ == "--debug")
+
+    val overrideLinkerMode = if (isRelease && isDebug) {
+      None
```
tgodzik, line 440: MAybe we should log an error in this case?

tgodzik (changes requested):
Looks good! Just two minor comments, otherwise LGTM

frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
KristianAN: I logged it as a warning as we still run the linking, we just ignore the dynamic --debug or --release arguments.

tgodzik (approved):
Great work! LGTM

Issue #1631, opened 2021-12-21
Resources file not visible in classes directory 


lwronski:
Hello, I have an issue with `bloop run`, that resource files are not copied to the classes directory.

Using this source:
```
$ cat src/main/resources
1 2 3 4
$ cat src/main/scala/Main.scala
import scala.io.Source

object Main extends App {
  val inputs = Source.fromResource("input").getLines.toSeq
  println(inputs.mkString(""))
}
```

With SBT:
```
$ sbt run
...
$ ls target/scala-2.13/classes/
Main$.class                     Main.class
Main$delayedInit$body.class     input
```

`classes directory` contains `input` file

With bloop:

After running `sbt bloopInstall`
```
$ bloop compile hello-world
...
$ ls .bloop/hello-world/bloop-internal-classes/classes-Metals-V0ZUSdL_Q7KLGmw0GcCopw\=\=-NfPLLT32T_C-HitNIi5aLQ\=\=/
META-INF                        Main$delayedInit$body.class
Main$.class                     Main.class
```

but with `bloop` classes directory doesn't contain `input` file

tgodzik:
Thanks for reporting! We might not be correctly copying the results to the client class directories.

Arthurm1:
As far as I understand - Bloop intentionally doesn't copy resources into the classes dir.  Instead it puts the resources dir on the classpath (and expects it to be on the classpath of dependent projects).  I guess this is an optimisation to save on file copying.  This works fine for compile/test/run but not packaging.  It might be easier to change the package code to include the resources dir.

tgodzik:
Thanks for pointing that out! Bloop wasn't doing any packaging yet, so that would explain it.

kpodsiad:
@tgodzik It's worth noting that the issue is not about packaging, @lwronski provided an example with `sbt run` which compiles and runs code.

For me, the main issue is that once compiled stuff shouldn't change its behavior. If bloop puts _**current** **resources**_ from the working directory on the classpath then after compilation I may do some changes in resources. Suddenly the program is working differently and it could be very hard to spot why.

Every build tool which I know is copying resources and I expect that Bloop behaves similarly.

crt-31:
Bump. This is still an issue. 

I agree Bloop should copy the files into the classpath directory.
Also, the classloaders only support jarfiles and directories on the classpath, so if I put a.txt as a resource, bloop currently adds it to the classpath which is incorrect.
[...]

PR #2553, opened 2025-01-02
bugfix: Don't add single files to classpath

tgodzik:
Only jars and directories are excepted. In case of resources it will most likely be always a directory.

### DIFF
--- a/backend/src/main/scala/bloop/BloopClassFileManager.scala
+++ b/backend/src/main/scala/bloop/BloopClassFileManager.scala
@@ -201,6 +201,34 @@ final class BloopClassFileManager(
 
   }
 
+  private def copyResources(
+      resources: List[AbsolutePath],
+      copyTo: AbsolutePath,
+      config: ParallelOps.CopyConfiguration
+  ): Task[Unit] = {
+    val (singleFiles, classpathEntries) =
+      resources.partition(path => path.exists && path.isFile)
+
+    val singleFilesToCopy =
+      for (file <- singleFiles)
+        yield file.underlying -> copyTo.underlying.resolve(file.underlying.toFile().getName())
+
+    val classpathEntriesCopy =
+      for (entry <- classpathEntries) yield {
+        ParallelOps
+          .copyDirectories(config)(
+            entry.underlying,
+            copyTo.underlying,
+            inputs.ioScheduler,
+            enableCancellation = false,
+            inputs.logger,
+            singleFilesToCopy
+          )
+      }
+
+    Task.gatherUnordered(classpathEntriesCopy).map(_ => ())
+  }
+
   def complete(success: Boolean): Unit = {
 
     val deleteAfterCompilation = Task { BloopPaths.delete(AbsolutePath(backupDir)) }
@@ -220,7 +248,7 @@ final class BloopClassFileManager(
             val config =
               ParallelOps.CopyConfiguration(5, CopyMode.ReplaceExisting, Set.empty, Set.empty)
 
-            ParallelOps
+            val copyClassFiles = ParallelOps
               .copyDirectories(config)(
                 newClassesDir,
                 clientExternalClassesDir.underlying,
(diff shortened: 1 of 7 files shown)


crt-31:
Suggest throwing an error (or perhaps a warning) if you filter something out

adpi2 (commented):
I agree with @crt-31 that a warning would be nice to have here.

tgodzik:
> I agree with @crt-31 that a warning would be nice to have here.

I thinking that maybe we could just create a temp directory instead, so that would work nicely. I will work on that later

tgodzik:
Actually, it might be easier to just copy those resources, let me try work on that.

crt-31:
> Actually, it might be easier to just copy those resources, let me try work on that.

@tgodzik, are you then planning to edit the config format so that it has a way to map the resources to specific places in the 'jar'?  (I hope so :))

adpi2:
> @tgodzik, are you then planning to edit the config format so that it has a way to map the resources to specific places in the 'jar'? (I hope so :))

Do we have any build tool that supports this mapping from files to resource path? If not, then I don't think we need it in Bloop. In the original issue (#1631) it seems that sbt puts the resource file at the root of the jar.

backend/src/main/scala/bloop/BloopClassFileManager.scala
```diff
@@ -201,6 +201,38 @@ final class BloopClassFileManager(
...
 
+  private def copyResources(
+      resources: List[AbsolutePath],
+      copyTo: AbsolutePath,
+      config: ParallelOps.CopyConfiguration
+  ): Task[Unit] = {
+    val (singleFiles, classpathEntries) =
+      resources.partition(path => path.exists && path.isFile)
+
+    val singleFilesCopy = Task {
```
tgodzik, line 212: TODO: Let's add this to ParallelOps to take advantage of locking we have

frontend/src/test/scala/bloop/RunSpec.scala
```diff
@@ -474,6 +475,42 @@ class RunSpec extends BloopHelpers {
...
+  def runSeesSingleFileResources(): Unit = {
+    TestUtil.withinWorkspace { workspace =>
+      object Sources {
+        val `a/A.scala` =
+          """/a/A.scala
+            |object A {
+            |  def main(args: Array[String]): Unit = {
+            |    val res = getClass.getClassLoader.getResourceAsStream("resource.txt")
+            |    val content = scala.io.Source.fromInputStream(res).mkString
+            |    //assert("goodbye" == content)
```
tgodzik, line 488: TODO: This is commented, let's make sure it works

crt-31:
> Do we have any build tool that supports this mapping from files to resource path? If not, then I don't think we need it in Bloop. In the original issue (#1631) it seems that sbt puts the resource file at the root of the jar.

I'll try sbt later today, but I could have sworn it allowed single file resource mapping.

backend/src/main/scala/bloop/BloopClassFileManager.scala
```diff
@@ -201,6 +201,34 @@ final class BloopClassFileManager(
 
   }
 
+  private def copyResources(
+      resources: List[AbsolutePath],
+      copyTo: AbsolutePath,
+      config: ParallelOps.CopyConfiguration
+  ): Task[Unit] = {
+    val (singleFiles, classpathEntries) =
+      resources.partition(path => path.exists && path.isFile)
```
adpi2, line 210: Are we sure there is no jar in here?

adpi2 (approved):
LGTM

backend/src/main/scala/bloop/BloopClassFileManager.scala
tgodzik: Even if there is, we want to copy it since it's set as a resource. Though, I have never seen a case like this.

adpi2: Okay that's fair

PR #2567, opened 2025-01-11
Better support for linking SN

KristianAN:
Includes new options for SN linking with new configuration options added to bloop-config. blocked by PR https://github.com/scalacenter/bloop-config/pull/104 in bloop-config.

One thing I do not understand at all here is that  in the tests linking with SN version 0.5 takes 5+ times longer than SN (often timing out the tests). This makes no sense to me. If anyone has any good idea about why that could be I would be very happy (it might just be some config in the generated project?). 

This probably needs the eyes of someone with a bit more intimate knowledge than me of SN


tgodzik (approved):
LGTM!

PR #2581, opened 2025-01-21
docs: Add release notes for Bloop 2.0.8

tgodzik:


tgodzik:
Ach, maybe we should also do https://github.com/scalacenter/bloop/pull/2582

tgodzik:
Will go ahead to include in new scala cli

PR #2597, opened 2025-02-04
Best-effort: fix issue with frequent file saves leading to leftover errors

jchyb:
This was caused by the custom hashing mechanism used for best effort compilations where we would needlessly hash the inputs twice - once before compilation (to see if it is necessary), and once after (to save the results). If a file was resaved during the compilation, by the end of it the hash input would be outdated and unreleted to the hash output, so bloop would cease recompilations until next change in the source file, leading to incoherent, leftover errors.

So now, we decouple the output hashing from the input hashing, and only hash the outputs after the compilation, and inputs always before the compilation. In theory, this could cause an unnecessary recompilation if a file is re-saved after reading inputs, but before compiling, leading to a needless compilation, but this is the case where too much is better then not enough.

I am no longer able to reproduce the leftover autosave errors (https://github.com/scalameta/metals/issues/6628#issuecomment-2543169442) after the changes here. Outside of that I am unsure if there is a good way to test it here, so for now I did not add any unit tests.

### DIFF
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -387,7 +387,7 @@ object Compiler {
 
     // Manually skip redundant best-effort compilations. This is necessary because compiler
     // phases supplying the data needed to skip compilations in zinc remain unimplemented for now.
-    val noopBestEffortResult = compileInputs.previousCompilerResult match {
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM! thanks for looking into this!

Issue #821, opened 2019-02-24
NoSuchFileException when deleting cached jars


olafurpg:
Steps to reproduce
- `sbt bloopInstall` a project
- compile works as expected
- `rm -rf` the cache location for downloaded jars that are referenced in `.bloop/*.json` files
- `bloop compile proje<TAB>`

```
$ bloop compile fron<TAB>
java.nio.file.NoSuchFileException: /Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.10.7/scala-library-2.10.7.jar
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
	at sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
	at sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:144)
	at java.nio.file.Files.readAttributes(Files.java:1737)
	at bloop.ScalaInstance.$anonfun$hashCode$1(ScalaInstance.scala:65)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:237)
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:39)
	at scala.collection.TraversableLike.map(TraversableLike.scala:237)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:230)
	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
	at bloop.ScalaInstance.<init>(ScalaInstance.scala:61)
	at bloop.ScalaInstance$.newInstance$1(ScalaInstance.scala:109)
	at bloop.ScalaInstance$.$anonfun$apply$7(ScalaInstance.scala:114)
	at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1660)
	at bloop.ScalaInstance$.apply(ScalaInstance.scala:114)
	at bloop.data.Project$.$anonfun$fromConfig$1(Project.scala:110)
[...]

PR #2604, opened 2025-02-12
bugfix: Redownload artifacts if missing

tgodzik:
Fixes https://github.com/scalacenter/bloop/issues/821

This might cause loading configuration a bit, but that should not be an issue compared to potential user trouble when the artifacts are missing.

The alternative is to only react to missing main scala jars, which would indicate that a cache was cleaned, but not if it was deleted only partly

### DIFF
--- a/backend/src/main/scala/bloop/DependencyResolution.scala
+++ b/backend/src/main/scala/bloop/DependencyResolution.scala
@@ -16,7 +16,12 @@ object DependencyResolution {
    * @param module       The module's name.
    * @param version      The module's version.
    */
-  case class Artifact(organization: String, module: String, version: String)
+  case class Artifact(
+      organization: String,
+      module: String,
+      version: String,
+      classifier: Option[String] = None
+  )
 
   /**
    * Resolve the specified modules and get all the files. By default, the local Ivy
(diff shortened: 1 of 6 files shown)


adpi2 (commented):
I am not sure about this. Cleaning the cache is not something we can or should do blindly. When we decide to do it anyway, for whatever reason, maybe it is reasonable that we have to re-import each project.

Also technically I have a few concerns:
- Resolving each artifact one after the other can be very slow: Coursier needs to do the conflict resolution of transitive dependencies again and again. Can we bypass the resolution and directly download the missing JAR? Or, should we do one resolution of all the missing libraries at once?
- We don't even know if Coursier is going to put the JAR in the right folder. Maybe the project does not use Coursier, or it has some custom cache configuration. Should we check the output files and copy them if necessary?

adpi2:
> And sometimes those artifacts are removed by the build tool like in case of gradle.

Do you know why Gradle does such thing?

tgodzik:
I think it was local cache inside the project, but not sure if that is a standard behaviour even

adpi2:
> I think it was local cache inside the project, but not sure if that is a standard behaviour even

Then I think it would be weird for Bloop to write files inside Gradle local cache.

Maybe it is safer for Bloop not to interact too much with project dependencies. So I would opt for a better error message.

tgodzik:
> > I think it was local cache inside the project, but not sure if that is a standard behaviour even
> 
> Then I think it would be weird for Bloop to write files inside Gradle local cache.
> 
> Maybe it is safer for Bloop not to interact too much with project dependencies. So I would opt for a better error message.
> 
> Although I would not be mad if you disagree :)

Not super sure if I disagree 😅 Potentially we could only do it for things in coursier cache, but yeah it all kind of falls apart

PR #2606, opened 2025-02-13
feat: add sbt.testing.Framework implementation for TestNG on classpath when needed

kasiaMarek:
connected to: https://github.com/scalameta/metals/issues/6826
Should bump mill version after https://github.com/com-lihaoyi/mill/pull/4506 to get single test cases support.


kasiaMarek:
Not sure if this is the best approach or if the it should be added to the classpath only for running test... But then I'm not sure where to best cache the result.
CC: @tgodzik

tgodzik (changes requested):
> Not sure if this is the best approach or if the it should be added to the classpath only for running test... But then I'm not sure where to best cache the result.

I think we are fine not caching it for now, this will only show up now when TestNG is used, right? One thing we could potentially do is not resolve it every time, so we can use `val provider = BloopComponentCompiler.getComponentProvider(Paths.getCacheDirectory("testng"))` similar to SemanticDBCache though a bit easier.

Could you add a test case to BspTestSpec ?

tgodzik (approved):
LGTM!

Issue #2414, opened 2024-08-29
disable noise when printing test results


svalaskevicius:
I spent a while to figure out how to disable noisy specs2 results duplication only to find that it's done by bloop: https://github.com/scalacenter/bloop/blob/34840d485a56abc9d388796392df46151b3ac93b/frontend/src/main/scala/bloop/testing/LoggingEventHandler.scala#L103C16-L135

is there a way to get rid of it (an option or smthg)?

Thanks!

svalaskevicius:
to explain, the result like from the below is already better:

```
bloop test ... | awk '/===============================================/{h=!h} {if (!h) print $0}' 
```

tgodzik:
Thanks for reporting! Could post how it looks for you right now and how you would want it to look like?

I am not sure I will have time to work on it, but it is worth to put as much information as possible. Besides that we are always open to a contribution.

svalaskevicius:
hi, right now, it prints all specs2 output, and then repeatedly prints errors and failures, but the bloop failure printing is inferior as it omits the line numbers.

when there is more than a few errors, I have to always scroll past the bloop failure printout, to find what specs2 reported it - to find the line numbers..

I don't think bloop's failure reprinting adds much value, and it could simply be removed - the test frameworks print their outputs - don't they all? essentially that's what the hacky awk snippet is doing there

tgodzik:
you might be right here, I think we can remove it.

PR #2636, opened 2025-03-14
Formatting test output based on screen length of 80

tambonbon:
This PR is made from reading issue #2414 .
"tput cols" is number of columns (width) in the terminal window. In Scala it seems to always return 80 (I tried it in scala-cli). So this PR is to solve the TODO in LoggingEventHandler.


frontend/src/test/scala/bloop/TestSpec.scala
```diff
@@ -91,10 +96,10 @@ object JsTestSpec extends BaseTestSpec("test-projectJS-test", "cross-test-build-
        |1 tests, 1 passed
        |All tests in hello.UTestTest passed
        |
-       |===============================================
+       |$delimiter
```
tgodzik, line 99: This needs a string interpolation at the start of the multiline string, that's why the tests are failing

frontend/src/main/scala/bloop/testing/LoggingEventHandler.scala
```diff
@@ -132,7 +133,15 @@ class LoggingEventHandler(logger: Logger) extends BloopTestSuiteEventHandler {
       }
     }
 
-    logger.info("===============================================")
+    logger.info(delimiter)
+  }
+
+  def getTerminalWidth: Int = {
```
tgodzik, line 139: We should put it as a value in an object, otherwise we'll be invoking this a lot.

I also wonder if it's really what we want, if we have a terminal for an entire screen, we would get a lot of unneccessary additional '='

Maybe instead we should make it match the longest part of the produced string? OR half a screen?

tgodzik (changes requested):

frontend/src/main/scala/bloop/testing/LoggingEventHandler.scala
tambonbon: Hi excuses for the late response.
I think it's pre-set to have length 80 regardless. The change is merely to reduce the manual repeat ====== in the test cases

frontend/src/test/scala/bloop/TestSpec.scala
tambonbon: Fixed.

tgodzik (approved):
LGTM! Thanks!

PR #2671, opened 2025-05-13
add test that tries to reproduce an issue

wiacekm:
tests to check if we can reproduce an issue


frontend/src/test/scala/bloop/util/TestUtil.scala
```diff
@@ -661,4 +661,19 @@ object TestUtil {
       DependencyResolution.Artifact("org.testng", "testng", "7.9.0")
     DependencyResolution.resolve(List(testNG), logger)
   }
+
+  def getMunitDep(logger: Logger) = {
+    // TODO: figure out clean way
```
tgodzik, line 666: There is clean way to do it would be to use BuildInfo data from sbt. We should be able to add binary version if it's not there yet.

tgodzik (approved):
Turns out the issue was in the code (though caused by bloop directory structure), but worth adding the test anyway

PR #2680, opened 2025-05-30
Try get rid of CompletableFuture start using TaskApp

wiacekm:
preliminary changes to get rid of CompletableFuture from Cli and use Deffered isntead and start using TaskApp instead of App

One compile test is failing, not sure if end conditions are ok. It would be good to add some tests


frontend/src/main/scala/bloop/Cli.scala
```diff
@@ -44,8 +49,9 @@ object Cli {
       out: PrintStream,
       err: PrintStream,
       props: java.util.Properties,
-      cancel: CompletableFuture[java.lang.Boolean]
-  ): Int = {
+      cancel: Deferred[MonixTask, Boolean],
+      activeCliSessions: Ref[MonixTask, Map[Path, List[CliSession]]]
```
wiacekm, line 53: I am not sure about hits change.

PR #2689, opened 2025-06-11
Follow-up: Add nested tracing for debugging

wiacekm:
Follow-up: Add additional tracing to investigate race conditions

This commit adds more detailed tracing instrumentation, properly nested within the existing structure. It was used to investigate potential race conditions introduced in PR #2680.


tgodzik (approved):
Thanks! Look pretty useful to get some more insights.

PR #2692, opened 2025-06-24
Fix #2678: Handle projects with only resources

### LINKED ISSUE #2678: bloop does not see a dependent project that contains only resources
Hello! I have build.sbt file

```
ThisBuild / scalaVersion := "3.3.6"

lazy val projectA = project.in(file("project-a"))

lazy val projectB = project.in(file("project-b")).dependsOn(projectA)
```

projectA have only one file text.txt in src/main/resources.
I  want to open this file inside projectB.

```scala
import scala.io.Source

@main
def main() = 
  val text = Source.fromResource("text.txt").getLines().mkString
  println(text)
```

It works if I run "sbt projectB/run". But if i run "bloop run projectB", project crashes with error "java.io.FileNotFoundException".
The error disappears if I create an empty class in projectA. I don't want to create a useless class just to make bloop work.
Can you help me solve this problem?
Thank you!

wiacekm:
Bloop did not recognize projects that contain only resources and no Scala or Java sources. This change ensures that such projects are correctly handled by copying resources even when no sources are present.

- Add test case covering resource-only project scenario
- Ensure resources are copied when both Java and Scala sources are missing

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
@@ -113,7 +113,7 @@ object CompileTask {
             compileProjectTracer.terminate()
             Task.now(earlyResultBundle)
           case Right(CopyResourcesOnly) =>
-            val denylist = Set.empty[Path] // TODO: verify if it shouldn't be computed somehow
+            val denylist = Set.empty[Path]
             val config = ParallelOps.CopyConfiguration(5, CopyMode.NoReplace, denylist, Set.empty)
             val copyResourcesTask: Task[Unit] =
               ParallelOps.copyResources(


frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
```diff
@@ -113,6 +112,25 @@ object CompileTask {
           case Left(earlyResultBundle) =>
             compileProjectTracer.terminate()
             Task.now(earlyResultBundle)
+          case Right(CopyResourcesOnly) =>
+            val denylist = Set.empty[Path] // TODO: verify if it shouldn't be computed somehow
```
tgodzik, line 116: I don't think we have anything that we don't want to copy, this is usually useful for invalidated sources. It's safe to set it to empty for this use case

I wonder if we should delete anything no longer available, but that might be a separate thing altogether.

frontend/src/test/scala/bloop/BaseCompileSpec.scala
```diff
@@ -1506,7 +1506,20 @@ abstract class BaseCompileSpec extends bloop.testing.BaseSuite {
...
 
-  test("compiler plugins are cached automatically") {
+  test("resources only downstream project resolved correctly") {
+    TestUtil.withinWorkspace { workspace =>
+      val logger = new RecordingLogger(ansiCodesSupported = false)
+      val build = loadBuildFromResources("resources-only-downstream-project", workspace, logger)
+      val projectB = build.projectFor("projectB")
+      val compiledState = build.state.compile(projectB)
+      val runState = compiledState.run(projectB)
+      println(logger.render)
```
tgodzik, line 1516:
```suggestion
```

tgodzik (changes requested):
Great work! Two minor comments from me

PR #2722, opened 2025-09-09
bugfix: Save initial source hashes when compiling

tgodzik:


backend/src/main/scala/bloop/util/HashedSource.scala
```diff
@@ -0,0 +1,35 @@
...
+import bloop.reporter.Problem
+
+import sbt.internal.inc.HashUtil
+import sbt.internal.inc.PlainVirtualFileConverter
+import xsbti.BasicVirtualFileRef
+import xsbti.FileConverter
+import xsbti.PathBasedFile
+import xsbti.VirtualFile
+
+case class HashedSource(content: Array[Byte], bloopHash: Int, path: Path)
```
tgodzik, line 16: This is the biggest change, we use our own VirtualFile with eagerly read input so that we don't compile files with a hash that doesn't correspond to them

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
```diff
@@ -168,21 +167,22 @@ private final class BloopNameHashing(
...
-      val hashesMap = uniqueInputs.sources.map(kv => kv.source -> kv.hash).toMap
       val sourceChanges = tracer.traceVerbose("source changes") { _ =>
         lookup.changedSources(previousAnalysis).getOrElse {
           val previousSources = previous.allSources.toSet
           new UnderlyingChanges[VirtualFileRef] {
             private val sourceRefs = sources.map(f => f: VirtualFileRef)
-            private val inBoth = previousSources & sourceRefs
-            val removed = previousSources -- inBoth
-            val added = sourceRefs -- inBoth
+            private val inBoth = sourceRefs.filter(previousSources)
```
tgodzik, line 175: Otherwise we get PlainFile from previous sources and we want the current sources

backend/src/main/scala/bloop/Compiler.scala
```diff
@@ -939,7 +940,7 @@ object Compiler {
       newClassesDir: Path
   ): CompileOptions = {
     // Sources are all files
-    val sources = inputs.sources.map(path => converter.toVirtualFile(path.underlying))
+    val sources = inputs.uniqueInputs.sources
```
tgodzik, line 943: We can now just use the sources, which are VirtualFile now

frontend/src/main/scala/bloop/engine/caches/LastSuccessfulResult.scala
```diff
@@ -11,21 +11,18 @@ import bloop.task.Task
 
 import monix.execution.atomic.AtomicInt
 import xsbti.compile.CompileAnalysis
-import xsbti.compile.FileHash
 import xsbti.compile.MiniSetup
 import xsbti.compile.PreviousResult
 
 case class LastSuccessfulResult(
-    sources: Vector[UniqueCompileInputs.HashedSource],
```
tgodzik, line 19: This was unused, so no need to keep in memory especially that we now have also file contents

backend/src/main/scala/bloop/util/HashedSource.scala
rochala: I'd double check whether it should be a case class. BasicVirtualFileRef implements an equals method

backend/src/main/scala/bloop/util/HashedSource.scala
```diff
@@ -0,0 +1,43 @@
...
+  override def toPath(): Path = path
+
+  override def input(): InputStream = {
+    new ByteArrayInputStream(content)
+  }
+  override def toString: String = s"VirtualSourceFile($Problem.id@${bloopHash})"
+}
+
+object HashedSource {
+  val converter: FileConverter = PlainVirtualFileConverter.converter
```
rochala, line 42: This can be improved, because FileConverter is used to convert 
VirtualFileRef into VirtualFile, and HashedSource implement boths. Here you will probably create new instance on each conversion while you can just upcast

backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
```diff
@@ -168,21 +167,22 @@ private final class BloopNameHashing(
...
             val (changed, unmodified) = inBoth.partition { f =>
               // We compute hashes via xxHash in Bloop, so we adapt them to the zinc hex format
-              val newStamp = hashesMap
-                .get(f)
-                .map(bloopHash => BloopStamps.fromBloopHashToZincHash(bloopHash))
-                .getOrElse(BloopStamps.forHash(f))
+              val newStamp = f match {
+                case hashed: HashedSource => BloopStamps.fromBloopHashToZincHash(hashed.bloopHash)
+                case other =>
+                  log.debug(s"Expected file to be hashed previously: ${other}")
```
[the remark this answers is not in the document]
rochala: I'd log error if you don't want to throw exceptions

frontend/src/main/scala/bloop/io/SourceHasher.scala
```diff
@@ -150,8 +149,9 @@ object SourceHasher {
         val hashSourcesInParallel = observable.mapParallelUnordered(parallelUnits) {
           (source: Path) =>
             monix.eval.Task.eval {
-              val hash = ByteHasher.hashFileContents(source.toFile)
-              HashedSource(PlainVirtualFileConverter.converter.toVirtualFile(source), hash)
+              val content = Files.readAllBytes(source)
```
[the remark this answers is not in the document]
rochala: If you see performance issues here you can try hashing the content during the read via input stream. Should be faster.

backend/src/main/scala/bloop/util/HashedSource.scala
rochala: Yeah, you can even go step further and
```
        case virtualFile: VirtualFile => virtualFile
```

backend/src/main/scala/bloop/util/HashedSource.scala
tgodzik: Changed to normal class

frontend/src/main/scala/bloop/io/SourceHasher.scala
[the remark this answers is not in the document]
tgodzik: right, we can do that anyway. Changed to save the output at the same time

backend/src/main/scala/bloop/util/HashedSource.scala
tgodzik: done

backend/src/main/scala/bloop/util/HashedSource.scala
```diff
@@ -39,5 +40,17 @@ case class HashedSource(content: Array[Byte], bloopHash: Int, path: Path)
...
 
 object HashedSource {
-  val converter: FileConverter = PlainVirtualFileConverter.converter
+  private val baseConverter = PlainVirtualFileConverter.converter
+  class HashedSourceConverter extends FileConverter {
+    override def toPath(file: VirtualFileRef): Path = baseConverter.toPath(file)
+    override def toVirtualFile(path: Path): VirtualFile = baseConverter.toVirtualFile(path)
+    override def toVirtualFile(ref: VirtualFileRef): VirtualFile = {
+      ref match {
+        case source: HashedSource => source
```
rochala, line 49:
```suggestion
```

HashedSource implements VirtualFile

rochala (approved):
One small nitpick, Can fix it later in the future. LGTM

Issue #1309, opened 2020-06-10
Scala.js module kinds are not being generated by sbt-bloop with Scala.js 1.x


MaximeKjaer:
## Bug description

Running `sbt bloopInstall` on a Scala.js 1.0.1 project generates a Bloop config that has `platform.kind` set to `none`, regardless of the module kind in the sbt config.

## Reproduction

I created a simple project with a small `build.sbt` config:

```scala
lazy val root = project
  .in(file("."))
  .enablePlugins(ScalaJSPlugin)
  .settings(
    scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) },
    scalaJSUseMainModuleInitializer := true
  )
```

The `project/plugins.sbt` file adds the Scala.js 1.0.1 sbt plugin:

```scala
addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.1")
```

Running `sbt bloopInstall` produces `.bloop/root.json` with the following `platform` field:

```json
"platform": {
    "name": "js",
    "config": {
        "version": "1.0.1",
        "mode": "debug",
        "kind": "none",
        "emitSourceMaps": false,
        "jsdom": false,
        "toolchain": [
            
        ]
    },
    "mainClass": [
        
    ]
},
```

<details><summary><code>.bloop/root.json</code></summary>

```json
{
    "version": "1.4.0",
    "project": {
        "name": "root",
        "directory": "/home/maxime/code/tmp",
        "workspaceDir": "/home/maxime/code/tmp",
        "sources": [
            "/home/maxime/code/tmp/src/main/scala-2.12",
            "/home/maxime/code/tmp/src/main/scala",
            "/home/maxime/code/tmp/src/main/java",
            "/home/maxime/code/tmp/target/scala-2.12/src_managed/main"
        ],
        "dependencies": [
            
        ],
        "classpath": [
            "/home/maxime/.sbt/boot/scala-2.12.10/lib/scala-library.jar",
            "/home/maxime/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-js/scalajs-library_2.12/1.0.1/scalajs-library_2.12-1.0.1.jar"
        ],
        "out": "/home/maxime/code/tmp/.bloop/root",
        "classesDir": "/home/maxime/code/tmp/.bloop/root/scala-2.12/classes",
        "resources": [
            "/home/maxime/code/tmp/src/main/resources",
[...]

Issue #1308, opened 2020-06-10
Support Scala.js 1.1.0


MaximeKjaer:
Bloop is currently able to compile and link Scala.js 0.6.x and 1.0.x, but not 1.1.0. It fails with the following error:

```console
> bloop compile root
> bloop link root
[E] Expected compatible Scala.js version [0.6, 1.0], 1.1.0 given
```

Based on the [release notes](https://www.scala-js.org/news/2020/05/18/announcing-scalajs-1.1.0/) of 1.1.0, I think Scala.js 1.1.0 could largely reuse the Scala.js 1.0 bridge.

jvican:
Is Scala.js 1.1.0 not binary compatible with Scala.js 1.0? If they are both binary compatible, we should just upgrade to Scala.js 1.1.0.

jvican:
I've just checked the release notes as they have the answer to my question. I think it is safe to just upgrade the Scala.js version in the 1.x bridge series.

sjrd:
At least it won't break anything that is not already broken today.

But the way bloop handles its dependency on the Scala.js linker has serious flaws already now:
* It should not depend statically on a version of scalajs-linker. It can statically depend on a version of scalajs-linker-interface, but should load the correct scalajs-linker depending on the `scalaJSVersion` used by the project. Problems with the current scheme include:
  - issues like this one will keep popping up every time there is a new version of Scala.js
  - if the project uses an older version of Scala.js, the linker used by bloop will be different than the one used by sbt/Maven/the original build tool, resulting in different produced .js files depending on whether bloop is used or not
* It should not outsmart the build tool's actual config of `scalaJSLinkerConfig` and `scalaJSModuleInitializers`, instead it should just take the value of those two settings from the build and roll with those. bloop currently reconstructs its own `StandardConfig` and `Seq[ModuleInitializer]` based on some arbitrary config that is specific to bloop, and is not in sync with what the original build tool does. Problems with the current scheme include:
  - #1304 Builds without a main class, but it generalizes to builds with arbitrary `scalaJSModuleInitializers`
  - #1309 and #715 mishandling of module kinds (including currently no support for ES modules at all)
  - any other configuration of `scalaJSLinkerConfig` done in users' builds and that is not handled by bloop

In general, a Scala.js codebase is entirely defined by (per sbt Configuration / Mill project):

* the `fullClasspath`
* the value of `scalaJSLinkerConfig` and `scalaJSModuleInitializers`
[...]

PR #2745, opened 2025-09-30
Support WebAssembly: Correctly detect Scala.js 1.x scalaJSLinkerConfig#...

tanishiking:
This PR supports `useExperimentalWebAssembly` option in Scala.js (based on https://github.com/scalacenter/bloop-config/pull/129).

---

Also, fixes https://github.com/scalacenter/bloop/issues/1309

The sbt-bloop integration was failing to detect the
`scalaJSModuleKind` for projects using Scala.js 1.x. This was
because the previous implementation relied on old API
that have changed in Scala.js 1.x, through reflection.
This commit updates the `findOutScalaJsModuleKind`
to use the new Scala.js 1.x linker interface APIs
through reflection.

Now, sbt-bloop loses Scala.js ~0.6 supports in detecting `moduleKind`,
but I guess it's no problem since Scala.js is already 1.21 (?)

Note that:
As commented in https://github.com/scalacenter/bloop/issues/1308#issuecomment-645263171
current implementation isn't ideal. It should statically depend on `scalajs-linker-interface` instead of dynamically accessing those APIs.

> It can statically depend on a version of scalajs-linker-interface, but should load the correct scalajs-linker depending on the scalaJSVersion used by the project. Problems with the current scheme include:
> issues like this one will keep popping up every time there is a new version of Scala.js
> if the project uses an older version of Scala.js, the linker used by bloop will be different than the one used by sbt/Maven/the original build tool, resulting in different produced .js files depending on whether bloop is used or not

### DIFF
--- a/bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
+++ b/bridges/scalajs-1/src/main/scala/bloop/scalajs/JsBridge.scala
@@ -81,6 +81,7 @@ object JsBridge {
         .withModuleKind(scalaJSModuleKind)
         .withSourceMap(config.emitSourceMaps)
         .withMinify(isFullLinkJS)
+        .withExperimentalUseWebAssembly(config.useWebAssembly)
 
       (config.kind, scalaJSModuleKindSplitStyle) match {
         case (ModuleKindJS.ESModule, Some(value)) =>
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -620,8 +620,16 @@ object Interpreter {
                     None
                   )
                   .flatMap { state =>
+                    // TODO: We should use `jsEnvInput` to obtain the JS files to run
+                    // (or even use https://github.com/scala-js/scala-js-js-envs to run ideally),
+                    // https://github.com/scala-js/scala-js/blob/6a145af4dc575340a40b80459d1bf15184c3a2da/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala#L252-L255
+                    // and get command line options from `jsEnv` tasks for Node.js options.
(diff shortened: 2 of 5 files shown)


tanishiking:
Hmm, it looks like `bloop.LInkSpec` is failing, but not sure where can we fix it.

```
X bloop.LinkSpec.can run scalaJS project 881ms 
  utest.AssertionError: assert(resultingState.status.isOk)
resultingState: bloop.engine.State = State(Build(/Users/runner/work/bloop/bloop/frontend/target/scala-2.12/test-classes/cross-test-build-scalajs-1.x/bloop-config,List(RawProject(test-project-test), RawProject(test-project), RawProject(cross-test-build-scalajs-1-x-test), RawProject(test-projectJS), RawProject(cross-test-build-scalajs-1-x), RawProject(test-projectJS-test)),None),PrettyPrintedResultsCache(
...
```

And I couldn't reproduce in my local environment, because when I run `sbt frontend/test`, it fails at the other point and can't reach to the LinkSpec. (even in `main` branch).

```
sbt> test
...
[error] Failed to generate bloop config for resource project: /Users/tanishiking/ghq/github.com/tanishiking/bloop/frontend/src/test/resources/compiler-plugin-allowlist.
[error] (frontend / Test / resources) Failed to generate bloop config for resource project: /Users/tanishiking/ghq/github.com/tanishiking/bloop/frontend/src/test/resources/compiler-plugin-allowlist.
```

tgodzik:
Ok, looks like the failure is actually valid. `bloop.LinkSpec.can run scalaJS project` is failing on. linux also, which normally doesn't happen

tanishiking:
Okay, this is happening because the `target` directory is created under `frontend/src/test/resources/target` when I launched `sbt` at `frontend/src/test/resources` directory for some reason, and failed running at https://github.com/scalacenter/bloop/blob/bcf77a703def20f66f46f771c9894485d1c5f42d/project/BuildPlugin.scala#L289-L331
After deleting unnecessary directories, I could reproduce the tests.

Issue #1578, opened 2021-10-05
Bloop should delete domain socket files used for BSP communication


alexarchambault:
When opening a BSP connection with Bloop via `bloop bsp --protocol local --socket /path/to/socket`, Bloop doesn't delete the socket file when it stops listening on it.

Even though its path is provided by clients, the socket file itself is created by Bloop. So I assume it should be deleted by Bloop itself, when it stops listening on it.

arnavsharma990:
@tgodzik  I’ve opened a PR that addresses this issue by cleaning up the local BSP UNIX
domain socket on shutdown and on JVM termination.

PR: https://github.com/scalacenter/bloop/pull/2832
Refs #1578

jozanek:
This looks resolved. #2832 (merged 2026-01-14, shipped in v2.0.19) makes the BSP server delete the socket file when it stops listening, plus a shutdown hook for abnormal termination. The issue stayed open because that PR said `Refs #1578` rather than `Fixes #1578`.

Verified against v2.1.0 with a real BSP session over `bloop bsp --protocol local --socket …`:

| Scenario | Socket file |
|---|---|
| `build/shutdown` + `build/exit` | removed |
| client drops the connection without shutting down | removed |
| `bloop bsp` killed before any client connects | still present, but the server is still listening on it — removed as soon as a client connects and disconnects |

The only case that still leaves a genuinely stale file is the bloop server being `SIGKILL`ed, where no shutdown hook can run; `ServerHandle.UnixLocal` then fails to bind on that path. I've opened #3069 to unlink a stale socket file before binding, along with regression tests for that and for the #2832 behaviour.

Happy for this to be closed.

arnavsharma990:
sure @jozanek , i will open a follow up pr to resolve that and add regression tests

PR #2832, opened 2026-01-06
Delete local BSP socket file on shutdown

arnavsharma990:
When running BSP with `--protocol local`, Bloop creates a UNIX domain socket file
but did not remove it on shutdown, leaving stale sockets that block subsequent
BSP sessions.

This change ensures the socket file is deleted when the BSP server stops
listening and also registers a JVM shutdown hook to handle abnormal termination.

Testing:
- Compiled backend successfully


arnavsharma990:
Refs #1578

PR #2834, opened 2026-01-06
Implement resource mapping support (#2592)

### LINKED ISSUE #2592: Need to add support for file mapping of resources
Bloop should add support for file mapping of resource files (directories and single files) so they are available at custom 'jar' locations at runtime.  Essentially be able to do something like make the file at  src/main/resources/myfile.txt  be available as a jar resource at /custompath/myfilerenamed.txt. 

Rationale: 
SBT and other Build Tools support this mapping of files to specific and arbitrary places into the jar. SBT does it using the "mapping" field; see https://www.scala-sbt.org/1.x/docs/Mapping-Files.html.  If Bloop supported this, then the resources would be available in proper places during bloop run, bloop debug and bloop test.

Adding this functionality will make bloop more comprehensive and correct during bloop run/debug/test.

Suggested implementation:
* In the bloop project.json file schema, add a resourceMapping field that allows you map from source (file or dir) to a jar-relative location.
*  Sbt-bloop should be able to read sbt's "mapping" field and put it into the project.json
*  At bloop run/test/debug, bloop would probably need to copy the files into the classes directory.  (But it should be able to use its change detection to see if the original file changed to reduce copying).  
* But, it would only need to do the copy during run/test/debug (not necessarily , and probably ideally not, for every compile)

My main desire is for Bloop to handle single-file arbitrary mapping, which we currently rely on.

Note: This is kind of related to #1631.

krrish175-byte:
## Description
This PR implements support for file mapping of resources in Bloop, allowing resources to be copied to custom locations in the classes directory (jar-relative paths). This feature enables Bloop to support build configurations similar to SBT's mappings setting, where resources need to be placed at specific paths in the output jar/directory.

## Changes

- **Project Model**: Added resourceMappings field to the Project data structure.
- **ResourceMapper**: Implemented ResourceMapper utility to handle efficient copying of files and directories with validation for duplicate targets and path traversal.
- **Compile Integration**: Integrated resource copying into CompileTask. Mapped resources are copied:
    - During CopyResourcesOnly tasks (when no sources change).
    - As a background task after successful compilation.

## Testing:

- Added ResourceMapperSpec for comprehensive unit testing of the utility.
- Added ResourceMappingIntegrationSpec for end-to-end verification of the feature during compilation.

## Verification:

- **Unit Tests**: Verified recursive directory copying, empty mapping handling, and validation logic.

- **Integration Tests**: Verified that:
    - Single files are copied to the correct target path.
    - Entire directories are recursively copied.
    - Resources are updated even during no-op compilations.

- **Manual Verification**: Confirmed that resources appear in the client-visible classes directory (bloop-bsp-clients-classes).

## Dependencies / Notes:

- This implementation relies on a schema update in bloop-config (to version 2.3.4 or higher) to parse the resourceMapping field from the JSON configuration.

- The parsing logic in Project.scala is currently commented out and marked with a TODO, pending the release and upgrade of bloop-config. The internal infrastructure is fully functional and tested via manual injection in tests.

## Closing Issue:

Closes #2592 

@tgodzik

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
@@ -195,7 +195,8 @@ object CompileTask {
                 ExecutionContext.ioExecutor,
                 bundle.dependenciesData.allInvalidatedClassFiles,
                 bundle.dependenciesData.allGeneratedClassFilePaths,
-                project.runtimeResources
+                project.runtimeResources,
+                project.resourceMappings
               )
             }
 
@@ -258,15 +259,8 @@ object CompileTask {
                       )
 
                   // Copy mapped resources after compilation
-                  val copyMappedResourcesTask =
-                    bloop.io.ResourceMapper.copyMappedResources(
-                      project.resourceMappings,
-                      bundle.clientClassesObserver.classesDir,
-                      logger
-                    )
-
                   val allTasks = Task
-                    .gatherUnordered(List(postCompilationTasks, copyMappedResourcesTask))
+                    .gatherUnordered(List(postCompilationTasks))
                     .map(_ => ())
                     .doOnFinish(_ => Task(compileProjectTracer.terminate()))
 
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -74,7 +74,8 @@ case class CompileInputs(
     ioExecutor: Executor,
     invalidatedClassFilesInDependentProjects: Set[File],
     generatedClassFilePathsInDependentProjects: Map[String, File],
-    resources: List[AbsolutePath]
+    resources: List[AbsolutePath],
+    resourceMappings: List[(AbsolutePath, String)]
 )
 
 case class CompileOutPaths(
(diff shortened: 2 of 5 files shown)


frontend/src/main/scala/bloop/io/ResourceMapper.scala
```diff
@@ -0,0 +1,147 @@
...
+    if (!dir.exists) return false
+
+    val stream = Files.walk(dir.underlying)
+    try {
+      stream.anyMatch(path => Files.getLastModifiedTime(path).toMillis > lastModified)
+    } finally {
+      stream.close()
+    }
+  }
+}
```
tgodzik:
```suggestion
  private def hasDirectoryChanged(dir: AbsolutePath, lastModified: Long): Boolean = {
    if (dir.exists) {

    val stream = Files.walk(dir.underlying)
    try {
      stream.anyMatch(path => Files.getLastModifiedTime(path).toMillis > lastModified)
    } finally {
      stream.close()
    }
  }
  } else {
  false
}
```

frontend/src/main/scala/bloop/engine/tasks/CompileTask.scala
```diff
@@ -245,8 +256,21 @@ object CompileTask {
                         compileProjectTracer,
                         logger
                       )
-                      .doOnFinish(_ => Task(compileProjectTracer.terminate()))
-                  postCompilationTasks.runAsync(ExecutionContext.ioScheduler)
+
+                  // Copy mapped resources after compilation
+                  val copyMappedResourcesTask =
```
tgodzik: We should probably put it alongside the normal copyResources in updateExternalClassesDirWithReadOnly

PR #2855, opened 2026-01-22
Fix input stream piping in Forker (Issue #882)

### LINKED ISSUE #882: Reading input from application doesn't work
To reproduce, create a simple app such as:

```
object Main extends App {

  val x = readInt()
  val y = readInt()
  println(x + y)
}
```

When run and supply numbers (1, 2) you get the following unexpected error:

[E] Exception in thread "main" java.lang.NumberFormatException: For input string: "2"
[E] 	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)

The same does not occur if running with sbt.

(p.s. Loving Bloop)

dorodb-web22:
Fixes #882.

## Problem
Reading input from `stdin` (e.g. using `readInt`) was flaky. The issue was that we were using `InputStream.available()` to poll for data. This method isn't reliable; it often returns 0 even when data is waiting, causing the process to drop input or wait indefinitely.

## Solution
I've replaced the polling mechanism with a standard blocking reader thread.
- **Removed**: The `gobbleInput` loop that relied on `available()`.
- **Added**: A dedicated thread that blocks on a standard `read()` call. As soon as any bytes appear, they are immediately piped to the forked process's standard input.

This ensures input is forwarded immediately on all platforms.

## Verification
- Verified locally that the blocking approach forwards input correctly without flakiness.
- `Forker.scala` changes follow standard I/O patterns.

### DIFF
--- a/frontend/src/test/scala/bloop/Issue882Spec.scala
+++ b/frontend/src/test/scala/bloop/Issue882Spec.scala
@@ -52,4 +52,69 @@ class Issue882Spec extends BloopHelpers {
       )
     }
   }
+
+  @Test
+  def verifyReadLine(): Unit = {
+    object Sources {
+      val `B.scala` =
+        """object Main extends App {
+          |  val l1 = scala.io.StdIn.readLine()
+          |  val l2 = scala.io.StdIn.readLine()
+          |  println(l1 + " " + l2)
+          |}
+        """.stripMargin
+    }
+
+    val logger = new RecordingLogger
+    val structure = Map("B" -> Map("B.scala" -> Sources.`B.scala`))
+    TestUtil.testState(structure, Map.empty) { (state0: State) =>
+      val input = "Hello\nWorld\n"
+      val ourInputStream = new ByteArrayInputStream(input.getBytes(StandardCharsets.UTF_8))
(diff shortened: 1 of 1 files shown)


frontend/src/main/scala/bloop/exec/Forker.scala
```diff
@@ -69,51 +69,36 @@ object Forker {
...
-    }
-
     val runTask = run(
       Some(cwd.underlying.toFile),
       cmd,
       logger,
       opts.env.toMap,
       writeToStdIn = outputStream => {
-        val mainCancellable = goobleInput(outputStream)
+        val thread = new Thread {
```
tgodzik: Are you able to add some tests? I can confirm that this is fixing the issues at hand

dorodb-web22: @tgodzik 

I've added the tests as requested.
The test cases verify that the blocking reader thread approach fixes the stdin readline issue, covering:

1) Input forwarding without data loss
2) No indefinite waiting when reading from stdin
3) Compatibility with forker-realib patterns
 4)Standard I/O patterns work correctly

Could you approve the workflow run so CI can execute the tests? Thanks!

PR #2923, opened 2026-05-07
bugfix: Accept asm jars for compiler jars

tgodzik:


SethTisue:
nitpick: `startsWith("asm-")` would be a little safer, I think

tgodzik:
> nitpick: `startsWith("asm-")` would be a little safer, I think

Isn't there asm jar as well? That's what I understood from your PR 🤔 

Anyway, we should be fine since it's really unlikely that someone would add a jar that might break something and it's not the normal classpath either, but a separate classpath for the compiler.

SethTisue:
> Isn't there asm jar as well? 

yes, but its filename is e.g. `asm-9.9.1.jar`, not plain `asm.jar`

anyway, as you see fit

tgodzik:
> > Isn't there asm jar as well?
> 
> yes, but it's filename is e.g. `asm-9.9.1.jar`, not plain `asm.jar`
> 
> anyway, as you see fit

Ach ok, added! Thanks for explaining

Issue #2566, opened 2025-01-10
Favor NIO over IO to avoid FileSystemExceptions on Windows


alexarchambault:
On Windows, when Bloop compiled a project using a locally published JAR, it seems it still holds an exclusive access to the file. This is a problem when one wants to overwrite the locally published artifact from a build tool. One gets errors like
```text
native.publishLocal java.nio.file.FileSystemException: C:\Users\Alex\.ivy2\local\io.github.alexarchambault.native-terminal\native-terminal\0.0.7-SNAPSHOT\jars\native-terminal.jar: The process cannot access the file because it is being used by another process
    java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
    java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
    java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
    java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:273)
    java.base/sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:104)
    java.base/java.nio.file.Files.delete(Files.java:1152)
    os.remove$all$.apply(FileOps.scala:341)
    os.copy$over$.apply(FileOps.scala:300)
    mill.scalalib.publish.LocalIvyPublisher.$anonfun$publishLocal$2(LocalIvyPublisher.scala:46)
    scala.collection.immutable.List.map(List.scala:247)
    scala.collection.immutable.List.map(List.scala:79)
    mill.scalalib.publish.LocalIvyPublisher.publishLocal(LocalIvyPublisher.scala:44)
    mill.scalalib.PublishModule.$anonfun$publishLocalTask$2(PublishModule.scala:231)
```

Restarting the Bloop server solves that problem, but isn't very handy.

Favoring Java NIO over Java IO seems to solve that problem, like done [here](https://github.com/com-lihaoyi/mill/pull/3764) or [here](https://github.com/coursier/coursier/pull/3120).

Arthurm1:
This might help with semanticdb on Windows as well.  Bloop clashing with Metals....
```
[...]

PR #2932, opened 2026-05-28
Replace usages of `java.io.File` with `java.nio.file.Path` in `bloop-rifle`

vasilmkd:
Related to #2566.

The Scala Plugin for IntelliJ IDEA depends on `bloop-rifle` for managing the bloop BSP server instance. We're in the process of migrating code which uses `java.io.File` to use `java.nio.file.Path` instead.

The changes in this PR are **binary incompatible** as `case class` definitions are being changed. From what I can see, there is no `mima` set up in this repository.

`BloopRifleConfig` has already been changed previously in https://github.com/scalacenter/bloop/commit/4c8fcde4. I guess this happened with a major version bump?

Thanks in advance.

### DIFF
--- a/bloop-rifle/src/main/scala/bloop/rifle/BloopRifle.scala
+++ b/bloop-rifle/src/main/scala/bloop/rifle/BloopRifle.scala
@@ -75,7 +75,7 @@ object BloopRifle {
           config.address,
           bloopJava,
           config.javaOpts,
-          cp.map(_.toPath),
+          cp,
           config.workingDir,
           scheduler,
           config.startCheckPeriod,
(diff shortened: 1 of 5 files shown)


tgodzik:
We mostly treat Bloop as an application, so that's why MiMa is not turned on for the repository, so in reality we have not compat guarantees, didn't seem neccessary.

tgodzik (approved):
LGTM!

tgodzik:
Let me know if you need a new release. We'll probably do 2.1.0 just to mark the changes.

vasilmkd:
A release would be nice, yes. Thanks! 😁

PR #2935, opened 2026-06-02
Serve testClasses/mainClasses from compile gatekeeper after taskFinish

### LINKED ISSUE #1067: Allow `buildTarget/testClasses` after `taskFinish`
BSP clients can make the sensible choice of sending `buildTarget/testClasses` or `buildTarget/mainClasses` BSP requests after they have received `taskFinish` notifications for a given build target. However, Bloop doesn't ensure that the compile products are available until the *whole* compile transaction has been executed. Therefore, there can be race conditions where the clients ask for it before or after the transaction is registered by the bloop compile gatekeeper.

As this behavior doesn't make sense, I'll look into how we can change the internals to ensure fast and reliable results to these BSP requests. /cc @marek1840 who reported this issue and @jastice who might have run into it.

jozanek:
Fixes #1067.

BSP clients send `buildTarget/testClasses` / `buildTarget/mainClasses` right after a per-target `taskFinish`, but sometimes get an empty result. These endpoints serve classes from the per-connection `state.results`, which is only refreshed at the compile request's terminal `saveState` - after `taskFinish` is emitted. So a client reacting to `taskFinish` reads a stale, empty cache.

## Note

This makes the race negligibly rare, not impossible - there's no enforced happens-before between gatekeeper registration and `taskFinish`. A hard guarantee would need sequencing in the compile engine, out of scope here.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
+++ b/frontend/src/main/scala/bloop/engine/tasks/Tasks.scala
@@ -11,6 +11,8 @@ import bloop.data.JdkConfig
 import bloop.data.Project
 import bloop.engine.Dag
 import bloop.engine.State
+import bloop.engine.caches.LastSuccessfulResult
+import bloop.engine.tasks.compilation.CompileGatekeeper
 import bloop.exec.Forker
 import bloop.exec.JvmProcessForker
 import bloop.io.AbsolutePath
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
Makes sense, thanks!

PR #2936, opened 2026-06-02
Forward run stdin via blocking read instead of polling available()

### LINKED ISSUE #882: Reading input from application doesn't work

jozanek:
Fixes #882.

## Root cause

`bloop run` never delivered usable stdin to the forked application, and it broke on two layers:

- **Server side** — `Forker` forwarded stdin with a 50ms poll loop gated on `opts.in.available() > 0`. On the CLI path `opts.in` is nailgun's `NGInputStream`, whose `available()` returns 0 while data is in flight, so the loop never read — and never drove nailgun's demand protocol, so the server never even asked the client for input.
- **Client side** — the nailgun client (the external, unmaintained scala-cli fork of snailgun) read stdin with `readLine()` and forwarded each line *without* its newline, so the forked app's `readInt`/`readLine`/`Scanner` never saw a delimiter and blocked forever. It also NPE'd on EOF (`line.length()` was checked before `line == null`).

## Changes

- **`Forker`**: replace the `available()` poll with a blocking read on a monix `Task` run on `ExecutionContext.ioScheduler` (the same blocking-I/O idiom used by `BspServer`/`TestServer`). This drives nailgun's demand protocol so the server actually requests and reads stdin.
- **Drop the external `snailgun` dependency** (unmaintained, no release since 2023) and replace it with a minimal, bloop-native nailgun client under `bloop.rifle.internal.nailgun`, trimmed to what bloop uses (daemon-thread path only, `BloopRifleLogger` directly). Its stdin forwarder sends each line *with* its newline and handles EOF. Adapted from snailgun (Apache-2.0, attributed in-file).
- Remove the now-unused `BloopRifleLogger.nailgunLogger` adapter.

## Verification

- `RunSpec.canRunApplicationThatReadsStdinWithoutAvailableBytes`: a regression test using an `InputStream` whose `available()` always returns 0 (simulating `NGInputStream`). Fails on the old poll loop (15s timeout), passes on the blocking read.
- End-to-end: `bloop run` of an app doing `println(StdIn.readInt() + StdIn.readInt())` now reads `1`/`2` and prints `3` (previously hung indefinitely).


jozanek:
@tgodzik feel free to rerun, should not be connected to my changes.

frontend/src/main/scala/bloop/exec/Forker.scala
```diff
@@ -69,52 +67,37 @@ object Forker {
...
-        val mainCancellable = goobleInput(outputStream)
-        Cancelable { () =>
-          shutdownInput = true
-          mainCancellable.cancel()
+        /* Block on stdin in a daemon thread and forward bytes to the forked process
+         * as soon as they arrive. We must NOT poll `available()`: on the CLI path
+         * `opts.in` is nailgun's NGInputStream, whose `available()` returns 0 even
+         * while data is in flight, which silently drops input. Mirrors the blocking
+         * `readOutput` threads used below for stdout/stderr. */
+        val thread = new Thread("bloop-forker-stdin") {
```
tgodzik: I wonder whether we could reuse any of the monix utilities instead of using Threads directly.

`ExecutionContext.ioScheduler.scheduleWithFixedDelay(duration, duration) {` wouldn't block as well?

jozanek: updated

tgodzik:
Were you able to confirm that it works when using `bloop run` ? It seems to not read anything at the moment with the example:
```
import scala.io.StdIn

@main def hello(): Unit =
  val x = StdIn.readInt()
  val y = StdIn.readInt()
  println(x + y)
```

jozanek:
Hey Tomek! You are right, the test wasn't sufficient. The issue is in snailgun, but since it is not maintained anymore, I am thinking about moving it completely under bloop (500 LOC) and then try to reduce it as much as possible.

tgodzik:
> Hey Tomek! You are right, the test wasn't sufficient. The issue is in snailgun, but since it is not maintained anymore, I am thinking about moving it completely under bloop (500 LOC) and then try to reduce it as much as possible.

Makes perfect sens, it would be cool to do that.

Edit: Previously I wrote from scala-steward account, since I needed to fix one repo and forgot to switch back 😅

jozanek:
@tgodzik I already did it, see my last commit, so this is now dropping snailgun dependency. Please review 🙏

tgodzik (changes requested):
Looks great, it seems it now works correctly reading from commandline. 

The only issue remaining is that it needs additional input to finish. Even if the program uses just println it will require you now to type enter at the end. Any idea why that started to happen?

jozanek:
Yes, I just fixed it. Please let me know if you see any other issues!

tgodzik (approved):
Looks to be working now! Great work!

PR #2938, opened 2026-06-03
Support JUnit 5 (Jupiter) tests

### LINKED ISSUE #996: Support JUnit 5 tests
Currently there is support for JUnit 4 testing in Bloop and SBT by including `"com.novocode" % "junit-interface" % "0.11"`

This library will run JUnit 5 tests only if they are annotated with `@RunWith(JUnitPlatform.class)`.  Gradle directly supports JUnit 5 so non of my projects' tests are marked with this and therefore no classes are recognised as tests.

There is `"net.aichler" % "sbt-jupiter-interface" % "0.8.2"` which supports JUnit5 without any need for additional annotations but this uses specific code in its SBT plugin (not the Fingerprint implementation) to identify tests and so it doesn't work in Bloop.
The reasoning behind this (stated [here](https://github.com/maichler/sbt-jupiter-interface)) is that SBT (and Bloop) doesn't recognise package private test classes.
I'm not even sure that checking package private classes would work as JUnit 5 allows you to annotate just class methods and it seems that SBT/Bloop purely check for class annotations only - is that correct?

Can this be changed?  So then (hopefully) this library can be changed to then work with Bloop?  I haven't spoken to the library author about this so I may have misinterpreted what's going on.

jozanek:
Closes #996.

Bloop discovers tests through the sbt test-interface fingerprint mechanism, but JUnit 5 can't be found that way: the `com.github.sbt.junit:jupiter-interface` adapter deliberately reports a *fake* fingerprint so fingerprint discovery finds nothing, and does its real discovery in an **sbt plugin** (via the JUnit Platform launcher) — which Bloop doesn't run. As a result, JUnit 5 tests (including method-level `@Test` and package-private classes) were never discovered.

## How

- Bundle `jupiter-interface 0.19.0` and run its `JupiterTestCollector` (the JUnit Platform launcher) to discover Jupiter tests, merging the results into Bloop's normal discovery for **both** the `test` run path and the BSP test-listing path (Metals' test explorer).
- Discovery uses Bloop's **bundled** JUnit Platform launcher/engine; the project's own platform/jupiter jars are kept out of the discovery classloader — otherwise the platform's ServiceLoader sees two `TestEngine` copies and fails with "not a subtype"/duplicate-engine.
- The tests themselves run in the forked JVM with the **project's own** engine. So a project on Jupiter 5.x is discovered by the bundled 6.x launcher and executed by its 5.x engine - one bundle, both lines.
- If a project depends on `junit-jupiter` but lacks the adapter (typical for Maven/Gradle), Bloop auto-resolves and appends `jupiter-interface`, mirroring the existing TestNG handling.

## ⚠️ Requires the build server on JDK 17+

`jupiter-interface 0.19.0` pulls JUnit Platform 6.x, whose baseline is **Java 17**. JUnit 5 *discovery* therefore requires the Bloop **server** to run on JDK 17+. This is guarded: on older JVMs discovery is skipped with a clear warning (no crash). The native CLI is unaffected, and Bloop already de-facto requires JDK 16+ (`UnixDomainSocketAddress`).

## Out of scope / follow-ups

- `scalacenter/bloop-config`: a `Config.TestFramework.Jupiter` constant + `DefaultFrameworks` entry.
- `bloop-maven-plugin` / `gradle-bloop`: emit the Jupiter framework FQN in generated config (Bloop's classpath-based auto-detect mitigates, but the generators should still emit it).

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -11,6 +11,12 @@ Bloop's build server is a long-running process designed to provide the fastest
 compilation possible to users. As such, users are responsible for managing its
 lifecycle and minimizing the amount of times it's restarted.
 
+<blockquote class="grab-attention">
+Discovering and running <b>JUnit 5</b> (Jupiter) tests requires the build server to run on
+<b>JDK 17 or newer</b>, because Bloop discovers them with the JUnit Platform launcher (which targets
+Java 17). The native CLI is unaffected. On older JVMs JUnit 5 discovery is skipped with a warning.
+</blockquote>
+
 ## Start the build server
 
 At the end of the day, the build server is an artifact in Maven Central.
--- a/build.sbt
+++ b/build.sbt
@@ -142,7 +142,8 @@ lazy val frontend: Project = project
       "jsBridge06" -> (jsBridge06Name + "_" + Keys.scalaBinaryVersion.value),
       "jsBridge1" -> (jsBridge1Name + "_" + Keys.scalaBinaryVersion.value),
       "lastSupportedSemanticdb" -> build.SemanticDbSupport.last,
-      "millVersion" -> Dependencies.millVersion
+      "millVersion" -> Dependencies.millVersion,
+      "jupiterInterfaceVersion" -> Dependencies.jupiterInterfaceVersion
     ),
     (run / javaOptions) ++= jvmOptions,
     (Test / javaOptions) ++= jvmOptions,
@@ -159,7 +160,8 @@ lazy val frontend: Project = project
       Dependencies.scalaDebugAdapter,
       Dependencies.bloopConfig,
       Dependencies.logback,
-      Dependencies.libdaemonjvm
+      Dependencies.libdaemonjvm,
+      Dependencies.jupiterInterface
     ),
     // needed for tests and to be automatically updated
     Test / libraryDependencies += Dependencies.semanticdb intransitive (),
--- a/docs/assets/bloop-schema.json
+++ b/docs/assets/bloop-schema.json
@@ -636,7 +636,8 @@
                 "examples": [
                   "org.scalacheck.ScalaCheckFramework",
                   "org.scalatest.tools.Framework",
-                  "org.scalatest.tools.ScalaTestFramework"
+                  "org.scalatest.tools.ScalaTestFramework",
+                  "com.github.sbt.junit.jupiter.api.JupiterFramework"
                 ]
               }
             },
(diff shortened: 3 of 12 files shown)


docs/server.md
```diff
@@ -11,6 +11,12 @@ Bloop's build server is a long-running process designed to provide the fastest
 compilation possible to users. As such, users are responsible for managing its
 lifecycle and minimizing the amount of times it's restarted.
 
+<blockquote class="grab-attention">
+Discovering and running <b>JUnit 5</b> (Jupiter) tests requires the build server to run on
```
tgodzik, line 15: Bloop only runs anyway on JDK 17 and up currently.

jozanek: Yeah, thanks for confirmation. I could only find note about jdk16 apis in the release notes.

tgodzik: To be exact we do support 16 also, but since it's not LTS we omit it.

tgodzik (approved):
LGTM! Seems to work on a test projects I found.

jairovsky:
Thank you guys for this addition. I'm new to/learning Scala and was getting bummed that VSCode/Metals/scala-cli weren't detecting tests based on junit-jupiter.

I've been testing Bloop on the main branch the past few days and it seems be working fine with junit-jupiter tests now, with just a minor problem that Bloop is not yet detecting individual `@Test` methods or `@Nested` classes, so it's not possible to "hit play"/run an individual test from VSCode. I've set up a test project [here](https://github.com/jairovsky/bloop-2938-tests) to illustrate.

However, the issue seems to be on sbt-jupiter-interface's side. [JupiterTestCollector](https://github.com/sbt/sbt-jupiter-interface/blob/ff6dd9ee93eb0c38731f8bad52c5d46a0113a382/src/library/src/main/java/com/github/sbt/junit/jupiter/api/JupiterTestCollector.java#L455-L457) does not traverse through all levels of the TestPlan tree. I'll open an issue there later and see if it can be fixed. Just leaving this comment here for now in case someone else runs into this problem.

Leon0402:
@jairovsky I don't know if there is an issue in bloop, but in metals there is one as well, see https://github.com/scalameta/metals/pull/8457. With this and sbt bsp it now works, with bloop I haven't tested it so far.

tgodzik:
If you mean detecting it in Metals, that should be fixed by the PR https://github.com/scalameta/metals/pull/8457 and Scala CLI itself has ongoing work linked to this PR.

If you mean running `bloop test...` then that would need to be fixed here. Are you having issues running Bloop CLI and testing individual test cases?

jairovsky:
I'm using Metals on VSCode. I'll check the work being done on the Metals PR. Thanks.

PR #2940, opened 2026-06-06
Run BSP targets in the project working directory

### LINKED ISSUE #1210: working directory is not properly set for sub-projects
Hi, 

The change in #1175 doesn't actually change the working directory when you run from a sub-project. Relative paths are still relative to the project root; it's only when they're canonicalized (`getAbsolutePath` for example) that they are reinterpreted as being rooted at `user.dir`. Here is an example.

Here is a little program, in subproject `main` under `modules/main` in my project:

```scala
object Test {

  def main(args: Array[String]): Unit = {
    val cwd = new java.io.File(".")
    println(cwd.getAbsolutePath)
    println(cwd.listFiles().toList)
  }

}
```

When when I run it I see this:

```
itac (cli *)$ bloop run main -m Test
/Users/rnorris/Scala/itac/.
List(./target, ./src)
```

However they don't correspond as expected.

```
itac (cli *)$ ls /Users/rnorris/Scala/itac/.
arch/      build.sbt  modules/   project/   src/       target/    work/
```

Running from sbt gives the expected result.

```
tac (cli *)$ sbt "main/runMain Test"
[info] Loading global plugins from /Users/rnorris/.sbt/1.0/plugins
...
[info] running Test 
/Users/rnorris/Scala/itac/.
List(./.metals, ./.DS_Store, ./target, ./project, ./.gitignore, ./.bloop, ./work, ./arch, ./.git, ./modules, ./build.sbt, ./src)
[success] Total time: 1 s, completed Mar 23, 2020 12:17:00 PM
itac (cli *)$
```

There is a [JDK issue](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4117557) about this and the evaluation begins:

> "user.dir", which is initialized during jvm startup, should be used as an 
informative/readonly system property, try to customize it via command line
-Duser.dir=xyz will end up at implementation dependend/unspecified behavior.

So I think bloop shouldn't do this. Instead is it possible to set cwd in the process builder that's used to for the JVM?

Thanks!

jozanek:
Fixes #1210

### DIFF
--- a/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
+++ b/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala
@@ -1035,7 +1035,7 @@ final class BloopBspServices(
         state: State
     ): Task[State] = {
       import bloop.engine.tasks.LinkTask.{linkJS, linkNative}
-      val cwd = state.commonOptions.workingPath
+      val cwd = project.workingDirectory
 
       parseMainClass(project, state) match {
         case Left(error) =>
--- a/frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
+++ b/frontend/src/test/scala/bloop/bsp/BspProtocolSpec.scala
@@ -5,6 +5,7 @@ import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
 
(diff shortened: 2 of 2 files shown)


tgodzik (approved):
LGTM!

PR #2947, opened 2026-06-06
Combine --cascade with --include-dependencies in test and clean

### LINKED ISSUE #1229: Test's `--include-dependencies` is ignored when `--cascade` is set
Looking at `bloop.engine.Interpreter`, it looks like `--include-dependencies` is ignored when `--cascade` is set in a test run (this is the only reference to `includeDependencies`):  https://github.com/scalacenter/bloop/blob/f69dd3ba74d81e71f6f1ca73a0b557055138d0b2/frontend/src/main/scala/bloop/engine/Interpreter.scala#L342-L346

I don't see why `--cascade` should exclude `--include-dependencies`. This should be either fixed or documented.

jozanek:
Fixes #1229.

- `--cascade` adds reverse dependencies — projects depending on the target.
- `--include-dependencies` adds forward dependencies — projects the target depends on.
- Neither flag implicitly enables the other.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/Interpreter.scala
+++ b/frontend/src/main/scala/bloop/engine/Interpreter.scala
@@ -401,7 +401,16 @@ object Interpreter {
           (userSelectedProjects, projectsToTest)
         } else {
           val result = Dag.inverseDependencies(state.build.dags, userSelectedProjects)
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
LGTM

PR #2948, opened 2026-06-06
Add end-to-end BSP test for debug session environment variables

### LINKED ISSUE #1264: Support environment variables parameters for debug session to start

jozanek:
Adds an end-to-end test for starting a debug session with environment variables through the BSP `debugSession/start` request.

Environment-variable support for debug sessions (#1264) was implemented in #1268, but the only existing coverage builds the debuggee directly via `BloopDebuggeeRunner.forMainClass`, bypassing the BSP JSON layer. The new test in `DebugProtocolSpec` sends `environmentVariables` in the `ScalaMainClass` JSON and drives the real `debugSession/start` endpoint, so a future bsp4s bump or codec change can't silently drop the field.

- Extends the `mainClassParams` test helper to optionally carry arguments, jvm options, and environment variables (defaults keep existing callers unchanged).
- Adds the `"starts a debug session with environment variables"` test, asserting the debuggee observes the env var, jvm property, and program argument.

Closes #1264.

### DIFF
--- a/frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
+++ b/frontend/src/test/scala/bloop/dap/DebugProtocolSpec.scala
@@ -45,6 +45,51 @@ object DebugProtocolSpec extends DebugBspBaseSuite {
     }
   }
 
+  test("starts a debug session with environment variables") {
+    TestUtil.withinWorkspace { workspace =>
+      val main =
+        """|/main/scala/Main.scala
+           |object Main {
+           |  def main(args: Array[String]): Unit = {
+           |    println(args(0))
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
Thanks!

PR #2949, opened 2026-06-07
Support multi-jvm source directories in sbt

### LINKED ISSUE #1280: multi-jvm source directory not added to configuration
sbt-bloop does not recognise multi-jvm sources when using https://doc.akka.io/docs/akka/current/multi-jvm-testing.html.

**To Reproduce**

1. Add multi-jvm plugin in build.sbt
2. Enable MultiJvm plugin for one of the sbt module
3. Add scala file in multi-jvm directory
4. The file will not be compiled

I suspect this has something to do with the way those sources are being added in the multi-jvm plugin, most likely they are somehow not added to sourceDirectories. Current workaround would be to add them manually to sbt via:
```scala
    unmanagedSourceDirectories.in(Compile) += baseDirectory
      .in(ThisBuild)
      .value / "src" / "multi-jvm" / "scala",
```

jozanek:
Closes #1280.

Builds using Akka's [sbt-multi-jvm](https://doc.akka.io/docs/akka/current/multi-jvm-testing.html) plugin keep their sources under a custom `multi-jvm` configuration (`config("multi-jvm") extend Test`). sbt-bloop only wired its settings into `Compile`, `Test` and `IntegrationTest`, so `multi-jvm` had no `bloopGenerate` task and was invisible to `bloopInstall` — those sources were never exported or compiled, and IDEs couldn't see them.

This scopes bloop's `configSettings` into a `multi-jvm` configuration matched by name, so the plugin's config is exported automatically with no `build.sbt` changes — mirroring the existing special-casing of Scala.js / Scala Native. Projects that don't enable the plugin define no products for it, so `bloopGenerate` short-circuits and no target is produced.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,7 @@ jobs:
             sbt-bloop/bsp-enabled \
             sbt-bloop/meta-builds \
             sbt-bloop/no-compile-incremental \
+            sbt-bloop/multi-jvm-configuration \
             sbt-bloop/non-compiling"
         shell: bash
 
--- a/docs/build-tools/sbt.md
+++ b/docs/build-tools/sbt.md
@@ -131,8 +131,13 @@ $ sbt bloopInstall
 ### Enable custom configurations
 
 By default, `bloopInstall` exports projects for the standard `Compile`,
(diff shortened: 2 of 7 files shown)


tgodzik (approved):
LGTM

PR #2950, opened 2026-06-08
Print the about banner for `bloop --version` and `bloop version`

### LINKED ISSUE #1283: bloop --version is broken
Expected: should print the same thing as `bloop about` (which works)
Actual:
```
➜ bloop --version
The handling of `about` does not happen in the `Interpreter`
```

jozanek:
Closes #1283

`bloop --version` printed an internal error instead of version info. Now `bloop --version`, `bloop -v`, `bloop version`, and `bloop about` all print the same banner.


tgodzik (changes requested):
Looks good! Only a small fix in AutoCompleteSpec is needed since we actually have a new method.

tgodzik (approved):
LGTM, rerunning the tests to make sure

PR #2956, opened 2026-06-08
Recompile retained sources when a re-added source reintroduces a package object

### LINKED ISSUE #1347: Compilation results are not updated when the source directory that had been removed is added back
### Environment details
```
$ bloop about
bloop v1.4.3

Using Scala v2.12.8 and Zinc v1.3.0-M4+45-d4354be3
Running on Java JDK v11.0.6 (~/.sdkman/candidates/java/11.0.6.hs-adpt)
  -> Supports debugging user code, Java Debug Interface (JDI) is available.
Maintained by the Scala Center (Jorge Vicente Cantero, Martin Duhem)
```

#### Download the example project
```
ipfs get --output bloop-bug-example QmaRTWDyJp3GeiAQQdLrjSXcE99sn4GVMAJ2QEgDQVXKBv
```
or:
```
curl --silent --output bloop-bug-example.tar.gz 'https://gateway.pinata.cloud/api/v0/get?arg=QmaRTWDyJp3GeiAQQdLrjSXcE99sn4GVMAJ2QEgDQVXKBv&archive=true&compress=true'
mkdir bloop-bug-example
tar -xvf bloop-bug-example.tar.gz -C bloop-bug-example --strip-components=1
```
#### Compile and run for the first time
```
cd bloop-bug-example
bloop run -p hello-world_app
```

*Expected output: 14*

#### Exclude one of the source directories
```
vim .bloop/hello-world_app.json
```
Find the source entry called `"/tmp/fury-testing/src2"`, and remove it from the configuration file.

#### Compile and run for the second time
```
bloop run -p hello-world_app
```
The code will still compile, but the application will behave differently.

*Expected output: 5*

#### Bring back the source directory that was excluded:
```
vim .bloop/hello-world_app.json
```
Edit the configuration file again to restore the source entry that was removed previously.

#### Compile and run for the third time
```
bloop run -p hello-world_app
```
Since the project configuration is now the same as it was during the first run, the output should also be the same, shouldn't it?

*Expected output: 14*
**Actual output: 5**

#### Clean and recompile
The problem disappears if the project is cleaned:
```
bloop clean -p hello-world_app
bloop run -p hello-world_app
```
*Expected output: 14*

jozanek:
Fixes #1347.

## Root cause
Zinc's incremental invalidation is edge-based. When a source is removed, the dependency edges from retained sources onto its symbols are erased, and those retained sources recompile against a fallback. Re-adding the source does **not** restore those edges, so nothing reinvalidates the retained sources — they keep the stale compilation.

This reproduces with stock Zinc in a plain sbt project (delete a `package.scala`, compile, restore it, compile), so the underlying gap is in Zinc. `BloopNameHashing` delegates invalidation precision to Zinc by policy, so a precise fix belongs upstream; this PR is a pragmatic Bloop-side guard for the source-set-churn case #1347 reports.

## Fix
When a recompiled source reintroduces a **package object** absent from the previous analysis, invalidate every class in that package (each has the package object in its implicit scope) and every class that references one of them. Guarded to fire only for added package objects, so ordinary and clean compiles are unaffected.

### Tradeoffs / limitations (documented in code)
- **Conservative:** re-adding a package object recompiles its whole package. Acceptable — source-set churn is rare and correctness outweighs incrementality here.
- **Scala 2 package objects only:** keyed on the `*.package` class name; Scala 3 top-level `$package` wrappers are not matched.
- **Edge-bounded coverage:** a source that imports the package solely for package-object members (without referencing any of the package's classes) holds no recorded edge in the stale analysis and is not detected.

## Test
Adds a regression test mirroring the original report: package `p` with retained `p.A` (same package as the togglable package object `p`) plus a cross-package importer `q.Main`. It runs the app and asserts the user-visible output **28 → 10 → 28** across compile / remove-root / re-add-root. Verified to fail without the fix (**28 → 10 → 10**), so it cannot silently pass while the bug persists.

### DIFF
--- a/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
+++ b/backend/src/main/scala/sbt/internal/inc/bloop/internal/BloopNameHashing.scala
@@ -115,13 +115,23 @@ private final class BloopNameHashing(
               current.apis.internalAPI
             )
           debug("\nChanges:\n" + newApiChanges)
+
+          // Zinc invalidates dependents through recorded dependency edges, but a
+          // source removed from the build and later re-added with unchanged
+          // contents has no such edges from retained sources: they were erased
+          // when it was removed. When a (re-)added source reintroduces a package
+          // object, invalidate the retained sources that reference that package
+          // so their implicit resolution is recomputed against it.
+          val readdedPackageObjectInvalidations =
+            invalidationsFromAddedPackageObjects(invalidatedSources, previous, current)
+
           val nextInvalidations = invalidateAfterInternalCompilation(
             current,
             newApiChanges,
             recompiledClasses,
             cycleNum >= options.transitiveStep,
             IncrementalCommon.comesFromScalaSource(previous.relations, Some(current.relations))
-          )
+          ) ++ readdedPackageObjectInvalidations
           debug(s"Next invalidations: $nextInvalidations")
 
           val continue = lookup.shouldDoIncrementalCompilation(nextInvalidations, current)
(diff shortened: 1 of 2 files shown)


jozanek:
Feel free to rerun failed tests, I verified and it's not related to my changes, but rather flaky tests.

tgodzik (approved):
LGTM

I shortened a bit the comments, I think we should try to keep them shorter overall.

PR #2958, opened 2026-06-09
Change-detect and quiet the automatic sbt meta-build export

### LINKED ISSUE #1357: Bloop generate is automatically called during sbt project loading
I've noticed with the latest metals that when I start sbt (in the sbt project) I see:
```
[info] welcome to sbt 1.4.0-M2 (Azul Systems, Inc. Java 1.8.0_252)
[info] loading settings for project sbt-build-build-build from metals.sbt ...
[info] loading project definition from ${SBT_LOCATION}/project/project/project
[info] loading settings for project sbt-build-build from metals.sbt ...
[info] loading project definition from ${SBT_LOCATION}/project/project
[success] Generated .bloop/sbt-build-build.json
[success] Generated .bloop/sbt-build-build.json
[success] Total time: 0 s, completed Aug 20, 2020 11:30:31 AM
[info] loading settings for project sbt-build from local.sbt,metals.sbt,plugins.sbt ...
[info] loading project definition from ${SBT_LOCATION}/project
[success] Generated .bloop/sbt-build.json
[success] Generated .bloop/sbt-build.json
[success] Total time: 1 s, completed Aug 20, 2020 11:30:32 AM
```
This makes sbt noticeably slower to start up. Without the bloop plugin, I find the sbt project loads in about 9 seconds. With the bloop plugin, it increases to about 11 seconds. I also would prefer not having the success messages displayed during project loading (though they did help make it obvious why startup got slower).

It would be great if the old behavior could be restored where the plugin did not automatically call `bloopGenerate` on load.

jozanek:
Closes #1357.

## Problem

The sbt-bloop plugin exports the sbt **meta-build** (the build definition under `project/`) to bloop on every load, so Metals can compile and navigate `project/*.scala` / `*.sbt`. This makes `sbt` startup slower and prints confusing (and duplicated) `[success] Generated …json` lines on every load.

Fully reverting the auto-export isn't an option (Metals depends on it) so this keeps it on by default but removes the cost and the noise.

## What changed

- **Change detection.** The on-load export is skipped while a fingerprint of the files under `project/` (path + mtime + size, with `target`/`.bloop` subtrees pruned) is unchanged. The fingerprint is stamped in `.bloop` whenever `bloopInstall` exports the meta-build, so a steady-state `sbt` startup is fast and silent, while edits / additions / deletions regenerate. Change detection fails safe: any IO error falls back to exporting, so it can never break load.
- **Quieter logging.** The automatic export now logs at `debug`; a manual `bloopInstall` keeps its usual `[success]` output. Quietness is keyed on the on-load path rather than on "is this a meta-build", so manual runs are unaffected.
- **Opt-out.** New `bloopExportMetaBuild`. Disable globally with `-Dbloop.export-meta-build=false` or `BLOOP_EXPORT_META_BUILD=false` (read at every meta-build layer, so it also covers nested `project/project` builds), or per layer with `Global / bloopExportMetaBuild := false`. Documented in `docs/build-tools/sbt.md`.

## Scope / caveat

The fingerprint tracks files under `project/` only. Inputs read from elsewhere — system properties, environment variables, or files outside `project/` — are not detected; a `bloopInstall` (or any edit under `project/`) forces a refresh. This is documented in the new docs section.

Sorry for long PR, wanted to split/stack it, but wasn't very successful.


integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -909,6 +918,88 @@ object BloopDefaults {
...
+
+  /**
+   * Hash of every file under the meta-build (path, mtime, size), with the
+   * `target` and `.bloop` subtrees pruned so they are never traversed during
+   * load. We fingerprint all files, not just `*.{sbt,scala,java}`, so build
+   * definitions that read versions/settings from e.g. `dependencies.properties`
+   * are still invalidated. Inputs that are not files under `project/` (system
+   * properties, environment variables, …) are out of scope.
+   */
+  private def metaBuildFingerprint(buildBase: File): String = {
```
tgodzik: We shouldn't calculate the fingerprints ourselves, this something that sbt is doing and we should delegate to it if needed.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
```diff
@@ -1279,6 +1372,10 @@ object BloopDefaults {
             )
           }
           throw fail.head
+        } else if (isMetaBuild) {
+          // Stamp the change-detection fingerprint so the next load can skip,
+          // regardless of whether this `bloopInstall` was automatic or manual.
+          writeMetaBuildFingerprint(metaBuildBase)
```
tgodzik: Maybe instead of writting we could use a java property? -Dexport-meta=true or something along those lines? We could add it by default in Metals when running bloopInstall.

tgodzik (changes requested):

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
jozanek: The meta-build export is now opt-in via the `bloop.export-meta-build` system property (default off), so there's nothing to change-detect: a plain `sbt` never exports the meta-build, and when the property is set it always regenerates fresh.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
jozanek: Added `-Dbloop.export-meta-build=true` (default off), so Metals can set it when running `bloopInstall`. I namespaced the property to match the existing `bloop.export-jar-classifiers`. Happy to rename to `export-meta` or whatever you'd prefer. Heads-up for the changelog: any non-Metals consumer that relied on the old default auto-export will now need to set this property.

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
tgodzik: Perfect, thanks!

tgodzik (approved):
LGTM! Thanks!

tgodzik:
Added the property to metals https://github.com/scalameta/metals/pull/8538

PR #2966, opened 2026-06-11
Support sbt internal configurations in exported project dependencies

### LINKED ISSUE #1373: bloopInstall omits dependency for internal dependency
In my project Kiama (https://github.com/inkytonik/kiama) I have a multi-project structure that includes a base project with macros that are used by other sub-projects. I am using the approach from the sbt reference manual (https://www.scala-sbt.org/1.x/docs/Macro-Projects.html, section Distribution) to set it up so that the macro sub-project is published as part of the main sub-project, rather than being published on its own.

I've been using Bloop via VSCode + Metals to develop Kiama for at least a couple of years with no problems, but unfortunately I can't build Kiama in Metals in the current version (Metals 1.93 which uses Bloop 1.4.3-23-550c6c0a). I've also tested the problem outside Metals using Bloop 1.4.4 and I see the same issue.

The issue seems to be that I use an internal dependency of the main project in the macro sub-project as described in the sbt reference manual. When that dependency is used, bloopInstall omits the dependency in the generated project JSON file. As you would expect, this means that the macro sub-project is not built first, so I get lots of spurious errors in Metals. Building the same project using sbt 1.3.13 outside VSCode works fine.

Rather than have you deal with Kiama and its complexity, I've cut the problem down to a simple project ([simple.zip](https://github.com/scalacenter/bloop/files/5228895/simple.zip)). There are two sub-projects util and core with core depending on util. If it's a regular dependency then everything works fine. If it's an internal dependency it fails, so the dependency is omitted from the generated core.json file and I see a compile error in VSCode on the use of the Util object from the main Core object.

jozanek:
Fixes #1373

`bloopInstall` silently dropped inter-project dependencies declared with sbt's internal configurations, e.g. the macro-project pattern from the sbt reference manual:

```scala
val core = project.dependsOn(util % "compile-internal, test-internal")
```

The generated `core.json` had `util`'s classes directory on the classpath, but `util` was missing from `dependencies`, so bloop never compiled it first and users saw spurious "not found" errors in Metals.

sbt resolves these through hidden ivy configurations (`compile-internal` extends `compile`/`optional`/`provided`) that every project has. `projectDependencyName` filtered those names out of the lists handed to `sbt.Classpaths.mapped` and looked the parsed mapping up only by the base configuration name, never by `compile-internal` — the key it is actually stored under — so the mapping came back empty and the dependency was dropped.

Now the internal variant of a supported configuration counts as supported, the mapping is also looked up under `<configuration>-internal`, and dep-side internal names resolve to their base configuration.

### DIFF
--- a/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
+++ b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala
@@ -559,9 +559,10 @@ object BloopDefaults {
    * Creates a project name from a classpath dependency and its configuration.
    *
    * This function uses internal sbt utils (`sbt.Classpaths`) to parse configuration
-   * dependencies like sbt does and extract them. This parsing only supports compile
-   * and test, any kind of other dependency will be assumed to be test and will be
-   * reported to the user.
+   * dependencies like sbt does and extract them. Internal configurations such as
+   * `compile-internal` are resolved like their base configuration because they only
+   * differ in publishing semantics. Any other unsupported dependency will be assumed
+   * to be test and will be reported to the user.
    *
    * Ref https://www.scala-sbt.org/1.x/docs/Library-Management.html#Configurations.
    */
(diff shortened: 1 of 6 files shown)


tgodzik (approved):
LGTM

PR #2967, opened 2026-06-11
Report a crash from the compiler instead of hanging the compilation

### LINKED ISSUE #1434: provide a complete stacktrace when the compiler crashes
I had an issue today where a `java.lang.NoSuchMethodError` was thrown from a compiler plugin, and all bloop reported was the error's text:

```
$ bloop --verbose --debug all compile core
Compiling core (1 Scala source)
Compiled core (51ms)
[E] Unexpected error when compiling core: ''scala.tools.nsc.reporters.Reporter scala.tools.nsc.Global.reporter()''
[E] Failed to compile 'core'
```

It would be nice if bloop were able to provide the stacktrace in these cases.

jozanek:
Fixes #1434

The original report asked for the complete stack trace when a compiler plugin crashes the compiler with a `NoSuchMethodError`. The reporting side was largely fixed in the meantime (since v1.5.12 the consumer of a failed compile result prints the whole trace), but reproducing the scenario revealed something worse: the compilation never finishes at all.

`NoSuchMethodError` is a `LinkageError`, which `NonFatal` treats as fatal: it escapes the monix task running the compilation and the result is never completed, leaving the client waiting forever. `StackOverflowError` and `NoClassDefFoundError` were already special-cased in `BloopHighLevelCompiler`, but every other `LinkageError` fell through, and the crash reporting was split between the catch sites and the consumer of the result, which printed the trace twice.

Now all fatal compiler crashes are wrapped in a new non-fatal `bloop.CompilerCrash` carrying type-specific advice, so the compilation fails normally through `Result.Failed(..., Some(t))` and the consumer of the result reports the full stack trace exactly once:

```
Unexpected error when compiling b: bloop.CompilerCrash: The compiler crashed with a NoSuchMethodError. A compiler plugin or something else on the compiler classpath may be incompatible with the used Scala version
	at sbt.internal.inc.bloop.internal.BloopHighLevelCompiler.compileSources$1(...)
	...
Caused by: java.lang.NoSuchMethodError: boom
	at crash.CrashingPlugin$Component$$anon$1.apply(CrashingPlugin.scala:19)
	at scala.tools.nsc.Global$GlobalPhase.applyPhase(Global.scala:465)
	...
```

### Notes

`CompilerCrash` is deliberately public in the `bloop` package even though it is just a transport between the compiler integration and the result reporting: the only instantiation site lives in `sbt.internal.inc.bloop.internal`, outside the `bloop` package hierarchy, so `private[bloop]` would not compile, and the class's fully-qualified name is the first line of the user-visible report, where `bloop.CompilerCrash` reads better than a `sbt.internal.inc.*` name.

### DIFF
--- a/backend/src/main/scala/bloop/CompileExceptions.scala
+++ b/backend/src/main/scala/bloop/CompileExceptions.scala
@@ -2,6 +2,13 @@ package bloop
 
 import scala.util.control.NoStackTrace
 
+/**
+ * A fatal error thrown by the compiler, wrapped in a non-fatal exception so that it can
+ * be turned into a failed compilation result and reported back to the client.
+ */
+final class CompilerCrash(message: String, cause: Throwable)
+    extends RuntimeException(message, cause)
+
 private[bloop] object CompileExceptions {
   abstract class CompileException(msg: String) extends RuntimeException(msg) with NoStackTrace
   object FailedOrCancelledPromise
--- a/backend/src/main/scala/bloop/Compiler.scala
+++ b/backend/src/main/scala/bloop/Compiler.scala
@@ -1,8 +1,6 @@
 package bloop
 
 import java.io.File
-import java.io.PrintWriter
-import java.io.StringWriter
 import java.nio.file.Files
 import java.nio.file.NoSuchFileException
 import java.nio.file.Path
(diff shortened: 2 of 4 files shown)


tgodzik (approved):
LGTM

PR #2973, opened 2026-06-13
Honor sbt's includeFilter/excludeFilter when exporting to Bloop

### LINKED ISSUE #1440: Bloop ignores excludeFilter in build.sbt
As far as I can tell bloop ignores the sbt setting `excludeFilter`.

Simple setup:

```sh
> cat build.sbt
ThisBuild / organization := "foo"
ThisBuild / scalaVersion := "2.12.12"

lazy val root = (project in file ("."))
  .settings(
    excludeFilter := new SimpleFileFilter((f: File) =>
      f.getAbsolutePath.endsWith("bar.scala")
    )
  )


> cat src/main/scala/foo.scala
trait Foo

> cat src/main/scala/bar.scala
DOES NOT COMPILE

> cat project/build.properties
sbt.version=1.4.6

> cat project/metals.sbt
// DO NOT EDIT! This file is auto-generated.
// This file enables sbt-bloop to create bloop config files.

addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.4.6-15-209c2a5c")
```

`sbt compile` compiles successfully. 

But I get the following error mesage in VSCode for the file `bar.scala`

```
bar.scala
 * expected class or object definition bloop [1, 1]
```

Is there a way to get bloop to ignore this file as well?

What I expect: No error messages in VSCode

jozanek:
Fixes #1440.

## Approach

No config-schema change: the model already has `sourcesGlobs` (directory + includes + excludes), honored by the server everywhere. At `bloopGenerate` time, each unmanaged source directory is diffed against what sbt actually compiles; a directory with filtered files is exported as a `sourcesGlobs` entry (and dropped from plain `sources`, since the server unions the two).

Filter translation is done as faithfully as is statically possible:

- **Name-based filters** (glob/`*`, prefix, suffix, extension, and `||` unions) are translated to glob patterns — they keep working for files created *after* the export. `includeFilter` is translated only when it is *entirely* name-based, because a partial translation could make Bloop compile less than sbt.
- **Arbitrary predicate filters** (`SimpleFileFilter`, `&&`/`--`) cannot be evaluated against files that do not exist yet, so the files and directory subtrees they reject are captured as an export-time snapshot.

Directories whose filters are left at their defaults produce byte-identical configs.

### Companion server fix

Once directories are exported as globs, two pre-existing file-watching gaps become user-visible, so this PR also watches `sourcesGlobs` directories in `--watch` mode (new files under a glob-only directory now trigger a rebuild) and teaches the watch event filter to respect glob excludes so filtered files no longer cause no-op iterations.

## Limitations (documented in `docs/build-tools/sbt.md`)

A predicate filter's exclusion set is a snapshot: a file created after export that the predicate would reject is compiled until the next `bloopInstall` or build re-import. Name-based filters do not have this limitation.


integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SourcesGlobsExport.scala
```diff
@@ -0,0 +1,241 @@
...
+ * that include everything the server would pick up by default and exclude what sbt filters out.
+ *
+ * Name-based filters are translated into glob patterns, so they also apply to files created
+ * after the export. Filters that are arbitrary predicates cannot be evaluated against files
+ * that do not exist yet; their effect is exported as the files and directories they reject at
+ * export time. A file is only excluded this way when it is absent from what sbt compiles *and*
+ * the filters reject it, so that a file created while the export runs is never excluded by
+ * accident.
+ */
+object SourcesGlobsExport {
```
tgodzik: I think it would be better in these cases to just list all the files, it seems very fragile to translate the exclude filters to globs. Is it even possible with 100% certainty?

I think we should rather apply those eagerly and print a warning during import like:
```
Due to the presence of exclude and include filters you will need to reimport the build every time you add files to the project with those specified.
```

tgodzik (changes requested):

integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SourcesGlobsExport.scala
jozanek: Done — now lists the files explicitly (`Keys.sources`), no glob translation. Triggered by the presence of non-default filters + a re-import warning; default-filter projects unchanged.

frontend/src/main/scala/bloop/io/SourceWatcher.scala
```diff
@@ -213,6 +218,36 @@ final class SourceWatcher private (
...
 object SourceWatcher {
+
+  /**
+   * Decides whether a file watching event should trigger a new compilation iteration.
+   *
+   * A file inside a sources-glob directory is a source only if some glob matches it; otherwise
+   * it is filtered out and must not wake compilation, unless a plain source directory also
+   * covers it. Files outside glob directories keep the default non-hidden source-file rule.
+   */
+  def isWatchedSourceFile(
```
tgodzik: Do we need this change now? Maybe we could separate it?

tgodzik (changes requested):
LEt's separate the file watcher changes, since they don't seem needed now.

frontend/src/main/scala/bloop/io/SourceWatcher.scala
jozanek: Removed from this PR, will open new one with these changes.

tgodzik (approved):
LGTM!

PR #2974, opened 2026-06-14
Add --jvm-debug flag to run and test for attaching a debugger

### LINKED ISSUE #1457: -jvm-debug flag
Hi! I think it would be pretty valuable to allow passing a flag to `run` / `test`, like [we can do in sbt](https://github.com/sbt/sbt-launcher-package/commit/211ffa3e3#diff-64087abe6d4b2df660baa360f5041e2e3c05ed9d406e6d297d13c314c3f3ded5R285).

I saw a `--debug` flag in `bloop run`, but it's not very well documented and I'm not sure it can even be used for this.

jozanek:
Closes #1457.

`bloop run` / `bloop test` can now fork the JVM with a JDWP agent via `--jvm-debug <port>`, so a debugger (IntelliJ, jdb) can attach. Add `--jvm-debug-suspend` to make the JVM wait for the debugger before running (default: don't wait, matching sbt's `-jvm-debug`).

### Notes
- Reuses the existing `RunMode.Debug` / `JvmDebuggingForker` machinery that was previously only reachable through BSP/DAP; `RunMode.Debug` now carries the debug address + suspend flag. The DAP agent string is unchanged.
- Port range and a dangling `--jvm-debug-suspend` are validated at the CLI layer (like the bsp port validation). `--jvm-debug` with `--parallel` is rejected only when more than one JVM test project would fork at once; it's a no-op (ignored with a warning) for non-JVM projects.
- Docs: `reference.md` gets the flag entries; `tutorial.md` now points to `--jvm-debug` instead of the raw `-J-agentlib:jdwp=...` workaround.

### DIFF
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -316,6 +316,10 @@ sidebar_label: CLI --help
   <dd><p>If set, do not color output. Defaults to false.</p></dd>
   <dt><code>--debug</code> (type: <code>"all" | "file-watching" | "compilation" | "test" | "bsp" | "link"*</code>)</dt>
   <dd><p>Debug the execution of a concrete task.</p></dd>
+  <dt><code>--jvm-debug</code> (type: <code>int?</code>)</dt>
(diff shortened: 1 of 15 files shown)


tgodzik (approved):
LGTM!

PR #2941, opened 2026-06-06
Add regression test for the Ammonite console on JDK > 8

jozanek:
Closes #1226

## Summary

`bloop console <project>` historically failed on JDK > 8 with:

```
scala.reflect.internal.MissingRequirementError: object java.lang.Object in compiler mirror not found
```

This was an upstream Ammonite / JDK 9+ issue (com-lihaoyi/Ammonite#1035): on JDK > 8 the Scala compiler can't locate the platform classes (`rt.jar` / `sun.boot.class.path` were removed). It was never bloop-specific — even a direct `coursier launch com.lihaoyi:ammonite_2.13.3:2.3.8` failed identically.

It no longer reproduces on current `main`: the console emits a `coursier launch …` command that the CLI runs with the user's own coursier (#2856 / #2568), defaults to a `latest.release` Ammonite that supports JDK > 8, and no longer passes the old `--scala-version` flag.

Verified on JDK 17.

PR #2975, opened 2026-06-14
improvement: Resolve console REPL artifacts via coursier API instead of the CLI

jozanek:
## Summary

Follow-up to #2941. Reworks how `bloop console` launches REPLs: it no longer depends on a `coursier`/`cs` launcher on the user's PATH, works for Scala 3, and runs under the project's configured JDK. Previously the server emitted a `coursier launch …` command for the CLI to run — which required a coursier binary on PATH, used an ambient `java`, and only supported the Scala 2 REPL.

## Notable decisions

- **`--exclude-root` had no prior behavior.** The option existed on `Commands.Console` but was never wired to anything, so there was nothing to preserve — I implemented it from its help text ("compile only the target's dependencies"): compile the dependency projects and build the REPL classpath from them.
- **The Scala 3 `≥ 3.8.0` → `scala3-repl_3` split is forward-looking** (the REPL moves to its own artifact at 3.8.0; `< 3.8.0` still ships it in `scala3-compiler_3`), following the review guidance and Coursier's metadata.


frontend/src/main/scala/bloop/engine/Interpreter.scala
```diff
@@ -294,23 +311,40 @@ object Interpreter {
...
+                      .fullRuntimeClasspath(dag, state.client)
+                      .map(_.syntax)
+                      .mkString(java.io.File.pathSeparator)
+
+                    val artifact =
+                      DependencyResolution.Artifact(
+                        "org.scala-lang",
+                        "scala-compiler",
+                        scalaVersion
+                      )
```
tgodzik: I realized that for Scala 3.8.x and up we have a separate artifact for repl `scala3-repl`, then I realized that we don't support Scala 3 here at all. Could you add the handling here:

For Scala 3 below 3.8.0 we would need `scala3-compiler` artifact,
for above `scala3-repl`. And we would need to use another name `dotty.tools.repl.Main` for the runner.

tgodzik: This should work for ammonite already, since that's a separate artifact.

tgodzik: Example how it's handled in coursier for context https://github.com/coursier/apps/blob/main/apps/resources/scala3-repl.json

jozanek: Done. `scalaReplArtifactAndMain` now picks the REPL per Scala version:

- Scala 2 → `scala-compiler` + `scala.tools.nsc.MainGenericRunner`
- Scala 3 `< 3.8.0` → `scala3-compiler_3` + `dotty.tools.repl.Main`
- Scala 3 `≥ 3.8.0` → `scala3-repl_3` (+ `scala3-compiler_3`, since `scala3-repl` excludes the stdlib) + `dotty.tools.repl.Main`

Per the `scala3-repl.json` you linked, it also passes `-Dscala.usejavacp=true` so the REPL finds the JDK classes on JDK > 8. Verified both Scala 2 and Scala 3 (3.7.4 and 3.8.4) REPLs boot and evaluate. Ammonite left untouched, as you noted.

frontend/src/test/scala/bloop/ConsoleSpec.scala
```diff
@@ -1,15 +1,32 @@
...
 import bloop.util.TestProject
 import bloop.util.TestUtil
 
 object ConsoleSpec extends BaseSuite {
-  test("default ammonite console works in multi-build project") {
+
+  // The console resolves REPL artifacts with coursier's API, so these tests need a working
+  // coursier setup; gate on the launcher being present as a proxy and skip otherwise.
+  private def consoleTest(name: String)(fun: => Any): Unit =
+    if (TestUtil.hasCoursier) test(name)(fun)
```
tgodzik: Actually, do we need coursier if we use API, we should be ok without coursier being installed separately.

tgodzik (changes requested):
One minor questions, otherwise looks good!

frontend/src/test/scala/bloop/ConsoleSpec.scala
jozanek: Good point, I removed `hasCoursier`. The API needs no binary, so these now run everywhere (they were being skipped on Windows/macOS-Intel).

tgodzik (approved):
LGTM!

PR #2976, opened 2026-06-14
Fully reset incremental state on clean

### LINKED ISSUE #1466: Converting package to object gives "<object> is already defined as package <package>"
Converting a package to a top level object with the same name gives "<object> is already defined as package <package>"

To reproduce start with the follow setup:

```
 > cat build.sbt
ThisBuild / organization := "foo"
ThisBuild / scalaVersion := "2.12.12"

lazy val root = (project in file ("."))

> cat project/build.properties
sbt.version=1.4.6

> cat src/main/scala/foo.scala
package foo.bar

object Bar {
  val myFooBar: Int = 1
}

> cat project/metals.sbt
// DO NOT EDIT! This file is auto-generated.
// This file enables sbt-bloop to create bloop config files.

addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.4.6-21-464e4ec4")
```

After metals has loaded and built the project change `foo.scala` to

```scala
package foo

object bar {
  object Bar {
    val myFooBar: Int = 1
  }
}
```

Get the error message `bar is already defined as package bar`.

The error message remains even through doing the following commands exposed by VSCode:
 * Clean compiling workspace
 * Restarting the build server
 * (Re-)Importing the build

I can get rid of the error message by doing 

```
 > rm -fr .bloop
```

and reimporting the build.

jozanek:
Fixes #1466.

Converting a `package foo.bar` declaration into a top-level `object bar` of the same name reported `bar is already defined as package bar`. The error survived "Clean compiling workspace", restarting the build server, and reimporting the build — only `rm -rf .bloop` cleared it.

### Cause

`clean` deleted the class directories but kept two pieces of incremental state:
- the persisted Zinc analysis file (`<project.out>/<name>-analysis.bin`), and
- the in-memory `CompileGatekeeper` last-successful result.

After a restart the stale analysis was reloaded and the next compile ran incrementally against it, so the old `package foo.bar` definition kept colliding with the new `object bar`. `rm -rf .bloop` was the only fix because it also removes the analysis file.

> **Note for reviewers:** the analysis file is now deleted for *all* cleaned targets, not only those with an in-memory successful result — the file can exist on disk even when the in-memory cache has none (e.g. right after a restart), so it can't reuse the existing `successful`-keyed deletion loop.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
+++ b/frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
@@ -79,7 +79,7 @@ final case class ResultsCache private (
       logger: Logger
   ): Task[ResultsCache] = {
     def delete(path: AbsolutePath): Task[Unit] = Task {
-      logger.debug(s"Deleting classes directory: $path")(DebugFilter.All)
+      logger.debug(s"Deleting $path")(DebugFilter.All)
       Paths.delete(path)
     }
     // Remove all the successful results from the cache.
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
LGTM! I was sure we remove the analysis, but must have missed it.

PR #3000, opened 2026-07-04
Source globs watcher

jozanek:
## Summary

Improve file watching for projects that use `sourcesGlobs`.

This makes `bloop compile -w` watch source-glob directories themselves, so newly-created matching files can trigger a new compile iteration. It also tightens event matching so watched files follow the same glob semantics as source discovery: globs only apply inside their configured directory and respect `walkDepth`.

## Details

- Watches `sourcesGlobs` directories in addition to already-expanded source files.
- Uses only the reachable watched project closure when deciding whether an event is a source event.
- Prevents unrelated loaded projects' globs from influencing the active watcher.
- Respects `sourcesGlobs.walkDepth` for both project source globs and source-generator input globs.
- Adds unit coverage for glob include/exclude matching, directory containment, and walk-depth behavior.
- Adds a file-watching regression test for creating a new matching file inside a `sourcesGlobs` directory.

## Notes

This was split out from the sbt source filter fix (#2973) because issue #1440 is now handled by exporting filtered sbt sources as explicit files. This PR keeps only the independent watcher improvement for existing `sourcesGlobs` support.

### DIFF
--- a/frontend/src/main/scala/bloop/io/SourceWatcher.scala
+++ b/frontend/src/main/scala/bloop/io/SourceWatcher.scala
@@ -31,6 +31,9 @@ final class SourceWatcher private (
     projectNames: List[String],
     dirs: Seq[Path],
     files: Seq[Path],
+    plainSources: Seq[Path],
+    projectGlobs: Seq[SourcesGlobs],
+    sourceGeneratorGlobs: Seq[SourcesGlobs],
     logger: Logger
 ) {
   import java.nio.file.Files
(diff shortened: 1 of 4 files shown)


frontend/src/main/scala/bloop/io/SourceWatcher.scala
```diff
@@ -30,6 +31,9 @@ final class SourceWatcher private (
     projectNames: List[String],
     dirs: Seq[Path],
     files: Seq[Path],
```
tgodzik, line 33: Is this still used? How different it is to plainSources?

tgodzik (commented):
One quick question from me, otherwise looks good.

frontend/src/main/scala/bloop/io/SourceWatcher.scala
jozanek: Yes, it's still used — `files` and `dirs` tell the watcher which paths to watch. `plainSources` is separate: it only helps decide whether a changed file should trigger a rebuild.

tgodzik (approved):
Thanks for explaining!

PR #3007, opened 2026-07-09
Detect JSDOMNodeJSEnv when generating Scala.js configs

### LINKED ISSUE #1489: Bug: jsdom hardcoded to false in Scala.js projects
## Bug 

Seems like (unintentionally?) jsdom setting in the generated config is set to false: https://github.com/scalacenter/bloop/blob/3eab1f8079cfaf228ec68740daadf5854070c89b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala#L779

Which means the tests don't run from IDEs.

Manually changing `jsdom` to `true` in the generated json config leads to tests running successfully via both bloop and from IDE.

## Reproduce

It's not minimal, but not involved:

1. Clone https://github.com/keynmol/http4s-laminar-stack
2. Attempt to run `FrontendSpec`
    It will fail
3. modify `.bloop/frontend-test.json`, changing `jsdom` property to true
4. run `FrontendSpec` again - it should succeed (as long as jsdom is installed on the machine)

jozanek:
Closes #1489. The plugin hardcoded `jsdom = false`, ignoring a project's `jsEnv := new JSDOMNodeJSEnv()`. Now detected reflectively and set accordingly.

Notable: `jsEnv` is a `TaskKey` in Scala.js 1.x, so this adds a task proxy (`toAnyRefTaskKey`) alongside the existing setting proxies; wired in `configSettings` so `Test / jsEnv` is honored. Detects `JSDOMNodeJSEnv` only — other DOM envs can force it via `Test / bloopScalaJSJsdom := Some(true)`.

### DIFF
--- a/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/Compat.scala
+++ b/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/Compat.scala
@@ -7,6 +7,7 @@ import sbt.Def
 import sbt.Exec
 import sbt.Keys
 import sbt.SettingKey
+import sbt.TaskKey
 import sbt.io.syntax.File
 import sbt.librarymanagement.ScalaModuleInfo
 import sbt.util.CacheStore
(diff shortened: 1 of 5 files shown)


tgodzik (approved):
Looks a bit complex, but should be fine as a fallback, thanks. LGTM

PR #3014, opened 2026-07-16
fix: Do not let the native launcher consume -D arguments

### LINKED ISSUE #1490: Passing arguments from CLI to Scalatest runner does not work
CLI runner [arguments](https://scalacenter.github.io/bloop/docs/cli/tutorial#pass-test-arguments-to-test) are not passed to [Scalatest](https://www.scalatest.org/scaladoc/3.2.7/org/scalatest/ConfigMap.html). Previously reported (and not fixed) in  #1288.

```
$ bloop test root --only MySpec -- -Dkey=value
MySpec:
configMap=Map()
```

It works perfectly fine with `sbt`:
```
$ sbt "root/testOnly MySpec -- -Dkey=value"
[info] compiling ...
MySpec:
configMap=Map(key -> value)
```

Tried and not worked:
- The `--args` flag and/or variations of quotations
-  bloop "1.4.8" and "v1.4.8-19-4d9f966b"
- JDK 8 and 11
- `bloop test xyz -- -J-Dkey=value`

jozanek:
Fixes #1490. `bloop test foo -- -Dkey=value` never reached the test framework (ScalaTest's ConfigMap stayed empty), though sbt handles it fine.

**Cause:** GraalVM native images consume `-D`/`-X`/`-XX:` from argv at startup by default, even after `--`, so the CLI binary ate `-Dkey=value` before Bloop ran. Broken since bloopgun became the first native client (1.4.1); the server side was always correct.

**Fix:** build with `-H:-ParseRuntimeOptions` so args reach the server verbatim. `Bloop.main` still lifts the launcher's own `-D` properties (those before the command) so `bloop.java-opts` etc. keep working; everything from the command onward, including after `--`, is forwarded untouched. Client-side `-X`/`-XX:` tuning is no longer honored (a native image can't apply it once started).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,6 +154,9 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
       - uses: sbt/setup-sbt@v1
+      - uses: coursier/setup-action@v3
+        with:
+          apps: scala-cli
 
       - name: Publish GraalVM Native artifacts
         run: sbt "cli/graalvm-native-image:packageBin"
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -184,13 +184,31 @@ jobs:
...
+        shell: bash
+        run: |
+          root=$(pwd)
+          if [ "${{ matrix.os }}" = "windows-latest" ]; then
+            bin=$root/cli/target/graalvm-native-image/bloop-cli.exe
+          else
+            bin=$root/cli/target/graalvm-native-image/bloop-cli
+          fi
+          "$bin" about
+          # A -D before the command is a launcher property that Bloop.main must extract
```
tgodzik: I don't think it's worth putting these comments here. Ideally, we should extract them to standalone file and have the comments there

tgodzik: Also, maybe we could create a Scala CLI test, run compile on it to create json bloop files. We can then invoke it easily in .scala-build directory, which itself contains .bloop directory.

Could be easier to maintain and would actually be a real life usage

tgodzik (changes requested):
Thanks for working on this! One thing I would change is maybe try to create test using scala cli.

.github/workflows/ci.yml
jozanek: Done, moved to `bin/test-native-cli.sh` with the comments, can also be run locally against a freshly built binary.

jozanek: Done. Scala CLI generates the project and the binary is invoked in `.scala-build`. With a real project available, the check now covers the full scenario: `test <project> -- -Dkey=value` must print `configMap=Map(key -> value)` from ScalaTest.

tgodzik (approved):
LGTM!

PR #3029, opened 2026-07-25
Let caller-supplied JVM options override config javaOptions

jozanek:
The forked JVM command put `config.javaOptions` last, so for single-valued options the config won (last occurrence wins). Since `sbt-bloop` injects `-Duser.dir` into every config, a caller's `-Duser.dir` — e.g. sent via `debugSession/start` — was silently ignored (scalameta/metals#2888).

Reorder to `config.javaOptions ++ jargs` so caller options (DAP `jvmOptions`, CLI/test `-J` args) win.


tgodzik (approved):
LGTM

Issue #1351, opened 2020-07-31
Caching artifacts in bloop


Alexey-NM:
It would be very useful to have the ability to share compilation cache between different computers. 
There is prototype for such feature, it is [remotebloop](https://github.com/global-system/remotebloop)
It has been said that it is quite fragile and it can be fixed by adding built-in support for such feature. 

see also:
[Reload analysis files when they change](https://github.com/scalacenter/bloop/issues/1091)

Alexey-NM:
They say in https://www.scala-sbt.org/1.x/docs/Remote-Caching.html
> sbt 1.4.0 / Zinc 1.4.0 virtualizes the file paths tracked during incremental compilation, and uses content hash for change detection. With these combination, we can realize repeatable build, also known as build as function

It can be done similar

Alexey-NM:
There is said in https://www.scala-sbt.org/1.x/docs/Remote-Caching.html
sbt 1.4.0 / Zinc 1.4.0 virtualizes the file paths tracked during incremental compilation, and uses content hash for change detection. With these combination, we can realize repeatable build, also known as build as function.

jozanek:
Hi @tgodzik, asking first this time, before writing any code.

The saved analysis files contain absolute paths, so if you move a workspace or build in docker, everything recompiles even though no sources changed. Zinc already has mappers to fix this (sbt uses them for remote caching), and using them in Bloop would be a small internal change and useful on its own.

Should I write a short design proposal for this here, or would you rather wait until users ask for it?

tgodzik:
If it's a small enough change we could include it as an option. Overall not relying on absolute paths would be a positive change.

However, if this proves to be a large refactor, I think we should wait until it's more heavily requested.

PR #3044, opened 2026-08-11
Add reload command to pick up externally modified analysis files

### LINKED ISSUE #1091: Reload analysis files when they change
It will be very useful to have the ability to refresh bloop without restarting. 
Would it be possible to add a bloop command for that purpose, it could be something like 'refresh project'

It is very important for large project to have the ability reuse CI compilation cache. 
[jvican has writen](https://github.com/scalacenter/bloop/issues/673#issuecomment-516186219):
_any tool that wants remote compilation can implement it outside of bloop_

Unfrotunatly if we replace analysis files the bloop server just will not be able to see changes, We have to stop\start bloop server manually. It is very inconvenient.

jozanek:
Bloop reads each project's persisted compilation state once and keeps it in memory, so state replaced on disk — restoring a cache saved from an earlier run, say — is invisible until a restart.

Adds a way to re-read it:

- **CLI:** `bloop reload`, all projects or `--project`
- **BSP:** `bloop/reloadAnalysis`, given targets or all

A reload replaces every requested project or nothing. It fails, untouched, while a project compiles or is cleaned, if state changes mid-read, or if what's on disk can't be used; BSP errors carry a stable `reason` and `retryable` flag. Persisted state is only adopted when the output directory it names is one Bloop manages for that project — compilation later deletes that directory, so an unvalidated path would eventually be removed recursively.

`workspace/reload` is implemented as configuration-only (Bloop already reloads config per request) and never imports state, so it can't fail for unrelated reasons.

**Closes #1091.** The remaining half of that discussion — sharing state across machines or checkouts — is blocked by absolute paths in the persisted state and is tracked in #1351.

**Known issue:** after a reload, an earlier-started request can leave a stale *latest* result, affecting BSP diagnostic carry-over only — the successful result compilation builds on is authoritative.

Tests: `ReloadAnalysisSpec` (12) and `BspReloadSpec` (3 per transport).

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -109,6 +109,38 @@ to it.
 You can exit the server by running `bloop exit` from the CLI. You can also kill
 it with `kill`, the Activity Monitor in your machine or `htop`.
 
+## Reload compilation state from disk
+
+The server reads the compilation state it persists for each project once and
+then keeps it in memory, so a server that is already running does not notice
+when that state changes on disk. Reloading makes the server read it again:
+
+- CLI: `bloop reload` reloads the compilation state of all projects, or a
+  selection with `--project`.
+- BSP: the `bloop/reloadAnalysis` request does the same, for the given build
+  targets, or for all of them when the request carries no target list. (The
+  standard `workspace/reload` request reloads the *build configuration* only —
+  Bloop already does that on every request.)
+
+When a reload does not happen, the BSP request fails with a machine-readable
+reason and a `retryable` flag in the error data: a build that is busy compiling
+or being cleaned settles on its own, whereas state that cannot be used needs the
+caller to fix what is on disk.
+
+A reload either replaces the state of every requested project or changes
+nothing at all. It fails, leaving the server exactly as it was, when a
+requested project is compiling, when its state changes while the reload is in
+progress, or when the state on disk cannot be used. Projects with no persisted
+state simply start their next compilation from scratch.
+
+Two limits are worth knowing before building on this. Compilation state records
+absolute paths, so state produced under a different workspace path — on another
+machine, or in another checkout — is not usable: the reload is rejected and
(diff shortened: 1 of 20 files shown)


tgodzik (changes requested):
I wonder if we shouldn't just watch the analysis instead to see if it changes from outside? Ideally we wouldn't need the user to do anything.

Looks like the CI is failing currently.

jozanek:
Watching was my first thought too, but Bloop rewrites analysis files itself after every successful compile. A watcher can't tell our own writes from external ones, and a false positive would silently replace newer in-memory state with older state from disk. Restores are also not atomic, since the analysis and its class directories land over a window of time, so tools need a way to say "done, read now" in any case.

That's why the command is the base building block. I'm happy to add opt-in automatic detection on top of it in a follow-up PR.

CI was failing because AutoCompleteSpec didn't know the new command. Fixed now.

tgodzik:
> Watching was my first thought too, but Bloop rewrites analysis files itself after every successful compile. A watcher can't tell our own writes from external ones, and a false positive would silently replace newer in-memory state with older state from disk. Restores are also not atomic, since the analysis and its class directories land over a window of time, so tools need a way to say "done, read now" in any case.
> 
> That's why the command is the base building block. I'm happy to add opt-in automatic detection on top of it in a follow-up PR.
> 
> CI was failing because AutoCompleteSpec didn't know the new command. Fixed now.

Thanks for explaining! I wonder though if we want this feature at all currently. The last discussions were 6 years ago and we might want to hold off on adding it until someone let's us know that it's something they would need. This is a huge change, so I wouldn't want to add it without a reason to do it. I don't want to waste your work here, but complexity in Bloop is something we do struggle with and I wouldn't want to add more.

jozanek:
Hey @tgodzik, feel free to close the issue. I completely get your point. Don't worry about my effort, I am gaining much more by learning/contributing!

PR #3054, opened 2026-09-04
Add opt-in portable analysis files

jozanek:
Addresses the path-portability part of #1351 as a small, opt-in change.

Persisted analysis files contain absolute paths, so moving a workspace (Docker layer, CI checkout, git worktree) forces a full recompile. With `-Dbloop.analysis.portable=true` Bloop passes a `ReadWriteMappers` pair to Zinc's `FileAnalysisStore` that rewrites paths under known roots (workspace, coursier cache, ivy, sbt boot, java home) to `${KEY}/relative` tokens on write and resolves them on read. In-memory analysis, stamps, reporters and BSP are untouched.

- Default off: on-disk format unchanged. Reading tokens is always on and is a no-op for absolute paths, so existing files keep loading.
- Unresolvable tokens fail closed: the analysis is ignored with a warning and a full compile follows.
- Extra roots via `-Dbloop.analysis.roots=KEY=/path,...`
- Known limitation: Zinc does not map diagnostic positions, so replayed warnings can point at the old path until the file recompiles.

### DIFF
--- a/frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
+++ b/frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
@@ -14,6 +14,7 @@ import bloop.CompileOutPaths
 import bloop.CompileProducts
 import bloop.Compiler
 import bloop.Compiler.Result
+import bloop.PortableAnalysis
 import bloop.UniqueCompileInputs
 import bloop.data.ClientInfo
 import bloop.data.Project
@@ -242,7 +243,12 @@ object ResultsCache {
       val analysisFile = p.analysisOut
       if (analysisFile.exists) {
         Task {
-          val contents = FileAnalysisStore.binary(analysisFile.toFile).get().toOption
+          // Tokens in a portable analysis resolve against this build's roots; a token that
+          // cannot be resolved makes the store return nothing and records the reason
+          val roots = PortableAnalysis.Roots.derive(p.workspaceRoot.underlying)
+          val readSide = PortableAnalysis.readMappers(roots)
+          val contents =
+            FileAnalysisStore.binary(analysisFile.toFile, readSide.mappers).get().toOption
           contents match {
             case Some(res) =>
               logger.debug(s"Loading previous analysis for '${p.name}' from '$analysisFile'.")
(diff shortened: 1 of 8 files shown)


frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
```diff
@@ -242,7 +243,12 @@ object ResultsCache {
       val analysisFile = p.analysisOut
       if (analysisFile.exists) {
         Task {
-          val contents = FileAnalysisStore.binary(analysisFile.toFile).get().toOption
+          // Tokens in a portable analysis resolve against this build's roots; a token that
+          // cannot be resolved makes the store return nothing and records the reason
+          val roots = PortableAnalysis.Roots.derive(p.workspaceRoot.underlying)
+          val readSide = PortableAnalysis.readMappers(roots)
+          val contents =
+            FileAnalysisStore.binary(analysisFile.toFile, readSide.mappers).get().toOption
```
tgodzik, line 251:
```suggestion
          
          val readSide = if (PortableAnalysis.enabled) {
             val roots = PortableAnalysis.Roots.derive(p.workspaceRoot.underlying)
             Some(PortableAnalysis.readMappers(roots)) 
          } else None
          val contents =
            FileAnalysisStore.binary(analysisFile.toFile, readSide.map(_.mappers).getOrElse(ReadWriteMappers.getEmptyMappers())).get().toOption
```
Let's avoid any changes if the portable analysys is disabled maybe?

tgodzik (changes requested):
One small change from me, otherwise looks like a nicely isolated feature. Worth letting people test it out.

frontend/src/main/scala/bloop/engine/caches/ResultsCache.scala
jozanek: Done. The read side now runs only when `PortableAnalysis.enabled` is true, so nothing changes when the option is off.

I changed your snippet a little so it compiles: `readSide` is now an `Option`, so the line below it is `readSide.flatMap(_.failure)`.

One effect: the option is now needed on both sides. If you read a portable analysis with the option off, Bloop ignores it and does a full compile, which writes the file again with absolute paths. I updated the docs.

tgodzik (approved):
LGTM!

Issue #3024, opened 2026-07-23
CommandsDocGenerator produces broken output since the case-app 2.x migration, and CI can't notice


vimalaguti:
## Summary

While working on #2076 I tried to regenerate `docs/cli/reference.md` the same way CI does:

```
sbt "frontend/runMain bloop.util.CommandsDocGenerator --out ../docs/cli/reference.md"
```

The regenerated file is visibly broken compared to the checked-in one. It looks like the generator has been producing broken output since the case-app 2.x migration (a077d7cbc, "Update to mainline case-app 2.x") — the checked-in `docs/cli/reference.md` predates that migration and was never regenerated since. Three related problems:

## 1. Command headings render as `## bloop List(about)`

In `frontend/src/main/scala/bloop/util/CommandsDocGenerator.scala`, `generateHTML` interpolates the command name directly:

```scala
Commands.RawCommand.help.messages.map {
  case (commandName, messages) =>
    ...
    b ++= s"## `$progName $commandName$argsOption`"
```

With case-app 2.x, `help.messages` yields the command name as a `Seq[String]` rather than a `String`, so every heading renders as `## bloop List(about)`, `## bloop List(compile)`, etc.

## 2. `(type: ...)` annotations disappeared from option listings

The checked-in docs render options as:

```
--config-dir or -c (type: path?)
```

The current `optionsMessage` only renders option names and the `@HelpMessage` text, so regenerated docs lose all type information:

```
--config-dir or -c
```

## 3. CI runs the generator but never checks its output

`.github/workflows/ci.yml` runs:

```
frontend/runMain bloop.util.CommandsDocGenerator --test; \
frontend/runMain bloop.util.CommandsDocGenerator --out ../docs/cli/reference.md
```

`--test` only asserts the generation is non-empty, and the `--out` result is written into the CI workspace and then discarded — there is no diff check against the checked-in file and no commit step, so drift between the generator output and the committed `docs/cli/reference.md` goes undetected. That's how problems 1 and 2 stayed invisible: the website kept serving the last file generated before the migration.

## Reproduction

On current `main`:

```
sbt "frontend/runMain bloop.util.CommandsDocGenerator --out ../docs/cli/reference.md"
git diff docs/cli/reference.md | head -20
```

shows `-## `bloop about`` / `+## `bloop List(about)`` style changes on every command section.

tgodzik:
Thanks for reporting! I haven't looked into that one in a long while.

PR #3062, opened 2026-09-09
Add --reverse-order flag to control CLI error output order

### LINKED ISSUE #1502: Flag to in inverse output in the cli
Bloop's CLI currently prints out errors last-to-first. Though this is great when you are running the cli next to code, it makes it hard to pipe the errors into simple text-based jumping mechanisms (like vim's quickfix list). 

Bloop should include a way to inverse the output back to the traditional first-to-last order.

jozanek:
Fixes #1502.

Bloop's default CLI reporter buffers compiler diagnostics and prints them last-to-first. That is good interactive UX — the first error lands nearest the prompt — but it breaks piping output into consumers that expect the traditional order.

This adds `--reverse-order` to every compiling command (`compile`, `test`, `run`, `console`, `link`):

- **unset** — current behavior, unchanged. The reporter decides: bloop reverses, scalac does not.
- **`--reverse-order=false`** — diagnostics stream first-to-last, as the compiler reports them.
- **`--reverse-order`** — forces the reversed summary, including under `--reporter scalac`.

### Why a flag rather than changing the default

`ReporterConfig.reverseOrder` already existed but was welded to the reporter choice: picking `--reporter scalac` got you first-to-last order *and* a different output format. This decouples the two axes, so you can keep bloop's format and choose the order. Unset is byte-identical to the previous expression, so no existing output changes.

### Scope

Implemented at a single site. `Interpreter.runCompile` is the only place the CLI builds a `LogReporter` config, and all five commands reach it — `compile` directly, the rest through `compileAnd`:

```scala
val baseConfig = ReporterKind.toReporterConfig(cmd.reporter)
val config = baseConfig.copy(
  colors = !noColor,
  reverseOrder = cmd.reverseOrder.getOrElse(baseConfig.reverseOrder)
)
```

`reverseOrder` is declared on the sealed `CompilingCommand` trait, so the compiler enforces that no command silently omits it. BSP is untouched — it builds its own config and already forces first-to-last.

The `docs/cli/reference.md` entries are hand-written rather than regenerated, because running `CommandsDocGenerator` on current `main` corrupts the file (#3024). The added entries follow the file's existing style, including the `?` suffix it uses for optional types.

### One caveat, stated plainly

The issue's motivation was vim's quickfix list, and this flag only fixes the ordering half of that. Bloop's default format is multi-line with source snippets and error IDs, and diagnostics go to stderr — so a quickfix workflow still wants `2>&1` and realistically `--reporter scalac` for a one-line-per-error format. Ordering is the part users could not work around on their own; the format axis stays where it was.

### DIFF
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -117,6 +117,8 @@ sidebar_label: CLI --help
   <dd><p>Pipeline the compilation of modules in your build. By default, false.</p></dd>
   <dt><code>--reporter</code> (type: <code>reporter</code>)</dt>
   <dd><p>Pick reporter to show compilation messages. By default, bloop's used.</p></dd>
+  <dt><code>--reverse-order</code> (type: <code>bool?</code>)</dt>
+  <dd><p>Print errors newest-first. Pass --reverse-order=false to print them as the compiler reports them. By default, bloop reverses and scalac does not.</p></dd>
   <dt><code>--watch</code> or <code>-w</code> (type: <code>bool</code>)</dt>
   <dd><p>Run the command when projects' source files change. By default, false.</p></dd>
   <dt><code>--cascade</code> (type: <code>bool</code>)</dt>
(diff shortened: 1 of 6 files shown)


tgodzik (approved):
LGTM

Issue #1504, opened 2021-05-19
`sbt bloopInstall` does not share resource files if there are no scala files in a project


pjrt:
The title might be confusing, here's what's happening. 

We have a project (sbt) that has nothing in it but a resource file. This file is shared with other projects via "depends" method on SBT projects. 

When we call `sbt bloopInstall`, the projects that have a dependency on it seem to not get access to those resource files. 

Adding a dummy scala class to that resources-only project fixes the issue.

tgodzik:
Thanks for reporting, this should not be that hard to fix I think. We most likely treat the module as empty is there are no Scala files.

tgodzik:
Fixed by https://github.com/scalacenter/bloop/pull/2692

PR #3067, opened 2026-09-13
Fix #1503: resources of source-less dependencies on compile classpath

### LINKED ISSUE #1503: Inaccurate Scalaz-deriving Errors
Hey guys,

Looks like there may be some compatibility issues with the latest scalaz-deriving and bloop. Under my multi-project mono-repo I'm using the following libs:

- Bloop v1.4.8-19-4d9f966b
- scala 2.12.10
- scalaz-deriving 3.0.0-M2
- scalaz 7.3.3
- pureconfig 0.13.0

The issue I'm facing is that using those libs together will cause persistent errors to show up, despite not showing up when compiling within SBT. Here a code snippet as an example...

```scala
import pureconfig.ConfigReader
import scalaz.annotation.deriving

@deriving(ConfigReader)
final case class Foo(
    name: String,
    phone: String)
```
This will result in...
```
could not find implicit value for parameter F: scalaz.DerivingProducts[pureconfig.ConfigReader]
```
This issue seems to prevent Metals and Bloop from properly compiling dependent files, and breaks IDE functionality downstream.

Thanks for the help!

jozanek:
Fixes #1503.

## Problem

A dependency project with only resources and no sources never has its resource directories
placed on a dependent's **compile** classpath. `CompileDependenciesData.compute` rewrote the raw
classpath from `dependentProducts`, and a source-less project yields `Compiler.Result.Empty`, so it
is never a key of that map. A macro expanding in the dependent cannot read the resources; for
scalaz-deriving that means no `deriving.conf`, a fallback to `scalaz.Deriving.gen`, and the
reported `could not find implicit value for parameter F: scalaz.DerivingProducts[...]`.

#2692 fixed the sibling report (#1504) by copying such resources into the client classes
directory, which the **run** classpath and the BSP-reported classpath resolve to. So `bloop run`
and the editor already work; the compile classpath never references that directory, so compiling
still failed. This completes the compile side.

## Fix

Resources are inputs, not products, so they are now derived from the project graph: `compute`
takes the compiled project's transitive dependencies and prepends each one's existing resource
directories in front of its classes directory whenever that directory is a classpath entry.
Ordering and cross-dependency de-duplication are unchanged.

## Reproduction

Scala 2.13.5, scalaz-deriving 3.0.0-M2 with `deriving-plugin`, pureconfig 0.13.0; a `conf`
project holding only `deriving.conf` and an `app` project depending on it with the snippet from
the issue. `sbt compile` succeeds. `bloop compile app` on 2.1.1 fails with the error from the
issue, succeeds after adding one dummy source to `conf` (the workaround from the issue), and
succeeds on a server built from this branch.

## Tests

Four cases in `BaseCompileSpec`, each with a def macro that reads `deriving.conf` off its own
classloader during expansion and aborts with a positioned diagnostic when it is missing: a
resources-only dependency and a transitive one (both fail before this change), a dependency
with sources (control), and a resources-only project the consumer does not depend on (negative
control). The harness gives every project in a workspace the same `resources` directory, which
would put the file on every classpath, so the cases hand one project a private directory via
`additionalResources`.

## Behaviour to be aware of

- Dependents of a resources-only project recompile once after upgrading, since the new
  directories enter the classpath hashes.
- Editing a resource afterwards still does not invalidate dependents (directories hash to a
  constant); this adds first-compile visibility, not resource-content tracking.

## Verification

`bloop.CompileSpec` 49 pass (the one local failure, `resources only downstream project resolved
correctly`, points at a JDK 11 path that does not exist on this machine); the reproducers fail
and the controls pass with the engine change reverted; `bloop.DeduplicationSpec` 7 pass;
`./bin/scalafmt --test` clean; end to end with the real libraries above.


tgodzik (approved):
LGTM

Issue #1524, opened 2021-06-27
Bloop `run --watch` doesn't appear to reload when Spring Boot "Spring-Fu" functional routing Netty app source changes 


GavinRay97:
I have a Scala 3 app with the buildfile in SBT, using the above mentioned setup.

Doing `bloop run <appname> --watch` doesn't appear to reload it on changes. Is there a trick to getting this to work 👀 

https://user-images.githubusercontent.com/26604994/123532152-bb626580-d6d8-11eb-94e3-281a3516b5c2.mp4

```scala
package org.example

import scala.beans.BeanProperty
import reactor.core.publisher.Mono
import org.springframework.fu.jafu.Jafu.reactiveWebApplication
import org.springframework.fu.jafu.webflux.WebFluxServerDsl.webFlux
import org.springframework.web.reactive.function.server.ServerRequest
import org.springframework.web.reactive.function.server.ServerResponse
import org.springframework.web.reactive.function.server.ServerResponse.ok
import org.springframework.fu.jafu.webflux.WebFluxServerDsl

case class Sample(@BeanProperty message: String)

class SampleService:
  def generateMessage(): String = "Hello world!"

class SampleHandler(private var sampleService: SampleService):

  def hello(request: ServerRequest): Mono[ServerResponse] =
    ServerResponse.ok().bodyValue(sampleService.generateMessage())

  def json(request: ServerRequest): Mono[ServerResponse] =
    ServerResponse.ok().bodyValue(new Sample(sampleService.generateMessage()))
    
object Application {
  val app = reactiveWebApplication((a) =>
    a.beans((b) => b.bean(classOf[SampleHandler]).bean(classOf[SampleService]))
      .enable(
        webFlux((server) =>
          server
            .port(if (server.profiles().contains("test")) 8181 else 8080)
            .router((app) => {
              val handler = server.ref(classOf[SampleHandler])
              app.GET("/", handler.hello)
              app.GET("/api", handler.json)
            })
            .codecs(_.string().jackson())
        )
      )
  )

  def main(args: Array[String]): Unit = app.run(args)
}
```
```scala
ThisBuild / organization := "com.example"
ThisBuild / version := "0.0.1-SNAPSHOT"
ThisBuild / scalaVersion := "3.0.0"

val SPRING_VERSION = "2.6.0-SNAPSHOT"

lazy val root = (project in file("."))
  .settings(
    name := "demo",
    resolvers ++= List(
[...]

PR #3068, opened 2026-09-15
docs: Note that run --watch does not restart long-running applications

### LINKED ISSUE #558: Add restarting mode to `bloop run --watch`
Currently, `bloop run --watch` only works for self-terminating apps so that it's not applicable for non-terminating apps such as servers.

What about adding an option to kill the forked vm and restart it when a source file change is detected? This feature would be drop-in replacement of https://github.com/spray/sbt-revolver.

jozanek:
`bloop run -w` only reacts to a source change once the application exits, so for a server or daemon the first run blocks the watch loop and nothing is ever recompiled or restarted. The CLI tutorial currently advertises `-w` for `run` with no caveat.

This documents the limitation and points at the supervisor-based workarounds people already use. Docs only, no behaviour change.

Root-cause analysis and a reproduction: #1524 (a duplicate of the closed #558).


tgodzik (approved):
LGTM

PR #3069, opened 2026-09-15
Unlink stale BSP socket files before binding

jozanek:
Follow-up hardening for #1578, which #2832 already fixed for the normal cases.

`ServerHandle.UnixLocal` binds without unlinking first. When the bloop server is killed with `SIGKILL` the shutdown hook added in #2832 never runs, so the socket file survives and the next server on that path fails with `BindException: Address already in use`.

This removes a leftover socket file before binding. To avoid trading one problem for a worse one, the file is only removed when it is a socket **and** nothing is listening on it — if a live server owns the address, `bind` is allowed to fail as it does today rather than silently orphaning that server.

Tests:
- `ServerHandleSpec` — binds over a dead server's leftover file; refuses to bind over a live server's file; leaves a regular file at the socket path untouched.
- `LocalBspConnectionSpec` — regression guard that the server removes the socket file when a session ends, which #2832 shipped without. The test client's own deletion of that file was removed, since it made such an assertion vacuous.

Both new tests were confirmed to fail without their respective fix.


tgodzik (approved):
LGTM

Issue #2076, opened 2023-06-13
`bloop exit` is undocumented in CLI help and completions


ches:
Quite minor usability thing here, but I occasionally want to stop the server for one reason or another and I fumble with a task manager, or finding a working `pkill` invocation, or attempting `bloop stop`, `bloop quit`, `bloop kill`…

Clearly there are workarounds, but `bloop exit` is convenient, if only I remember the command.

And that's the rub: `bloop exit` is documented only by mention in [the server reference page][1], it is not listed in `bloop --help` nor the built-in shell completions.

This is probably low-hanging fruit, and I keep meaning to try making a patch but have a sneaking suspicion I'll need to invest time into learning the architecture of why the `frontend` `bloop.Cli` and `bloopgun` are distinct things 😅 

### expectation

- `bloop --help` includes `exit` when listing available commands
- `bloop exit --help` gives usage like any other subcommand
- Shell completion works for `exit`

[1]: https://scalacenter.github.io/bloop/docs/server-reference

tgodzik:
Thanks for reporting! We should add that for sure

jozanek:
The scripts ask the *server* for the command list — `bloop autocomplete --mode=commands` — which returns:

```
about autocomplete bsp clean compile configure console help link projects run test
```

The CLI's own commands (`exit`, `output`, `start`, `status` in `cli/src/main/scala/bloop/cli/Bloop.scala`) aren't in that list and can't be, since they never reach the server. So the hunch about `frontend`'s `bloop.Cli` and the launcher being distinct things was exactly right.

One related detail: all three scripts' `_notStarted()` fallback suggests `server`, which no longer exists as a command anywhere. Mostly unreachable now that the CLI auto-starts a server, but it confirms the scripts predate the split.

(Came across this while fixing #1551 in #3070, which doesn't touch the scripts.)

PR #3070, opened 2026-09-15
[DOCS] Explain how to obtain the command-line completion scripts

### LINKED ISSUE #1551: Autocompletion instructions don't explain how to get autocomplete scripts
I've run through the steps for installing bloop, but the instructions for command completions seem to be missing any info explaining how to actually get the scripts downloaded or initialized in `~/.bloop`. 

Currently, on https://scalacenter.github.io/bloop/setup#command-line-completions, it says:

```
Command-Line Completions

Bloop supports command-line completions in bash, zsh and fish. The use of command-line autocompletions is recommended as it significantly improves the user experience. The installation of autocompletions via curl requires you to configure the completions manually.

    Note that the following instructions assume that the bloop installation directory is the default $HOME/.bloop.
```

And then gives instructions for how to source the relevant rc files.  But this `.bloop` directory has not been created automatically for me, and I've not been able to find any instructions explaining where the completions should be curled from etc.

---

This is my first issue on this repo, so I'd like to also note that finding bloop has been a life saver: It's reduced the pain of working on our codebase by at least %70. Many thanks for for your work!

jozanek:
Closes #1551.

The completions section tells readers to source files under `$HOME/.bloop/{bash,zsh,fish}`, but
nothing has created them since `213e98c87` moved installation to coursier in 2020. Only the
package-manager tabs install completions; the default coursier tab was never updated, so following
it end to end gives you no completions and no error.

Each shell now gets the missing download step, pointing at the scripts published with every
release. Two non-obvious bits:

- zsh only autoloads the script when the file is named `_bloop`, so the page now says so —
  otherwise even a reader who finds the URL gets nothing, silently.
- bash leads with a plain `source` line; `~/.local/share/bash-completion/completions/` is inert
  unless bash-completion 2.x is installed (hence the Homebrew formula's `depends_on`).

The closing advice about starting the server before reloading the shell is removed — the scripts
call `bloop autocomplete` at completion time, and the CLI starts a server when none is running.
The Scoop page now links the two assets it referred to but never named.

Every command was run as rendered against a sandboxed `$HOME`, and all three shells register. The
AUR page was checked and is accurate, so it is unchanged.

Note when testing: completions cover only the server's commands, so `bloop exit<TAB>` is still
empty — that is #2076, not this change.


tgodzik (approved):
Thanks for looking into this one LGTM

PR #3071, opened 2026-09-15
Fix vacuous sbt scripted tests for the runtime classpath export

jozanek:
Three sbt scripted tests never run their assertions.

`runtime-dependency`, `provided-runtime-classpath` and `non-compiling` each wrap their `checkBloopFiles` body in `if (Keys.name.value != "<projectName>")`. That guard exists to run the check once from the aggregate root rather than once per sub-project, and it does that correctly in tests whose sub-projects live in subdirectories. But these three define their only project at `file(".")`, so the root *is* that project, the guard is always false, and nothing is asserted.

`runtime-dependency` was added in 8132093 to guard the runtime classpath export for #1606, and has never executed.

This unwraps the three checks and switches the two whose assertions read a `-test.json` from `bloopGenerate` (Compile scope only) to `bloopInstall`.

To confirm the tests now have teeth, I temporarily changed `emulateRuntimeDependencyClasspath` to read `Compile` instead of `Runtime` — `sbt-bloop/runtime-dependency` fails, as it should.

Also included: the `Global / onLoad` pre-push hook installer assumes `.git` is a directory. In a git worktree it is a file, so `Files.createDirectories` throws `FileSystemException: .git/hooks: Not a directory` and the build cannot load at all. Now guarded on `.git` being a directory.


tgodzik (approved):
LGTM

jozanek:
Heads-up: this broke `main`. The `Build integrations tests` job now fails on both ubuntu and macos.

Removing the always-false guard made these assertions run for the first time, and they only hold under sbt 1 — the `Sbt 2` step runs `runtime-dependency` with `++3.8.4`, where the classpath picks up `scala3-library_3-3.8.4.jar` and a different ordering. I verified only against sbt 1 before opening this, which is how it slipped through.

Fix in #3075: assert the property rather than an exact classpath listing, so it holds under both sbt versions.

Issue #1693, opened 2022-02-23
Fix and restore benchmark step in CI


kpodsiad:
Benchmark step was failing at least for 2 months - https://github.com/scalacenter/bloop/runs/4675462587?check_suite_focus=true
Because it was also blocking PR which bumps sbt #1679 I decided to remove it temporarily from CI in #1692 until it'll be fixed.

The goal of this issue is to investigate problems, fix them and then add again benchmark to the CI.

PR #3073, opened 2026-09-15
fix: Repair two bugs in the Bloop compile benchmark

jozanek:
Found while investigating #1693. Doesn't close it — just makes `HotBloopBenchmark` work for anyone following the [performance guide](docs/performance-guide.md).

1. **Classpath separator was hardcoded to `:`** — on Windows it's `;`, so the forked JVM got one bogus entry and couldn't find `bloop.Bloop`. Likely why the old CI step was gated on `matrix.os != 'windows-latest'`. Now uses `File.pathSeparator`.

2. **`awaitPrompt` aborted on any `[E]`, including before compiling** — correct for the compile command, but `spawn()` reuses it for the startup prompt, where the output is just Bloop loading the build. A config pointing at an uninstalled JDK killed the run before a single measurement. `awaitPrompt` now takes `failOnError` (default `true`); `spawn()` passes `false` and prints to stderr instead.

Verified locally: a healthy build now measures (144 samples, 69.453 ± 23.505 ms/op); a genuinely failing compile still fails the run.

Note for restoring the CI step: JMH exits **0** when `@Setup` throws, so the step removed in #1692 was passing vacuously. It needs `-foe true`.


tgodzik (approved):
LGTM

Issue #1765, opened 2022-07-24
Bloop is no able to detect custom source sets


slovvik:
Given the following configuration:
```scala
lazy val SystemTest = config("system") extend(Test)

lazy val root = (project in file(".")).
configs(IntegrationTest, SystemTest)
  .settings(
    inConfig(SystemTest)(Defaults.testSettings),
    Defaults.itSettings,
    inThisBuild(List(
      organization := "com.example",
      scalaVersion := "2.13.6"
    )),
    name := "scalatest-example"
  )

libraryDependencies += "org.scalatest" %% "scalatest" % "3.2.9" % Test
libraryDependencies += "org.scalatest" %% "scalatest" % "3.2.9" % IntegrationTest
libraryDependencies += "org.scalatest" %% "scalatest" % "3.2.9" % SystemTest
```
and when running `sbt bloopInstall` the following files are generated:
```
[success] Generated .bloop/root.json
[success] Generated .bloop/root-test.json
[success] Generated .bloop/root-it.json
```
`bloop projects` also show only:
```
root
root-it
root-test
```
It looks like bloop is not able to detect the system-test source set.

dos65:
Thanks for reporting! Yep it doesn't support non build-in scopes.

I took a quick look so here a bit details:

`bloopGenerate` task is defined only for `Compile`, `Test`, `IntegrationTest` scopes:
https://github.com/scalacenter/bloop/blob/c1fd321b3f62da4fab79b19794616990c7b656c8/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala#L226-L228

And `bloopInstall` filter scopes where this task is defined:
https://github.com/scalacenter/bloop/blob/c1fd321b3f62da4fab79b19794616990c7b656c8/integrations/sbt-bloop/src/main/scala/bloop/integrations/sbt/SbtBloop.scala#L1109-L1114

Also, we can use this method imp from sbt-bsp as a reference: https://github.com/sbt/sbt/blob/1f29c9053cdc52128fa45cfe0b22e9a4d085cc1b/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala#L520

slovvik:
@dos65 thank you for taking a look. Is it something that can be taken by someone? :) We are frustrated with using Intellij, but without this feature, we can't switch. For me, it will take like a whole eternity to do it :(

Arthurm1:
[...]

PR #3074, opened 2026-09-15
Tag custom sbt configurations that extend Test as test targets

jozanek:
A custom sbt configuration that extends `Test` was exported with the `library` tag:

```scala
lazy val SystemTest = config("system") extend Test
```

Bloop only runs tests for targets tagged `test` or `integration-test`
(`TestTask.isTestProject`), which gates the `test` CLI, `Interpreter` and BSP
`buildTarget/test`, so the target compiled and then silently ran nothing:

```
$ bloop test root-system
Compiled root-system (137ms)
No test suites were run.
```

It now reports `All 1 test suites passed.` Metals keys off the same tag, so the suites
become runnable from the editor too.

### What changed

The tag was chosen by matching the configuration object itself; it is now chosen by walking
the configuration's lineage, checking `it` before `test` because `IntegrationTest` usually
extends `Test`.

One subtlety: `Keys.configuration.value.extendsConfigs` is **empty** when a project extends
the configuration at the `.configs(...)` site but passes the bare object to `inConfig`, e.g.
`.configs(Custom.extend(Test))` + `inConfig(Custom)(...)`. Resolving the project's own
configuration by name recovers it, the same workaround `eligibleDepsFromConfig` already
documents. The scripted test covers both spellings, since only the first one works without
that fallback.

### Scope

This fixes the second half of #1765. It does **not** make `bloopInstall` export custom
configurations automatically — that still needs
`inConfig(MyCustom)(BloopDefaults.configSettings)` as documented. I looked into automating it
and it needs a design decision, so I left #1765 open and wrote up what I found there.


tgodzik (approved):
LGTM!

